PR #1, opened 2017-05-31
update argument name in async example

weargoggles:
It seems like the `message` local used later in `sleepy_hello_world` isn't declared. I guess that the async version takes `message` like the sync version in the first example?


lovelydinosaur:
Apologies, I missed this! Good catch yup, tho all sorted now, happily.

Issue #8, opened 2017-06-13
SSL handling


lovelydinosaur:
Set up ssl context in the worker

PR #26, opened 2017-06-25
SSL support

maximilianhurl:
Testing command (with self signed certficiate hence 'not secure' warning)`uvicorn --certfile=server.crt --keyfile=server.key basic_app:hello_world`

screen shot 2017-06-25 at 12 33 46

Closes https://github.com/encode/uvicorn/issues/8


lovelydinosaur:
Rad! 😎

PR #36, opened 2017-06-27
Add BodyChannel.close()

38elements:
When `keep alive` is enabled, user can not stop  incoming data of `request body chunks`.
I think a function of  stopping incoming data is necessary.


lovelydinosaur:
The protocol class handles flow control, ensuring that reading data stops once the request buffer reaches a high watermark. There shouldn't be any need for a client to close the body stream - all it has to do is stop reading from it. We could support honouring the "Connection: close" response header tho.

38elements:
Thank you for your response.
I think that the flag of `Connection: close` is better.

Issue #2, opened 2017-06-07
Configurable http parser & event loop.


lovelydinosaur:
Eg. support h11, as used in https://github.com/jeamland/guvnor/blob/master/guvnor/asyncio_worker.py

In particular it'd be good to have a pure-python configuration available to us.

If anyone's interested in helping move this forward I'd suggest that a good first approach would be a pull request that *removes* http_parser, and replaces it with h11. That'll then help us figure out where their interfaces are different, and help us work towards what we'd like the generalized interface to be.

/cc @jeamland

lovelydinosaur:
Interestingly PyPy3 performance is looking pretty nifty: https://github.com/squeaky-pl/zenchmarks

Simplest tack onto a pure-python worker would be to use aiohttp

We'd want to override RequestHandler.start(message, payload, handler)

* https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_protocol.py#L397

Also worth checking out picohttpparser, as used by japronto.

leobuskin:
Probably we can try [PicoHTTPParser](https://github.com/h2o/picohttpparser) (h2o project) instead of httptools? I'm making 99% compatible drop-in replacement for httptools based on [httpparser](https://github.com/dontcare/httpparser) project (it's based on PicoHTTPParser but has familiar to httptools interface, just no tests, a bit another naming). 

Probably uvicorn can be more faster with such replacement, @tomchristie did you research this area already? Is httptools affects speed somehow or it's far not a potential bottleneck and replacement will change nothing?

lovelydinosaur:
@leobuskin Not sure. It'd be interesting to find out, yup.

jordaneremieff:
Is there still interest `h11` support? If so, how should it be implemented alongside the other protocols?

lovelydinosaur:
Resolved in 0.2.

`--loop [asyncio|uvloop]` and `--http [h11|httptools]`

uvloop and httptools remain the default.

For gunicorn use `gunicorn -k uvicorn.workers.UvicornWorker` for the defaults, and `gunicorn -k uvicorn.workers.UvicornH11Worker` for pypy compatible.

PR #44, opened 2017-06-30
h11 implementation

lovelydinosaur:
Refs https://github.com/encode/uvicorn/issues/2

We'd like to support a pure-python configuration.
This is a first-pass attempt towards that, using the rather nice `h11` library.

Testing with pypy3 this still appears to be significantly slower that the C-implemented `http_parser`/`uvloop` combination, but it's still a worthwhile thing to pursue.

Possibly add to https://github.com/squeaky-pl/zenchmarks if this gets completed sometime.


tomchristie:
Refs https://github.com/encode/uvicorn/pull/44

PR #51, opened 2017-07-22
Added draft plain Serve functionality

### LINKED ISSUE #7: .serve() interface
We should expose a plain python `.serve(host, port)` style interface, for when you don't want fully-fledged gunicorn process management.

maximilianhurl:
This is my initial pass at a plain serve interface #7 

### Usage:

```
import uvicorn

async def hello_world(message, channels):
    content = b'Hello, world'
    response = {
        'status': 200,
        'headers': [
        ],
        'content': content
    }
    await channels['reply'].send(response)


if __name__ == '__main__':
    uvicorn.serve(hello_world)
```

screen shot 2017-07-26 at 14 36 55

### Notes:

I've kept the structure similar to the gunicorn worker so could refactor it into mixin if you approve of the approach (plus add tests for them both).

Other things to consider:

- Using sockets
- Running multiple server processes using`multiprocessing`
- SSL Support
- any other important settings?


lovelydinosaur:
Neato! 👍 

First thing to mention is that we *won't* have a command line app for this, rather just `python app.py`, and use this pattern in the user's application code...

    if __name__ = '__main__':
        uvicorn.run(...)

Once that's sorted will have a think about if there's any refactoring or suchlike required.

maximilianhurl:
Let me know any refactoring points you have and whether it is even worth using a mixin to make this and the gunicorn worker reuse code.

lovelydinosaur:
> using a mixin

n.e.v.e.r mixins.

This is great as it stands. It's waiting for me to get the next release of apistar with asyncio support, so that I can figure if there's any interface tweaking we'll want to do as a result. Eg. we might want things like reloading and the debugger to be options on the serve command.

Reckon we should leave the PR in its current state, and pull it in once I start looking at integrating this using apistar. Thanks again - crackin work!

maximilianhurl:
> n.e.v.e.r mixins.

Heh, noted. Well I'll leave this for now and have a look at the next thing - postgres support on uvitools rather than adding tests for this and the worker.

lovelydinosaur:
If you mean LISTEN/NOTIFY broadcast support, then yeah that'd be really nice!

lovelydinosaur:
Let's 'av it!
Nice one - will be integrating this with https://github.com/tomchristie/apistar/pull/235 in short order.

PR #52, opened 2017-07-27
support py3 wheels

graingert:
will need to be run with "python3 setup.py sdist bdist_wheel upload"

And because you're only supporting brand new Python 3 installs it might even not be worth uploading the sdist at all:

will need to be run with "python3 setup.py bdist_wheel upload"

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[bdist_wheel]
+python-tag = py3


lovelydinosaur:
Great thanks!
I'll consider pulling this in next time I'm about to roll a release. Thanks so much!

lovelydinosaur:
👍  Thanks!

PR #59, opened 2018-01-29
Don't override path if it was not provided yet

### LINKED ISSUE #58: Unhandled exception when httptools.parse_url() returns a None path
I encountered the following exception while running an apistar application using uvicorn 0.0.15 on CPython 3.6.4 with SSL enabled:

```
Unhandled exception in event loop
  File "httptools/parser/parser.pyx", line 241, in httptools.parser.parser.cb_on_url
  File "/home/omer/.local/share/virtualenvs/myproject--RkZn49j/lib/python3.6/site-packages/uvicorn/protocols/http.py", line 278, in on_url
    'path': parsed.path.decode('ascii'),
AttributeError: 'NoneType' object has no attribute 'decode'

  File "uvloop/handles/stream.pyx", line 784, in uvloop.loop.__uv_stream_on_read_impl
  File "uvloop/handles/stream.pyx", line 563, in uvloop.loop.UVStream._on_read
  File "/home/omer/.pyenv/versions/3.6.4/lib/python3.6/asyncio/sslproto.py", line 514, in data_received
    self._app_protocol.data_received(chunk)
  File "/home/omer/.local/share/virtualenvs/myproject--RkZn49j/lib/python3.6/site-packages/uvicorn/protocols/http.py", line 238, in data_received
    self.request_parser.feed_data(data)
  File "httptools/parser/parser.pyx", line 189, in httptools.parser.parser.HttpParser.feed_data
httptools.parser.errors.HttpParserCallbackError: the on_url callback failed
```

The exception occurs when the empty HTTP message `b' HTTP/1.1\r\n'` is provided in https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/http.py#L238

Unfortunately I'm unable to reproduce this problem with requests or curl.
So far the only HTTP client that reproduces this bug is [OSQuery](http://osquery.readthedocs.io/) 2.11.2 with the [remote API](http://osquery.readthedocs.io/en/stable/deployment/remote/) enabled. OSQuery 2.8.0 does not present the same problem.

I'm running uvicorn with the following parameters:
```
uvicorn myproject.app:app \
            --workers=8 \
            --bind=localhost:8080 \
            --keyfile=./server.pem \
            --certfile=./server.pem \
            --ca-certs=./ca.pem \
            --log-level DEBUG \
            --access-logfile=- \
            --do-handshake-on-connect \
            --ciphers TLSv1.2
```
and OSQuery with the following parameters:
```
sudo osqueryd --pidfile /tmp/osqueryd.pid --verbose --debug \
    --database_path /tmp/osquery.db/ \
    --tls_hostname localhost:8080 \
    --config_plugin tls \
    --tls_server_certs /usr/local/share/ca-certificates/ca.crt \
    --config_tls_endpoint /collection/configuration \
    --logger_tls_endpoint /collection/log_event \
    --logger_plugin tls \
    --enroll_tls_endpoint /collection/enroll \
    --disable_distributed=false --distributed_plugin=tls --distributed_interval=60 \
    --distributed_tls_read_endpoint /collection/queries \
    --distributed_tls_write_endpoint /collection/results \
    --watchdog_level -1 \
    --enroll_secret_path ./etc/test_enroll_secret.txt
```

In order to reproduce the problem you'll need to generate your own certificates.
I recommend [trustme](https://github.com/python-trio/trustme) for that purpose.

I'm going to try to reverse proxy with Nginx next to figure out if OSQuery is at fault here but in any case, the uvicorn should handle malformed (or incomplete? not sure about that) HTTP requests more gracefully.

thedrow:
Fixes #58.

### DIFF
--- a/uvicorn/protocols/http.py
+++ b/uvicorn/protocols/http.py
@@ -272,13 +272,16 @@ class HttpProtocol(asyncio.Protocol):
         parsed = httptools.parse_url(url)
         method = self.request_parser.get_method()
         http_version = self.request_parser.get_http_version()
-        self.message.update({
+        message_attributes = {
             'http_version': http_version,
             'method': method.decode('ascii'),
-            'path': parsed.path.decode('ascii'),
             'query_string': parsed.query if parsed.query else b'',
             'headers': self.headers
-        })
+        }
+        if parsed.path:
+            message_attributes['path'] = parsed.path.decode('ascii')
+
+        self.message.update(message_attributes)
 
     def on_header(self, name: bytes, value: bytes):
         name = name.lower()


lovelydinosaur:
Not obvious to me that this would be a correct fix. When would 'path' ever get filled, with this code?

I think we'd need to replicate the issue in the tests, by explicitly feeding in the exact data that causes the issue. Eg. https://github.com/encode/uvicorn/blob/master/tests/test_http.py#L61

Issue #56, opened 2017-11-10
Error using websockets==4.0.1


viniciuschiele:
Hi,

I was trying to run this example using uvicorn + websockets 4.0.1 (latest version):

```python
async def chat_server(message, channels):
    """
    ASGI-style 'Hello, world' application.
    """
    if message['channel'] == 'websocket.connect':
        await channels['reply'].send({'accept': True})
```

And I get this error after sending the accept message.

```
Starting worker [9001] serving at: 0.0.0.0:8000
Task exception was never retrieved
future: <Task finished coro=<reader() done, defined at /home/vinicius/projects/dealer/venv/lib/python3.6/site-packages/uvicorn/protocols/websocket.py:65> exception=TypeError('An asyncio.Future, a coroutine or an awaitable is required',)>
Traceback (most recent call last):
  File "/home/vinicius/projects/dealer/venv/lib/python3.6/site-packages/uvicorn/protocols/websocket.py", line 72, in reader
    data = await protocol.recv()
  File "/home/vinicius/projects/dealer/venv/lib/python3.6/site-packages/websockets/protocol.py", line 309, in recv
    loop=self.loop, return_when=asyncio.FIRST_COMPLETED)
  File "/usr/lib/python3.6/asyncio/tasks.py", line 311, in wait
    fs = {ensure_future(f, loop=loop) for f in set(fs)}
  File "/usr/lib/python3.6/asyncio/tasks.py", line 311, in <setcomp>
    fs = {ensure_future(f, loop=loop) for f in set(fs)}
  File "/usr/lib/python3.6/asyncio/tasks.py", line 526, in ensure_future
    raise TypeError('An asyncio.Future, a coroutine or an awaitable is '
TypeError: An asyncio.Future, a coroutine or an awaitable is required
```
It works fine with websockets 3.4.0

I was looking into it and I noticed that in the protocol.py (from websockets) the variable ```transfer_data_task``` is not initialized so a ```None``` is being added to the queue of events causing the error.

It could be a bug in websockets 4.0.* or they have changed the logic and uvicorn is not compatible anymore.

jordaneremieff:
I had this same issue working on my fork - using 3.4.0 allowed websockets to work normally.

lovelydinosaur:
Thanks for the confirmation @erm. Hopefully I’ll get back to this given the renewed work on API Star. Bunch of other work I’ll need to get through first before coming back to websockets, tho.

jordaneremieff:
[...]

PR #62, opened 2018-04-03
Refactor for updated ASGI spec and Django/Channels integration

jordaneremieff:
Refs https://github.com/encode/uvitools/issues/2

Still WIP, but the worker runs both standalone examples and with normal Django views and Channels consumers.

I've ended up with a pattern of creating ASGI application instances from the protocols and associating them with specified handlers - I'm not quite certain if this ideal or not yet, but it seems to be the cleanest approach I've tried thus far. The handlers include a websocket handler and http request handler.

The flow control methods and channel classes that existed before have been removed. I will still need to work out how much of the previous plumbing and various protocol methods need to be re-implemented in the handlers. Also, cookies/headers/misc request information needs to be handled properly - right now things are in a relatively bare-minimum state to get the examples running/debugging.

There was also a bug mentioned in https://github.com/encode/uvicorn/issues/56 that is resolved.

Example apps available in: https://github.com/erm/uvicorn-examples


jordaneremieff:
When I have time this evening I will be pulling out the examples into a separate repo to make the diff easier to read.

jordaneremieff:
Given the work in https://github.com/encode/uvicorn/pull/63 I'll be closing this PR.

PR #63, opened 2018-04-06
ASGI support

lovelydinosaur:
Since the ASGI specification has been updated it now fulfils the requirements that the "Uvicorn Messaging Interface" was intended to bridge.

This pull request updates Uvicorn to become an ASGI-compliant webserver.

* I've currently dropped the flow control, which will need to be updated.
* I've currently dropped the websocket handling, which will need to be updated.
* After this change, Uvicorn will no longer support API Star 0.3, which uses the old "Uvicorn Messaging Interface". Users wanting to run API Star 0.3 async applications will need to stick with version 0.0.15. There is an [API Star 0.4 version](https://github.com/encode/apistar/pull/400) shortly due to be released, which updates API Star async apps to use ASGI.


uvicorn/protocols/http.py
```diff
@@ -237,86 +194,82 @@ def data_received(self, data):
...
+        self.body = b''
+        self.queue = None
 
     def on_url(self, url):
         parsed = httptools.parse_url(url)
         method = self.request_parser.get_method()
         http_version = self.request_parser.get_http_version()
-        self.message.update({
+        self.scope = {
+            'type': 'http.request',
```
jordaneremieff: The connection scope type needs to be set to `'type': 'http'`. 

https://github.com/django/asgiref/blob/master/specs/www.rst#connection-scope

lovelydinosaur: Thanks!

PR #64, opened 2018-04-06
ASGI Websocket support

jordaneremieff:
Updating websocket protocol for use with ASGI changes. 
Example app added. 
Still need to add tests and handle handshake failure case.

Resolves https://github.com/encode/uvicorn/issues/56


lovelydinosaur:
Fab, thanks! I should be able to start reviewing stuff on Monday.

uvicorn/protocols/http.py
```diff
@@ -247,10 +249,11 @@ def on_body(self, body: bytes):
         self.body = body
 
     def on_message_complete(self):
-        self.body_queue({
-            'type': 'http.request',
-            'body': self.body
-        })
+        if self.active_request:
```
lovelydinosaur: Note that ‘body_queue’ and ‘active_request’ don’t (necessarily) point to the same thing. (The request currently being parsed by the server vs. the request currently being handled by a client)

I guess if we value clarity rather than microoptimizations then I could make that clearer in the codebase.

uvicorn/protocols/websocket.py
```diff
@@ -0,0 +1,138 @@
...
+        # }))
+        return
+
+    protocol = WebSocketProtocol(http, response_headers)
+    protocol.connection_open()
+    protocol.connection_made(http.transport, http.scope)
+    http.transport.set_protocol(protocol)
+
+
+class Request():
```
auvipy: just curious, why it is Req(): not Req: ?

jordaneremieff: The code in the parent branch defines classes in this way, so I was following the existing convention to keep things consistent - I am not sure if there was a particular reason.

uvicorn/protocols/websocket.py
```diff
@@ -135,4 +135,4 @@ def accept(self):
     def reject(self):
         rv = b'HTTP/1.1 403 Forbidden\r\n\r\n'
         self.transport.write(rv)
-        self.transport.close()
+        self.transport.close()
```
auvipy: newline

uvicorn/protocols/websocket.py
```diff
@@ -0,0 +1,149 @@
...
+    http.transport.set_protocol(protocol)
+
+
+async def reader(protocol):
+    close_code = None
+    order = 1
+    path = protocol.scope['path']
+    loop = protocol.loop
+    try:
+        async for data in protocol:
```
jordaneremieff: Async iteration will not work with Python < 3.6, so I will change it back to using a loop/await when I have time tonight.

lovelydinosaur:
@erm I've pulled in the ASGI changes now and released as 0.1.
Wanna re-issue this against master?

jordaneremieff:
@tomchristie Sure, will do so now.

Btw, great work!  🎉

PR #65, opened 2018-04-09
Websocket support for new ASGI spec

jordaneremieff:
[WIP] Support for websockets using new ASGI spec. Currently have it working with Django and standalone ASGI consumers.  It still needs tests and to have its opening/closing/timeout behavior refined and some specific cases handled.

Resolves https://github.com/encode/uvicorn/issues/56


erm:
I am not sure if the reader being created with the session the way it is now is ideal, it needs some more thought. Probably won't have time until the weekend to look at this further.

jordaneremieff:
I've tested the current state of this with the multichat example [here](https://github.com/andrewgodwin/channels-examples/tree/master/multichat/), if I comment out the line that does the initial check for anonymous user it runs fine.

If I do not comment out that check then I receive the error:`Exception inside application: You cannot use AsyncToSync in the same thread as an async event loop - just await the async function directly.`. This occurs on both daphne and uvicorn, and I believe it may be related to https://github.com/django/channels/issues/944. (I haven't investigated this too much specifically, but figured it was worth mentioning.)

lovelydinosaur:
Fab to see so much work going into this. Planning to take a look later on in the week! 😄

uvicorn/protocols/websocket.py
```diff
@@ -82,8 +82,8 @@ def put_message(self, message):
 
         if self.protocol.state == websockets.protocol.State.OPEN:
             if not self.protocol.accepted:
-                accept = bool(message_type == 'websocket.accept')
-                close = bool(message_type == 'websocket.close')
+                accept = not not message_type == 'websocket.accept'
+                close = not not message_type == 'websocket.close'
```
lovelydinosaur: `not not message_type  == 'websocket.close'`

🤔

lovelydinosaur: `close = (message_type == 'websocket.close')` ?

sirex: `not not` sounds like a JavaScript.

```python
close = message_type == 'websocket.close'
```

This is totally enough and there is no need for parenthesis.

```python
>>> message_type = 'websocket.close'
>>> close = message_type == 'websocket.close'
>>> close
True
```

lovelydinosaur: Yes, tho possibly more readable with the explicit parenthesis. It's a bit hard to otherwise tell if it's intended to be like that otherwise, or an error.

sirex: It's python, explicit parenthesis are not needed, for example, no one writes:

```python
if (condition):
    ...
```

But I guess need for explicit parenthesis comes from experience with other programming languages where parenthesis are required.

Nevermind.

jordaneremieff: I had initially just went with `not not` because I was replacing the `bool` call based on the information on this page https://wiki.python.org/moin/PythonSpeed, but it doesn't really matter here.

lovelydinosaur:
Mate, this looks *great*!
Really nice PR - top-notch stuff.

lovelydinosaur:
Okay great, so that's all in now, and released as 1.1.
We'll need to update the docs too.

I think the first thing to tackle is updating the index.md page so that it includes both HTTP and WebSocket sections. Basically updating https://github.com/encode/uvicorn/blob/0.0.15/docs/index.md#websockets for the new ASGI spec.

Really this is more walking users through ASGI, rather than walking them through uvicorn as such, but that's still kinda necessary.

Thinking aloud about other stuff we'll want to move on to...

* Considering broadcast support.
* Perhaps documenting usage with channels. (Let's not *overly* focus on that tho, since it's only one of the use-cases. Might also be using it to run API Star, or to run stuff directly on top of)
* Documenting command line options etc...
* HTTP/2
* I reckon move away from gunicorn for running from the commandline, just run directly, perhaps with eg. click, but *do* include a gunicorn worker class, and document using that *properly* with gunicorn.
* Update [the uvicorn cases in the techempower benchmarks](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Python/uvicorn), perhaps also include a quick comparison against `node`,`sanic`, and `go` in this repo.

PR #67, opened 2018-04-24
Allow consumers to negotiate websocket subprotocols during handshake

jordaneremieff:
Subprotocol support for websockets. If subprotocols are listed in the request header then they are added to the response headers and the scope to allow the consumer to specify a protocol. Conforms to the spec information here: https://github.com/django/asgiref/blob/master/specs/www.rst#accept.

### DIFF
--- a/tests/test_websocket.py
+++ b/tests/test_websocket.py
@@ -230,3 +230,24 @@ def test_send_after_protocol_close():
         assert data == '123'
         assert not is_open
         loop.close()
+
+
+def test_subprotocols():
+    class App:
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
Okay, looks good. Is there anywhere in particular I should read up for context on this, or a particular use-case/example?

jordaneremieff:
Relevant channels issue: https://github.com/django/channels/issues/930

An example use-case is using this GraphQL subscriptions library: https://github.com/graphql-python/graphql-ws.

Issue #70, opened 2018-05-05
AttributeError: 'uvloop.loop.TCPTransport' object has no attribute 'drain'


parisk:
Hi there,

First of all, congratulations for Uvicorn — it's truly amazing and pleasant to use!

Today I was messing around with Django 2.0 + Channels and thought about giving Uvicorn a try as a replacement for Daphne.

Although everything seems to work like a charm, I see an `AttributeError` in my server's log, without this exception surfacing to the browser. It seems to be caused by the following piece of code:

https://github.com/encode/uvicorn/blob/24bf63f5c6e1f3fa226e417b89f4e4d32ffa26e4/uvicorn/protocols/http.py#L154-L157

Providing some more details below (full traceback + Python package + system details)

## Traceback

```python
Not Found: /favicon.ico
Task exception was never retrieved
future: <Task finished coro=<AsgiHandler.__call__() done, defined at /mnt/data/.python-3.6/lib/python3.6/site-packages/channels/http.py:172> exception=AttributeError("'uvloop.loop.TCPTransport' object has no attribute 'drain'",)>
Traceback (most recent call last):
  File "/mnt/data/.python-3.6/lib/python3.6/site-packages/channels/http.py", line 190, in __call__
    await self.handle(body)
  File "/mnt/data/.python-3.6/lib/python3.6/site-packages/asgiref/sync.py", line 110, in __call__
    return await asyncio.wait_for(future, timeout=None)
  File "/usr/local/lib/python3.6/asyncio/tasks.py", line 339, in wait_for
    return (yield from fut)
  File "/usr/local/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/mnt/data/.python-3.6/lib/python3.6/site-packages/asgiref/sync.py", line 125, in thread_handler
    return self.func(*args, **kwargs)
  File "/mnt/data/.python-3.6/lib/python3.6/site-packages/channels/http.py", line 226, in handle
    self.send(response_message)
  File "/mnt/data/.python-3.6/lib/python3.6/site-packages/asgiref/sync.py", line 64, in __call__
    return call_result.result()
  File "/usr/local/lib/python3.6/concurrent/futures/_base.py", line 432, in result
    return self.__get_result()
  File "/usr/local/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
[...]

PR #71, opened 2018-05-16
Awaitable event for write flow control and drain method

jordaneremieff:
Refers to https://github.com/encode/uvicorn/issues/70

Calling `transport.drain()` when writing is paused raises an exception because uvloop doesn't implement a drain method. I found this solution implemented in the following places:

https://gitlab.com/pgjones/quart/commit/cf82d539c67125eb3c3b96e2decb65966f6d174b
https://github.com/channelcat/sanic/pull/1179/commits/89df1b427e66097b5ac4c7d69f720ca9afbfb2c7

I was able to re-create the error described in the issue with an http chunking consumer:

`future: <Task finished coro=<App.__call__() done, defined at /Users/erm/Documents/GitHub/asgi/chunking.py:10> exception=AttributeError("'uvloop.loop.TCPTransport' object has no attribute 'drain'",)>
  File "/Users/erm/Documents/GitHub/asgi/chunking.py", line 40, in __call__
    "more_body": bool(i != x),
  File "/Users/erm/Documents/GitHub/asgi/uvicorn/uvicorn/protocols/http.py", line 156, in send
    await self.transport.drain()
AttributeError: 'uvloop.loop.TCPTransport' object has no attribute 'drain'`

After implementing the drain method this no longer occurs. Please advise any changes that may be necessary as my current understanding of the flow control mechanisms is limited.

### DIFF
--- a/uvicorn/protocols/http.py
+++ b/uvicorn/protocols/http.py
@@ -152,8 +152,7 @@ class RequestResponseCycle:
         else:
             raise Exception('Unexpected message type "%s"' % message_type)
 
-        if self.protocol.is_writing:
-            await self.protocol.drain()
+        await self.protocol.drain()
 
         if self.state == RequestResponseState.CLOSED:
             self.protocol.on_response_complete(keep_alive=self.keep_alive)
@@ -186,11 +185,12 @@ class HttpProtocol(asyncio.Protocol):
         # Flow control
         self.buffer_size = 0
         self.read_paused = False
+        self.write_paused = False
         self.high_water_limit = HIGH_WATER_LIMIT
         self.low_water_limit = LOW_WATER_LIMIT
         self.max_pipelined_requests = MAX_PIPELINED_REQUESTS
-        self.is_writing = asyncio.Event(loop=loop)
-        self.is_writing.set()
+        self.drain_waiter = asyncio.Event()
+        self.drain_waiter.set()
 
     # The asyncio.Protocol hooks...
     def connection_made(self, transport):
@@ -212,21 +212,26 @@ class HttpProtocol(asyncio.Protocol):
             websocket_upgrade(self)
 
     # Flow control
+
     def pause_writing(self):
-        self.is_writing.clear()
+        self.write_paused = True
+        self.drain_waiter.clear()
 
     def resume_writing(self):
-        self.is_writing.set()
+        self.write_paused = False
+        self.drain_waiter.set()
 
     async def drain(self):
-        await self.is_writing.wait()
+        if not self.write_paused:
+            return
+        await self.drain_waiter.wait()
 
     def check_pause_reading(self):
         if self.read_paused:
             return
 
         if (self.buffer_size > self.high_water_limit or
-            len(self.pipelined_requests) > self.max_pipelined_requests):
+                len(self.pipelined_requests) > self.max_pipelined_requests):
             self.transport.pause_reading()
             self.read_paused = True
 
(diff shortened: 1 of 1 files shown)


jordaneremieff:
To provide a bit more context, this issue has some relevant discussion: https://github.com/channelcat/sanic/issues/1176

danigosa:
@erm could you post the chunked consumer code/tool?

jordaneremieff:
@danigosa I was using a chunked consumer code initially, but have since been able to reproduce by triggering the drain method with a larger request that does not chunk, here are two examples:

```python
import asyncio


class App:

    def __init__(self, scope):
        self.scope = scope

        await send({
            'status': 200,
            'headers': [
                (b"content-type", b"text/plain"),
            ]
        })

        body = b"Hello World." * 65536

        await send({
            'type': 'http.response.body',
            'body': body,
            'more_body': False
        })
        await asyncio.sleep(0.1)

```

and a chunking consumer:

```python
import asyncio


class App:

    def __init__(self, scope):
        self.scope = scope

        await send({
            'status': 200,
            'headers': [
                (b"cache-control", b"no-cache"),
                (b"content-type", b"text/event-stream"),
                (b"transfer-encoding", b"chunked"),
            ]
        })

        body = b"Hello world." * 65536
        max_ = 5
        for i in range(1, max_ + 1):
            if i == max_:
                more_body = False
            else:
                more_body = True
            await send({
                'type': 'http.response.body',
                'body': body,
                'more_body': more_body
            })
            await asyncio.sleep(0.1)

```

danigosa:
@erm awesomeness, really appreciate the sharing 👍  asyncio based servers (and even worse uvloop ones) they are still having problems compared to the oldie pure WSGI. I will investigate.

jordaneremieff:
@danigosa what other problems are you running into? If you open any issues I will help where possible.

danigosa:
Mainly problems with flow control and uvloop/asyncio discrepancies, nowadays the most stable framework by far is `aiohttp`.

jordaneremieff:
@danigosa Other than the flow control issue described here, are there other issues you have encountered?

danigosa:
@erm I experienced SSL nightmares with `uvloop` in earlier versions but in production we do SSL termination so I did not have to investigate it a lot, and I think in the new version (0.10) they have focused on improving this

danigosa:
@erm BTW we are using a patched version of `uvicorn` in production with your PR and we cannot find issues (using your consumers example), whilst the master version does for the specific tests

lovelydinosaur:
Closing in favor of #81.

lovelydinosaur:
Actually, reviewing this, I'm not so sure I do want to close in favor of #81.

jordaneremieff:
After experimenting with the protocol behavior quite a bit more, I've realised this solution is actually incomplete - it needs to also handle the drain event when the connection is lost as well. Closing this now since it seems a different solution will likely be implemented in the next release.

danigosa:
@erm 
 is this what you refer to:

```Exception
'Unexpected ASGI message type 'http.disconnect'.
apistar/server/asgi.py in resolve at line 115
            return body
        while True:
            message = await receive()
            if not message['type'] == 'http.request':
                error = "'Unexpected ASGI message type '%s'."
                raise Exception(error % message['type'])
            body += message.get('body', b'')
            if not message.get('more_body', False):
                break
        return http.Body(body)
```

It's happening from time to time in our servers, going back to master

PR #77, opened 2018-06-02
Minor changes to match ASGI Specification

pgjones:
I think these two are required to match the ASGI specification (and I need them for Uvicorn to serve Quart :P).


lovelydinosaur:
Yup both look right to me after a quick check against the spec. Nice one!

Issue #79, opened 2018-06-11
Block of upload file a big file.


ShichaoMa:
```
  def on_body(self, body: bytes):
        if self.body:
            self.parsing_request.put_message({
                'type': 'http.request',
                'body': self.body,
                'more_body': True
            })
            self.check_pause_reading()
        self.body = body
```
because of  this method, when upload a file bigger than 1M, method `check_pause_reading` will block
 with  self.read_paused set to True。

lovelydinosaur:
Flow control's rejigged in 0.2 so going to close this off, tho please re-raise if you're still seeing issues.

PR #80, opened 2018-06-12
fixed big file upload block

ShichaoMa:
#79

### DIFF
--- a/uvicorn/protocols/http.py
+++ b/uvicorn/protocols/http.py
@@ -71,6 +71,8 @@ class RequestResponseCycle:
     async def receive(self):
         message = await self.receive_queue.get()
         self.protocol.buffer_size -= len(message.get('body', b''))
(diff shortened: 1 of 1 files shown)


uvicorn/protocols/http.py
```diff
@@ -71,6 +71,8 @@ def put_message(self, message):
     async def receive(self):
         message = await self.receive_queue.get()
         self.protocol.buffer_size -= len(message.get('body', b''))
+        if self.protocol.buffer_size <= 0 and message.get("more_body"):
```
lovelydinosaur, line 74: I think this check should probably just be `if self.protocol.read_paused`

ShichaoMa: ok, just fix it

lovelydinosaur:
Closing as per #79

PR #84, opened 2018-06-27
Version 0.2

lovelydinosaur:
* *Extensive* improvements to error handling.
* Adds an `h11` implementation alongside the existing `httptools` - Can now run under PyPy.
* Fixes access logs.
* `uvicorn` command line now runs stand-alone, rather than under gunicorn. Gunicorn worker class documented.
* Nice throughput bump.

TODO: WebSocket tests are hanging under travis.


lovelydinosaur:
Going to take another go at resolving the websocket tests hanging under travis. They run to completion locally. I don’t want the release blocked on that so I might have to proceed anyway and come back to them.

Probably worth trying to remove the multi-threading and run them all under asyncio with pytest-asyncio or similar.

Any insights from anyone also much appreciated here!

lovelydinosaur:
One other note on this... both the h11 and httptools implementations share exactly the same structure. I’m really happy with that as I think it’ll help keep maintainance lower, and it gives me a higher confidence in correctness.

I’ve also tackled it with later cythonisation in mind - sticking to basic data structures throughout.

Reminder to self: Break the cycle/protocol circular reference. Improve the read flow control so it blocks less often, using high/low watermarks instead of existing read/wait/read/wait style.

lovelydinosaur:
Downgrade websockets version to a known passing case.
Worth investigating, since we *will* want to upgrade to the latest, but this should do for now.

lovelydinosaur:
Also hanging on 4.0.1

lovelydinosaur:
Okay, lets do this!

Issue #86, opened 2018-06-29
Uvicorn 0.2.0 package missing README.md


Axik:
```Collecting uvicorn
  Using cached https://files.pythonhosted.org/packages/c7/a7/92333ff7e11f703af974485247ae1231bed96d3e9b8e5a48f4009d985761/uvicorn-0.2.0.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/4g/1rsbx7zd74z2pz2x_qhbhwlm0004z1/T/pip-install-92dplo46/uvicorn/setup.py", line 44, in <module>
        long_description=get_long_description(),
      File "/private/var/folders/4g/1rsbx7zd74z2pz2x_qhbhwlm0004z1/T/pip-install-92dplo46/uvicorn/setup.py", line 23, in get_long_description
        return open('README.md', 'r').read()
    FileNotFoundError: [Errno 2] No such file or directory: 'README.md'

```

I believe, include those to MANIFEST could help (not sure, I always had troubles with python packaging)

lovelydinosaur:
Thanks for notifying me about this - after a bunch of messing about I've now resolved the issue by upgrading twine.

Threw me a bit, since it's almost identical to `starlette`, but was failing in this case.

PR #87, opened 2018-06-29
Add manifest file to include readme to dist

Axik:
that would help with https://github.com/encode/uvicorn/issues/86
@tomchristie


lovelydinosaur:
Resolved by upgrading twine.
Thanks! :)

Issue #103, opened 2018-06-29
Upgrade `websockets` package


lovelydinosaur:
Currently on 3.3. Unresolved issue with tests hanging on higher versions, see: https://github.com/encode/uvicorn/pull/84

Needs looking into.

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/uvicorn/issues/103">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/103/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/uvicorn/issues/103/pledge.svg">
</picture>
</a>

PR #108, opened 2018-06-30
Websockets version, test

jordaneremieff:
Refs: https://github.com/encode/uvicorn/issues/103

I think the issue with the tests hanging may have to do with `protocol.connection_open()` not being re-added when testing against the higher versions with `thread.join()`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ click
 h11
 httptools
 uvloop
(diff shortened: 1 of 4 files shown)


lovelydinosaur (commented):
Amazing! We oughta unpin the version in setup.py too

PR #119, opened 2018-07-07
Allow receipt of http.disconnect messages after body consumption

pgjones:
This allows ASGI applications to know that the connection has
disconnected after consuming the entire request body. This is
important if the application is streaming a response as the send will
become a no-op and hence not indicate the connection has closed.


lovelydinosaur:
Yup - I'm in favor of this. Thanks!

Calls to `receive()` after the end of the response body will hang until any disconnect message. Perfectly reasonable.

lovelydinosaur:
Released as 0.2.8

Issue #90, opened 2018-06-29
Support websocket upgrades from `h11` implementation


lovelydinosaur:
The `httptools` implementation (default) supports websockets upgrades, but the `h11` implementation doesn't yet.

Add this support, and then ensure that the websocket tests run against both protocol classes.

PR #121, opened 2018-07-12
Support websocket upgrades from `h11` implementation

jordaneremieff:
Closes https://github.com/encode/uvicorn/issues/90

Raises an exception when parsing the initial connection event within the `H11Protocol` to upgrade to the `WebSocketProtocol`. The attributes`headers` and `scope` have been added to the protocol class constructor in order to duplicate the upgrade behavior of the `HttpToolsProtocol`.

Tests have been updated accordingly, following the example in the http tests.

### DIFF
--- a/uvicorn/protocols/http/h11.py
+++ b/uvicorn/protocols/http/h11.py
@@ -32,10 +32,6 @@ DEFAULT_HEADERS = _get_default_headers()
 HIGH_WATER_LIMIT = 65536
 
 
-class WebSocketUpgrade(Exception):
-    """Raised during a WebSocket upgrade"""
-
-
 class H11Protocol(asyncio.Protocol):
     def __init__(self, app, loop=None, state=None, logger=None):
         self.app = app
@@ -97,10 +93,7 @@ class H11Protocol(asyncio.Protocol):
 
     def data_received(self, data):
         self.conn.receive_data(data)
-        try:
-            self.handle_events()
-        except WebSocketUpgrade:
-            websocket_upgrade(self)
+        self.handle_events()
 
     def handle_events(self):
         while True:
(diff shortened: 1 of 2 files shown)


uvicorn/protocols/http/h11.py
```diff
@@ -126,9 +137,14 @@ def handle_events(self):
...
                     "path": unquote(path.decode("ascii")),
                     "query_string": query_string,
-                    "headers": event.headers,
+                    "headers": self.headers,
                 }
-                self.cycle = RequestResponseCycle(scope, self)
+                for header in self.headers:
+                    name, value = header[0].lower(), header[1].lower()
+                    if name == b"upgrade" and value == b"websocket":
+                        raise WebSocketUpgrade()
```
lovelydinosaur: Let's just call `websocket_upgrade(self)` here.

uvicorn/protocols/http/h11.py
```diff
@@ -90,7 +97,10 @@ def eof_received(self):
 
     def data_received(self, data):
         self.conn.receive_data(data)
-        self.handle_events()
+        try:
+            self.handle_events()
+        except WebSocketUpgrade:
```
lovelydinosaur: We can drop this exception flow.

lovelydinosaur:
Awesome stuff, yas!

lovelydinosaur:
> The attributes headers and scope have been added to the protocol class constructor in order to duplicate the upgrade behavior of the HttpToolsProtocol.

Fair enough, yup. I guess we might finesse the websocket implementation at some point and pass things in explicitly, but this seems pretty sensible for now.

lovelydinosaur:
Rad. :)

PR #126, opened 2018-07-14
Updates to the http receive functions

pgjones:
These are required for Uvicorn to work with Quart, as Quart has a different interpretation of the ASGI specification. Firstly in a29ddab it assumes the request.body message with more_body equal to False should be received only once and then the receive should block until something else happens. Secondly in e23058d it assumes it can call the receive after sending the response and get a disconnect message. In think this is closer to the Daphne and Hypercorn interpretations.

I've suggested a clarification to the [ASGI spec](https://github.com/django/asgiref/pull/53) that is relevant to the second commit.

### DIFF
--- a/uvicorn/protocols/http/h11.py
+++ b/uvicorn/protocols/http/h11.py
@@ -325,20 +325,13 @@ class RequestResponseCycle:
             protocol.conn.start_next_cycle()
 
     async def receive(self):
-        # If a client calls recieve once they've already sent the response
-        # then raise an error. Allows us to stop buffering any more request
-        # body to memory once the response has been sent.
-        if self.response_complete:
-            msg = "Response already sent. Receive channel no longer available."
-            raise RuntimeError(msg)
-
         protocol = self.protocol
         protocol.resume_reading()
 
         await protocol.client_event.wait()
         protocol.client_event.clear()
 
-        if self.disconnected:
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
Great stuff, thanks!!!
I’ll need to take a detailed look at each.

uvicorn/protocols/http/h11.py
```diff
@@ -325,21 +325,13 @@ def __init__(self, scope, protocol):
...
         protocol.resume_reading()
 
-        if self.more_body and not self.body and not self.disconnected:
-            await protocol.client_event.wait()
-            protocol.client_event.clear()
+        await protocol.client_event.wait()
+        protocol.client_event.clear()
 
-        if self.disconnected:
+        if self.disconnected or self.response_complete:
```
lovelydinosaur, line 334: Yup, this make sense for sure.

lovelydinosaur:
Okay, I've followed this up with https://github.com/encode/uvicorn/pull/147

I'm not convinced that part (1) of this, so that'd need further explanation of the rationale for the change. Right now the behavior is "block if we don't yet have anything to send", which seems pretty straight-forward.

pgjones:
Block if we don't yet have anything to send, is what I was trying to achieve with (1), so thanks for #147.

PR #127, opened 2018-07-15
Updated subprotocols behaviour

JohnDoee:
The websocket subprotocol header handling was a wee bit messy.

Things changed:
* Stripping spaces from each subprotocol name
* Changed default to list (as per ASGI specifications)
* Changed encoding to ascii to align with Daphne behaviour (was bytes before)
* Allow multiple `Sec-WebSocket-Protocol` headers as per https://tools.ietf.org/html/rfc6455#section-11.3.4

Things I can't seem to fix
* Test miltiple `Sec-WebSocket-Protocol` headers, the clients used seem to suffer from the same problem uvicorn does of making everything into a dict

### DIFF
--- a/uvicorn/protocols/websockets/websockets.py
+++ b/uvicorn/protocols/websockets/websockets.py
@@ -24,10 +24,12 @@ def websocket_upgrade(http):
         return
 
     # Retrieve any subprotocols to be negotiated with the consumer later
-    subprotocols = request_headers.get(b"sec-websocket-protocol", [])
-    if subprotocols:
-        subprotocols = [subprotocol.strip(b" ").decode("ascii")
-                        for subprotocol in subprotocols.split(b",")]
+    subprotocols = []
+    for header_key, header_value in http.headers:
+        if header_key != b"sec-websocket-protocol":
+            continue
+        for subprotocol in header_value.split(b","):
+            subprotocols.append(subprotocol.decode("ascii").strip(" "))
     http.scope.update({"type": "websocket", "subprotocols": subprotocols})
     asgi_instance = http.app(http.scope)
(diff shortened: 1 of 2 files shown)


uvicorn/protocols/websockets/websockets.py
```diff
@@ -24,9 +24,12 @@ def set_header(key, val):
...
     # Retrieve any subprotocols to be negotiated with the consumer later
-    subprotocols = request_headers.get(b"sec-websocket-protocol", None)
-    if subprotocols:
-        subprotocols = subprotocols.split(b",")
+    subprotocols = []
+    for header_key, header_value in http.headers:
+        if header_key != b"sec-websocket-protocol":
+            continue
+        for subprotocol in header_value.split(b","):
+            subprotocols.append(subprotocol.decode("ascii").strip(" "))
```
lovelydinosaur: I guess just `.strip()` might make sense here. (Shouldn't be other whitespace characters, but we may as well drop 'em if there are, right?)

lovelydinosaur:
Good stuff, thanks!

Out of interest - What use-case motivated the change? / What subprotocol(s) are you using?

JohnDoee:
Autobahn WAMP clients require routers to have a working subprotocol system and I wrote a WAMP router for Django. First and only time I've seen it used.

My own ASGI implementation had problematic subprotocols handling and so did uvicorn :smile:

uvicorn/protocols/websockets/websockets.py
JohnDoee: Sure thing!

lovelydinosaur:
Thanks @JohnDoee!

Marginally related to your use-case: #128

PR #130, opened 2018-07-18
Improve PyPy and Windows support

lovelydinosaur:
* Don't attempt to log `signal.name` under windows.
* Use explicit `encoding=` when opening files in `setup.py`
* Requirements depend on runtime/platform - PyPy and Windows should no longer install `httptools` and `uvloop`.
* Defaults depend on runtime/platform - PyPy and Windows should no longer run with `httptools` and `uvloop` by default.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@
 import os
 import re
 import sys
+import platform
 
 from setuptools import setup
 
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
Yas!

PR #131, opened 2018-07-18
Revert "Updated subprotocols behavior"

lovelydinosaur:
Reverts encode/uvicorn#127

Builds have started failing since that PR.

I don't *think* that it's actually the issue, since it passed the tests, and it's only since the merge-to-master that it's started failing, but let's see if they run properly against a revert....


lovelydinosaur:
See #133 for the required fix.
(`websockets 6.0` had some API changes)

PR #137, opened 2018-07-19
Support single process mode

lovelydinosaur:


lovelydinosaur:
Single process is now the default, except with `--debug`.

Possible we might introduce `--workers` again, but I didn't like the style it had - let's keep any worker spawning out of the main code, and make sure to document `gunicorn`/`supervisor`/`circus` as the production options.

PR #140, opened 2018-07-20
Added --debug for in-browser tracebacks and autoreload.

lovelydinosaur:
When `--debug` is provided, two things happen:

* We switch on autoreload. (Occurs when any python files in or below the current working directory change.)
* We include `DebugMiddleware` (Displays tracebacks in the browser when exceptions occur within the ASGI app.)


cgarciae:
Hey @tomchristie , thanks for adding this feature. However, I am not getting the traceback, I only get `"server error"` with code `500`. When I switch back to the normal `app.serve` with `debug=True` on the standard `App` class, I see the traceback, but not with the ASyncApp + uvicorn.

Issue #95, opened 2018-06-29
Timeouts & resource limits


lovelydinosaur:
- [x] Task timeouts (default to 60s, for complete request/response plus any deferred processing.)
- [x] Keep-Alive timeouts (default to something like 2-4s.)
- [x] Max concurrency
- [x] Max tasks

Possibly also shutdown timeouts, but may need more thinking about.

PR #148, opened 2018-07-23
Add `--max-connections` setting

lovelydinosaur:
Adds a new `--max-connections` setting.

If the maximum number of allowable concurrent exceptions is exceeded then uvicorn will start responding with 503 responses, rather than invoking the ASGI application.

This helps you to bound the resources that each `uvicorn` process can use. You'll still want to ensure that you're adequately provisioning for expected usage, but you'll be able to ensure that severely overloaded servers are less likely to degrade the system.


lovelydinosaur:
Refs #95

PR #167, opened 2018-07-27
The application should share the same loop as uvicorn

leetreveil:
By deferring module load until after the loop has been configured allows `app.py` to use the same event loop as uvicorn.

For my purposes this allows me to initialize an `aioredis` pool inside of `app.py`.

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -190,14 +190,6 @@ def run(
     install_signal_handlers=True,
     ready_event=None,
 ):
-    try:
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Good call, yup! 😎

lovelydinosaur:
Released as 0.2.22

> For my purposes this allows me to initialize an aioredis pool inside of app.py.

If you're writing any ASGI middleware or any other tooling like that would love to see where you get to! :)

Issue #169, opened 2018-07-30
Trio support.


lovelydinosaur:
Revisiting some thoughts on this following on from #118

I'm not against having trio-support as one of the built-in options. Presumably that'd mean using `--loop trio`, and having that switch out:

* The loop implementation.
* The server class implementation.
* Using a new trio h11 based protocol implementation for the default `--http auto`.

The `asyncio`/`trio`/`curio` split *is* problematic, but it's an ecosystem issue, rather than an ASGI issue (ASGI is just an async/await interface, and is agnostic what event loop implementation is used to call into that interface.)

The problem is that whatever event loop implementation you're running in must *also* match any async primitives used in the application codebase.

If you're just using `async`/`await` and `send`/`receive`, then it doesn't matter, but as soon as you start using things like `asyncio.sleep()/trio.sleep()`, or creating concurrent tasks, or accessing network or file I/O directly, then you're locked into whichever system you're in.

I'd say that Trio is indisputably more refined and more properly constrained than `asyncio`, but the ecosystem split is problematic.

miracle2k:
For the version I did in #118, rather than implementing a new protocol, I instead built a bridge between the asyncio protocol interface and trio.

Specifically, the reason trio (and curio) do not have a protocol class, as I understand, is that the asyncio protocol implementation is not fully async/await (https://vorpus.org/blog/some-thoughts-on-asynchronous-api-design-in-a-post-asyncawait-world/#review-and-summing-up-what-is-async-await-native-anyway). That is, both `data_received` and `transport.send` are sync-functions, not async, and as such do not communicate backpressure. Rather, in asyncio, you need to use `pause_reading` and `pause_writing` to handle backpressure.

So, no "asyncio protocols" in trio. But in asyncio proper, a server code that is inside asyncio is just reading from the socket and calling the `Protocol.data_received` class, and taking the data from `transport.send` and writing it to the socket. So my solution was to reproduce this server logic in trio, such that the existing protocol classes in `uvicorn` can be used.

This means that we lose the purported advantage of trio's "async/await all the way down" in some way, but I don't think this matters here - it might be a better developer experience in general, but it's not necessary to implement a correctly functioning server.

I packaged up this bridge code in https://github.com/miracle2k/trio-protocol. 

[...]

PR #170, opened 2018-08-01
Trio support

miracle2k:
See #169.

Some notes:

- In addition to a gunicorn worker, the integration point on part of the CLI is a custom `Server` class. I made it so that a custom class is used when the loop is selected as trio, otherwise the default.

- The `TrioServer` class looks similar to the regular `Server`, but also has a bunch of differences; notable in the way signals are setup. We also call `trio_protocol.create_server` rather than `loop.create_server`. Now, it would probably be possible to cramp all of this inside `trio-protocol` in such a way that the original `Server` class can be used without changes, but I think that goes to far. The `Server` class to me seems like a nice integration point where some trio-specific code can be placed, but not too much, and nicely segregated; and there is some code duplication between `TrioServer` and `Server`, but not too much either.

- trio needs to create it's own Loop-like object, and it won't be available until the server starts, so for this reason the `create_protocol` helper now takes an argument to allow overriding the loop, while `loop_setup` for trio returns nothing.

The main thing that is missing here is websockets support. My stumbling block here was the dependency on the websockets library, which is used, but in a kind of strange way?

I'm not actually a 100% sure if `websockets` is a stumbling block here (due to the limited way in which it seems to be used, I don't know if it runs any asyncio code), but if it is, it would be a problem for supporting trio. I judge the chances of getting trio-support into websockets to be slim.

I happen to have been working on a trio-based websocket library as well (https://github.com/miracle2k/trio-websockets), but I don't think this is required. Rather, what about replacing the dependency on websockets with `wsproto`? It itself is based on `h11`, which we already use, and is sans-io, so should fit more naturally with our usage here.

If we don't switch to `wsproto`, I can look into getting the current `websockets`-based code to run using `trio-protocol`.


lovelydinosaur:
> what about replacing the dependency on websockets with wsproto?

Indeed, yup - I'd actually already started doing that, so I've pushed up the work in progress that I currently have: https://github.com/encode/uvicorn/pull/171

Ideally I'd like us to move to having two websocket protocol implementations. There's a big bunch of clean-up that needs to happen on the existing implementation, and having a `wsproto` implementation should help push that all in the right direction. I'd *guess* that I probably want `wsproto` to become the default - working with the sans-io library it's been *immediately* clear to me that it'll present a more obvious integration.

> The TrioServer class looks similar to the regular Server, but also has a bunch of differences; notable in the way signals are setup.

Great. You'll see that the server interface is a bit heavily overloaded at the moment. I think the right thing to do would be to introduce a `Config` class that stores basically everything passed from the command line, and pass that around, rather than the multitude of arguments we currently have on both the server and protocol classes.

lovelydinosaur:
@miracle2k The `wsproto` implementation is now incoming in #174.

lovelydinosaur:
Alrighty, `wsproto` is now available in the 0.3.0 release.

I've kept the default as `websockets` for the moment, just because I'd like to first see a newer `wsproto` release, with the latest `h11` as a dependency.

Once that happens I'd probably prefer to switch to `wsproto` as the default. (Perhaps also pending on if the package gets some docs love.)

miracle2k:
I intend to continue working on this soon. Is it enough to support trio with wsproto?

lovelydinosaur:
> Is it enough to support trio with wsproto?

Absolutely, yup. No reason for us to support the `websockets` implementation in that case.

miracle2k:
A challenge that I previously overlooked are uses of `asyncio.Event` or `asyncio.Queue`. Because the websocket protocol relies on them more heavily, I had to pay attention now!

I don't see any way around passing some kind of object to the protocol classes which allows access to `Event` and `Queue` helpers that are compatible with the selected loop. Maybe like this:

```python
class WSProtocol(asyncio.Protocol):
    def __init__(self, app, library, **kwargs):
        ....
       self.queue = library.create_queue()
       # or:
       self.queue = library.Queue()
       # instead of:
       # self.queue = asyncio.Queue()
       ....
```

lovelydinosaur:
@miracle2k Are we using a subset of the `Event` and `Queue` interface that's compatible with both `trio` and `asyncio`? (eg. I know the interfaces differ a little in some places, but I think we may only be using bits that are the same in either case)

miracle2k:
Yes, they should be compatible, with the exception that the trio Queue requires a size to to specified. trio does not allow an unbounded Queue.

miracle2k:
I've rebased this and implemented the kind of "library" concept discussed in the comment above, to allow the protocols to use `trio.Event` and `trio.Queue` when running on trio. I'm now gonna try to run it in a side project to see how it works, but... I think this is roughly how trio support would have to look like.

lovelydinosaur:
This seems pretty good! Great work on progressing it.

The most obvious thing we'd need next is to have the existing HTTP and WebSocket tests run against trio too.

I think we might want to change `aiolib` to `asyncio`, or else perhaps pass `event_class` and `queue_class`?

I'm not sure about including the trio worker class, since it's adding a large footprint of untested code. I guess my ideal here would be that we add multi-worker and process monitoring directly into uvicorn, drop the recommendation to use gunicorn, and that hopefully gunicorn gets it's own ASGI support. That we we can focus on being our own thing, and gunicorn can do it's thing.

dekoza:
Any progress on this? This (or using `anyio`) would be a great step to make uvicorn aio-agnostic.

florimondmanca:
Hey all,

Going to close this off for now as "obviously stale", but it's _great_ that we now have an initial proof-of-concept for Trio support in this repo, that we can use as reference in the future. :-) Thanks for your initial work on this, @miracle2k!

PR #174, opened 2018-08-12
WebSocket refactor

### LINKED ISSUE #100: Add `--ws` flag
Allow users to switch the websockets implementation. Initially should just be `--ws [websockets|none]`, and should disable websocket upgrades if turned off.

### LINKED ISSUE #123: Finesse "Connection: Upgrade" case.
* Return graceful response when "Upgrade" is an unsupported option.
* Check for comma separated cases as in "Connection: Upgrade, Keep-Alive"

Related to #121.

I suspect we probably just hang on h11 for this case right now.
Probably return a 403 on httptools.

Could be that simplest way to support thing is just push *all* requests with *any* upgrade header to `websocket_upgrade` and allow it to fail non-websocket cases. More graceful would be for both protocols to explicitly catch any non-websocket upgrades.

Also related: gracefully handle "CONNECT" cases. (Don't care if that means just closing the connection on them, but make sure we don't just pause and block)

lovelydinosaur:
Refactor the websocket implementation, so that:

* We're now subclassing `websockets.WebSocketServerProtocol`, rather than dealing with the handshake ourselves.
* The HTTP protocol passes the entire HTTP request data through to the WS protocol, rather than passing over the scope.
* Add an alternate `wsproto` implementation.
* Support `--ws [none|auto|websockets|wsproto]`

TODO:

- [x] Add reason text to websocket 400/403 responses.
- [x] Handle shutdown.
- [x] Flow control on wsproto.
- [x] Support `--ws none`.

Closes #100
Closes #123

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,11 @@
 click
 h11
+
+# Optional
 httptools
 uvloop
 websockets>=6.0
+wsproto
 
 # Testing
 pytest
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -1,5 +1,6 @@
 from uvicorn.protocols.http.h11_impl import H11Protocol
 from uvicorn.protocols.http.httptools_impl import HttpToolsProtocol
+from uvicorn.protocols.websockets.wsproto_impl import WSProtocol
 import asyncio
 import h11
 import pytest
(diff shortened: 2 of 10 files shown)


lovelydinosaur:
Still using `websockets` as the default for now.

Will consider switching to `wsproto` if it gets a new release and bumps to using the latest `h11` version. See https://github.com/python-hyper/wsproto/pull/53 and https://github.com/python-hyper/wsproto/pull/54

PR #181, opened 2018-08-24
giving logger a name

freakypie:
This will allow django to configure the output

i believe giving it a name will allow non-django users to configure it as well, but i'm only familiar with django

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -51,7 +51,7 @@ HANDLED_SIGNALS = (
 def get_logger(log_level):
     log_level = LOG_LEVELS[log_level]
     logging.basicConfig(format="%(levelname)s: %(message)s", level=log_level)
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
> This will allow django to configure the output

Could you give some more context / an example. Thanks!!! :)

(Just as a comment on here, for my own reference)

freakypie:
If the logger is given a name, users can configure uvicorn's output with django via the LOGGING setting https://docs.djangoproject.com/en/dev/topics/logging/#configuring-logging

A brief example 

    import os

    LOGGING = {
        'version': 1,
        'disable_existing_loggers': True,
        'handlers': {
            'console': {
                'class': 'logging.StreamHandler',
            },
        },
        'loggers': {
            'uvicorn': {
                'handlers': ['console'],
            },
        },
    }

PR #182, opened 2018-08-26
bug(Doc): Missing `'type'` key in `send()` dict

Gyllsdorff:
The example in the documentation sometimes missed the `type` key, but not the value, so the example code had syntax errors.


lovelydinosaur:
Fab, thanks!

PR #186, opened 2018-08-29
feat: Allow caller to provide `loop` instance

### LINKED ISSUE #183: `uvicorn.run` event loop behaviour when running programmatically.
I am trying to run Uvicorn in the same asyncio loop as some other components like the [asyncpg](https://github.com/MagicStack/asyncpg) DB connection pool.

As far as I can tell we can specify the callable that receives the scope and creates the request coroutine two ways:

- Passing a `str` containing the path to `uvicorn.run` but then we can't setup our dependencies (db, redis) since that function is called each new request. We could load the dependencies by creating them when the file is imported but there is no way that will pass codereview.

- Pass a callable to `uvicorn.run` with the dependencies in the scope. Unfortunately `uvicorn` [destroy the current event loop and replaces the event loop policy](https://github.com/encode/uvicorn/blob/master/uvicorn/loops/uvloop.py#L6) during the setup phase.

What is the recommended way to do things like this? If I could pass a loop instance in the `loop` parameter to `run` and have uvicorn use that directly without destroying it then it then I could use alternative 2 as long as we don't call `unicorn.run` from a async function [due to the `server.run` behaviour.](https://github.com/encode/uvicorn/blob/master/uvicorn/main.py#L337).

@tomchristie

Gyllsdorff:
Closes #183

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -221,12 +221,13 @@ def run(
         sock = socket.fromfd(fd, socket.AF_UNIX, socket.SOCK_STREAM)
 
     logger = get_logger(log_level)
-    loop_setup = import_from_string(LOOP_SETUPS[loop])
     http_protocol_class = import_from_string(HTTP_PROTOCOLS[http])
     ws_protocol_class = import_from_string(WS_PROTOCOLS[ws])
 
-    loop = loop_setup()
-
+    if isinstance(loop, str):
+        loop_setup = import_from_string(LOOP_SETUPS[loop])
+        loop = loop_setup()
+    
     try:
         app = import_from_string(app)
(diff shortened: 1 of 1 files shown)


Gyllsdorff:
@tomchristie 
Is this a good enough solution? If that's the case I can write some tests for it.

lovelydinosaur:
Certainly something like this, yup.

Would it be nicer to support `loop = None` supported instead, and have that call `asyncio.get_event_loop()`, rather than having the caller pass it in?

lovelydinosaur:
Obviously this isn't supported from the command line, but it'd be good to have it in the docs for the programmatic invocation case.

Issue #194, opened 2018-09-21
Uvicorn does not respect gunicorn's keepalive timeout


leetreveil:
```
gunicorn --worker-class uvicorn.workers.UvicornWorker myapp:App --keep-alive=75
```

If we sleep 10 seconds between requests we can see that uvicorn ignores gunicorn's keep-alive setting:

```
while :;do echo -e "GET / HTTP/1.1\nhost: 127.0.0.1\n\n";sleep 10;done|telnet 127.0.0.1 8000
```
```
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
HTTP/1.1 200 OK
server: uvicorn
date: Fri, 21 Sep 2018 12:15:40 GMT
content-type: text/plain
transfer-encoding: chunked

d
Hello, world!
0

Connection closed by foreign host.
```

PR #195, opened 2018-09-21
Respect Gunicorn's keep-alive timeout

leetreveil:
Fixes https://github.com/encode/uvicorn/issues/194


lovelydinosaur:
Good call!

PR #196, opened 2018-09-23
Include tag to disable favicon request in debug middleware

jordaneremieff:
Minor tweak to the debug middleware HTML response. 

I am including this middleware in https://github.com/erm/fikki/issues/3, and I was able to just drop it in for the most part, but in the case of HTML responses the favicon request prevents the entire stack trace from being rendered in my server implementation. 

It did not occur when running the app with `uvicorn`, so I imagine this is handled here somehow, but this change seems minor enough to make it on this end.

### DIFF
--- a/uvicorn/middleware/debug.py
+++ b/uvicorn/middleware/debug.py
@@ -85,7 +85,7 @@ class _DebugResponder:
             if "text/html" in accept:
                 exc_html = html.escape(traceback.format_exc())
                 content = (
-                    "<html><body><h1>500 Server Error</h1><pre>%s</pre></body></html>"
+                    "<html><head><link rel='icon' href='data:,'></head><body><h1>500 Server Error</h1><pre>%s</pre></body></html>"
                     % exc_html
                 )
                 response = HTMLResponse(content, status_code=500)


lovelydinosaur:
Let's look to prior art here: What does the default Django 500 template look like?

lovelydinosaur:
Similarly Werkzeug's debug middleware - does the template include a no-data favicon, or does it simply omit the tag?

jordaneremieff:
Werkzeug includes a dummy favicon here https://github.com/pallets/werkzeug/blob/master/werkzeug/debug/tbtools.py#L48 for a similar purpose.

Django _appears_ to handle this using some setting https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/docs/howto/error-reporting.txt#L97, but I didn't dig too deeply if this is how it actually uses it in the debug middleware.

Though at this point I've opted to swap in the Starlette debug middleware which did not have this issue - probably not necessary to include here at this point if there is no issue for Uvicorn specifically, but I'll leave this for you to close in case it's still worth considering.

lovelydinosaur:
Going to drop this for now. In any case I *think* we probably want to eventually just move to `--reload`, with debug middleware being an application-level concern (eg. Starlette's debug middleware)

PR #202, opened 2018-09-27
Fix sample nginx conf block

matteosimone:
After basing my nginx conf on the sample nginx conf in the docs, I encountered some problems:

1. Missing semicolon in the upstream block causes nginx to crash
2. $proxy_x_forwarded_port and $proxy_x_forwarded_proto are not valid variables. This might have been copied from a different nginx conf that used map to define these? I am not sure about this one.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -178,8 +178,8 @@ http {
     location / {
       proxy_set_header Host $http_host;
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-      proxy_set_header X-Forwarded-Port $proxy_x_forwarded_port;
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Great thanks!

> $proxy_x_forwarded_port and $proxy_x_forwarded_proto are not valid variables. This might have been copied from a different nginx conf that used map to define these?

Okay. `$scheme` looks good here. I'm going to modify the `ProxyHeadersMiddleware` to drop `X-Forwarded-Port` since (1. It's not sufficiently referenced as a de-facto header) and (2. I think it's used in a different manner to what I'd thought.)

PR #205, opened 2018-09-28
Send HTTP 400 response for invalid request

markbreedlove:
Given an invalid request, respond with an HTTP 400 error instead of closing the connection without a response.

Before this change, a request with invalid characters would result in the HTTP connection being dropped, as in this example:

```
$ curl -i -X GET 'http://localhost:8000/example?coordinates=76.916664, 8.483333'
```

Such a request should result in an HTTP 400 (Bad Request).

With a reverse-proxied application, this would typically be presented to the user as a 502 (Bad Gateway) error.

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -57,6 +57,12 @@ UPGRADE_REQUEST = b"\r\n".join([
     b""
 ])
 
+INVALID_REQUEST = b"\r\n".join([
+    b"GET /?x=y z HTTP/1.1",  # bad space character
+    b"Host: example.org",
+    b"",
+    b""
+])
 
 class MockTransport:
     def __init__(self, sockname=None, peername=None, sslcontext=False):
@@ -651,3 +657,14 @@ def test_supported_upgrade_request(protocol_cls):
     protocol.data_received(UPGRADE_REQUEST)
     assert b"HTTP/1.1 400 Bad Request" in protocol.transport.buffer
     assert b"Missing Sec-WebSocket-Version header" in protocol.transport.buffer
+
+
+@pytest.mark.parametrize("protocol_cls", [HttpToolsProtocol, H11Protocol])
+def test_invalid_http_request(protocol_cls):
+    def app(scope):
+        return Response("Hello, world", media_type="text/plain")
+
+    protocol = get_connected_protocol(app, protocol_cls)
+    protocol.data_received(INVALID_REQUEST)
+    assert b"HTTP/1.1 400 Bad Request" in protocol.transport.buffer
+    assert b"Invalid HTTP request received." in protocol.transport.buffer
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -179,7 +179,7 @@ class H11Protocol(asyncio.Protocol):
             except h11.RemoteProtocolError as exc:
                 msg = "Invalid HTTP request received."
                 self.logger.warning(msg)
-                self.transport.close()
+                self.send_400_response(msg)
                 return
             event_type = type(event)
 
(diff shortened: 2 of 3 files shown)


lovelydinosaur:
Great work, thanks!

Before we decide on behaviour here I’d like to have a better idea of existing implementation behaviours.

Any of the following would be really useful to know:

* How do gunicorn or node deal with that request, and does it make it to the application or not?
* How does nginx deal with a request like that?

markbreedlove:
Hi, Tom,

I've been running `uvicorn` with `gunicorn` as follows.
* In development: `gunicorn -k uvicorn.workers.UvicornWorker --log-level=debug --reload dplaapi:app`
* In production: `gunicorn -w 2 -b 0.0.0.0:8000 -k uvicorn.workers.UvicornWorker dplaapi:app`
... both exhibiting the same behavior from that `curl` request.

I'm running in AWS Elastic Beanstalk in production. The application is Dockerized and reverse-proxied with `nginx`.

Here's a sample line from `nginx`'s `access.log`:
```
172.30.0.141 - - [28/Sep/2018:12:21:00 +0000] "GET /v2/items?api_key=[redacted]&sourceResource.spatial.coordinates=76.916664, 8.483333 HTTP/1.1" 502 173 "-" "-"
```

Here's the corresponding `error.log` line:
```
2018/09/28 12:21:00 [error] 22285#0: *15872 upstream prematurely closed connection while reading response header from upstream, client: 172.30.0.141, server: , request: "GET /v2/items?api_key=[redacted]&sourceResource.spatial.coordinates=76.916664, 8.483333 HTTP/1.1", upstream: "http://172.17.0.2:8000/v2/items?api_key=[redacted]&sourceResource.spatial.coordinates=76.916664, 8.483333", host: "api.dp.la"
```

Here's what `gunicorn` was printing on stdout; having it bubble up from `uvicorn`:
```
[2018-09-28 12:21:00 +0000] [10] [WARNING] Invalid HTTP request received.
```

The request did not make it past `gunicorn` and did not get down to my application, so to speak. As far as I can tell, `gunicorn` (or `uvicorn`?) never got to the point of making a call on my Api Star `ASyncApp` object. I'm not sure if that's the best way to describe it -- just let me know if that isn't clear!

lovelydinosaur:
Sorry I wasn’t clear. I meant what response do we get running just regular gunicorn and a plain old hello world wsgi app. That’d be a really useful point of comparison. (Node’s behaviour would be interesting too)

markbreedlove:
Aha, of course, makes sense. Here's what I observe with gunicorn 19.9.0 and the "hello world" app at https://gunicorn.org.

```
(venv) uvicorn-pull205-test$ cat gunicornapp.py 
def app(environ, start_response):
    data = b"Hello, World!\n"
    start_response("200 OK", [
        ("Content-Type", "text/plain"),
        ("Content-Length", str(len(data)))
    ])
    return iter([data])
(venv) uvicorn-pull205-test$ 
(venv) uvicorn-pull205-test$ gunicorn -w 2 gunicornapp:app
[2018-10-02 17:11:52 -0400] [73491] [INFO] Starting gunicorn 19.9.0
[2018-10-02 17:11:52 -0400] [73491] [INFO] Listening at: http://127.0.0.1:8000 (73491)
[2018-10-02 17:11:52 -0400] [73491] [INFO] Using worker: sync
[2018-10-02 17:11:52 -0400] [73494] [INFO] Booting worker with pid: 73494
[2018-10-02 17:11:52 -0400] [73495] [INFO] Booting worker with pid: 73495
^C[2018-10-02 17:12:30 -0400] [73491] [INFO] Handling signal: int
```

... other terminal ...

```
~$ curl -i 'http://localhost:8000/?x=y z'
HTTP/1.1 400 Bad Request
Connection: close
Content-Type: text/html
Content-Length: 198

<html>
  <head>
    <title>Bad Request</title>
  </head>
  <body>
    <h1><p>Bad Request</p></h1>
    Invalid HTTP Version &#x27;Invalid HTTP Version: &#x27;z HTTP/1.1&#x27;&#x27;
  </body>
</html>
```

I'll check Node.js a bit later -- looks like tomorrow morning at this point ...

markbreedlove:
Actually, this is easy enough in Node and I can do it now rather than later.  I'm using this Hello World from https://nodejs.org/en/about/ ...
```
uvicorn-pull205-test$ cat nodeapp.js 
const http = require('http');

const hostname = '127.0.0.1';
const port = 3000;

const server = http.createServer((req, res) => {
  res.statusCode = 200;
  res.setHeader('Content-Type', 'text/plain');
  res.end('Hello World\n');
});

server.listen(port, hostname, () => {
  console.log(`Server running at http://${hostname}:${port}/`);
});
uvicorn-pull205-test$ 
uvicorn-pull205-test$ node nodeapp.js 
Server running at http://127.0.0.1:3000/
^C
```
... in another terminal ...
```
~$ curl -i 'http://localhost:3000/?x=y z'
```

So Node just closes the connection, whereas gunicorn responds with a 400.

I'm using Node v8.11.3.

lovelydinosaur:
Okay, I like this, yup. Looks like there's a test failure for us to dig into.

gvbgduh:
This could possibly be related to https://github.com/encode/uvicorn/issues/344

Kludex:
I'm interested on this. How Hypercorn deals with it: https://gitlab.com/pgjones/hypercorn/-/blob/main/src/hypercorn/protocol/h11.py#L147-153

Kludex:
@markbreedlove Sorry for taking so long. Would you mind rebasing it?

euri10:
closed by https://github.com/encode/uvicorn/pull/1352

Issue #201, opened 2018-09-27
ASGI callable returned without completing response & nginx redirect


matteosimone:
Hi, I am trying to figure out how to resolve this error and if I am doing something incorrectly. It is triggered when I use the [X-Accel-Redirect](https://www.nginx.com/resources/wiki/start/topics/examples/x-accel/) feature of nginx. This view is working fine as far as I can tell, but after each request I get that exception logged.

My view looks like this:
```
def view(request):
    filename = "some_file"
    path = "/path/to/image.png"
    response = HttpResponse()
    response["Content-Disposition"] = "attachment; filename={}".format(filename)
    response["X-Accel-Redirect"] = "/protected/{}".format(path)
    return response
```

Example log:
```
INFO: / - "GET /app/image/35/ HTTP/1.0" 200
ERROR: ASGI callable returned without completing response.
```

lovelydinosaur:
Perhaps try `HttpResponse('')` or `HttpResponse('', status_code=204)`?
I assume that you're returning a response that *looks* as if it has a response body but that then never includes one.

It's not impossible that there's a corner-case bug there for ASGI-over Django that's not been thrashed out there?

I guess you also ought to drop the `Content-Disposition` since I'd assume that nginx would send whatever headers correspond to the media that the internal redirect points at.

matteosimone:
Hmm, none of these methods worked. Even including fake content and dropping the content disposition did not help. I will have to continue investigating

lovelydinosaur:
Okay. I’m going to close this off for now, since it *looks* like uvicorn is behaving correctly in response to an application issue. What would be helpful tho would be if we had some proper debug-level logging that showed all the ASGI messages, so that this could be debugged more thoroughly.

matteosimone:
I have reproduced this with only Django, channels, and uvicorn installed without any custom code. I am still debugging. Do you have any pointers of where to add debug logs to show ASGI messages?

The only thing I am doing to trigger it is refreshing the admin/ page using standard http request. I have been able to reproduce in some situations in both Chrome and Safari, although they are not consistent.

lovelydinosaur:
Okay, given that you can reproduce this, and are looking into it, I’m going to reopen, with a view to adding sufficient logging.

There’s a number of different layers where the issue might be - I’d *guess* you’re seeing the issue because uvicorn is (perhaps) stricter than daphne in logging this error case.

[...]

PR #206, opened 2018-10-03
Set response_complete if connection is already disconnected

matteosimone:
Fixing #201


matteosimone:
Is it possible to get into a state where the client is disconnected and `self.response_started` is not set either? In the current code, this will also cause a false error log.

lovelydinosaur:
Good stuff, but we should deal with this the other way around... Don’t log an error condition for #201 if the connection is disconnected.

lovelydinosaur:
Closing in favour of #208

PR #209, opened 2018-10-04
Fix log error on uds

matteosimone:
Currently uvicorn 0.3.8 crashes while logging when run with a unix domain socket so it is unusable. Unix domain sockets return a peername of `""` and the sockname is also a string. [Docs](https://docs.python.org/3/library/socket.html#socket-families)

```
server_1         | ERROR: Exception in callback UVTransport._call_connection_lost
server_1         | handle: <Handle UVTransport._call_connection_lost>
server_1         | Traceback (most recent call last):
server_1         |   File "uvloop/cbhandles.pyx", line 72, in uvloop.loop.Handle._run
server_1         |   File "uvloop/handles/basetransport.pyx", line 181, in uvloop.loop.UVBaseTransport._call_connection_lost
server_1         |   File "/venv/lib/python3.7/site-packages/uvicorn/protocols/http/httptools_impl.py", line 163, in connection_lost
server_1         |     self.logger.debug("%s - Disconnected", self.connection_id)
server_1         |   File "/venv/lib/python3.7/site-packages/uvicorn/protocols/http/httptools_impl.py", line 140, in connection_id
server_1         |     return self.client[0]
server_1         | IndexError: string index out of range
```

### DIFF
--- a/uvicorn/middleware/message_logger.py
+++ b/uvicorn/middleware/message_logger.py
@@ -39,7 +39,7 @@ class MessageLoggerResponder:
         self.app = app
         self.logger = logger
         self.task_counter = task_counter
-        self.client_addr = scope['client'][0]
+        self.client_addr = scope['client']
 
     async def __call__(self, receive, send):
         self._receive = receive
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -118,6 +118,7 @@ class H11Protocol(asyncio.Protocol):
         self.server = None
         self.client = None
         self.scheme = None
+        self.connection_id = None
 
         # Per-request state
         self.scope = None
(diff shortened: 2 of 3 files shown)


lovelydinosaur:
I don’t really get it. “peername” should be the remote end of the connection - shouldn’t have anything to do with what it’s bound to?

matteosimone:
When I run it with uds in Docker attached to nginx, peername returns an empty string. I looked online and could only find [this from 2007](https://utcc.utoronto.ca/~cks/space/blog/python/UnixDomainSockets): `Unix domain sockets have no peername. Don't bother trying`.

Regardless, the python docs do mention that the socket family returns a string for sockname and peername, so the indexing by 0 needs to be worked around otherwise it can crash on empty string and in other cases it will be useless (returning the first character of the socket path for sockname)

matteosimone:
peername can also return `None` in the case of error

lovelydinosaur:
Okay, gotcha. Actually you’ll probably need to be using the proxy headers in that case, tho at the point of connection we don’t yet have those. I’d suggest we just log `self.client` whatever that is. (Or scope[‘client’] once we’re within an ASGI cycle, so that the proxy middleware can which sets that ends up being used.)

lovelydinosaur:
Looking good. I think the new MessageLoggingMiddeware needs the same treatment too, then I reckon it’s good to go.

lovelydinosaur:
Ace, thanks!

lovelydinosaur:
Released as 0.3.9 - thanks for your time on this!

Issue #135, opened 2018-07-19
Use environment markers in `setup.py` rather than explicitly switching requirements.


lovelydinosaur:
Recommended by @uranusjr, here: https://github.com/encode/uvicorn/issues/82#issuecomment-406148569

> I would strongly recommend using PEP 508 environment markers to specify dependencies instead. This allows for the same effect (in relatively recent Setuptools and pip), prevents the packager from needing to maintain platform-specific artifacts (e.g. wheels), and has immensely better compatibility with the new trend of locking-aware dependency management tools like Pipenv and Poetry.

> The following should be functionality-wise equivalent to the if-elif-else clause:

```python
requirements = [
    'click',
    'h11',
    'websockets>=6.0',
    "httptools; platform_system != 'Windows' and 'CYGWIN' not in platform_system and platform_python_implementation != 'PyPy'",
    "uvloop; platform_system != 'Windows' and 'CYGWIN' not in platform_system and platform_python_implementation != 'PyPy'",
]
```

lovelydinosaur:
This would be a nice simple one for a contributor to jump in on! 😄

lovelydinosaur:
Closing in favor of using optional dependencies #219

vlcinsky:
PR #309 resolves this issue, released in uvicorn==0.4.6

This will change in next major version, where use of extras is to be implemented.

PR #210, opened 2018-10-05
Use environment markers in `setup.py`

songww:
Use environment markers in `setup.py` rather than explicitly switching requirements.  #135

uvloop: https://github.com/MagicStack/uvloop/blob/8d5a9a3a4a5a74b1caad12423305b23049c00ab9/setup.py#L241

httptools: https://github.com/MagicStack/httptools/blob/14043bd7cb46fb879aaa1072dccf9a3817afaf32/setup.py#L22

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,26 +34,13 @@ def get_packages(package):
             if os.path.exists(os.path.join(dirpath, '__init__.py'))]
 
 
-if platform.python_implementation() == 'PyPy':
-    requirements = [
-        'click',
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Okay, I'm undecided on this just now.
It's possible that we'll make a different jump with 0.4, and instead force either `pip install uvicorn` for minimal dependencies, or `pip install uvicorn[full]` to include all optional dependencies.

PR #212, opened 2018-10-05
Add ssl support

songww:
#146


lovelydinosaur:
Great stuff! 😎

Any chance we could get this separated from the `Exception as exc` stuff?
What're you using as a guide for this? (eg. does this look the same as daphne, sanic, or something else?)

PR #215, opened 2018-10-08
Bring MessageLoggerMiddleware support to the gunicorn worker

leetreveil:


lovelydinosaur:
Good stuff, yup!
I've had to adapt this slightly, and pull it into #216

lovelydinosaur:
Available now in 0.3.10.

Thanks!

PR #218, opened 2018-10-09
Return `server` instance from run

Gyllsdorff:
In some cases a application needs to be able to shutdown the server on command from a outside signal or coordinate the shutdown process.

In this case I also need to shutdown Kafka consumer. By allowing the application to get the server instance it can register its own signal handlers and shutdown everything in the correct order.

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -285,6 +285,7 @@ def run(
         ready_event=ready_event,
     )
     server.run()
+    return server
 
 
 class Server:


Gyllsdorff:
This is just a suggestion, I can also handle it by adding a application callback the line after `uvicorn.run`.

songww:
Why not use `atexit` module that in stdlib.

lovelydinosaur:
The ASGI lifespan messages provide for this kind of behavior.

If you're writing a raw ASGI app, then this kind of thing:

```python
class App:
    def __call__(self, scope):
            return self.lifespan
        ... # Other application logic

    async def lifespan(self, receive: Receive, send: Send) -> None:
        message = await receive()
        await self.startup()

        message = await receive()
        await self.shutdown()

    async def startup(self):
        ...

    async def shutdown(self):
        ...
```

ASGI frameworks ought to give you an API to add startup/shutdown events.
The Starlette package also has some helpers here too.

PR #227, opened 2018-10-16
Rejig message logger so that ASGI apps are instantiated immediately

### LINKED ISSUE #222: Error integrating with Channels if 'lifespan' is not specified in router
I'm not entirely sure if I should be posting this here or on `channels`.

I'm using v0.3.12 which I believe has already introduced the new `lifespan` protocol defined in asgiref. But this causes an error with `channels`' router

```bash
   File "/usr/local/lib/python3.6/site-packages/uvicorn/lifespan.py", line 29, in run
     await self.asgi(self.receive, self.send)
   File "/usr/local/lib/python3.6/site-packages/uvicorn/middleware/message_logger.py", line 51, in __call__
     inner = self.app(self.scope)
   File "/usr/local/lib/python3.6/site-packages/channels/routing.py", line 58, in __call__
     raise ValueError("No application configured for scope type %r" % scope["type"])
ValueError: No application configured for scope type 'lifespan'
```

My `routing.py` file looks like this:

```python
application = ProtocolTypeRouter({
    # Empty for now (http->django views is added by default)
    'websocket': JWTWebsocketMiddleware(
        URLRouter(urlpatterns)
    )
})
```

**EDIT**: Sorry my workaround wasn't actually working as you'll need at least one `path` in the `URLRouter`, so I've removed it.

To temporarily get around this, I had to downgrade to `v0.3.9`.

lovelydinosaur:
Should resolve #222

### DIFF
--- a/tests/middleware/test_message_logger.py
+++ b/tests/middleware/test_message_logger.py
@@ -19,7 +19,8 @@ def test_message_logger(caplog):
     response = client.get("/")
     assert response.status_code == 200
     messages = [record.msg % record.args for record in caplog.records]
-    assert sum(['ASGI [1] Started' in message for message in messages]) == 1
+    assert sum(['ASGI [1] Initialized' in message for message in messages]) == 1
+    assert sum(['ASGI [1] Started task' in message for message in messages]) == 1
     assert sum(['ASGI [1] Sent' in message for message in messages]) == 1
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
This way around we'll get an error immediately if lifespan instantiation is not supported. That will gracefully disable it, and should resolve #222.

Issue #225, opened 2018-10-16
Log errors inside ASGI applications using exc_info, not as strings


daisylb:
There's a few places in uvicorn where it seems to be calling `traceback.format_exc()` on an exception produced by the ASGI application being run, and then log the resulting string using `logger.error()`. This is fine when logging to the console, but when logging to tools like Sentry, the parts of the stack trace that are actually important at the bottom get truncated.

Logging using `logger.error(message, exc_info=True)` instead would pass the actual exception through to Sentry and other similar tools, which means we'd not only get to see the full exception, but we'd also get some of the niceties that the tool provides, such as being able to see locals in each stack frame. The console output should be more or less the same.

The specific case we've been hitting is [this one, on `websockets_impl.py`](https://github.com/encode/uvicorn/blob/7977b95eae7a01fab2e390af120eb3dec47fd03e/uvicorn/protocols/websockets/websockets_impl.py#L143), but it looks like there's a couple of other usages of `.format_exc()` where the same pattern is being used too.

lovelydinosaur:
Yes good call. A pull request for this would be appreciated.

What'd also be good as part of that would be to make sure we're not using bare exceptions in those cases, but instead `except BaseException as exc` and pass the exc info to the logger explicitly, rather than just passing a boolean `True`.

This prevents Python implicitly creating a cyclical reference.

almarklein:
This should do the trick: `logger.error(..., exc_info=(type(exc ), exc , exc .__traceback__))`

jessebraham:
I was just working on this issue along with #191, and was looking for a bit of clarification on @almarklein's comment before I make a pull request.

To my understanding, Python 3.5+ allows passing the `exc_info` keyword argument either:
- a Tuple in the format you have provided (ie. what is returned by `sys.exc_info()`  
- the exception instance (ie. `exc`).

Is there a benefit I'm unaware of to providing the Tuple you demonstrated, or should I just provide the exception instance to keep things clean?

Thanks :)

almarklein:
@jessebraham Good catch! Apparently, you can also simply pass the exception object, which leads to much cleaner code. I just dived in Python's `logging` source and found:
```py
            if isinstance(exc_info, BaseException):
                exc_info = (type(exc_info), exc_info, exc_info.__traceback__)
```

lovelydinosaur:
Resolved in 0.3.19

PR #232, opened 2018-10-19
Use `except Exception as some_name` consistently and log errors using exc_info

### LINKED ISSUE #191: Use `except Exception as some_name` consistently.
All our exception handling should use the `except Exception as some_name` syntax, as the unnamed version creates a reference cycle and unnecessarily invokes garbage collection. When a name is used the reference cycle is scoped to the `except` block, and broken on exit.

jessebraham:
Closes #191 and #225.

As outlined in #191, all caught exceptions should be named in order to avoid a garbage collector invocation. This is now the case.

#225 suggests providing the `exc_info` keyword argument to logging function calls rather than converting the tracebacks to strings. The changes required to accomplish this have been made, and the extraneous imports removed.

If there are any changes or tweaks you would like made, just let me know.

### DIFF
--- a/uvicorn/middleware/debug.py
+++ b/uvicorn/middleware/debug.py
@@ -78,7 +78,7 @@ class _DebugResponder:
         try:
             asgi = self.app(self.scope)
             await asgi(receive, self.send)
-        except:
+        except BaseException as exc:
             if self.response_started:
                 raise
             accept = get_accept_header(self.scope)
--- a/uvicorn/middleware/debug.py
+++ b/uvicorn/middleware/debug.py
@@ -80,7 +80,7 @@ class _DebugResponder:
             await asgi(receive, self.send)
         except BaseException as exc:
             if self.response_started:
-                raise
+                raise exc from None
             accept = get_accept_header(self.scope)
             if "text/html" in accept:
                 exc_html = html.escape(traceback.format_exc())
@@ -93,7 +93,7 @@ class _DebugResponder:
                 content = traceback.format_exc()
                 response = PlainTextResponse(content, status_code=500)
             await response(receive, send)
-            raise
+            raise exc from None
 
     async def send(self, message):
         if message["type"] == "http.response.start":
--- a/uvicorn/middleware/message_logger.py
+++ b/uvicorn/middleware/message_logger.py
@@ -45,7 +45,7 @@ class MessageLoggerResponder:
         self.logger.debug(log_text, self.client_addr, self.task_counter, logged_scope)
         try:
             self.inner = app(scope)
-        except:
+        except BaseException as exc:
             log_text = '%s - ASGI [%d] Raised exception'
             self.logger.debug(log_text, self.client_addr, self.task_counter)
             raise
@@ -57,7 +57,7 @@ class MessageLoggerResponder:
         self.logger.debug(log_text, self.client_addr, self.task_counter)
         try:
             await self.inner(self.receive, self.send)
-        except:
+        except BaseException as exc:
             log_text = '%s - ASGI [%d] Raised exception'
             self.logger.debug(log_text, self.client_addr, self.task_counter)
             raise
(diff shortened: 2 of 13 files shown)


uvicorn/middleware/debug.py
```diff
@@ -78,7 +78,7 @@ def __init__(self, app, scope):
         try:
             asgi = self.app(self.scope)
             await asgi(receive, self.send)
-        except:
+        except BaseException as exc:
```
lovelydinosaur, line 81: I think there’s some adjustements needed in this block - eg `raise` May need to be `raise exc from None`. I’d double check how the behaviour looks in a browser with a simple app that raises an exception: possible our test cases are dealing with ensuring the correct behaviour fully. Similarly I *expect* `traceback.format_exc()` will now need to be passed the exception instance explicitly or something similar.

uvicorn/middleware/message_logger.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, scope, app, logger, task_counter):
         self.logger.debug(log_text, self.client_addr, self.task_counter, logged_scope)
         try:
             self.inner = app(scope)
-        except:
+        except BaseException as exc:
             log_text = '%s - ASGI [%d] Raised exception'
             self.logger.debug(log_text, self.client_addr, self.task_counter)
             raise
```
lovelydinosaur: Again, do we need an explicit `raise exc from None` here now?

uvicorn/middleware/message_logger.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, scope, app, logger, task_counter):
         self.logger.debug(log_text, self.client_addr, self.task_counter)
         try:
             await self.inner(self.receive, self.send)
-        except:
+        except BaseException as exc:
             log_text = '%s - ASGI [%d] Raised exception'
             self.logger.debug(log_text, self.client_addr, self.task_counter)
             raise
```
lovelydinosaur: Same question.

lovelydinosaur (commented):
Loving this PR! Looking good - some questions worth addressing, now that we’re using explicit exception info everywhere.

jessebraham:
Thanks for the quick response!

> I think there’s some adjustements needed in this block - eg raise May need to be raise exc from None.

I am not super knowledgeable when it comes to exception handling unfortunately, so I had to refer to the [the `raise` statement documentation](https://docs.python.org/3.5/reference/simple_stmts.html#the-raise-statement), and also found [this answer on StackOverflow](https://stackoverflow.com/a/24752607).

The docs state that a bare `raise` statement will re-raise the previously active exception in the current scope. If no active exception is present, a `RuntimeError` is raised instead. The `from` clause will attach the raised exception as the `__cause__` attribute, and if the raised exception is not handled both exceptions will be printed.

As you suggested I created a simple app that simply threw an exception. Using the `from` clause causes an additional pair of lines to be logged compared to the bare `raise` statement. Below is a diff of the two tracebacks.

```
  ERROR: Exception in ASGI application
  %s
    File "/home/jesse/Repos/uvicorn/uvicorn/protocols/http/httptools_impl.py", line 385, in run_asgi
      result = await asgi(self.receive, self.send)
+   File "/home/jesse/Repos/uvicorn/uvicorn/middleware/message_logger.py", line 63, in __call__
+     raise exc from None
    File "/home/jesse/Repos/uvicorn/uvicorn/middleware/message_logger.py", line 59, in __call__
      await self.inner(self.receive, self.send)
    File "./app.py", line 6, in __call__
      x = 1 / 0
  ZeroDivisionError: division by zero
```

I haven't noted any other changes in behaviour at this point, and all tests continue to pass. If this is the behaviour we're after, I can commit the changes whenever. If not, I guess we'll have to decide how to move forward.

> Similarly I expect traceback.format_exc() will now need to be passed the exception instance explicitly or something similar.

According to the [`traceback` documentation](https://docs.python.org/3.5/library/traceback.html#traceback.format_exc), `format_exc` doesn't seem to accept an exception as a parameter, so I think it's fine to leave it?

**Edit:** I also seem to have broken a few logging lines in my previous commit, leaving an extraneous `%s` trailing on some strings. Those will be fixed in the next commit as well.

lovelydinosaur:
Thanks!

PR #240, opened 2018-11-02
Decode percent escapes for path in connection scope

alexbotello:
Per ASGI spec, `scope["path"]` should be:
`Unicode string HTTP path from URL, with percent escapes decoded and UTF-8 byte sequences decoded into characters.`

`protocols/http/h11_impl.py` already follows this. I've gone ahead and added it to `httptools_impl.py` as well.

Closes encode/starlette#175


alexbotello:
All local tests are passing :man_shrugging:

lovelydinosaur:
Resolved in 0.3.18. Thanks!

Issue #260, opened 2018-12-30
Option to disable lifespan protocol


ferndot:
Since the lifespan protocol is new, some ASGI frameworks have not implemented it. It would be nice to have a configuration option to disable lifespan so that no errors are thrown.

simonw:
The example in the README doesn't work any more - if I try to run it I get this error:

```
$ uvicorn app:App --port 8005
INFO: Started server process [31803]
INFO: Waiting for application startup.
ERROR: Exception in 'lifespan' protocol
Traceback (most recent call last):
  File "/Users/simonw/.local/share/virtualenvs/datasette-oJRYYJuA/lib/python3.7/site-packages/uvicorn/lifespan.py", line 29, in run
    await self.asgi(self.receive, self.send)
  File "./app2.py", line 10, in __call__
    [b'content-type', b'text/plain'],
  File "/Users/simonw/.local/share/virtualenvs/datasette-oJRYYJuA/lib/python3.7/site-packages/uvicorn/lifespan.py", line 50, in send
    raise RuntimeError(error % message['type'])
RuntimeError: Got invalid message type on lifespan protocol "http.response.start"
ERROR: Application startup failed. Exiting.

```

elmcrest:
I could work around the issue the README has with `uvicorn==0.3.22`.

lovelydinosaur:
@simonw Thanks yup. I've updated that now.

(It should raise an error on anything except HTTP messsges, which will cause the lifespan protocol to be ignored.)

lovelydinosaur:
Agree that we should have an option for this. #262 gets most of the way there, except that it's important that we *default* to having it on.

lovelydinosaur:
Now available as `--disable-lifespan`

nashaad:
Update: currently, the flag is `--lifespan off`

PR #262, opened 2019-01-01
Include 'enable-lifespan' run option for lifespan support

jordaneremieff:
Refs https://github.com/encode/uvicorn/issues/260

Introduces an option to make lifespan support configurable. This allows ASGI applications that do not support the lifespan protocol to run with the defaults.

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -150,7 +150,7 @@ def get_logger(log_level):
     show_default=True,
 )
 @click.option(
-    "--enable-lifespan",
+    "--disable-lifespan",
     is_flag=True,
     default=False,
     help="Enable lifespan events such as startup and shutdown within an ASGI application.",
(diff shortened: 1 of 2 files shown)


jordaneremieff:
Also, this change doesn't consider the `UvicornWorker` classes - just the main server run case. If this is a reasonable approach to lifespan configuration, then I think allowing configuration options generally in the worker classes would be a separate issue.

uvicorn/main.py
```diff
@@ -149,6 +149,13 @@ def get_logger(log_level):
     help="Close Keep-Alive connections if no new data is received within this timeout.",
     show_default=True,
 )
+@click.option(
+    "--enable-lifespan",
+    is_flag=True,
+    default=False,
```
lovelydinosaur, line 155: We should default to using lifespan, rather than having it off by default.
I'm not sure how that would work with click's flags, or what we should use as the option for disabling support

jordaneremieff:
I've updated the PR to include a `--disable-lifespan` option instead of `--enable-lifespan` so that the default behaviour will be to support lifespan events. I've also included a docs entry for the setting.

lovelydinosaur:
Fab, thanks!

ferndot:
I split the worker config issue into #266

PR #274, opened 2019-01-20
[RFC] run: add use_reloader

blueyed:
This is useful when not using `main` (e.g. from an own `__main__` function),
but you want to have the reloading feature.

I could imagine to enable this just with `debug=True` for `run()` already, but that would require to move the body of the existing `run` to `run_inner` or something similar.

Of course, using `main()` in this case might also do, but it requires `args` to be provided for click, and I've thought that `click` should really be used from CLI only (and not your own `__main__`).

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -191,11 +191,16 @@ def main(
     }
 
     if debug:
-        logger = get_logger(log_level)
-        reloader = StatReload(logger)
-        reloader.run(run, kwargs)
+        run_with_reloader(**kwargs)
     else:
         run(**kwargs)
+
+
+def run_with_reloader(**kwargs):
+    log_level = kwargs.get("log_level", "info")
+    logger = get_logger(log_level)
+    reloader = StatReload(logger)
+    reloader.run(run, kwargs)
 
 
 def run(
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -291,7 +291,7 @@ def run(
 
     if use_reloader:
         reloader = StatReload(logger)
-        reloader.run(server.run, {})
+        reloader.run(server.run)
     else:
         server.run()
 
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -183,6 +183,7 @@ def main(
         "access_log": not no_access_log,
         "wsgi": wsgi,
         "debug": debug,
+        "use_reloader": debug,
         "proxy_headers": proxy_headers,
         "root_path": root_path,
         "limit_concurrency": limit_concurrency,
@@ -190,17 +191,7 @@ def main(
         "timeout_keep_alive": timeout_keep_alive,
     }
 
-    if debug:
-        run_with_reloader(**kwargs)
-    else:
-        run(**kwargs)
-
-
-def run_with_reloader(**kwargs):
-    log_level = kwargs.get("log_level", "info")
-    logger = get_logger(log_level)
-    reloader = StatReload(logger)
-    reloader.run(run, kwargs)
+    run(**kwargs)
 
 
 def run(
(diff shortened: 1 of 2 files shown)


blueyed:
This was a bit overzealous.. should also handle `args` and be added to `__all__`.
I'll wait for your feedback.. I think a `reload` kwargs for `run` makes the most sense, which would default to `debug` then.

mindey:
When will this be merged? I definitely need it T_T. And, looks good to use... But maybe --debug itself could have a parameter, which, if empty, would use default reloader, as in `main.py`:

```
    if debug:
        logger = get_logger(log_level)
        reloader = StatReload(logger)
        reloader.run(run, kwargs)
    else:
        run(**kwargs)
```

One way or another, it would be nice to choose reloader.

blueyed:
@mindey 
This is not about chosing another reloader, but just to control if it gets used / use it from `run` also.

I've noticed that it could probably also be improved in general, because it currently loops there all the time (with 0.2s sleeps), and inotify would be better suited - but that could be another reloader class then.  Given that it might make sense to have an argument that would allow to specify the one to use.

OTOH I could/should probably just wrap running uvicorn in something like inotify-watch then.

lovelydinosaur:
> I definitely need it 

Could you be specific about *what exactly* it is that you need? That'd help me with prioritizations here.

A configurable reloader class? A `--reload` argument, seperate to `--debug`? `reload` available in `run()`? Or something else?

mindey:
@tomchristie Yes, a configurable reload class. In our specific case we just needed a pre-reload hook to be able to run a custom `os.system` task before each reload while in debugging mode. :) However, a generally configurable reloader class would solve the problem.

>  `--reload` argument, seperate to `--debug`? reload available in `run()`? 

Both yes. Currently we use the command `uvicorn src.package:app --debug`, and would like to have the argument like the @blueyed  suggested `use_reloader`, to customize how it is reloaded under debug.

mindey:
> OTOH I could/should probably just wrap running uvicorn in something like inotify-watch then.

Thought of that, however, few systems have it `inotify-watch` installed by default... @blueyed

lovelydinosaur:
> In our specific case we just needed a pre-reload hook to be able to run a custom os.system task before each reload while in debugging mode.

You probably want to make use of the `lifespan` messages then.
You can fire events on startup, which should include reloading.

```python
class App:
    def __call__(self, scope):
            return self.lifespan
        ... # Other application logic

    async def lifespan(self, receive: Receive, send: Send) -> None:
        message = await receive()
        await self.startup()

        message = await receive()
        await self.shutdown()

    async def startup(self):
        ...

    async def shutdown(self):
        ...
```

blueyed:
@mindey 
Re `inotify-watch`: that's just a wrapper script I am using around `inotifywait` from inotify-tools, i.e. it was meant as an example.  I am also using this to update my terminal/tmux title etc, so it is certainly more custom then what should ship with uvicorn by default.

@tomchristie 
Given that I'd like to have an option to disable the reloader when running `uvicorn`, although `--debug` is used.  Does that make sense?

lovelydinosaur:
@blueyed Well, yes we ought to be separating out the debug and the reload functionality, really.

Note that the debug tracebacks are implemented as ASGI middleware,
so you actually don't have to use `--debug` at all, you can use `app = DebugMiddleware(app)`. (Either the middleware from this package, or the one in Starlette)

lovelydinosaur:
I actually somewhat think that it might make sense for Uvicorn to eventually drop `debug` completely, and *just* implement reload. Debug Tracebacks then become solely an application concern, and we issue a nice big pointer towards Starlette's middleware class.

It ends up being confusing when it's available in both places, and becomes less clear what "debug" does and doesn't do within uvicorn.

mindey:
@tomchristie the `lifespan` wouldn't work, because the system task has to pre-configure the app itself (not other application), before reload. What works for us, is modifying `run()` method in `reloaders.statreload.StatReload`.

lovelydinosaur:
We have `--reload` and `uvicorn.run(app_string, reload=[True|False])` now.

PR #275, opened 2019-01-20
reloader: handle (re-open) stdin in subprocess

blueyed:
This is required for `pdb.set_trace()` to work when using the reloader.

### DIFF
--- a/uvicorn/reloaders/statreload.py
+++ b/uvicorn/reloaders/statreload.py
@@ -20,6 +20,12 @@ class StatReload:
     def handle_exit(self, sig, frame):
         self.should_exit = True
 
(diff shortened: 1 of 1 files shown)


0x1ee7:
This is a great PR and it adds tremendous value to be able to set breakpoints without sacrificing reloader. When do you think you can merge this? @tomchristie
Thank you very much @blueyed 🌟

lovelydinosaur:
Sure thing - thanks for the bump @0x1EE7 - happy to get this in this once the merge conflicts are resolved. That can either happen against this PR directly, or someone else could file a PR incorperating this work, but without the merge conflicts that we have now.

PR #284, opened 2019-01-21
Log root path

### LINKED ISSUE #253: Logging should use root path + path
Currently access logging just uses "path" from the scope. It ought to instead use root_path + path.

lovelydinosaur:
Closes #253


lovelydinosaur:
Codecov should not be commenting here, right? https://github.com/encode/uvicorn/blob/master/.codecov.yml

PR #287, opened 2019-01-22
Bit of slimming down of attributes within Server

lovelydinosaur:


lovelydinosaur:
Clean some stuff up now that we have a `Config` class

PR #288, opened 2019-01-22
Missing imports, flake8

matthiask:
I'm not sure whether you want this or not, but I noticed a few fixups because of missing imports in the history -- things which might be detected automatically before merging.

I propose adding a build stage which checks the coding style using flake8. Because I didn't want to put in too much work without some feedback first I have added an ignore list for some flake8 errors. It might be better to let https://github.com/ambv/black fix the remaining warnings anyway instead of tackling them by hand, at least if you like the black coding style. (I wasn't a big fan of some aspects at first, but an automatic an opinionated source code formatter *can* be helpful.)

Anyway, no hard feelings at all if you reject this pull request. The first commit (https://github.com/encode/uvicorn/commit/ec8e24f0fa3d4f2a3923cf3abb3dc82e7017ce1e) should be uncontroversial though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,11 @@ script:
 
 after_script:
     - codecov
+
+jobs:
+    include:
+        - stage: style
+          install: skip
+          script:
+              - pip install flake8
+              - flake8 .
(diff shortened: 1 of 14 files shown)


lovelydinosaur:
Thanks! Tackling this in #290.

(black for formatting, autoflake for detecting erroneous imports, isort for import sorting)
Matches what we have in starlette.

matthiask:
Nice! :)

PR #213, opened 2018-10-05
Add ssl support.

songww:
#146

PR #294, opened 2019-02-02
Add ssl support

woile:
Finished work from songww in PR #213 

I tried to keep it faithful to songww code + added documentation and the url is properly displayed with https now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .cache
 .coverage
+.mypy_cache/
 __pycache__/
 uvicorn.egg-info/
 venv/
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -58,14 +58,14 @@ Options:
   --timeout-keep-alive INTEGER    Close Keep-Alive connections if no new data
                                   is received within this timeout.  [default:
                                   5]
-  --keyfile TEXT                  SSL key file
-  --certfile TEXT                 SSL certificate file
+  --ssl-keyfile TEXT              SSL key file
+  --ssl-certfile TEXT             SSL certificate file
   --ssl-version INTEGER           SSL version to use (see stdlib ssl module's)
                                   [default: 2]
-  --cert-reqs INTEGER             Whether client certificate is required (see
+  --ssl-cert-reqs INTEGER         Whether client certificate is required (see
                                   stdlib ssl module's)  [default: 0]
-  --ca-certs TEXT                 CA certificates file
-  --ciphers TEXT                  Ciphers to use (see stdlib ssl module's)
+  --ssl-ca-certs TEXT             CA certificates file
+  --ssl-ciphers TEXT              Ciphers to use (see stdlib ssl module's)
                                   [default: TLSv1]
   --help                          Show this message and exit.
 ```
@@ -223,8 +223,22 @@ Content Delivery Networks can also be a low-effort way to provide HTTPS terminat
 ## Running with HTTPS
 
 To run uvicorn with https, a certificate and a private key are required.
-The recommended way to get them is using [Let's Encrypt](https://letsencrypt.org/)
+The recommended way to get them is using [Let's Encrypt][letsencrypt].
+
+For local development with https, it's possible to use [mkcert][mkcert]
+to generate a valid certificat and private key.
 
 ```bash
-$ uvicorn app:App --port 5000 --keyfile=./key.pem --certfile=./cert.pem
+$ uvicorn app:App --port 5000 --ssl-keyfile=./key.pem --ssl-certfile=./cert.pem
 ```
+
+### Running gunicorn worker
+
+It also possible to use certificates with uvicorn's worker for gunicorn
+
+```bash
+$ gunicorn --keyfile=./key.pem --certfile=./cert.pem -k uvicorn.workers.UvicornWorker app:App
+```
+
+[letsencrypt]: https://letsencrypt.org/
+[mkcert]: https://github.com/FiloSottile/mkcert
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -5,3 +5,10 @@ uvicorn.egg-info/
 venv/
 htmlcov/
 site/
+
+# mypy
+.mypy_cache/
+
+# editors
+.vscode/
+.idea/
```
lovelydinosaur: Editor ignores should stay in folks user specific gitignores.
The mypy one can go below .coverage without needing its own comment.

lovelydinosaur (changes requested):
Looking ace!
Ideally, UvicornWorker.run should pass through the Gunicorn ssl config options too.

.gitignore
woile: Oh didn't know about that, nice to learn something new. I'll update the PR

woile:
Added ssl for gunicorn worker based on: https://github.com/benoitc/gunicorn/blob/33025cf610bb7a6f1cb307644c1881863c2fddc4/gunicorn/workers/gtornado.py#L137

uvicorn/workers.py
```diff
@@ -23,6 +24,12 @@ def run(self):
             "timeout_notify": self.timeout,
             "callback_notify": self.callback_notify,
         }
+
+        if self.cfg.is_ssl:
+            _ssl_opt = copy.deepcopy(self.cfg.ssl_options)
```
lovelydinosaur: I'd probably rather we explicitly copy across the options we need, rather than using `deepcopy`.

docs/deployment.md
```diff
@@ -56,6 +58,15 @@ Options:
...
                                   5]
+  --keyfile TEXT                  SSL key file
+  --certfile TEXT                 SSL certificate file
+  --ssl-version INTEGER           SSL version to use (see stdlib ssl module's)
+                                  [default: 2]
+  --cert-reqs INTEGER             Whether client certificate is required (see
+                                  stdlib ssl module's)  [default: 0]
+  --ca-certs TEXT                 CA certificates file
+  --ciphers TEXT                  Ciphers to use (see stdlib ssl module's)
+                                  [default: TLSv1]
```
lovelydinosaur, line 69: I'm thinking it might make sense for us to prefix all our command line options here as "ssl-...".

uvicorn/config.py
```diff
@@ -72,6 +84,12 @@ def __init__(
         timeout_notify=30,
         callback_notify=None,
         install_signal_handlers=True,
+        keyfile=None,
+        certfile=None,
+        ssl_version=ssl.PROTOCOL_TLS,
+        cert_reqs=ssl.CERT_NONE,
+        ca_certs=None,
+        ciphers="TLSv1",
```
lovelydinosaur: Again, I think it might be a good idea for us to prefix all SSL options with "ssl_"

lovelydinosaur (changes requested):
Looking good - few more minor comments.
(Also I'm happy to pick up the last bits of this, if that helps)

uvicorn/workers.py
woile: gonna fix this

docs/deployment.md
woile: The only downside about doing this, is that it's no longer "compatible" with gunicorn.
Right now this way `uvicorn` and `gunicorn` share the same interface. So if you want to use ssl for development, and then go to production with gunicorn, your settings don't change. I think it might make it a bit annoying for uvicorn users. What do you think?

woile:
@tomchristie a minor note, not related to this ticket per se, but maybe you have some ideas about it. 
I've noticied that I had to repeat the default values in the `config.Config` and in the click decorators. 
This kind of makes sense, because you might want to use uvicorn as a command line or as code. But if someone makes a mistake, there might be 2 different default settings. Is it okay? Should we care? How can I improve this?

uvicorn/workers.py
woile: I just made it a dictionary. Just for reference this is how `self.cfg.ssl_options` looks:
```python
{
    "keyfile": "./key.pem",
    "certfile": "./cert.pem",
    "ssl_version": 2,
    "cert_reqs": 0,
    "ca_certs": None,
    "suppress_ragged_eofs": True,
    "do_handshake_on_connect": False,
    "ciphers": "TLSv1",
}
```

docs/deployment.md
lovelydinosaur: That's a good point.
I *think* I'm okay with that - we already differ on `--bind` vs. `--host`/`--port`.

lovelydinosaur:
> I've noticed that I had to repeat the default values in the `config.Config` and in the click decorators.

I think we just have to live with that yeah.

woile:
Oki, I'll leave it up to you then the decision to change the arguments to have a `ssl` prefix.

woile:
@tomchristie I see it reads "Changes requested" but I don't know what else needs to be updated, it shows nothing, what am I missing?

lovelydinosaur:
Nah, that's just because I hadn't got around to okay'ing it since your last update.

I *would* like to see "ssl_" argument prefixes and "ssl-" command line prefixes, but I'm perfectly happy to pick that bit up myself.

woile:
Ok, feel free to pick it up, otherwise I'll try to update the ssl prefix later tonight or tomorrow

woile:
Okay done, renamed variables to `ssl` and added a test :tada:

lovelydinosaur (approved):

lovelydinosaur:
Amazing work here! I’ll plan to merge and release tomorrow.

lovelydinosaur:
Released to PyPI as 0.4.3

PR #299, opened 2019-02-11
WIP make Uvicorn work with wsproto 0.13

almarklein:
This is a work in progress for making things work with wsproto 0.13. For reference, here's [wsprotos changelog](https://github.com/python-hyper/wsproto/blob/master/CHANGELOG.rst#0130-2019-01-24).

I started out assuming that it would need just some renaming in a few places, but it turned out to involve a bit more work. Not sure how to best proceed with this, e.g. trying to make both 0.12 and 0.13 work from one class, create an additional class, or maybe drop 0.12 support?

The tests currently pass (on my machine, with wsproto 0.13) but someone with more experience with this part of the codebase should do a thorough review, or (probably better) pick this up from here.

References for moving from 0.12 to 0.13:
* https://gitlab.com/pgjones/hypercorn/commit/5b635eba3914445e242457bba93e41e940aec34b
* https://github.com/python-hyper/wsproto/blob/master/CHANGELOG.rst#0130-2019-01-24

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -695,5 +695,8 @@ def test_supported_upgrade_request(protocol_cls):
 
     protocol = get_connected_protocol(app, protocol_cls, ws="wsproto")
     protocol.data_received(UPGRADE_REQUEST)
-    assert b"HTTP/1.1 400 Bad Request" in protocol.transport.buffer
-    assert b"Missing Sec-WebSocket-Version header" in protocol.transport.buffer
+    if b"HTTP/1.1 426 " in protocol.transport.buffer:
+        pass  # wsproto 0.13
+    else:
+        assert b"HTTP/1.1 400 Bad Request" in protocol.transport.buffer
+        assert b"Missing Sec-WebSocket-Version header" in protocol.transport.buffer
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -80,12 +80,16 @@ def test_invalid_upgrade(protocol_cls):
         response = requests.get(
             url, headers={"upgrade": "websocket", "connection": "upgrade"}, timeout=5
         )
-        assert response.status_code == 400
-        assert response.text in [
-            "Missing Sec-WebSocket-Key header\n",
-            "Missing Sec-WebSocket-Version header",  # websockets
-            "Missing or empty Sec-WebSocket-Key header\n",  # wsproto
-        ]
+        if response.status_code == 426:
+            # response.text == ""
+            pass  # ok, wsproto 0.13
+        else:
+            assert response.status_code == 400
+            assert response.text in [
+                "Missing Sec-WebSocket-Key header\n",
+                "Missing Sec-WebSocket-Version header",  # websockets
+                "Missing or empty Sec-WebSocket-Key header\n",  # wsproto
+            ]
 
 
 @pytest.mark.parametrize("protocol_cls", WS_PROTOCOLS)
--- a/uvicorn/protocols/websockets/wsproto_impl.py
+++ b/uvicorn/protocols/websockets/wsproto_impl.py
@@ -35,11 +35,20 @@ class WSProtocol(asyncio.Protocol):
         self.queue = asyncio.Queue()
         self.handshake_complete = False
         self.close_sent = False
-
-        self.conn = wsproto.connection.WSConnection(
-            conn_type=wsproto.connection.SERVER,
-            extensions=[wsproto.extensions.PerMessageDeflate()],
-        )
+        
+        if hasattr(wsproto, "WSConnection"):
+            self.conn = wsproto.WSConnection(
+                connection_type=wsproto.connection.SERVER,
+                #extensions=[wsproto.extensions.PerMessageDeflate()],
+            )
+        else:  # wsproto < 0.13
+            # https://github.com/python-hyper/wsproto/blob/master/CHANGELOG.rst#0130-2019-01-24
+            # todo: someday, remove this (and bit below) and depend on wsproto >=0.13
+            self.data_received = data_received_old
+            self.conn = wsproto.connection.WSConnection(
+                conn_type=wsproto.connection.SERVER,
+                extensions=[wsproto.extensions.PerMessageDeflate()],
+            )
 
         self.read_paused = False
         self.writable = asyncio.Event()
@@ -63,9 +72,37 @@ class WSProtocol(asyncio.Protocol):
 
     def eof_received(self):
         pass
-
+    
     def data_received(self, data):
-        self.conn.receive_bytes(data)
+        try:
+            self.conn.receive_data(data)
+        except wsproto.utilities.RemoteProtocolError as err:
+            if err.event_hint is not None:
+                #self.handle_no_connect(wsproto.events.CloseConnection(code=err.event_hint.status_code, reason=str(err)))
+                output = self.conn.send(err.event_hint)
+                self.transport.write(output)
+                self.transport.close()
+            else:
+                self.handle_no_connect(wsproto.events.CloseConnection())
+            return
+        for event in self.conn.events():
+            if isinstance(event, wsproto.events.Request):
+                self.handle_connect(event)
+            elif isinstance(event, wsproto.events.TextMessage):
+                self.handle_text(event)
+            elif isinstance(event, wsproto.events.BytesMessage):
+                self.handle_bytes(event)
+            elif isinstance(event, wsproto.events.RejectConnection):
+                self.handle_no_connect(event)
+            elif isinstance(event, wsproto.events.RejectData):
+                self.handle_no_connect(event)
+            elif isinstance(event, wsproto.events.CloseConnection):
+                self.handle_close(event)
+            elif isinstance(event, wsproto.events.Ping):
+                self.handle_ping(event)
+    
+    def data_received_old(self, data):
+        func = self.conn.receive_bytes(data)
         for event in self.conn.events():
             if isinstance(event, wsproto.events.ConnectionRequested):
                 self.handle_connect(event)
@@ -94,8 +131,9 @@ class WSProtocol(asyncio.Protocol):
 
     def shutdown(self):
         self.queue.put_nowait({"type": "websocket.disconnect", "code": 1012})
-        self.conn.close(1012)
-        output = self.conn.bytes_to_send()
+        # self.conn.close(1012)
+        output = self.conn.send(wsproto.events.CloseConnection(code=code))
+        # output = self.conn.bytes_to_send()
         self.transport.write(output)
         self.transport.close()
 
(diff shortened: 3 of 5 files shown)


uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -63,9 +72,37 @@ def connection_lost(self, exc):
...
+            self.conn.receive_data(data)
+        except wsproto.utilities.RemoteProtocolError as err:
+            if err.event_hint is not None:
+                #self.handle_no_connect(wsproto.events.CloseConnection(code=err.event_hint.status_code, reason=str(err)))
+                output = self.conn.send(err.event_hint)
+                self.transport.write(output)
+                self.transport.close()
+            else:
+                self.handle_no_connect(wsproto.events.CloseConnection())
+            return
```
almarklein: This corresponds to "Switch from Fail events being returned to RemoteProtocolError`s being raised." in the changelog.

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -35,11 +35,20 @@ def __init__(self, config, server_state):
...
-
-        self.conn = wsproto.connection.WSConnection(
-            conn_type=wsproto.connection.SERVER,
-            extensions=[wsproto.extensions.PerMessageDeflate()],
-        )
+        
+        if hasattr(wsproto, "WSConnection"):
+            self.conn = wsproto.WSConnection(
+                connection_type=wsproto.connection.SERVER,
+                #extensions=[wsproto.extensions.PerMessageDeflate()],
```
almarklein: Not sure where the extensions should gol the new class does not have an argument/attribute for it.

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -94,8 +131,9 @@ def resume_writing(self):
 
     def shutdown(self):
         self.queue.put_nowait({"type": "websocket.disconnect", "code": 1012})
-        self.conn.close(1012)
-        output = self.conn.bytes_to_send()
+        # self.conn.close(1012)
+        output = self.conn.send(wsproto.events.CloseConnection(code=code))
+        # output = self.conn.bytes_to_send()
```
almarklein: All these bits correspond to: "Switch from a bytes_to_send method to the send method returning the bytes to send directly. "

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -106,16 +144,20 @@ def on_task_complete(self, task):
...
         self.connect_event = event
-        request = event.h11request
-        headers = [(key.lower(), value) for key, value in request.headers]
-        path, _, query_string = request.target.partition(b"?")
+        # print([x for x in dir(event) if not x.startswith('__')])
+        request = event#.h11request
+        # headers = [(key.lower(), value) for key, value in request.headers]
+        headers = [(key.lower(), value) for key, value in request.extra_headers]
+        #path, _, query_string = request.target.partition(b"?")
+        path, _, query_string = request.target.partition("?")
```
almarklein: `target` and `path` are now a string. I *think* `extra_headers` is what previously was `headers`.

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -159,25 +201,31 @@ def handle_bytes(self, event):
...
 
     def send_500_response(self):
         headers = [
             (b"content-type", b"text/plain; charset=utf-8"),
             (b"connection", b"close"),
         ]
-        msg = h11.Response(
+        if self.conn.connection is None:
+            output = self.conn.send(wsproto.events.RejectConnection(status_code=500))
+        else:
```
almarklein, line 195: Cannot send below stuff during the handshake. I suspect this `if self.conn.connection is None` check should be done in a few other places (maybe where `_upgrade_connection` was previously used?).

lovelydinosaur:
> maybe drop 0.12 support

Yup, that'd be the thing to do.

lovelydinosaur:
Thanks for taking a crack at this - let me know if you need extra eyes on it.

almarklein:
> > maybe drop 0.12 support
>
> Yup, that'd be the thing to do.

What about users who need wsproto 0.12 somwhere else in their code? Or can we assume that wsproto is an implementation detail of Uvicorn so unlikely to happen?

lovelydinosaur:
> What about users who need wsproto 0.12 somwhere else in their code? Or can we assume that wsproto is an implementation detail of Uvicorn so unlikely to happen?

We can't reasonably be expected to maintain cross-version compatability if there are API changes in our dependencies. If someone *needs* wsproto 0.12 within the same codebase as uvicorn, then they can also stick with uvicorn 0.4.3 or below.

uvicorn/protocols/websockets/wsproto_impl.py
almarklein: These are now passed in the `AcceptConnection` event.

uvicorn/protocols/websockets/wsproto_impl.py
almarklein: Confirmed.

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -131,11 +155,11 @@ def handle_no_connect(self, event):
             (b"connection", b"close"),
```
almarklein, line 149: It seems like this `handle_no_connect()` method is not covered by the test suite ...

almarklein:
Ok, I made some more updates, and also took a look at Hypercorn's implementation for reference. The tests pass, but please have a thorough look at this. I also left some `todo`'s which should probably be resolved before merging.

I'll leave it to you to either review and request changes, or to take this over from here.

tests/protocols/test_http.py
```diff
@@ -695,5 +695,8 @@ def app(scope):
 
     protocol = get_connected_protocol(app, protocol_cls, ws="wsproto")
     protocol.data_received(UPGRADE_REQUEST)
-    assert b"HTTP/1.1 400 Bad Request" in protocol.transport.buffer
-    assert b"Missing Sec-WebSocket-Version header" in protocol.transport.buffer
+    if b"HTTP/1.1 426 " in protocol.transport.buffer:
```
lovelydinosaur, line 698: What's this switch for?

tests/protocols/test_websocket.py
```diff
@@ -80,12 +80,16 @@ def test_invalid_upgrade(protocol_cls):
         response = requests.get(
             url, headers={"upgrade": "websocket", "connection": "upgrade"}, timeout=5
         )
-        assert response.status_code == 400
-        assert response.text in [
-            "Missing Sec-WebSocket-Key header\n",
-            "Missing Sec-WebSocket-Version header",  # websockets
-            "Missing or empty Sec-WebSocket-Key header\n",  # wsproto
-        ]
+        if response.status_code == 426:
```
lovelydinosaur, line 83: Same question

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -59,25 +66,42 @@ def connection_made(self, transport):
...
+        if exc is not None:
+            self.queue.put_nowait({"type": "websocket.disconnect"})
         self.connections.remove(self)
 
     def eof_received(self):
         pass
 
     def data_received(self, data):
-        self.conn.receive_bytes(data)
+        # Push data. Can raise RemoteProtocolError (previously resulted in Fail event)
```
lovelydinosaur: Okay, that's helpful to know for review, but we should drop the comment now.

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -106,19 +129,20 @@ def on_task_complete(self, task):
 
     def handle_connect(self, event):
         self.connect_event = event
-        request = event.h11request
-        headers = [(key.lower(), value) for key, value in request.headers]
-        path, _, query_string = request.target.partition(b"?")
+        headers = [(key.lower(), value) for key, value in event.extra_headers]
+        # todo: Hypercorn adds (b"host", event.host.encode()), should Uvicorn do too?
```
lovelydinosaur: Yes looks like it to me. Had a quick look at the `wsproto` docs, and extra_headers + host seems to make sense here.

lovelydinosaur (changes requested):

tests/protocols/test_http.py
almarklein: It's related to the `RemoteProtocolError ` and its `event_hint` value: it responds with 426 (Upgrade Required). Unfortunately, there is no body.

tests/protocols/test_websocket.py
almarklein: Same answer :)

tests/protocols/test_http.py
lovelydinosaur: > Unfortunately, there is no body.

Just to double check: Do you mean there's no status phrase? (Eg. `HTTP/1.1 426 Upgrade Required`) or there's no body?

almarklein: Either, I think.

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -59,25 +66,42 @@ def connection_made(self, transport):
...
-            elif isinstance(event, wsproto.events.ConnectionFailed):
+            elif isinstance(event, events.RejectConnection):
+                self.handle_no_connect(event)
+            elif isinstance(event, events.RejectData):
                 self.handle_no_connect(event)
-            elif isinstance(event, wsproto.events.ConnectionClosed):
+            elif isinstance(event, events.CloseConnection):
                 self.handle_close(event)
-            elif isinstance(event, wsproto.events.PingReceived):
+                break  # todo: right?
```
almarklein: @tomchristie what do you think?

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -150,6 +174,7 @@ def handle_text(self, event):
 
     def handle_bytes(self, event):
         self.bytes += event.data
+        # todo: question (AK): should we guard the size of this (hypercorn does)
```
almarklein: @tomchristie any comment on this? Maybe something to look at in another PR.

lovelydinosaur: Yes, but I'd suggest we treat that as a separate issue.

uvicorn/protocols/websockets/wsproto_impl.py
lovelydinosaur: Let's add that and drop the comment.

uvicorn/protocols/websockets/wsproto_impl.py
lovelydinosaur: We probably don't need it there. `wsproto` ought to take take of giving us a consistent stream of messages. I'd drop the `break` assuming the tests still pass.

lovelydinosaur:
Looks good - just needs the various points you commented on cleaning up now.

almarklein:
I addressed the last points and cleaned up a bit.

lovelydinosaur (approved):

lovelydinosaur:
Thanks so much for this!

Issue #265, opened 2019-01-10
Low level server


JoseKilo:
Is it possible to create a server without launching the asyncio loop at the same time ? I've been checking the code in main.py, `Server.run()` seems to be what I need more or less, but it calls `loop.run_forever()` at the end.

I'm more familiar with `aiohttp`, which supports something like this: https://aiohttp.readthedocs.io/en/stable/web_lowlevel.html#run-a-basic-low-level-server

Basically, I would like to execute other coroutines on the same loop. Using the `main()` function from that example, I'm able to do something like this:

```python
async def foo():
    for i in itertools.count():
        print(i)
        await asyncio.sleep(1)


if __name__ == "__main__":
    loop = asyncio.get_event_loop()

    tasks = [
        loop.create_task(main()),
        loop.create_task(foo()),
    ]

    try:
        loop.run_forever()
    except KeyboardInterrupt:
        for task in tasks:
            task.cancel()

    loop.close()
```

What would be the equivalent of that `main()` function using `uvicorn` (if any) ?
Also, if you think this is not a good idea / good practice for some reason, please let me know.

JoseKilo:
Using the latest changes I was able to put a working example together.

Handling the signals is a bit tricky. If you need to integrate the server with a bigger application you probably want to control the signals yourself.

Note that it doesn't support a second Ctrl-C to force exit like the original Server. And I haven't added windows signals just to make it more simple. But it shouldn't be difficult to control those. 

Let me know what you think.

```python
import asyncio
import itertools
import signal
import sys

import starlette.applications
import starlette.responses
import uvicorn.config
from uvicorn.main import Server as OriginalServer


def install_signal_handlers(loop):
    for sig in (signal.SIGINT, signal.SIGTERM):
        loop.add_signal_handler(sig, sys.exit, sig)


class Server(OriginalServer):

    def install_signal_handlers(self):
        pass

    async def run(self):
        config = self.config
        if not config.loaded:
            config.load()

        self.loop = config.loop_instance
        self.logger = config.logger_instance
        self.lifespan = config.lifespan_class(config)

        await self.startup()
        self._task = self.loop.create_task(self.main_loop())

    async def stop(self):
        self.should_exit = True
        await self._task
        await self.shutdown()


def make_server(loop):
    app = starlette.applications.Starlette()

    @app.route('/')
    async def homepage(request):
        return starlette.responses.JSONResponse({'hello': 'world'})

    config = uvicorn.config.Config(
        app,
[...]

PR #300, opened 2019-02-11
Allow low-level use of Server

JoseKilo:
Close: #265 

Basic idea is being able to do something like this:

```
config = Config(app=make_app())
server = Server(config)
loop = asyncio.get_event_loop()
loop.run_until_complete(server.run())
```

This will allow to run other tasks in the same asyncio loop.

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -83,7 +83,6 @@ class Config:
         timeout_keep_alive=5,
         timeout_notify=30,
         callback_notify=None,
-        install_signal_handlers=True,
         ssl_keyfile=None,
         ssl_certfile=None,
         ssl_version=ssl.PROTOCOL_TLS,
@@ -113,7 +112,6 @@ class Config:
         self.timeout_keep_alive = timeout_keep_alive
         self.timeout_notify = timeout_notify
         self.callback_notify = callback_notify
-        self.install_signal_handlers = install_signal_handlers
         self.ssl_keyfile = ssl_keyfile
         self.ssl_certfile = ssl_certfile
         self.ssl_version = ssl_version
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -113,6 +113,7 @@ class Config:
         self.timeout_keep_alive = timeout_keep_alive
         self.timeout_notify = timeout_notify
         self.callback_notify = callback_notify
+        self.install_signal_handlers = install_signal_handlers
         self.ssl_keyfile = ssl_keyfile
         self.ssl_certfile = ssl_certfile
         self.ssl_version = ssl_version
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -258,7 +258,6 @@ class Server:
 
     def main(self):
         self.config.setup_event_loop()
-
         loop = asyncio.get_event_loop()
         loop.run_until_complete(self.run())
         loop.stop()
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -259,8 +259,9 @@ class Server:
     def main(self):
         self.config.setup_event_loop()
 
-        self.loop.run_until_complete(self.run())
-        self.loop.stop()
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(self.run())
+        loop.stop()
 
     async def run(self):
         process_id = os.getpid()
@@ -280,10 +281,6 @@ class Server:
         await self.shutdown()
         self.logger.info("Finished server process [{}]".format(process_id))
 
-    @property
-    def loop(self):
-        return asyncio.get_event_loop()
-
     async def startup(self):
         config = self.config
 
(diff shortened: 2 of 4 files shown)


uvicorn/loops/uvloop.py
```diff
@@ -4,6 +4,5 @@
 
 
 def uvloop_setup():
-    asyncio.get_event_loop().close()
```
JoseKilo: Not sure if this is actually needed. If I keep the line, it complains with `RuntimeError: Cannot close a running event loop` (using uvloop only).

uvicorn/main.py
```diff
@@ -256,26 +256,37 @@ def __init__(self, config):
...
-        self.loop = config.loop_instance
+        if config.loop_instance is None:
+            config.setup_event_loop()
+
         self.logger = config.logger_instance
         self.lifespan = config.lifespan_class(config)
 
-        self.install_signal_handlers()
+        if config.install_signal_handlers:
+            self.install_signal_handlers()
```
JoseKilo: This check was missing, but the config property was there

lovelydinosaur:
Okay, so I think we might want to get from A to B step by step here, as I think there's some related changes to make along the way.

First step: Let's *just* deal with splitting out the `Config.setup_event_loop`

I think we probably want to drop `loop_instance` on the config, and instead use `asyncio.get_event_loop()` everywhere instead. The only place where we can't do that is on the protocol classes, because for the sake of testing we want to be able to inject mock loop implementations there.

So, I think step one here is probably:

* Split out `Config.setup_event_loop()`
* Drop the `loop_instance` on the config completely.
* Use `asyncio.get_event_loop()` everywhere.
* In the protocol classes add an argument `_loop = None` on init, and use `self.loop = _loop or asyncio.get_event_loop()` - the test cases can then explicitly set the loop on the protocol.

Once we've got that in, then I think we should tackle the rest of the `main()` vs. `async run()` refactoring.

Does that seem reasonable?

JoseKilo:
Yes, it makes sense. Let's start with the first one: #301

uvicorn/config.py
```diff
@@ -83,7 +83,6 @@ def __init__(
         timeout_keep_alive=5,
         timeout_notify=30,
         callback_notify=None,
-        install_signal_handlers=True,
```
JoseKilo, line 86: Not sure if we want to remove it from the arguments (wouldn't be fully backwards compatible)

JoseKilo:
I have rebased on top of master and left the `install_signal_handlers` change out.

uvicorn/config.py
lovelydinosaur: It's not documented API, so we're good.
(We were just using it internally)

uvicorn/main.py
```diff
@@ -401,12 +407,15 @@ def install_signal_handlers(self):
...
 
-    def handle_exit(self, sig, frame):
+    def stop(self):
         if self.should_exit:
             self.force_exit = True
         else:
             self.should_exit = True
 
+    def handle_exit(self, sig, frame):
+        self.stop()
```
lovelydinosaur: Let's not add this extra indirection.

uvicorn/main.py
```diff
@@ -256,28 +256,34 @@ def __init__(self, config):
...
-        self.loop.run_until_complete(self.shutdown())
-        self.loop.stop()
+        await self.startup()
+        await self.main_loop()
+        await self.shutdown()
         self.logger.info("Finished server process [{}]".format(process_id))
 
+    @property
+    def loop(self):
+        return asyncio.get_event_loop()
```
lovelydinosaur: I think we should probably drop this, and just use `loop = asyncio.get_event_loop()` inside any methods that need it.

uvicorn/main.py
```diff
@@ -256,28 +256,34 @@ def __init__(self, config):
         self.should_exit = False
         self.force_exit = False
 
-    def run(self):
+    def main(self):
+        self.config.setup_event_loop()
+
```
lovelydinosaur, line 264: This newline probably isn't meant to be here.

lovelydinosaur (changes requested):
Looking really good!

I've put some minor comments in place to address. Other thing is: are we happy we have the naming right here?

Right now we have:

* `def uvicorn.run()`
* `def Server.main()`
* `async def Server.run()`

Which is a bit asymmetrical. Perhaps we should swap `run` and `main` here??? Or something else?

lovelydinosaur:
Or perhaps:

`uvicorn.run()`
`server.run()`
`await server.serve()`

lovelydinosaur:
Not sure exactly, but I guess it probably makes sense for the blocking server method name to match the blocking function name.

JoseKilo:
Going with `uvicorn.run -> Server.run -> Server.serve`.

That way the previous Server API doesn't change, and the tests don't need to change.

lovelydinosaur (approved):

lovelydinosaur:
Lovely, lovely stuff here! 🎉

I'm going to tweak the `__init__` imports to make `Server` and `Config` available from the main module. And also roll a new release.

At which point we will have the following:

Standard entry point:

```python
import uvicorn

uvicorn.run(app, **kwargs)
```

Run, blocking the current thread.

```python
import uvicorn

config = uvicorn.Config(app, **kwargs)
server.run()
```

Run using the event loop explictly.

```python
import asyncio
import uvicorn

config = uvicorn.Config(app, **kwargs)
loop = asyncio.get_event_loop()
loop.run_until_complete(server.serve())
```

In due course we should start documenting up the programatic interface.

JoseKilo:
That's great ! Thank you for your help :)

PR #301, opened 2019-02-12
Small refactors to the async loop initialization

JoseKilo:
This is a previous step before tackling #300 / #265

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -121,7 +121,7 @@ class Config:
         self.ssl_ciphers = ssl_ciphers
 
         self.loaded = False
-        self.loop_instance = None
+        self.loop_setup = False
 
     def load(self):
         assert not self.loaded
@@ -174,8 +174,10 @@ class Config:
         self.loaded = True
 
     def setup_event_loop(self):
+        assert not self.loop_setup
+
         if isinstance(self.loop, str):
             loop_setup = import_from_string(LOOP_SETUPS[self.loop])
-            self.loop_instance = loop_setup()
-        else:
-            self.loop_instance = self.loop
+            loop_setup()
+
+        self.loop_setup = True
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -121,7 +121,6 @@ class Config:
         self.ssl_ciphers = ssl_ciphers
 
         self.loaded = False
-        self.loop_setup = False
 
     def load(self):
         assert not self.loaded
@@ -174,9 +173,5 @@ class Config:
         self.loaded = True
 
     def setup_event_loop(self):
-        assert not self.loop_setup
-
         loop_setup = import_from_string(LOOP_SETUPS[self.loop])
         loop_setup()
-
-        self.loop_setup = True
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -121,6 +121,7 @@ class Config:
         self.ssl_ciphers = ssl_ciphers
 
         self.loaded = False
+        self.loop_instance = None
 
     def load(self):
         assert not self.loaded
@@ -141,12 +142,6 @@ class Config:
             self.ws_protocol_class = self.ws
 
         self.lifespan_class = import_from_string(LIFESPAN[self.lifespan])
-
-        if isinstance(self.loop, str):
-            loop_setup = import_from_string(LOOP_SETUPS[self.loop])
-            self.loop_instance = loop_setup()
-        else:
-            self.loop_instance = self.loop
 
         try:
             self.loaded_app = import_from_string(self.app)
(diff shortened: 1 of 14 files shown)


uvicorn/config.py
```diff
@@ -177,3 +172,8 @@ def load(self):
             self.ssl = None
 
         self.loaded = True
+
+    def setup_event_loop(self):
+        if isinstance(self.loop, str):
```
lovelydinosaur: It should always be a string now, so I think we can drop this branching.

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -74,13 +74,16 @@ def __init__(self, scope):
 
 
 class HttpToolsProtocol(asyncio.Protocol):
-    def __init__(self, config, server_state):
+    def __init__(self, config, server_state, _loop=None):
         if not config.loaded:
             config.load()
 
+        if config.loop_instance is None:
+            config.setup_event_loop()
```
lovelydinosaur: I guess we can drop these.

uvicorn/config.py
JoseKilo: Removed. Some tests were still passing the loop directly.

uvicorn/protocols/http/httptools_impl.py
JoseKilo: Do you mean the `loop_instance` reference or the call itself ?

JoseKilo: I have removed the reference

JoseKilo: Oh sorry, you mean the call in the protocols. Yes, I think you are right: the loop will always be already initialized at this point.

uvicorn/lifespan/on.py
```diff
@@ -9,6 +9,9 @@ def __init__(self, config):
         if not config.loaded:
             config.load()
 
+        if not config.loop_setup:
+            config.setup_event_loop()
```
lovelydinosaur: I'm not sure if we need these anymore - may be that we can get away with *only* making the check in the server?

uvicorn/lifespan/auto.py
```diff
@@ -6,6 +6,9 @@ def LifespanAuto(config):
     if not config.loaded:
         config.load()
 
+    if not config.loop_setup:
+        config.setup_event_loop()
```
lovelydinosaur: I'm not sure if we need these anymore - may be that we can get away with _only_ making the check in the server?

uvicorn/loops/uvloop.py
```diff
@@ -4,6 +4,5 @@
 
 
 def uvloop_setup():
-    asyncio.get_event_loop().close()
     asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
     return asyncio.get_event_loop()
```
lovelydinosaur: I guess we don't actually need to return the loop from our setup functions anymore, since we don't use the return value now.

uvicorn/config.py
```diff
@@ -121,6 +121,7 @@ def __init__(
         self.ssl_ciphers = ssl_ciphers
 
         self.loaded = False
+        self.loop_setup = False
```
lovelydinosaur: Okay, last bit. Let's remove this. It's safe to recall `Config.setup_loop()` and in any case it's *only* called from the `Server` class and isn't expected to be called elsewhere. So... We can just drop this bit of state.

uvicorn/config.py
```diff
@@ -177,3 +172,11 @@ def load(self):
...
 
         self.loaded = True
+
+    def setup_event_loop(self):
+        assert not self.loop_setup
+
+        loop_setup = import_from_string(LOOP_SETUPS[self.loop])
+        loop_setup()
+
+        self.loop_setup = True
```
lovelydinosaur: We'll drop this here, too.

uvicorn/main.py
```diff
@@ -263,7 +263,10 @@ def run(self):
         if not config.loaded:
             config.load()
 
-        self.loop = config.loop_instance
+        if not config.loop_setup:
+            config.setup_event_loop()
```
lovelydinosaur: We won't need to branch here, we'll just call `config.setup_event_loop()` whatever.

lovelydinosaur (changes requested):
Great! One last little set of tweaks to remove one more bit of state, and then we're good to go on this.

lovelydinosaur:
Ace!

JoseKilo:
Thanks ! I will start rebasing the other branch

lovelydinosaur:
👍 👍 👍 

I'm not sure if we want `install_signal_handlers` in the config or not. It kinda being phased out, since it doesn't match up with everything else in the config that can also be exposed in command line arguments. It's *possible* that we'd prefer overriding `Server` if folks want that level of customization.

Anyways - not 100% on it either way, but just mentioning it, since you might want to omit that bit from the PR so we can consider it seperately from the rest of the `def main()`/`async def run()` API change.

PR #305, opened 2019-02-22
Fixes #304 Default empty str is used for root path

### LINKED ISSUE #304: Issue getting root path from scope for access log
Based on changes from pr #284 for issue #253  the access log now is trying to get the root path from the scope and it doesn't have to be present.

perdy:
That should be enough to fix #304.

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -436,7 +436,7 @@ class RequestResponseCycle:
                     '%s - "%s %s HTTP/%s" %d',
                     self.scope["client"],
                     self.scope["method"],
-                    self.scope["root_path"] + self.scope["path"],
+                    self.scope.get("root_path", "") + self.scope["path"],
                     self.scope["http_version"],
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Thanks! Would need to be applied to the h11 implementation too.

How are you hitting this? It’s not required per-ASGI, but it oughta to actually always be populated in this case, no? (Since we setup scope with it included)

perdy:
I just updated it to apply the change to h11 implementation.

I found that because I'm modifying the wrapper used in `add_route` as you can see [here](https://github.com/PeRDy/starlette-api/blob/12ce0e616fbd9c4c0a3f659dc0e83b7695ef14c3/starlette_api/routing.py#L124). I could, as well, modify that wrapper to make sure that `root_path` is included in scope (as an empty str by default) but independently I'd say that having a default value in uvicorn is safer.

lovelydinosaur:
Great stuff, thanks! Ping me again if I’ve not pushed a release in the next few days.

PR #309, opened 2019-02-24
Added env markers for package requirements

### LINKED ISSUE #318: Problem to install responder on Windows using `poetry`
Status about options to install responder on MS Windows:
- `pip`: works well
- `pipenv`: no idea
- `poetry`: it fails on attempt to install `uvloop`

This issue deals with `poetry` problem.

The problem is with `uvicorn` building requirements in code instead of using environment markers. This is currently being worked on [using optional packages](https://github.com/encode/uvicorn/issues/219).

After that issue is resolved and relevant `uvicorn` package released, `responder` will probably have to update it's own environment markers to address new way of working with optional packages as installed by `uvicorn`.

Until then, we cannot do much about it.

vlcinsky:
Shall be temporary fix for problems with installing uvicorn on Windows without major release bump as probably required by issue #219.

Shall fix https://github.com/kennethreitz/responder/issues/318

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,26 +34,19 @@ def get_packages(package):
             if os.path.exists(os.path.join(dirpath, '__init__.py'))]
 
 
-if platform.python_implementation() == 'PyPy':
-    requirements = [
-        'click',
-        'h11',
-        'websockets>=6.0'
-    ]
-elif platform.system() == 'Windows' or platform.system().startswith('CYGWIN'):
-    requirements = [
-        'click',
-        'h11',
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Yup looks good to me.
Progressing #298 will need to wait til we're ready for a major version bump.

lovelydinosaur:
Yup looks good to me.
Progressing #298 will need to wait til we're ready for a major version bump.

lovelydinosaur:
Released as 0.4.6

PR #317, opened 2019-03-05
Version 0.5.1

lovelydinosaur:
* Auto-reloading only used when application is passed as an import string.
* Use default multiprocessing context for multi-workers.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -88,23 +88,18 @@ class App:
     ...
 
 if __name__ == "__main__":
-    uvicorn.run(App, host="127.0.0.1", port=5000, log_level="info", debug=True)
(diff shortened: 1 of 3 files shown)


florimondmanca (approved):
Sounds like a good solution if the pickling issue cannot be resolved. Just to be sure — this means we'll have to pass the app via an import string in `uvicorn.run()`?

lovelydinosaur:
Typically you'll want to just pass the app instance as normal.

It's only the case  that you'd need to pass it as an import string if you want programmatic `reload=True` support to work.

(It's never worked in the past - StatReload would reload, but you wouldn't actually have been seeing any code changes take effect.)

PR #318, opened 2019-03-05
Fix host header with wsproto

### LINKED ISSUE #316: Headers structure using wsproto
The headers structure gets an inconsistent `Host` header when using wsproto.

Using starlette, this causes a ValueError iterating the headers:

``` 
ValueError: too many values to unpack (expected 2)
```

For example:

```python


class App:
    def __init__(self, scope):
        assert scope['type'] == 'websocket'
        self.scope = scope

        print(self.scope["headers"])
        websocket = WebSocket(self.scope, receive=receive, send=send)
        print(websocket.headers)

        await websocket.accept()
        await websocket.send_text('Hello, world!')
        await websocket.close()
```

First print shows:

```
[b'host', b'127.0.0.1:8000', (b'upgrade', b'websocket'), (b'connection', b'Upgrade'), (b'sec-websocket-key', b'Bp5kfPi9iQnS9DGRoyfnFg=='), (b'sec-websocket-version', b'13'), (b'user-agent', b'Python/3.6 websock
ets/7.0')]
```

Second one triggers the `ValueError` because the `__repr__` implementation iterates the headers.

The same example works as expected using `websockets` protocol.

```
[(b'host', b'127.0.0.1:8000'), (b'upgrade', b'websocket'), (b'connection', b'Upgrade'), (b'sec-websocket-key', b'l8qeOR/VJx5BgPc+S0yhfQ=='), (b'sec-websocket-version', b'13'), (b'sec-websocket-extensions', b'pe
rmessage-deflate; client_max_window_bits'), (b'user-agent', b'Python/3.6 websockets/7.0')]
```

I think it's related to this line:
https://github.com/encode/uvicorn/blob/546a60496d886f78d9f4dc60bd0f4aa15484f706/uvicorn/protocols/websockets/wsproto_impl.py#L125

lovelydinosaur:
Closes #316

### DIFF
--- a/uvicorn/__init__.py
+++ b/uvicorn/__init__.py
@@ -1,5 +1,5 @@
 from uvicorn.config import Config
 from uvicorn.main import Server, main, run
 
-__version__ = "0.5.1"
+__version__ = "0.5.2"
 __all__ = ["main", "run", "Config", "Server"]
--- a/uvicorn/protocols/websockets/wsproto_impl.py
+++ b/uvicorn/protocols/websockets/wsproto_impl.py
@@ -122,7 +122,7 @@ class WSProtocol(asyncio.Protocol):
 
     def handle_connect(self, event):
         self.connect_event = event
-        headers = [b"host", event.host.encode()]
+        headers = [(b"host", event.host.encode())]
         headers += [(key.lower(), value) for key, value in event.extra_headers]
         path, _, query_string = event.target.partition("?")
         self.scope = {


lovelydinosaur:
Superseeded by #319

thomasrockhu:
Hi all, Tom from `Codecov` here.  I'm extremely grateful that you use Codecov, and I want to make sure that you are getting as much benefit as possible.

We recently received a message from GitHub that we were blocked due to content.  I want to understand your poor experience since we want to make sure we do everything in our power to fix this.

Would you mind sharing what went wrong and what steps we can take to prevent this from happening again?

lovelydinosaur:
Heya @thomasrockhu sure thing. Codecov is fab, but it does seem to keep commenting, even tho (as far as I can tell) I've disabled that. https://github.com/encode/uvicorn/blob/master/.codecov.yml#L11

It's perfectly possible that I've misunderstood the configuration, but I've not figured out what I'm doing wrong if so. Blocking @codecov-io seemed the only way to stop it.

If it's an issue my end do let me know and I'll sort out the config accordingly. 👍

Issue #243, opened 2018-11-09
Having a Hard Time Replacing Default Server Header


Jitsusama:
Hello;

I'm using uvicorn to front-end an ASGI application of mine and I'm loving it! There is only one slight marring of the beautiful steed which is uvicorn; I have been unable to find a means whereby I can over-ride the sending of the `Server: uvicorn` header on all outbound responses.

When using HTTP1.1 as the transport protocol, I see in the _uvicorn.protocols.http.h11_impl_ module a `_get_default_headers` function. I can see that this in turn gets called by `H11Protocol#tick`, which overrides whatever value that the global `DEFAULT_HEADERS` variable holds, which in turn is what gets augmented to the headers of every outbound request in the `RequestResponseCycle#send` method.

Is there a strong reason for this? If not, is there any desire with the existing devs to enable this to be disabled/changed? Or, is there a desired approach that someone else should take if they would wish to introduce a PR that would change this behaviour?

lovelydinosaur:
I think we should support a `-h/--header` interface to allow developers to explicitly set particular headers (and to override the default "Server" one)

Should all be much cleaner to do since https://github.com/encode/uvicorn/pull/302

Jitsusama:
@tomchristie; that sounds wonderful! Would you also supply an API method/parameter to do this?

lovelydinosaur:
Yup - it'd be available as part of the `Config`

antirais:
In the meantime, this [gist](https://gist.github.com/antirais/17f5d5c1b7b5ee7a1150bbea1dd9a802) will demonstrate how to monkey patch response headers.

Jitsusama:
Hey @antirais; I had thought about over-riding that method, but it makes me feel uncomfortable, given how intrinsic it is to the inner-workings of uvicorn. I feel like messing with that will require constant vigilance as the upstream changes.

antirais:
> I feel like messing with that will require constant vigilance as the upstream changes.

That is true, it's not perfect. Hopefully this issue will be resolved quickly and there is no need to maintain the monkey patch for too long.

lovelydinosaur:
If anyone wants to take a crack at a pull request adding `headers` to the config, and command line options that'd be most welcome!

Jitsusama:
Hey @tomchristie; I'm taking a look now. Should tests for this go in `tests/protocols/test_http.py`? I can't find any other place that really makes sense.

Jitsusama:
[...]

PR #321, opened 2019-03-09
Allow for Modification of Default HTTP Response Headers

Jitsusama:
As per issue #243, this PR allows a user to programmatically, or, through the passing of CLI arguments, modify the default HTTP response headers that uvicorn sends out. In the case of the `Server` header, user specified values will over-ride the default value of _uvicorn_. If multiple `Server` header values are passed, then they will all be kept and will over-ride the default value as well.

The progammatic method of specifying default server headers is to pass a list of tuples to the `Config` class instantiation under the kwarg `headers` like so:

```python
config = Config(
    app=App, headers=[('Server', 'Cool Name'), ('X-Candy', 'Is Yummy')])
```

This also introduces a new CLI argument named `--header`. This argument takes a single argument, which is a header name/value pair delineated by a `:`. IE: `uvicorn --header Server:Coolness`. This argument can be passed multiple times to supply multiple over-riding headers.

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -124,8 +124,8 @@ class Config:
         self.ssl_cert_reqs = ssl_cert_reqs
         self.ssl_ca_certs = ssl_ca_certs
         self.ssl_ciphers = ssl_ciphers
-        self._headers = headers if headers else []
-        self.headers = None  # type: List[Tuple[bytes, bytes]]
+        self.headers = headers if headers else []  # type: List[str]
+        self.encoded_headers = None  # type: List[Tuple[bytes, bytes]]
 
         if self.ssl_keyfile or self.ssl_certfile:
             self.ssl = create_ssl_context(
@@ -151,9 +151,9 @@ class Config:
 
         encoded_headers = [
             (key.lower().encode("latin1"), value.encode("latin1"))
-            for key, value in self._headers
+            for key, value in self.headers
         ]
-        self.headers = (
+        self.encoded_headers = (
             encoded_headers
             if b"server" in dict(encoded_headers)
             else [(b"server", b"uvicorn")] + encoded_headers
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -2,6 +2,7 @@ import logging
 import socket
 import ssl
 import sys
+from typing import List, Tuple
 
 from uvicorn.importer import ImportFromStringError, import_from_string
 from uvicorn.middleware.debug import DebugMiddleware
@@ -92,7 +93,7 @@ class Config:
         ssl_cert_reqs=ssl.CERT_NONE,
         ssl_ca_certs=None,
         ssl_ciphers="TLSv1",
-        custom_headers=None,
+        headers=None,
     ):
         self.app = app
         self.host = host
@@ -123,6 +124,8 @@ class Config:
         self.ssl_cert_reqs = ssl_cert_reqs
         self.ssl_ca_certs = ssl_ca_certs
         self.ssl_ciphers = ssl_ciphers
+        self._headers = headers if headers else []
+        self.headers = None  # type: List[Tuple[bytes, bytes]]
 
         if self.ssl_keyfile or self.ssl_certfile:
             self.ssl = create_ssl_context(
@@ -141,15 +144,20 @@ class Config:
         else:
             self.reload_dirs = reload_dirs
 
-        if custom_headers is None:
-            self.custom_headers = []
-        else:
-            self.custom_headers = custom_headers
-
         self.loaded = False
 
     def load(self):
         assert not self.loaded
+
+        encoded_headers = [
+            (key.lower().encode("latin1"), value.encode("latin1"))
+            for key, value in self._headers
+        ]
+        self.headers = (
+            encoded_headers
+            if b"server" in dict(encoded_headers)
+            else [(b"server", b"uvicorn")] + encoded_headers
+        )  # type: List[Tuple[bytes, bytes]]
 
         if isinstance(self.http, str):
             self.http_protocol_class = import_from_string(HTTP_PROTOCOLS[self.http])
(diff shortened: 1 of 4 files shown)


Jitsusama:
FYI; it looks like the failure is related to black. I’m running Python 3.7, and as per your test script, there is apparently an issue with it and Python 3.7 so I didn’t try to run it. If there are any formatting changes that you would like for me to make, just let me know.

davidism:
Is there a way to unset the default `Server` header so it's not sent at all? It's one way to fingerprint a server, which can be considered bad.

Jitsusama:
@davidism; not with these changes. I guess for that functionality, you’d probably want middleware to remove it at the last possible second in case the application added one?

If you’re more worried about just removing the default one, I would think maybe adding another config kwarg to specify just removing the default value of Server (however it is specified) could be an option. I purposefully didn’t make the new kwarg over-write all truly default values, as the Date header is quite useful and can’t really be passed as a kwarg due to its dynamic nature. I could change the functionality so that if the kwarg is passed, everything but Date is overridden.

@tomchristie, what do you think?

lovelydinosaur:
> Is there a way to unset the default Server header so it's not sent at all? It's one way to fingerprint a server, which can be considered bad.

We could consider removing it by default, or providing a way to remove it, tho let's keep that seperate from this issue at the moment. Welcome to open an issue for it and we can follow up in due course.

lovelydinosaur:
Need to run `scripts/lint` to resolve the linting errors: https://travis-ci.org/encode/uvicorn/jobs/503924223

uvicorn/main.py
```diff
@@ -182,6 +182,12 @@
     help="Ciphers to use (see stdlib ssl module's)",
     show_default=True,
 )
+@click.option(
+    "--default-header",
```
lovelydinosaur: Just `--header` would be fine.

uvicorn/main.py
```diff
@@ -210,6 +216,7 @@ def main(
     ssl_cert_reqs: int,
     ssl_ca_certs: str,
     ssl_ciphers: str,
+    custom_headers: typing.List[str],
```
lovelydinosaur: Just `headers`

uvicorn/config.py
```diff
@@ -92,6 +92,7 @@ def __init__(
         ssl_cert_reqs=ssl.CERT_NONE,
         ssl_ca_certs=None,
         ssl_ciphers="TLSv1",
+        custom_headers=None,
```
lovelydinosaur: Again, let's just go with `headers` here.

uvicorn/config.py
```diff
@@ -140,6 +141,11 @@ def __init__(
         else:
             self.reload_dirs = reload_dirs
 
+        if custom_headers is None:
+            self.custom_headers = []
+        else:
+            self.custom_headers = custom_headers
```
lovelydinosaur: I'd suggest that in the config, we store the headers as lowercased byte-pairs. It's more efficient for us to do that conversion here, just the once.

I guess we want something like this?...

```python
if headers is None:
    headers = []
self.headers = [
    (key.lower().encode('latin-1'), value.encode('latin-1'))
    for key, value in headers
]
if b'server' not in dict(self.headers):
    self.headers = [(b'server', b'uvicorn')] + self.headers
```

lovelydinosaur: Don't worry about `date` - it doesn't make sense for users to specify that, and if they want to do so. There's plenty of header values that it'd be erronous for users to setup here - if they're setting custom headers it's up to them to do so sensibly.

uvicorn/main.py
```diff
@@ -372,12 +381,7 @@ def run(self, sockets=None, shutdown_servers=True):
     async def on_tick(self, counter) -> bool:
         # Update the default headers, once per second.
         if counter % 10 == 0:
-            current_time = time.time()
-            current_date = formatdate(current_time, usegmt=True).encode()
-            self.server_state.default_headers = [
-                (b"server", b"uvicorn"),
-                (b"date", current_date),
-            ]
```
lovelydinosaur, line 380: If we're setting up the header values in the config, then I think we end up with something pretty simple here, eg?...


```python 
current_time = time.time()
current_date = formatdate(current_time, usegmt=True).encode()
self.server_state.default_headers = [b"date", current_date)] + self.config.headers
```

uvicorn/config.py
Jitsusama: I'm always reticent to store much business logic in `__init__`. The stuff I put in there already was because I couldn't think of another way to handle it. But, if you'd like it that way, I can accommodate.

Jitsusama: I found a compromise. I'm going to perform the logic in `config.load()`.

Jitsusama:
Hello @tomchristie. This should address all of the changes you requested. I do have a thought though; why don't we move all header generation into config and store it as a property method? We can still do the custom header merge once upon configuration loading, but we could move the date-stamp generation into the property logic so it is completely removed from main.Server. Does that sound good?

uvicorn/config.py
```diff
@@ -122,6 +124,8 @@ def __init__(
         self.ssl_cert_reqs = ssl_cert_reqs
         self.ssl_ca_certs = ssl_ca_certs
         self.ssl_ciphers = ssl_ciphers
+        self._headers = headers if headers else []
+        self.headers = None  # type: List[Tuple[bytes, bytes]]
```
lovelydinosaur: Thinking about this once more - How about we go for `.headers` as the string pairs, and `.encoded_headers` for the byte-encoded pairs?

uvicorn/main.py
```diff
@@ -373,11 +381,10 @@ def run(self, sockets=None, shutdown_servers=True):
         # Update the default headers, once per second.
         if counter % 10 == 0:
             current_time = time.time()
-            current_date = formatdate(current_time, usegmt=True).encode()
+            current_date = formatdate(current_time, usegmt=True)
```
lovelydinosaur: Let's leave this line as it currently is - always best to try to keep the footprint of pull requests as strictly minimal as possible.

Jitsusama:
The requested changes have been made.

lovelydinosaur:
Great - let's have it!
Released as 0.6.1

PR #323, opened 2019-03-09
Read settings from config file

### LINKED ISSUE #322: File-based configuration
Hi there,

Has file-based configuration ever been considered for uvicorn? By that, I mean starting the server via a config file which we contain values for any made available by the `uvicorn` command.

[gunicorn has this feature](http://docs.gunicorn.org/en/stable/configure.html#configuration-file) and I like how the config file is just a plain Python script, which allows to do any sorts of computation (e.g. deriving the number of workers from the number of CPUs). File-based configuration also allows the server config to be version-controlled and written *once and for all* for a given project.

As an illustration, I imagine that file-based configuration would allow to use:

```python
# config.py
import os
import multiprocessing

host = "0.0.0.0"
port = 5050
workers = 2 * multiprocessing.cpu_count() + 1
log_level = "debug"
reload = not os.environ.get("PRODUCTION")
```

```bash
uvicorn -c config.py app:app
```

Instead of:

```bash
uvicorn --host="0.0.0.0" --port=5050 --workers=4 --log-level=debug --reload app:app
```

florimondmanca:
Fixes #322 

I'm not married to the name nor the help text of the settings option, so feel free to suggest anything better. :+1:

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -2,6 +2,7 @@ import asyncio
 import functools
 import logging
 import os
+import runpy
 import signal
 import socket
 import ssl
@@ -40,8 +41,47 @@ HANDLED_SIGNALS = (
 logger = logging.getLogger("uvicorn.error")
 
 
+def _setup_defaults_from_config_script(
+    ctx: click.Context, param: click.Parameter, value: typing.Optional[str]
+):
+    if value is None:
+        return
+
+    members: dict = runpy.run_path(value)
+
+    try:
+        config: Config = members["config"]
+    except KeyError:
+        raise click.BadParameter(f"no 'config' object declared in {value}")
+
+    if not isinstance(config, Config):
+        raise click.BadParameter(
+            f"'config' must be a 'uvicorn.Config' object (got {type(config)})"
+        )
+
+    # Read parameter defaults from the config object.
+    # Values may still be overridden using command line arguments.
+    ctx.default_map = {
+        param.name: getattr(config, param.name)
+        for param in ctx.command.params
+        if hasattr(config, param.name)
+    }
+
+    return value
+
+
+def _require_app_if_not_set_in_config_script(
+    ctx: click.Context, param: click.Parameter, value: typing.Optional[str]
+):
+    if value is None:
+        raise click.MissingParameter()
+    return value
+
+
 @click.command()
-@click.argument("app")
+@click.argument(
+    "app", required=False, callback=_require_app_if_not_set_in_config_script
+)
 @click.option(
     "--host",
     type=str,
@@ -219,6 +259,13 @@ logger = logging.getLogger("uvicorn.error")
     multiple=True,
     help="Specify custom default HTTP response headers as a Name:Value pair",
 )
+@click.option(
+    "--config",
+    type=click.Path(exists=True, dir_okay=False),
+    is_eager=True,  # Ensures 'callback' is called before processing other parameters.
+    callback=_setup_defaults_from_config_script,
+    help="Read configuration from a Python script.",
+)
 def main(
     app,
     host: str,
@@ -251,6 +298,7 @@ def main(
     ssl_ca_certs: str,
     ssl_ciphers: str,
     headers: typing.List[str],
+    config: str,
 ):
     sys.path.insert(0, ".")
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -116,6 +116,7 @@ Options:
   --ssl-ca-certs TEXT             CA certificates file
   --ssl-ciphers TEXT              Ciphers to use (see stdlib ssl module's)
                                   [default: TLSv1]
+  --config FILE                   Read configuration from a Python script.
   --help                          Show this message and exit.
 ```
 
@@ -389,7 +390,6 @@ with application code still running in a standard threaded context.
 [**FastAPI**](https://github.com/tiangolo/fastapi) is an API framework based on **Starlette** and **Pydantic**, heavily inspired by previous server versions of **APIStar**.
 
 You write your API function parameters with Python 3.6+ type declarations and get automatic data conversion, data validation, OpenAPI schemas (with JSON Schemas) and interactive API documentation UIs.
-
 
 [uvloop]: https://github.com/MagicStack/uvloop
 [httptools]: https://github.com/MagicStack/httptools
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -2,7 +2,7 @@
 
 Use the following options to configure Uvicorn, when running from the command line.
 
-If you're running using programmatically, using `uvicorn.run(...)`, then use
+If you're running programmatically, using `uvicorn.run(...)`, then use
 equivalent keyword arguments, eg. `uvicorn.run(app, port=5000, debug=True, access_log=False)`.
 
 ## Application
@@ -65,3 +65,7 @@ Note that WSGI mode always disables WebSocket support, as it is not supported by
 ## Timeouts
 
 * `--timeout-keep-alive <int>` - Close Keep-Alive connections if no new data is received within this timeout. **Default:** *5*.
+
+## Other
+
+* `--config <path>` - Read configuration from a Python script. The script should declare an instance of `uvicorn.Config` as `config`, which will be used to pre-populate command line arguments.
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -104,7 +104,7 @@ def create_ssl_context(certfile, keyfile, ssl_version, cert_reqs, ca_certs, ciph
 class Config:
     def __init__(
         self,
-        app,
+        app=None,
         host="127.0.0.1",
         port=8000,
         uds=None,


lovelydinosaur:
Alrighty, so.

This is great, tho it eventually occured to me, that if we're supporting python-based config files, we could equally well just make sure that this is well supported in the programmatic API, so eg...

```python
import os
import multiprocessing
import uvicorn

config = {
    "app": "app:app",
    "host": "0.0.0.0",
    "port": 5050,
    "workers": 2 * multiprocessing.cpu_count() + 1,
    "log_level": "debug",
    "reload": not os.environ.get("PRODUCTION")
}

# Run with given config options. Allow command-line flag overrides if provided.
unvicorn.main(**config)
```

lovelydinosaur:
We don't quite provide that at the moment, but expanding out `.main(...)` slightly would allow us to support this use case.

florimondmanca:
Yes, we'd need to change a bit how `main()` behaves because it doesn't seem that Click supports passing arguments directly to a command:

```python
import click

@click.command()
@click.option("--debug", default=False)
def cli(debug: bool):
    pass
```

```python
>>> cli(debug=True)
TypeError: __init__() got an unexpected keyword argument 'debug'
```

Admitting that we make this work, as I understand the idea would be to support this use case by making the config script "runnable" instead, right? So if the script you listed was `server.py`:

```bash
python server.py
```

TBH I'm not particularly against this, but I will argue that:

a) It differs from the behavior of other popular Python web servers (e.g. gunicorn). In general I'm not against shaking the status quo but I've personally found the `-c` gunicorn API to be quite handy.
b) It makes usage drift away from the now-standard `uvicorn` command.
c) The dict (or kwargs) formatting required by this method may be against the intended simplicity of declaring variables in a plain Python script.
d) Ultimately, the `run()` function already fulfils the need of passing arguments programmatically, so I don't think there's a need to provide a duplicate solution via `main()`.

In total transparency, I was thinking of using such a feature to abstract uvicorn away in a CLI command for Queso (Bocadillo's CLI). We'd just run `queso start` and it would pass a conventionally named settings file to the `uvicorn` command (see also https://github.com/bocadilloproject/queso/issues/3).

So the use case here is more to have a single, version-controllable configuration that is also accessible to other software systems through the file system.

florimondmanca:
> ```python
> # Run with given config options. Allow command-line flag overrides if provided.
> ```

I just read this and realised that this is not possible in the current implementation: if a setting is defined in the config file, we can't override it via the command line. I think this should be the case too.

lovelydinosaur:
You might be right here.

One thing I'm a bit wary of is the possilibity of mistyped config going easily unnoticed.

I'm wondering if there'd be any sense at all in something like this?...

```python
import os
import multiprocessing
import uvicorn

config = uvicorn.Config(
    app="app:app",
    host="0.0.0.0",
    port=5050,
    workers=2 * multiprocessing.cpu_count() + 1,
    log_level="debug",
    reload=not os.environ.get("PRODUCTION")
)
```

Then run with...

`uvicorn -c config.py`

That would:

* Ensure that you've got the option names correct.
* Re-inforce the low-level API.

Any thoughts?

florimondmanca:
Very good point. Currently I'm just silently skipping variables from the config script that aren't ones we expect, but enforcing that we instanciate a `Config` object would prevent this problem. We'd also get autocompletion for parameters of `Config` from IDEs etc, which is :+1:.

To be sure: we'd enforce that the config file declares a `Config` object named `config`, right?

florimondmanca:
Caveat: `app` is a required argument to `Config`, which means we'd *have* to specify the app name in the config script. Is this desirable, or can I make it optional on `Config`? I'll also have to make the `APP` argument optional on the `uvicorn` command too, otherwise `uvicorn -c config.py` would fail with a complain from Click.

lovelydinosaur:
> To be sure: we'd enforce that the config file declares a Config object named config, right?

Yeah.

And yeah, there’s a bunch of edge bits I’d not considered there, tho I like the API for it otherwise.

florimondmanca:
Okay, I've got something up and running. Allows to run from settings as follows:

```python
# settings.py
from uvicorn import Config

config = Config(
    app="app:App",
    workers=1,
    port=5050
)
```

```bash
uvicorn --settings settings.py
```

And the `app` argument is still required if not set in the config. Happy to have your thoughts on the implementation!

Jitsusama:
Just a note; I'm used to CLI arguments always over-riding any other value. Generally, the order of operations when processing configuration in a UNIX like daemon goes like so:

1. CLI arguments
2. Environment variables
3. Configuration file
4. Default values

The idea being; the closer you are to the invoking user, the more imperative the desired value is.

uvicorn/main.py
```diff
@@ -33,8 +34,39 @@
...
+        raise click.BadParameter("no 'config' object declared in {}".format(value))
+
+    if not isinstance(config, Config):
+        raise click.BadParameter(
+            "'config' must be a 'Config' object, not '{}'".format(type(config).__name__)
+        )
+
+    # Read parameter defaults from the config.
+    # Values can still be overridden using command line arguments.
+    ctx.default_map = vars(config)
```
lovelydinosaur: I'm not convinced that `vars(config)` here would necessarily map exactly onto the `main()` arguments.

florimondmanca: Yes, I can see why. I provided a fix that explicitly reads the dict items from the command's parameters. :+1:

florimondmanca:
@Jitsusama Thanks for clarifying this order. In the current state of this PR, it is fulfilled : CLI arguments > config file > defaults.

florimondmanca:
On a side note — @tomchristie I would like to provide tests for this as this is non-trivial and potentially bug-prone. I'm used to using click's [CliRunner](http://click.palletsprojects.com/en/7.x/testing/), but the issue is that I was not able to monkey-patch the `run` function used within `main`. As a result, I'd need to run the actual server (and provide an actual app) just to test that CLI arguments are correctly filled in. Any idea on how to swap `run` with an empty implementation during tests?

florimondmanca:
@tomchristie Just stumbled upon this PR of mine that's been lurking around. :) I noticed #444 advocates for a "gunicorn-like" `-c config.py` CLI usage (cc @Hanaasagi). Should I update this PR in favor of this vs. `--settings settings.py`?

uvicorn/main.py
```diff
@@ -36,8 +37,47 @@
...
+def _setup_defaults_from_settings_script(
+    ctx: click.Context, param: click.Parameter, value: typing.Optional[str]
+):
+    if value is None:
+        return
+
+    members = runpy.run_path(value)  # type: dict
+
+    try:
+        config = members["config"]  # type: Config
```
Hanaasagi: We have dropped Python 3.5 support. PEP 526 style annotations is more better : )

Hanaasagi:
In fact, we need a file based config. It is not necessary that we should implement the same style config like gunicorn. I prefer `uvicorn -c config.py`, because it will not a mix of `settings` and `config`. Create a file(e.g. `config.py`) and define `Config` object, but pass this through `--settings` looks a litte strange.

uvicorn/main.py
florimondmanca: So much happened in 7 months! 😁

florimondmanca:
Sounds good! I updated the PR. I'll try to add a test for this as well soon. :)

gvbgduh:
Just out of curiosity, what would be the way to pass settings to the uvicorn worker from gunicorn?

Hanaasagi:
@gvbgduh  If it is used as a worker, gunicorn will control the config processing, pass it's own config. (inherit from `gunicorn.workers.base.Worker`, and retrieve the config through `self.cfg`)

gvbgduh:
Yeah, I see, thanks @Hanaasagi. I haven't tried to broadcast settings like `loop`, `lifespan`

uvicorn/main.py
florimondmanca: @Hanaasgi Actually it seems like there’s still a PyPy 3.5 step in our CI build…? I rebased from master, is there anything else I’m missing?

Hanaasagi: :sweat_smile: Pending https://github.com/encode/uvicorn/pull/424 . Master branch CI is using pyp3.5.

lovelydinosaur:
This looks *great*!
I'd suggest we do this, *but* change it so that we're just expecting top-level settings names, eg...

```python
workers = 4
loop = "asyncio"
ws = "wsproto"
```

Rather than...

```python
import uvicorn

config = uvicorn.Config(
    workers = 4,
    loop = "asyncio",
    ws = "wsproto"
)
```

I think it's nicer enough that I'd ignore my own "what about misspelled names"

uvicorn/main.py
```diff
@@ -37,8 +38,47 @@
...
+    except KeyError:
+        raise click.BadParameter(f"no 'config' object declared in {value}")
+
+    if not isinstance(config, Config):
+        raise click.BadParameter(
+            f"'config' must be a 'uvicorn.Config' object (got {type(config)})"
+        )
+
+    # Read parameter defaults from the config object.
+    # Values may still be overridden using command line arguments.
```
lovelydinosaur, line 63: How's this instead, for loading names at the top-level?...

```python
members: dict = runpy.run_path(value)

valid_param_names = inspect.signature(main).parameters.keys()
ctx.default_map = {
    param.name: members[param.name]
    for param in ctx.command.params
    if param.name in valid_param_names
}
```

tomchristie: I guess that maybe doesn't work right? Given that we still want type conversion to be applied?

florimondmanca:
> I'd suggest we do this, but change it so that we're just expecting top-level settings names, eg...

I actually like the Config-based usage a lot - gives you autocompletion, type checking, and thus reduces the need to check the documentation or CLI usage to see what options are available (your IDE will tell you).

*Maybe* we could support both?

lovelydinosaur:
I’m willing to be swayed on this!

florimondmanca:
Haven't felt the need for this feature in the past months, so closing this (stale) PR for now. :-)

PR #325, opened 2019-03-12
Support --interface [auto|asgi3|asgi2|wsgi]

lovelydinosaur:
Switches `--wsgi` to a more generic `--interface`.
Adds support for `--interface [auto|asgi3|asgi2|wsgi]`.
Since it'll stick with asgi2 for existing cases I don't see any reason not to get this in pre-emptively.

We'd need to jump to 0.6 for this, since it changes the programmatic API: `wsgi=True` to `interface="wsgi"` and the command line API `--wsgi` to `--interface wsgi`

```python
import uvicorn


    await send({
        'status': 200,
        'headers': [
        ]
    })
    await send({
        'type': 'http.response.body',
        'body': b'Hello, world!',
    })


if __name__ == "__main__":
    uvicorn.run(app)
```

See https://github.com/django/asgiref/issues/78 and https://github.com/django/asgiref/pull/80

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -45,8 +45,9 @@ Options:
                                   WebSocket protocol implementation.
                                   [default: auto]
   --lifespan [auto|on|off]        Lifespan implementation.  [default: auto]
-  --wsgi                          Use WSGI as the application interface,
-                                  instead of ASGI.
+  --interface [auto|asgi3|agsi2|wsgi]
+                                  Select ASGI3, ASGI2, or WSGI as the
+                                  application interface.
   --log-level [critical|error|warning|info|debug]
                                   Log level.  [default: info]
   --no-access-log                 Disable access log.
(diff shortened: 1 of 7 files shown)


lovelydinosaur:
ASGI3 app instances:

```python
class Response:
    def __init__(self, content):
        assert isinstance(content, bytes)
        self.content = content

        await send({
            'status': 200,
            'headers': [
                [b'content-length', str(len(self.content)).encode()],
            ]
        })
        await send({
            'type': 'http.response.body',
           'body': self.content,
        })

app = Response(b'Hello, world')
```

ASGI3 app classes:

```python
class App:
    def __init__(self, scope, receive, send):
        self.scope = scope
        self.receive = receive
        self.send = send

    def __await__(self):
        return self.asgi().__await__()

    async def asgi(self):
        await self.send({
            'status': 200,
            'headers': [
            ],
        })
        await self.send({
            'type': 'http.response.body',
            'body': b'Hello, world!',
        })
```

lovelydinosaur:
Let's get this one in now. It'll happily stick with ASGI 2 for existing apps, and we can make any minor finesses to it's behavior for ASGI 3 based on anything remaining that comes out of the discussion on `asgiref`.

pgjones:
@tomchristie I think uvicorn should set the correct `{"asgi": {"version": ??}}` value in the scope.

Issue #337, opened 2019-03-28
Feature request: unix socket owner and permision settings


em92:
By default, uvicorn sets 0755 permission flags to unix socket and using current user and group.
```
srwxr-xr-x 1 eugene eugene 0 мар 28 11:56 /tmp/uvicorn.sock
```

For example, if app is running on user `eugene` and nginx worker is running on user `www-data`, nginx won't be able to access socket and will return 502 error. One of the solutions is to change permission flags and group id of the file.
```
srwxrwxr-x 1 eugene www-data 0 мар 28 12:00 /tmp/uvicorn.sock
```
(assuming `eugene` is in group `www-data`)

Quick patch example:
```diff
diff --git uvicorn/config.py uvicorn/config.py
index 5200032..8f51e69 100644
--- uvicorn/config.py
+++ uvicorn/config.py
@@ -71,6 +71,9 @@ class Config:
         host="127.0.0.1",
         port=8000,
         uds=None,
+        uid=None,
+        gid=None,
+        chmod=None,
         fd=None,
         loop="auto",
         http="auto",
@@ -103,6 +106,9 @@ class Config:
         self.host = host
         self.port = port
         self.uds = uds
+        self.uid = uid
+        self.gid = gid
+        self.chmod = chmod
         self.fd = fd
         self.loop = loop
         self.http = http
diff --git uvicorn/main.py uvicorn/main.py
index ab88499..4d6a986 100644
--- uvicorn/main.py
+++ uvicorn/main.py
@@ -22,6 +22,22 @@ from uvicorn.config import (
 )
 from uvicorn.supervisors import Multiprocess, StatReload
 
+
+class BasedIntParamType(click.ParamType):
+    name = 'integer'
+
+    def convert(self, value, param, ctx):
+        try:
+            if value[:2].lower() == '0x':
+                return int(value[2:], 16)
+            elif value[:1] == '0':
+                return int(value, 8)
+            return int(value, 10)
+        except ValueError:
+            self.fail('%s is not a valid integer' % value, param, ctx)
+
+
+BASED_INT = BasedIntParamType()
 LEVEL_CHOICES = click.Choice(LOG_LEVELS.keys())
 HTTP_CHOICES = click.Choice(HTTP_PROTOCOLS.keys())
 WS_CHOICES = click.Choice(WS_PROTOCOLS.keys())
@@ -52,6 +68,9 @@ HANDLED_SIGNALS = (
     show_default=True,
 )
[...]

PR #339, opened 2019-03-29
chmod 0666 for unix socket

em92:
As discussed in https://github.com/encode/uvicorn/issues/337

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -360,6 +360,7 @@ class Server:
         elif config.uds is not None:
             # Create a socket using UNIX domain socket.
             server = await loop.create_unix_server(create_protocol, path=config.uds)
+            os.chmod(config.uds, 0o666)
             message = "Uvicorn running on unix socket %s (Press CTRL+C to quit)"
             self.logger.info(message % config.uds)
             self.servers = [server]
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -359,8 +359,11 @@ class Server:
 
         elif config.uds is not None:
             # Create a socket using UNIX domain socket.
+            uds_perms = 0o666
+            if os.path.exists(config.uds):
+                uds_perms = os.stat(config.uds).st_mode
             server = await loop.create_unix_server(create_protocol, path=config.uds)
-            os.chmod(config.uds, 0o666)
+            os.chmod(config.uds, uds_perms)
             message = "Uvicorn running on unix socket %s (Press CTRL+C to quit)"
             self.logger.info(message % config.uds)
             self.servers = [server]


uvicorn/main.py
```diff
@@ -360,6 +360,7 @@ def run(self, sockets=None, shutdown_servers=True):
         elif config.uds is not None:
             # Create a socket using UNIX domain socket.
             server = await loop.create_unix_server(create_protocol, path=config.uds)
+            os.chmod(config.uds, 0o666)
```
lovelydinosaur: I think we perhaps only do this if *we've* created the file.
If the UDS already exists, then we should stick with whatever existing permissions and ownership it has, right?
Perhaps this?...

```python
set_perms = not os.path.exists(config.uds)
server = await loop.create_unix_server(create_protocol, path=config.uds)
if set_perms:
    os.chmod(config.uds, 0o666)
```

Seem reasonable?

em92: > I think we perhaps only do this if we've created the file.

`create_unix_server` does it for us. [doc says](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_unix_server), that it always return Server object, or dies on exception.

> If the UDS already exists, then we should stick with whatever existing permissions and ownership it has, right?

We already do it.

em92: > If the UDS already exists, then we should stick with whatever existing permissions and ownership it has, right?

Pardon, ignore prevoius answer. We don't have to worry about permissions of existing files. If `create_unix_server` fails, it will throw exception.

lovelydinosaur: > We don't have to worry about permissions of existing files. If create_unix_server fails, it will throw exception.

I don't think it raises any exception if you're running it against an existing UDS.
(Eg. try running uvicorn --uds, then quitting it and running again)

em92: > I don't think it raises any exception if you're running it against an existing UDS.

It does not throw exception at least if user has permission to access to existing UDS. 
If existing UDS after successful execution of `create_unix_server` has some different permissions, it will change back to 0666 after `os.chmod(config.uds, 0o666)`

em92: I get it. You want to save UDS's permission if it runs on existing UDS. Editing...

uvicorn/main.py
```diff
@@ -359,7 +359,11 @@ def run(self, sockets=None, shutdown_servers=True):
 
         elif config.uds is not None:
             # Create a socket using UNIX domain socket.
+            uds_perms = 0o666
+            if os.path.exists(config.uds):
+                uds_perms = os.stat(config.uds).st_mode
```
lovelydinosaur, line 364: Rather thank do this, it'd make more sense simply to *not* perform `os.chmod` if `os.path.exists(...)` returned true.

Eg. Use `should_chmod = not os.path.exists(config.uds)` here.

(Eg. perhaps the user account might not have permissions to chmod it for whatever reason)

uvicorn/main.py
lovelydinosaur: > I get it. You want to save UDS's permission if it runs on existing UDS. Editing...
 
Exactly, yes. Or rather: Just don't chmod it at all in that situation.

lovelydinosaur: Eg. the UDS has been created by some completely separate supervisor process - we don't want to touch the permissions with uvicorn in that case.

uvicorn/main.py
lovelydinosaur: Thanks for your time in getting this ironed out!
It'll end up being a relatively small PR for quite a lot of chat, but it's worth getting it right.

em92: create_unix_server resets permissions (0755), so my code returns it back which it had before

lovelydinosaur: Ah gotcha

Issue #338, opened 2019-03-28
High CPU usage when using --reload


nymous:
Using the example code
```python
async def app(scope, receive, send):
    assert scope['type'] == 'http'

    await send({
        'type': 'http.response.start',
        'status': 200,
        'headers': [
            [b'content-type', b'text/plain'],
        ],
    })
    await send({
        'type': 'http.response.body',
        'body': b'Hello, world!',
    })
```

and running `uvicorn main:app`, CPU usage is unnoticeable (less than 1% of 1 core).
But when running with `uvicorn main:app --reload`, CPU usage jumps to 54% of 1 core.

*Environment:*
`python -V`: Python 3.7.2
`uvicorn`: 0.6.1
`uname -a`: `Linux architect 5.0.4-arch1-1-ARCH #1 SMP PREEMPT Sat Mar 23 21:00:33 UTC 2019 x86_64 GNU/Linux`

lovelydinosaur:
You should only use `--reload` in development, so I don't see this as a pressing issue.
I'd be very happy to see an alternative optional reload implementation based on `watchdog`, since that'd be far more efficient than our stat-based reloader. But the impetus for that would need to come from someone taking on a pull request.

nymous:
Yes I use it only in dev, I just wanted to know if it was normal to hear my fans ramp up.

lovelydinosaur:
We could bump down the frequency at which we check the file timestamps?
Happy to take any other suggestions?

Reloader implementation: https://github.com/encode/uvicorn/blob/master/uvicorn/supervisors/statreload.py

kevchentw:
Should we make reload check time configurable?
My fans is always running in full speed when I use `--reload`

lovelydinosaur:
> Should we make reload check time configurable?

No to configuring reload check time - that’s too many dials. Tho there’s more efficient file watching systems (eg gunciorn has an optional dependency, and only falls back to stat checks if needed) we should probably do the same. It could also be reasonable for us to allow controls into which files are being watched (eg perhaps it’s having to scan a very large, unchanging, virtualenv, as well as the actual source code?)

kevchentw:
This make sense. I think that we can choose which folder to watch will be great.

kevchentw:
I found that there are `--reload-dir` settings already. By applying app folder, CPU usage significant decrease.

wshayes:
[...]

PR #341, opened 2019-03-31
Changed reload check time to 1 sec from 100ms

wshayes:
Every 100ms seemed too fast and unnecessarily adding CPU load when debugging is turned on. If 1000ms is too much, maybe 500ms?

This is in response to #338

### DIFF
--- a/uvicorn/supervisors/statreload.py
+++ b/uvicorn/supervisors/statreload.py
@@ -34,7 +34,7 @@ class StatReload:
         process.start()
 
         while process.is_alive() and not self.should_exit:
-            time.sleep(0.1)
+            time.sleep(1)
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Thanks for the suggestion!
I guess somewhere in the 200-500ms range would probably be a reasonable trade-off.

lovelydinosaur:
That'd like be fast enough to not feel "sticky".

lovelydinosaur (approved):

lovelydinosaur:
Oops - needs black-compliant formatting. Switch it to 0.3, or run black and let it do that for you.

lovelydinosaur (approved):

Issue #346, opened 2019-04-13
Eventloop running at import time from CLI


thmsmlr:
Thanks for all the great work on uvicorn. This issue is more of a question than a bug. 

Consider the following `run.py`
```python
import asyncio
import uvicorn

print('Event loop running?: ', asyncio.get_event_loop().is_running())

async def app(scope, receive, send):
    pass

if __name__ == '__main__':
    uvicorn.run(app, host='127.0.0.1', port=5000, log_level='info')
```

When we run it programmatically we get,

```
➤ python run.py
Event loop running?:  False
INFO: Started server process [30067]
INFO: Waiting for application startup.
INFO: Uvicorn running on http://127.0.0.1:5000 (Press CTRL+C to quit)
```

Where as when we run it through the CLI

```
➤  uvicorn run:app
Event loop running?:  True
INFO: Started server process [29952]
INFO: Waiting for application startup.
INFO: Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
```

There seems to be inconsistent behavior here depending on how the application is run. It was my understanding that you don't want the event loop to be running when a module is imported. Since the `await` keyword isn't support at the module level you are forced to run any async code at import by executing `loop.run_until_complete(...)` but that will throw an error in the case of uvicorn running from the CLI with the error `RuntimeError: this event loop is already running.`. 

I dived into the code and it seems like it's happening because in `main.py` you are lazily importing the code via the `Config` object in the `async serve(...)` function. 

https://github.com/encode/uvicorn/blob/master/uvicorn/main.py#L310-L311

What do you think about hoisting that call into the `def run(...)` function before the event loop is started, but after the event loop is setup? That way the programmatic and the CLI will have the same behaviour.

Thanks again for the great project!

ohmeow:
I'm having a similar issue.  Ever get any resolution?

lovelydinosaur:
I think a reasonable way to tweak this would be to include config loading in the `run(...)` method.

```python
    config = self.config
    if not config.loaded:
        config.load()
```

However, I'd generally recommend avoiding having to carew about this at all, by using the `lifespan` ASGI event to startup any async services, rather than using `loop.run_until_complete` at import time.

lovelydinosaur:
I'll happily accept a pull request adding early config loading prior to calling into `.start()`

PR #347, opened 2019-04-13
main.py imports code before eventloop starts running

thmsmlr:
Before this patch there was an inconsistency where if you ran uvicorn
from the CLI versus programmatically the eventloop would either be or
not be running when the code was imported. It is generally good practice
to make sure that the event loop isn't running when code is imported
since async code can only be run at import time using the
loop.run_until_complete(...) command which requires the event loop to
not be running.

This patch makes sure that the code is loaded after the event loop is
configured, but before it starts running. This allows for consistent
behavior across deployment options.

Solves: https://github.com/encode/uvicorn/issues/346

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -300,6 +300,8 @@ class Server:
 
     def run(self, sockets=None, shutdown_servers=True):
         self.config.setup_event_loop()
+        if not self.config.loaded:
+            self.config.load()
         loop = asyncio.get_event_loop()
         loop.run_until_complete(self.serve(sockets=sockets))
 


lovelydinosaur:
Thanks for raising this.
Haven't had a chance to get around to reviewing it yet, but it is appreciated!
Will get around to it in due course 👍

kevinastone:
I'm currently running into this trying to migrate from `uvicorn my-app:app` to a console script that uses `uvicorn.run(app, ...)`.

kevinastone:
Never mind.  I was running into `aiohttp.ClientSession` fetching the event loop during `__init__` (https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L197).  I was using it as a context manager in the app lifecycle, so expected it to be fine until the `__aenter__`.

Kludex:
It's been 2 years, but I think we still need this. :thinking:

Kludex:
@thmsmlr I'm extremely sorry for taking so long. :disappointed: 

Thank you for the PR! :tada: 

Due to this PR being too old, and I couldn't solve the conflicts properly, I'm going to address those changes in #1110,

PR #349, opened 2019-04-18
Only set text_data or bytes_data but not both

### LINKED ISSUE #335: byes_data not working with Channels consumers
Uvicorn sends its data like this:

```
        is_text = isinstance(data, str)
        return {
            "type": "websocket.receive",
            "text": data if is_text else None,
            "bytes": None if is_text else data,
        }
```

Channels generic consumers work like this:
```
    def websocket_receive(self, message):
        """
        Called when a WebSocket frame is received. Decodes it and passes it
        to receive().
        """
        if "text" in message:
            self.receive(text_data=message["text"])
        else:
            self.receive(bytes_data=message["bytes"])
```

Since `text` is present but set to none `self.receive` is called with text_data discarding the bytes_data when it is present.

I don't know if this is a bug in Uvicorn or Channels. Its a simple fix either way. Either don't set it on the dict or the channels people should check if its a truthy value and not just present.

agronick:
This fixes #335

### DIFF
--- a/uvicorn/protocols/websockets/websockets_impl.py
+++ b/uvicorn/protocols/websockets/websockets_impl.py
@@ -88,7 +88,7 @@ class WebSocketProtocol(websockets.WebSocketServerProtocol):
         self.scope = {
             'type': 'websocket',
             'scheme': self.scheme,
-            'server': self.server or ('unknown', 0),
+            'server': self.server,
             'client': self.client,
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Thanks!

lovelydinosaur:
Now released in 0.7.0

Issue #333, opened 2019-03-26
AssertionError


gogobook:
Lack of `assert scope['type'] == 'http'` in the example of document will cause an `AssertionError`

lovelydinosaur:
Make sure you're running the latest version of uvicorn.

I see this...

```python
async def app(scope, receive, send):
    # assert scope['type'] == 'http'

    await send({
        'type': 'http.response.start',
        'status': 200,
        'headers': [
            [b'content-type', b'text/plain'],
        ],
    })
    await send({
        'type': 'http.response.body',
        'body': b'Hello, world!',
    })
```

```shell
$ uvicorn example:app
INFO: Started server process [57966]
INFO: Waiting for application startup.
INFO: ASGI 'lifespan' protocol appears unsupported.
INFO: Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
```

gogobook:
May try this again? please.
```py
class EchoBody():
    def __init__(self, scope):
        self.scope = scope

    async def read_body(self, receive):
        """
        Read and return the entire body from an incoming ASGI message.
        """
        body = b''
        more_body = True

        while more_body:
            message = await receive()
            body += message.get('body', b'')
            more_body = message.get('more_body', False)

        return body


    async def __call__(self, receive, send):
        body = await self.read_body(receive)
        await send({
            'type': 'http.response.start',
            'status': 200,
            'headers': [
                [b'content-type', b'text/plain'],
            ]
        })
        await send({
            'type': 'http.response.body',
            'body': body,
        })
```

lovelydinosaur:
Ah thanks. Docs were a bit out of date.
I've updated them to new ASGI-3 versions.

anentropic:
What does:
> ASGI 'lifespan' protocol appears unsupported 

mean?

lovelydinosaur:
ASGI has support for `http`, `websocket`, and `lifespan` message types.
Some servers may not support all message types.

https://asgi.readthedocs.io/en/latest/

anentropic:
I seem to get it when there is an error during start up

e.g. if there is an uncaught exception during Starlette `on_event('startup')` handler I don't see the exception logged but I see "ASGI 'lifespan' protocol appears unsupported" and the server starts up anyway 

if I handle the exception then that message goes away

lovelydinosaur:
Gotcha - you could probably open an issue for that.
[...]

PR #350, opened 2019-04-22
Add support for lifespan.startup.failed

florimondmanca:
This PR implements this part of the [ASGI spec](https://asgi.readthedocs.io/en/latest/specs/lifespan.html#scope):

> If you want to log an error that occurred during lifespan startup and prevent the server from starting, then send back `lifespan.startup.failed` instead.

and helps with https://github.com/encode/starlette/issues/486.

Unless I'm mistaken, contrary to what was mentioned in #333 uvicorn does not react to `lifespan.startup.failed` events yet.

It's the first time I dive into the lifespan internals, so very keen for any feedback on the code and tests. :+1:

### DIFF
--- a/uvicorn/lifespan/on.py
+++ b/uvicorn/lifespan/on.py
@@ -73,10 +73,9 @@ class LifespanOn:
             assert not self.startup_event.is_set(), STATE_TRANSITION_ERROR
             assert not self.shutdown_event.is_set(), STATE_TRANSITION_ERROR
             self.startup_event.set()
-
+            self.startup_failed = True
             if message.get("message"):
                 self.logger.error(message["message"])
-            self.startup_failed = True
 
         elif message["type"] == "lifespan.shutdown.complete":
             assert self.startup_event.is_set(), STATE_TRANSITION_ERROR
--- a/uvicorn/lifespan/on.py
+++ b/uvicorn/lifespan/on.py
@@ -14,6 +14,7 @@ class LifespanOn:
         self.shutdown_event = asyncio.Event()
         self.receive_queue = asyncio.Queue()
         self.error_occured = False
+        self.startup_failed = False
         self.should_exit = False
 
     async def startup(self):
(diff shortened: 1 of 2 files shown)


uvicorn/lifespan/on.py
```diff
@@ -58,13 +60,25 @@ def __init__(self, config):
...
             assert not self.startup_event.is_set(), STATE_TRANSITION_ERROR
             assert not self.shutdown_event.is_set(), STATE_TRANSITION_ERROR
             self.startup_event.set()
+
+        elif message["type"] == "lifespan.startup.failed":
+            assert not self.startup_event.is_set(), STATE_TRANSITION_ERROR
+            assert not self.shutdown_event.is_set(), STATE_TRANSITION_ERROR
+            self.startup_event.set()
+
+            if message.get("message", ""):
```
lovelydinosaur: I guess this could just be `message.get("message")`, since it'll return `None` off not present.

lovelydinosaur:
Really great work on these two tickets!
We *may* want to also check the case where the failure message is sent, but no exception is raised.

lovelydinosaur:
I've been a bit busy on `requests-async`/`httpcore` lately, but will def. aim to get these in and released this week - welcome to nudge me on here again if it drops off my radar.

uvicorn/lifespan/on.py
```diff
@@ -43,28 +43,41 @@ def __init__(self, config):
             scope = {"type": "lifespan"}
             await app(scope, self.receive, self.send)
         except BaseException as exc:
+            self.asgi = None
```
florimondmanca, line 46: @tomchristie Any reason that this particular line exists? I grep'ed the code base and couldn't find a `lifespan.asgi` being used anywhere. Since I was unsure I kept it, but maybe it's an ancient artefact that should be removed?

uvicorn/lifespan/on.py
florimondmanca: Fixed!

florimondmanca:
Hi there,

> welcome to nudge me on here again if it drops off my radar.

As you suggested, here's a kind reminder about the existence of this PR. :-) I believe I provided a fix for your comments from last time.

PR #351, opened 2019-04-24
Add extra context to access log records

Code0x58:
This uses the `extra` kwarg ([documented here](https://docs.python.org/3/library/logging.html#logging.Logger.debug)) to pass additional information which can be used by a logging handler.

We plan to use this in a log handler to produce structured (LDJSON) logs.

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -435,6 +435,7 @@ class RequestResponseCycle:
                     self.scope.get("root_path", "") + self.scope["path"],
                     self.scope["http_version"],
                     status_code,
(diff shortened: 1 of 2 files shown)


Code0x58:
@tomchristie can I get your thoughts on this? I'd be more than happy to talk about this, do any extra necessary work, or answer any questions you have on it.

The addition won't change how existing logs are rendered, so won't impact existing users. It just adds (using the standard `extra` kwarg of the logging methods) `record.scope` and `record.status_code` to the [`LogRecord`](https://docs.python.org/3/howto/logging.html#logging-levels) that [flows](https://docs.python.org/3/howto/logging.html#logging-flow) through the system when you make a logging call. We'd like to get this in to enrich our logging.

Issue #261, opened 2018-12-31
No longer possible to tell if path had escaped characters in it


simonw:
This commit introduced behaviour which is causing me problems: https://github.com/encode/uvicorn/commit/b420242ab7e5336cfca8a8bd16795e0ab21af889

For my application (Datasette) I need to be able to tell the difference between the following two URLs:

* https://v0-25-2.datasette.io/fixtures-dd88475/table%2Fwith%2Fslashes.csv
* https://v0-25-2.datasette.io/fixtures-dd88475/table/with/slashes.csv

This is because Datasette's URL design allows for the name of a SQLite database table to be incorporated into the URL path - and SQLite tables can (bizarrely) include slashes in their names. As you can see in the above example, this works: you can `%2F` encode those slashes and Datasette will know that they are part of the table name.

Prior to the b420242ab7e5336cfca8a8bd16795e0ab21af889 uvicorn handled this in a way that works for my application. Here's a debugging tool I deployed using an older version of uvicorn that illustrates the behaviour I need:

https://asgi-scope.now.sh/fixtures/table%2Fwith%2Fslashes.csv

This tool ( https://github.com/simonw/asgi-scope ) simply outputs the current ASGI scope. As you can see, the `path` component looks like this:
```
 'method': 'GET',
 'path': '/fixtures/table%2Fwith%2Fslashes.csv',
 'query_string': b'',
```
But... here's a new version of the above tool I just deployed using the latest uvicorn release:

https://asgi-scope-uvicorn-0-3-23.now.sh/fixtures/table%2Fwith%2Fslashes.csv
```
 'method': 'GET',
 'path': '/fixtures/table/with/slashes.csv',
 'query_string': b'',
```

The path has been decoded for me! This means that my application code has no way of telling if the initial input had `%2F` or `/` - that information has been lost entirely.

Unfortunately, this appears to be a behaviour that is encoded in the ASGI spec itself: 

https://github.com/django/asgiref/blob/5fe2e535e64f85ada8078ad6aabf5e418b77d26b/specs/www.rst#L63-L65

I'm going to open a similar bug against the spec, but assuming the spec doesn't change how about adding a new key to the scope called something like `raw_path` which exposes the original bytes?

Alternatively, how about making it so I can easily subclass the relevant part of uvicorn and get back the behaviour that I need?

simonw:
Huh... turns out this has been discussed in the asgi repo previously: https://github.com/django/asgiref/issues/51#issuecomment-403089109

[...]

PR #354, opened 2019-05-01
Do not unquote path

blueyed:
If `/foo%2Fbar` is requested, it should be passed on like that.

It is the responsibility of the app to do any encoding (if wanted), and
currently prevents to get the raw value.

Found only https://github.com/encode/uvicorn/commit/b420242ab7e5336cfca8a8bd16795e0ab21af889 quickly in this regard.

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -1,7 +1,6 @@
 import asyncio
 import http
 import logging
-from urllib.parse import unquote
 
 import h11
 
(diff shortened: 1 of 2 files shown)


blueyed:
hypercorn is also doing this.. I hope it is not in the spec - but could not find something about it in asgiref.

lovelydinosaur:
Spec is here: https://asgi.readthedocs.io/en/latest/specs/www.html#http

> `path` *(Unicode string)* – HTTP request target excluding any query string, with percent-encoded sequences and UTF-8 byte sequences decoded into characters.

Same as with WSGI.

blueyed:
Ok, created https://github.com/django/asgiref/issues/87.

What about a test covering this? :)

blueyed:
Also a duplicate of https://github.com/encode/uvicorn/issues/261 - so at least somebody else cares.. ;)

PR #357, opened 2019-05-05
Fix reload with custom logger

Hanaasagi:
It will resolve #328, backport `__reduce__` method from Python 3.7.

https://github.com/python/cpython/blob/29500737d45cbca9604d9ce845fb2acc3f531401/Lib/logging/__init__.py#L1730-L1736

```Python
    def __reduce__(self):
        # In general, only the root logger will not be accessible via its name.
        # However, the root logger's class has its own __reduce__ method.
        if getLogger(self.name) is not self:
            import pickle
            raise pickle.PicklingError('logger cannot be pickled')
        return getLogger, (self.name,)
```

PR #359, opened 2019-05-08
Fix repeatedly config logger

keakon:
There was a bug that it reconfiged a logger for each new connection.


Hanaasagi:
Seems you've duplicated some work in #357 :flushed:

keakon:
> Seems you've duplicated some work in #357 😳

Yes, the 2nd bug is same as #357.

lovelydinosaur:
Now resolved in master - thanks for this!

PR #361, opened 2019-05-11
Use wall clock to determine gap between notify calls

### LINKED ISSUE #360: The notify callback is not called on time
https://github.com/encode/uvicorn/blob/545a2a10859165db016d02a641f0c30d07476af1/uvicorn/main.py#L404

On this line we rely on the 0.1 second ticks that occur to notify gunicorn in time. Jitter or cpu bound or naughty blocking io can cause these ticks to gradually fall behind and the process gets killed.

We should use the call to time that we do a few lines above and store the last time we notified in a member variable rather than the ticks, and make sure we have a margin so we definitely don’t notify too late.

rjwittams:
Fixes #360

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -298,6 +298,7 @@ class Server:
         self.started = False
         self.should_exit = False
         self.force_exit = False
+        self.last_notified = 0
 
     def run(self, sockets=None, shutdown_servers=True):
         self.config.setup_event_loop()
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Yeah - that looks good to me. Thanks!

lovelydinosaur:
Released as 0.7.2 - cheers!

PR #367, opened 2019-06-01
Cast headers to list whichever iterable it is (list, tuple, other)

### LINKED ISSUE #366: Uvicorn should handle headers as any iterables, not just lists (ASGI spec compliance)
[ASGI Specification](https://asgi.readthedocs.io/en/latest/specs/www.html#response-start) in the `Response Start` says that if I send `http.response.start`, the `headers` part should be of the following form:

(Iterable[[byte string, byte string]])

>  A iterable of [name, value] two-item iterables, where name is the header name, and value is the header value. Order must be preserved in the HTTP response. Header names must be lowercased. Optional; defaults to an empty list.

So let's have an ASGI compliant application of the following form:

```
async def application(scope, receive, send):
    await send(
        {
            "headers": ([b"content-type", b"text/plain"],),
            "status": 200,
        }
    )
    await send({"type": "http.response.body", "body": b"dupa"})
```

After requesting this I get the following in the console:

```
ERROR: Exception in ASGI application
  File "/home/jerzy/venv/lib/python3.7/site-packages/uvicorn/protocols/http/httptools_impl.py", line 368, in run_asgi
  File "./asgi.py", line 9, in application
    "status": 200,
  File "/home/jerzy/venv/lib/python3.7/site-packages/uvicorn/protocols/http/httptools_impl.py", line 427, in send
    headers = self.default_headers + message.get("headers", [])
TypeError: can only concatenate list (not "tuple") to list

```

Looks like the headers have to be list of list, not simple iterable of 2-items iterable. Do you think this is important? If we want to be 100% compliant with ASGI spec uvicorn should handle just any iterables. If you agree, do you think I could prepare a patch?

JerzySpendel:
[In reference to this issue](https://github.com/encode/uvicorn/issues/366)

Uvicorn should be fine with just any iterable.

Closes #366

### DIFF
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -424,7 +424,7 @@ class RequestResponseCycle:
             self.waiting_for_100_continue = False
 
             status_code = message["status"]
-            headers = self.default_headers + message.get("headers", [])
+            headers = self.default_headers + list(message.get("headers", []))
 
             if self.access_log:
                 self.logger.info(


lovelydinosaur:
Thanks!

PR #372, opened 2019-06-16
Implement scope["raw_path"] for both HTTP and websockets

simonw:
As discussed in https://github.com/django/asgiref/issues/87 and implemented in https://github.com/django/asgiref/pull/92

The PR adding `raw_path` to the ASGI specification will hopefully be merging very shortly. This change to Uvicorn should be considered blocked until the specification has been updated.

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -58,6 +58,10 @@ FINISH_POST_REQUEST = b'{"hello": "world"}'
 
 HTTP10_GET_REQUEST = b"\r\n".join([b"GET / HTTP/1.0", b"Host: example.org", b"", b""])
 
+GET_REQUEST_WITH_RAW_PATH = b"\r\n".join(
+    [b"GET /one%2Ftwo HTTP/1.1", b"Host: example.org", b"", b""]
+)
+
 UPGRADE_REQUEST = b"\r\n".join(
     [
         b"GET / HTTP/1.1",
@@ -539,9 +543,7 @@ def test_raw_path(protocol_cls):
         await response(scope, receive, send)
 
     protocol = get_connected_protocol(app, protocol_cls, root_path="/app")
-    protocol.data_received(
-        b"\r\n".join([b"GET /one%2Ftwo HTTP/1.1", b"Host: example.org", b"", b""])
-    )
+    protocol.data_received(GET_REQUEST_WITH_RAW_PATH)
     protocol.loop.run_one()
     assert b"Done" in protocol.transport.buffer
 
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -532,10 +532,10 @@ def test_raw_path(protocol_cls):
     async def app(scope, receive, send):
         path = scope["path"]
         raw_path = scope.get("raw_path", None)
-        response = Response(
-            "Path: {}\nraw_path: {}".format(path, repr(raw_path)),
-            media_type="text/plain",
-        )
+        assert "/one/two" == path
+        assert b"/one%2Ftwo" == raw_path
+
+        response = Response("Done", media_type="text/plain")
         await response(scope, receive, send)
 
     protocol = get_connected_protocol(app, protocol_cls, root_path="/app")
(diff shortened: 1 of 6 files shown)


simonw:
Previously discussed in #261 and #354 - but that was before the change to the ASGI specification.

tests/protocols/test_websocket.py
```diff
@@ -52,15 +52,17 @@ def run_loop(loop):
...
     asyncio.set_event_loop(None)
     loop = asyncio.new_event_loop()
     config = Config(app=app, ws=protocol_cls)
     server_state = ServerState()
     protocol = functools.partial(H11Protocol, config=config, server_state=server_state)
     create_server_task = loop.create_server(protocol, host="127.0.0.1")
     server = loop.run_until_complete(create_server_task)
-    url = "ws://127.0.0.1:%d/" % server.sockets[0].getsockname()[1]
+    url = "ws://127.0.0.1:{port}{path}".format(
+        port=server.sockets[0].getsockname()[1], path=path
```
lovelydinosaur: I guess we may as well pull `port = server.sockets[0].getsockname()[1]` out to the line above, and then use `.format(port=port, path=path)`

tests/protocols/test_http.py
```diff
@@ -527,6 +527,28 @@ def test_root_path(protocol_cls):
     assert b"Path: /app/" in protocol.transport.buffer
 
 
+@pytest.mark.parametrize("protocol_cls", HTTP_PROTOCOLS)
+def test_raw_path(protocol_cls):
+    async def app(scope, receive, send):
+        path = scope["path"]
+        raw_path = scope.get("raw_path", None)
+        response = Response(
+            "Path: {}\nraw_path: {}".format(path, repr(raw_path)),
```
lovelydinosaur: Is the `repr` needed here?
Could we pull this out into two lines maybe:

```python
content = "path: {}\nraw_path: {}".format(path, raw_path)
response = Response(content, media_type="text/plain")
```

tests/protocols/test_http.py
```diff
@@ -527,6 +527,28 @@ def test_root_path(protocol_cls):
...
+            media_type="text/plain",
+        )
+        await response(scope, receive, send)
+
+    protocol = get_connected_protocol(app, protocol_cls, root_path="/app")
+    protocol.data_received(
+        b"\r\n".join([b"GET /one%2Ftwo HTTP/1.1", b"Host: example.org", b"", b""])
```
lovelydinosaur: Let's have a `GET_REQUEST_WITH_RAW_PATH` constant or something similar, just to keep the style consistent with the surrounding test cases.

lovelydinosaur (changes requested):
Crackin' stuff! Suggested a few tiny tweaks.

lovelydinosaur:
Really nice to see that supporting this actually needs almost no extra implementation.

tests/protocols/test_http.py
simonw: I used the repr as a cheap-and-cheerful way of "proving" that the value was bytes, not a string while still being able to run assertions against the unicode string content for the response.

How about I change this to be nested assertions instead?

tests/protocols/test_http.py
simonw: Shall do.

simonw:
The specification has now been updated to add `raw_path`: https://github.com/django/asgiref/pull/92

tests/protocols/test_http.py
lovelydinosaur: Sure thing

tests/protocols/test_websocket.py
simonw: Done: https://github.com/encode/uvicorn/pull/372/commits/4b61f2fbd97a469a25b790a3eb706f923fe9bcfe

tests/protocols/test_http.py
simonw: Done: https://github.com/encode/uvicorn/pull/372/commits/1c88abc91091a24fdc427ea8a9e973684f4c8e63

tests/protocols/test_http.py
simonw: Done: https://github.com/encode/uvicorn/pull/372/commits/1c88abc91091a24fdc427ea8a9e973684f4c8e63

tests/protocols/test_http.py
simonw: Fixed: https://github.com/encode/uvicorn/pull/372/commits/6893373768a4e296e920ef44c9a4da93749a064f

lovelydinosaur (approved):

lovelydinosaur:
Released as 0.8.0

PR #373, opened 2019-06-17
Fix using wsgi with unix socket results TypeError

silane:
Using wsgi interface with unix socket results TypeError.
It is caused by WSGIMiddleware does not expect scope['server'] or scope['client'] to be None.
So I fixed that.


lovelydinosaur:
Thanks!

PR #332, opened 2019-03-25
Fix: don't error if WS still connected during shutdown

JoseKilo:
Close: #326 

I started to work on the issues reported on #326, but I ended up making many changes.
Happy to discuss / split any changes.

To explain the original issues:
- using wsproto: the variable `code` is not defined in the `shutdown` method (and not covered by tests).
- using websockets: the ws implementation keeps trying to send messages if requested by the App even when the connection is already closed.

Issue #326, opened 2019-03-12
Server shutdown errors if WebSocket connection is still open


JoseKilo:
If I stop the server while a WS connection is still open, I get different issues depending on the ws protocol.

Test app:
```python
import asyncio

from starlette.websockets import WebSocket


class App:
    def __init__(self, scope):
        assert scope['type'] == 'websocket'
        self.scope = scope

    async def __call__(self, receive, send):
        websocket = WebSocket(self.scope, receive=receive, send=send)

        await websocket.accept()
        while True:
            await websocket.send_text('Hello, world!')
            await asyncio.sleep(1)
        await websocket.close()
```

Test client:
```python
import asyncio

import websockets
import websockets.exceptions
import websockets.protocol


async def recv(websocket):
    while websocket.state == websockets.protocol.State.OPEN:
        try:
            message = await websocket.recv()
        except websockets.exceptions.ConnectionClosed as exc:
            print('ConnectionClosed with code:', exc.code)
        else:
            print('Received message:', message)


async def main(uri):
    async with websockets.connect(uri) as websocket:
        try:
            await recv(websocket)
        finally:
            await websocket.close()


if __name__ == "__main__":
    loop = asyncio.get_event_loop()
    loop.run_until_complete(main("ws://127.0.0.1:8000"))
```

With `wsproto`
```
(.venv) > uvicorn app:App --ws wsproto
INFO: Started server process [25393]
INFO: Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO: ('127.0.0.1', 51014) - "WebSocket /" [accepted]
^CINFO: Shutting down
Traceback (most recent call last):
  File "/home/jose/workspace/uvicorn_test/.venv/bin/uvicorn", line 9, in <module>
    load_entry_point('uvicorn==0.5.2', 'console_scripts', 'uvicorn')()
  File "/home/jose/workspace/uvicorn_test/.venv/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/jose/workspace/uvicorn_test/.venv/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/jose/workspace/uvicorn_test/.venv/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #378, opened 2019-06-21
Fix: App sends message after client closes the connection

JoseKilo:
This is a subset of #332, sorry for the delay.

An example to reproduce the problem here (I'm using Python3.6):

Server:
```python
    while True:
        print('Before')
        await send({"type": "websocket.send", "text": "123"})
        print('After')
        await asyncio.sleep(0.3)
```

Client:
```python
async def main(uri):
    async with websockets.connect(uri) as websocket:
        await asyncio.sleep(2)
        await websocket.close()
```

The biggest issue here is that the two implementations (wsproto/websocket) behave differently:
- Websockets will print a stack trace: `ERROR: Exception in ASGI application`
- Wsproto doesn't finish the task, the app loop keeps running and printing to stdout

### DIFF
--- a/uvicorn/protocols/websockets/websockets_impl.py
+++ b/uvicorn/protocols/websockets/websockets_impl.py
@@ -40,6 +40,7 @@ class WebSocketProtocol(websockets.WebSocketServerProtocol):
         self.server = None
         self.client = None
         self.scheme = None
+        self.task = None
 
         # Connection events
         self.scope = None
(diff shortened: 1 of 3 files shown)


JoseKilo:
Hi @tomchristie , any chance you could provide some feedback on this one ?

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -66,6 +67,8 @@ def connection_made(self, transport):
     def connection_lost(self, exc):
         self.connections.remove(self)
         self.handshake_completed_event.set()
+        if self.task is not None:
+            self.task.cancel()
```
lovelydinosaur, line 71: I'm not sure we should just cancel the task here.

The web application needs to be notified via ASGI that the socket has shutdown, but it does need to be able to run anything else to completion without having the task cancelled at an arbitrary point in it's execution.

I reckon we'll want to think about some of this in more detail.

euri10:
Obsolete with master see https://github.com/encode/uvicorn/issues/326#issuecomment-743741982

PR #381, opened 2019-06-22
Fixed uvicorn.run() example to use host= and port=

simonw:
The previous example gave me the following error:

    TypeError: run() takes 1 positional argument but 3 were given


lovelydinosaur:
Thanks!

PR #383, opened 2019-06-26
Support AF_INET6 address family

### LINKED ISSUE #313: Uvicorn can't get client's ip address, and port when listening to ipv6 address
I have an app, which I need to bind to ipv6 host, so my run.py script looks like this

```
import uvicorn
from app import create_app


if __name__ == '__main__':
    uvicorn.run(create_app(), port=8001, host='::')
```
Also I need access logs from this app. If I look in my logs with this setup, I can see messages like this:
`INFO: None - "GET /docs HTTP/1.1" 200`

So, where client info should be seen (his ip address and port), there is just None value. If I use, for example, hypercorn with command like this
`hypercorn app.main:app --access-log - -b :::8001
`
In logs I can see client's ip address and port
`::1:65427 GET /docs 1.1 200 839 5501`

Is it possible to somehow support ipv6 clien't ip addresses in uvicorn?

Yolley:
Closes https://github.com/encode/uvicorn/issues/313 . 

As stated here https://docs.python.org/2/library/socket.html#socket.AF_INET6, "For AF_INET6 address family, a four-tuple (host, port, flowinfo, scopeid) is used", so there should be no possible problems with type conversion on row 12 of this PR. So I think, that there is no need for additional tests, already existing should be enough to cover this.

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -88,7 +88,7 @@ class MockTransport:
             "sockname": self.sockname,
             "peername": self.peername,
             "sslcontext": self.sslcontext,
-        }[key]
+        }.get(key)
 
     def write(self, data):
         assert not self.closed
--- a/tests/protocols/test_utils.py
+++ b/tests/protocols/test_utils.py
@@ -27,10 +27,14 @@ def test_get_local_addr_with_socket():
     transport = MockTransport({"socket": MockSocket(family=socket.AF_IPX)})
     assert get_local_addr(transport) == None
 
-    transport = MockTransport({"socket": MockSocket(family=socket.AF_INET6, sockname=["::1", 123])})
+    transport = MockTransport(
+        {"socket": MockSocket(family=socket.AF_INET6, sockname=["::1", 123])}
+    )
     assert get_local_addr(transport) == ("::1", 123)
 
-    transport = MockTransport({"socket": MockSocket(family=socket.AF_INET, sockname=["123.45.6.7", 123])})
+    transport = MockTransport(
+        {"socket": MockSocket(family=socket.AF_INET, sockname=["123.45.6.7", 123])}
+    )
     assert get_local_addr(transport) == ("123.45.6.7", 123)
 
 
@@ -38,10 +42,14 @@ def test_get_remote_addr_with_socket():
     transport = MockTransport({"socket": MockSocket(family=socket.AF_IPX)})
     assert get_remote_addr(transport) == None
 
-    transport = MockTransport({"socket": MockSocket(family=socket.AF_INET6, peername=["::1", 123])})
+    transport = MockTransport(
+        {"socket": MockSocket(family=socket.AF_INET6, peername=["::1", 123])}
+    )
     assert get_remote_addr(transport) == ("::1", 123)
 
-    transport = MockTransport({"socket": MockSocket(family=socket.AF_INET, peername=["123.45.6.7", 123])})
+    transport = MockTransport(
+        {"socket": MockSocket(family=socket.AF_INET, peername=["123.45.6.7", 123])}
+    )
     assert get_remote_addr(transport) == ("123.45.6.7", 123)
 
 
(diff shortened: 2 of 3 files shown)


lovelydinosaur:
Gotcha. Test cases currently failing, when looking up the “socket” extra info - not looked into it beyond that yet.

Yolley:
> Gotcha. Test cases currently failing, when looking up the “socket” extra info - not looked into it beyond that yet.

It seems to me, that MockTransport in tests misses 'socket' attribute. Here is a documentation for BaseTransport (which is being mocked in tests, as i understand) https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseTransport.get_extra_info - 'socket' is also an attribute among with sockname, peername and sslcontext . I guess, we need to fix MockTransport in tests. Should we do it in this pr or create another?

lovelydinosaur:
Fixing it in this PR would be fine. However we ought to be a bit cautious about which items might or might not be present with get_extra_info. If we can fallback gracefully when particular items are not included then we should do that.

Yolley:
> Fixing it in this PR would be fine. However we ought to be a bit cautious about which items might or might not be present with get_extra_info. If we can fallback gracefully when particular items are not included then we should do that.

Like, first try to get 'socket', then, if it is not present, just return peername/sockname?

lovelydinosaur:
@Yolley Exactly, yes. I'd also suggest leaving the existing layout of the `get_local_addr` and `get_remote_addr` functions in place so that the change has a minimal-as-possible footprint.

```python
def get_local_addr(transport):
    info = transport.get_extra_info("socket").  # Only these lines changed 
    if info is not None:                        # Only these lines changed 
        return info.getsockname()               # Only these lines changed 
    info = transport.get_extra_info("sockname")
    if info is not None and isinstance(info, (list, tuple)) and len(info) == 2:
    return None
```

Yolley:
> @Yolley Exactly, yes. I'd also suggest leaving the existing layout of the `get_local_addr` and `get_remote_addr` functions in place so that the change has a minimal-as-possible footprint.
> 
> ```python
> def get_local_addr(transport):
>     info = transport.get_extra_info("socket").  # Only these lines changed 
>     if info is not None:                        # Only these lines changed 
>         return info.getsockname()               # Only these lines changed 
>     info = transport.get_extra_info("sockname")
>     if info is not None and isinstance(info, (list, tuple)) and len(info) == 2:
>         return (str(info[0]), int(info[1]))
>     return None
> ```

I made changes in this PR like you suggested, but we still need to make some changes in MockTransport, we need to change logic of method `get_extra_info`. In documentation https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseTransport.get_extra_info it is stated, that `get_extra_info` should return `default`, if `name` is not present in transport. So, I guess we need (for example, here https://github.com/encode/uvicorn/blob/master/tests/protocols/test_http.py#L86) to do something like this 

```python
    def get_extra_info(self, key):
        return {
            "sockname": self.sockname,
            "peername": self.peername,
            "sslcontext": self.sslcontext,
        }.get(key)
```

lovelydinosaur:
> we still need to make some changes in MockTransport, we need to change logic of method get_extra_info

Yup, your suggestion there looks correct to me.

Yolley:
> > we still need to make some changes in MockTransport, we need to change logic of method get_extra_info
> 
> Yup, your suggestion there looks correct to me.

So, i fixed MockTransport, now it uses `get` on dictionary in get_extra_info. Also I added 2 new tests to test `socket` logic in utils. In travis I see, that it suggests some reformat for test_utils.py, but I don't understand, what needs to be reformatted exactly.

lovelydinosaur:
Running `./scripts/lint` will apply the correct formatting to the codebase.

Yolley:
> Running `./scripts/lint` will apply the correct formatting to the codebase.

Yep, it seems, that black follows pep8 line of length (which is questionable in 2019, but it is still an approved standart, so). Reformatted test_utils.py

lovelydinosaur:
Released as 0.8.3. Good stuff! 👍

Issue #382, opened 2019-06-25
Include query_string in access logs


simonw:
Uvicorn doesn't currently include the query_string in the logging output.

I need the query_string for my own projects - to analyze what people are searching for on the site for example.

Is there a reason not to include the query_string? Assuming there is, what's the most elegant way for me to reconfigure Uvicorn in my own projects to include this?

lovelydinosaur:
I guess it was either just a stylistic choice, or didn't get considered either ways around.
I'd probably be happy to have it added in there.
We *might* want to have a look at configurable access log styles. (?)

simonw:
Pull request: https://github.com/encode/uvicorn/pull/384

lovelydinosaur:
Resolved in #384

PR #384, opened 2019-07-04
Include query_string in logs

simonw:
Refs #382

### DIFF
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -427,11 +427,19 @@ class RequestResponseCycle:
             headers = self.default_headers + list(message.get("headers", []))
 
             if self.access_log:
+                path_and_query_string = (
+                    self.scope.get("root_path", "") + self.scope["path"]
+                )
+                if self.scope["query_string"]:
+                    path_and_query_string = "{}?{}".format(
+                        path_and_query_string,
+                        self.scope["query_string"].decode("ascii"),
+                    )
                 self.logger.info(
                     '%s - "%s %s HTTP/%s" %d',
                     self.scope["client"],
                     self.scope["method"],
-                    self.scope.get("root_path", "") + self.scope["path"],
+                    path_and_query_string,
                     self.scope["http_version"],
                     status_code,
                 )
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -5,7 +5,12 @@ import urllib
 
 import httptools
 
-from uvicorn.protocols.utils import get_local_addr, get_remote_addr, is_ssl
+from uvicorn.protocols.utils import (
+    get_local_addr,
+    get_remote_addr,
+    is_ssl,
+    get_path_with_query_string,
+)
 
 
 def _get_status_line(status_code):
@@ -429,19 +434,11 @@ class RequestResponseCycle:
             headers = self.default_headers + list(message.get("headers", []))
 
             if self.access_log:
-                path_and_query_string = (
-                    self.scope.get("root_path", "") + self.scope["path"]
-                )
-                if self.scope["query_string"]:
-                    path_and_query_string = "{}?{}".format(
-                        path_and_query_string,
-                        self.scope["query_string"].decode("ascii"),
-                    )
                 self.logger.info(
                     '%s - "%s %s HTTP/%s" %d',
                     self.scope["client"],
                     self.scope["method"],
-                    path_and_query_string,
+                    get_path_with_query_string(self.scope),
                     self.scope["http_version"],
                     status_code,
                 )
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -429,11 +429,19 @@ class RequestResponseCycle:
             headers = self.default_headers + list(message.get("headers", []))
 
             if self.access_log:
+                path_and_query_string = (
+                    self.scope.get("root_path", "") + self.scope["path"]
+                )
+                if self.scope["query_string"]:
+                    path_and_query_string = "{}?{}".format(
+                        path_and_query_string,
+                        self.scope["query_string"].decode("ascii"),
+                    )
                 self.logger.info(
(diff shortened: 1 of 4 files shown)


simonw:
I didn't see any existing unit tests for logging and I wasn't sure how best to add them.

Here's example output from my local development instance:

```
INFO: ('127.0.0.1', 62821) - "GET /fixtures/no_primary_key.json HTTP/1.1" 200
INFO: ('127.0.0.1', 62821) - "GET /fixtures/no_primary_key?_sort=content HTTP/1.1" 200
INFO: ('127.0.0.1', 62821) - "GET /-/static/app.css?5e037f HTTP/1.1" 200
INFO: ('127.0.0.1', 62824) - "GET /fixtures?sql=select+rowid%2C+%2A+from+no_primary_key+order+by+content+limit+101 HTTP/1.1" 200
```

simonw:
I'm a little bit puzzled by these test failures:
```
=================================== FAILURES ===================================
_____________________________ test_invalid_format ______________________________
    def test_invalid_format():
        with pytest.raises(ImportFromStringError) as exc:
            import_from_string("example:")
        expected = 'Import string "example:" must be in format "<module>:<attribute>".'
>       assert expected in str(exc)
E       assert 'Import string "example:" must be in format "<module>:<attribute>".' in '<ExceptionInfo ImportFromStringError tblen=2>'
E        +  where '<ExceptionInfo ImportFromStringError tblen=2>' = str(<ExceptionInfo ImportFromStringError tblen=2>)
tests/importer/test_importer.py:10: AssertionError
_____________________________ test_invalid_module ______________________________
    def test_invalid_module():
        with pytest.raises(ImportFromStringError) as exc:
            import_from_string("module_does_not_exist:myattr")
        expected = 'Could not import module "module_does_not_exist".'
>       assert expected in str(exc)
E       assert 'Could not import module "module_does_not_exist".' in '<ExceptionInfo ImportFromStringError tblen=2>'
E        +  where '<ExceptionInfo ImportFromStringError tblen=2>' = str(<ExceptionInfo ImportFromStringError tblen=2>)
tests/importer/test_importer.py:17: AssertionError
______________________________ test_invalid_attr _______________________________
    def test_invalid_attr():
        with pytest.raises(ImportFromStringError) as exc:
            import_from_string("tempfile:attr_does_not_exist")
        expected = 'Attribute "attr_does_not_exist" not found in module "tempfile".'
>       assert expected in str(exc)
E       assert 'Attribute "attr_does_not_exist" not found in module "tempfile".' in '<ExceptionInfo ImportFromStringError tblen=2>'
E        +  where '<ExceptionInfo ImportFromStringError tblen=2>' = str(<ExceptionInfo ImportFromStringError tblen=2>)
tests/importer/test_importer.py:24: AssertionError
```
I can't replicate them in my local environment.

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -429,11 +429,19 @@ def __init__(
             headers = self.default_headers + list(message.get("headers", []))
 
             if self.access_log:
+                path_and_query_string = (
```
simonw: I'm not delighted at this code being duplicated in both `httptools_impl.py` and `h11_impl.py` - maybe this should be a helper function in `uvicorn.protocols.utils`?

simonw: Fixed in https://github.com/encode/uvicorn/pull/384/commits/75c39c2ef41327e4569324498efaf76375c9a3c3 - I think this is better.

simonw:
Interesting thing I spotted while building the unit test: it looks like the ASGI specification makes it impossible to tell the difference between hits to `/` and hits to `/?` - both result in the same scope. I don't think it's a problem, just interesting.

lovelydinosaur:
Addressing the (unrelated) test failures you're seeing here: #385

lovelydinosaur:
And yes - ASGI can't see the '?' (WSGI either)

If I'd have remembered that at the time I guess I might(?) have advocated for `raw_path` to be both the path and query string, without any parsing or modification whatsoever.

lovelydinosaur:
Alrighy - looks like if you merge master then those tests oughta start passing again.

simonw:
Hmm... that fixed the previous failures but I now have this one, which is also puzzling to me:
```
___________________ test_request_logging[/?foo-H11Protocol] ___________________
protocol_cls = <class 'uvicorn.protocols.http.h11_impl.H11Protocol'>
path = '/?foo'
caplog = <_pytest.logging.LogCaptureFixture object at 0x0000021042B23390>
    @pytest.mark.parametrize("path", ["/", "/?foo", "/?foo=bar", "/?foo=bar&baz=1"])
    def test_request_logging(protocol_cls, path, caplog):
        get_request_with_query_string = b"\r\n".join(
            ["GET {} HTTP/1.1".format(path).encode("ascii"), b"Host: example.org", b"", b""]
        )
        caplog.set_level(logging.INFO, logger="uvicorn")
    
        protocol = get_connected_protocol(app, protocol_cls)
        protocol.data_received(get_request_with_query_string)
        protocol.loop.run_one()
    
>       assert '"GET {} HTTP/1.1" 200'.format(path) in caplog.records[0].message
E       assert '"GET /?foo HTTP/1.1" 200' in 'Task was destroyed but it is pending!\ntask: <Task pending coro=<WSGIResponder.sender() done, defined at C:\\Users\\t...icorn\\middleware\\wsgi.py:100> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x0000021042AC61F8>()]>>'
```

lovelydinosaur:
Strange - also it's only occurring on the windows build.
Could it be something odd to do with the interaction between the two parameterizations?
How's it run without the `path` parameterization decorator in place?

lovelydinosaur:
Or, eg. if you swap the two parameterization decorators around so that they're ordered the other way?

lovelydinosaur:
Or if you swap the second one so that it's only running `?foo`, which is the failing case?

simonw:
Swapping the order of the two parameterization decorators worked, great hunch there!

simonw:
Anything else we need to do before we can merge this?

lovelydinosaur:
Released! 0.8.4

Thanks 🙏

PR #395, opened 2019-07-28
[DOC] Resolve SyntaxError in usage 'streaming responses' section

nottheswimmer:


lovelydinosaur:
Thanks, yup!
We're a bit stuck just now since there's some unrelated build failures. See #396

gvbgduh:
also resolved in https://github.com/encode/uvicorn/pull/422

gvbgduh:
Fixed in https://github.com/encode/uvicorn/pull/422

Issue #102, opened 2018-06-29
Test coverage -> 100%


lovelydinosaur:
Yes please

Edit by @Kludex:
Report created on 12 July 2021:
| Name | Missing | PR |
| -------- | ---------- | ---- |
| tests/test_config.py | 12 | #1348 
| uvicorn/_handlers/http.py |  63-72, 83 |
| uvicorn/config.py                   | 15, 24-28, 250, 393-395, 406-408, 434-435, 441-443
| uvicorn/main.py |   39-48, 411-431, 435 | #1154
| uvicorn/middleware/proxy_headers.py |    44 | #1349
| uvicorn/protocols/http/flow_control.py |   25, 38-40, 44-45
| uvicorn/protocols/http/h11_impl.py |    124-125, 239-240, 260-261, 313, 319, 417, 420, 435
| uvicorn/protocols/http/httptools_impl.py |   124-125, 152-153, 157, 315, 321, 419, 422, 454, 456
| uvicorn/protocols/utils.py |   14-17 | #1132
| uvicorn/protocols/websockets/websockets_impl.py |   100-101
| uvicorn/protocols/websockets/wsproto_impl.py | 25, 75, 88, 98, 111, 113, 123, 129, 132-135, 167-178, 216-223
| uvicorn/server.py |   27-28, 64-66, 85, 89-91, 96-97, 112-128, 132-138, 142-151, 163-166, 185-186, 192, 205, 248-250, 254, 266, 277-280, 284-287, 296, 310-313
| uvicorn/subprocess.py                                    |   69-76 | #1122
| uvicorn/supervisors/basereload.py                 |   43-48, 86
| uvicorn/supervisors/watchgodreload.py          | 16, 59
TOTAL                                              |   96%

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/uvicorn/issues/102">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/102/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/uvicorn/issues/102/pledge.svg">
</picture>
</a>

Jitsusama:
Hello @tomchristie.

[...]

PR #399, opened 2019-08-07
Add config tests

cknv:
As mentioned in #102 there was a need for some more test coverage of the config class, which seems like a fairly lowing hanging fruit, possibly low enough that I could manage to pick it.

I used the latest codecov report of the class to see what lines was missing coverage and tried to target those as best I could. I went with the naive way of just asserting what the code returned, so it is not impossible (perhaps likely, even) that I am mistaken about the behaviour of the class.

Locally I have some failing tests in the `test_statreload.py` file, but as I have yet to make any changes to the code itself, I suspect there where there when I cloned the repo. I could have a look at what those are, but I have barely looked past the config class.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -65,3 +65,11 @@ def test_socket_bind():
     config.load()
 
     assert isinstance(config.bind_socket(), socket.socket)
+
+
+def test_ssl_config(certfile_and_keyfile):
+    certfile, keyfile = certfile_and_keyfile
+    config = Config(app=asgi_app, ssl_certfile=certfile.name, ssl_keyfile=keyfile.name)
+    config.load()
+
+    assert bool(config.is_ssl) is True
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -72,4 +72,4 @@ def test_ssl_config(certfile_and_keyfile):
     config = Config(app=asgi_app, ssl_certfile=certfile.name, ssl_keyfile=keyfile.name)
     config.load()
 
-    assert bool(config.is_ssl) is True
+    assert config.is_ssl is True
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -42,7 +42,7 @@ def test_proxy_headers():
 
 def test_app_unimportable():
     config = Config(app="no.such:app")
-    with pytest.raises(ModuleNotFoundError):
+    with pytest.raises(ImportError):
         config.load()
 
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -69,7 +69,7 @@ def test_socket_bind():
 
 def test_ssl_config(certfile_and_keyfile):
     certfile, keyfile = certfile_and_keyfile
-    config = Config(app=asgi_app, ssl_certfile=certfile.name, ssl_keyfile=keyfile.name)
+    config = Config(app=asgi_app, ssl_certfile=certfile, ssl_keyfile=keyfile)
     config.load()
 
     assert config.is_ssl is True
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,0 +1,67 @@
+import logging
+import socket
+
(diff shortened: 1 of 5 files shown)


tests/test_config.py
```diff
@@ -0,0 +1,75 @@
...
+
+    assert isinstance(config.bind_socket(), socket.socket)
+
+
+def test_ssl_config(certfile_and_keyfile):
+    certfile, keyfile = certfile_and_keyfile
+    config = Config(app=asgi_app, ssl_certfile=certfile.name, ssl_keyfile=keyfile.name)
+    config.load()
+
+    assert bool(config.is_ssl) is True
```
cknv: The `bool` call should be unnecessary here, but the `.is_ssl` property is defined as such:

```python
@property
def is_ssl(self) -> bool:
    return self.ssl_keyfile or self.ssl_certfile
```
Which of course evaluates as a boolean, but isn't _technically_ a boolean. I could move the bool coercion there if it's appropriate?

jlaine: I'd say fix the problem at the root, `is_ssl` should return an actual boolean, not a truthy / falsy value.

```python
@property
def is_ssl(self) -> bool:
    return bool(self.ssl_keyfile or self.ssl_certfile)
```

and the test should be:

```python
assert config.is_ssl is True
```

I think mypy usually catches these kinds of problems

cknv: Yeah, mypy would probably catch that.

I lean to actually fixing it too, but decided to keep it minimal at first and only add tests.

It's a tiny addition, so I should have time soon to make the change.

cknv:
I have rebased onto the latest master which should make the test suite pass.

cknv:
I see the previously failing tests was masking some actual problems, and have a bit more work cut out for me with regards to dealing with python 3.5 and the permissions of the certificates on windows.

cknv:
The cert + windows problems seems to stem from the fact that `NamedTemporaryFile` is not able to be re-opened on Windows NT and later (or so is my understanding of the [docs](https://docs.python.org/3.7/library/tempfile.html#tempfile.NamedTemporaryFile) at least) it _shouldn't_ be too hard to fix that too.

cknv:
The windows build is still failing, this time when getting `"/"` of `test_request_logging`.

Judging from the [logs on travis](https://travis-ci.org/encode/uvicorn/jobs/577496079#L169-L171), there is an extra task spawned onto the loop for some reason and the mock loop only runs a single coroutine for those tests. It started happening after the change from `NamedTemporaryFile` to using `tmp_path` to just build paths with.

I don't have a windows machine at hand to replicate this with, so if someone can see the connection between the changes I did and the test, I would be very happy to continue, but currently I am just hitting the wall here.

jlaine:
CI failures in `test_request_logging` may be a symptom of what I reported in #406 : this test is subject to "pollution" from exceptions raised by other tests. I'm starting to think that clearing the log capture at the beginning of this test would be a good idea.

cknv:
Oh yeah @jlaine that looks very similar, and indeed I can actually see a similar log line being spewed out when running the tests locally, although that does not make it fail for some reason (running Arch locally here with Python 3.7.4 and uvloop 0.13.0).
Could it be that _maybe_ its actually some stray asyncio tasks that have been left behind by a test only to be destroyed in another ? It feels a bit like wild speculation, but I cannot shake the thought.

jlaine:
I have submitted a tentative patch in #418. AFAICT the issue is that if the WSGI task raises an exception, the "sender" task (in charge of sending out response messages) is left dangling.

lovelydinosaur:
Okay, probably worth merging master into this PR, and seeing where that gets us, right?

cknv:
That seemed to do the trick. Very nice @jlaine!

Now that it is actually passing tests, we can

lovelydinosaur:
Lovely stuff!

PR #401, opened 2019-08-08
Added portable Uvicorn class

databasedav:
My use is specifically for loop extraction to other apps, but other functionality can be added.

Includes an example in the docs using `FastAPI` and `Faust`.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -135,6 +135,21 @@ if __name__ == "__main__":
     uvicorn.run(app, host="127.0.0.1", port=5000, log_level="info")
 ```
 
+#### Extracting Event Loop to Provide to other Applications
+
+```python
+from fastapi import FastAPI
+from uvicorn import Uvicorn
+from faust import App
+
+app = FastAPI()
+uvicorn = Uvicorn(app, ...)
+faust_app = App(..., loop=uvicorn.get_event_loop())
+
+if __name__ == '__main__':
+    uvicorn.run()
+```
+
 ### Running with Gunicorn
 
 [Gunicorn][gunicorn] is a mature, fully featured server and process manager.
--- a/docs/index.md
+++ b/docs/index.md
@@ -144,7 +144,7 @@ from faust import App
 
 app = FastAPI()
 uvicorn = Uvicorn(app, ...)
-faust_app = App(..., loop=uvicorn.get_event_loop())
+faust_app = App(..., loop=uvicorn.loop)
 
 if __name__ == '__main__':
     uvicorn.run()
--- a/uvicorn/__init__.py
+++ b/uvicorn/__init__.py
@@ -1,5 +1,5 @@
 from uvicorn.config import Config
-from uvicorn.main import Server, main, run
+from uvicorn.main import Server, main, run, Uvicorn
 
 __version__ = "0.8.6"
-__all__ = ["main", "run", "Config", "Server"]
+__all__ = ["main", "run", "Config", "Server", "Uvicorn"]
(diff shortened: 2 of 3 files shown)


florimondmanca:
Hi,

I quickly took a look at the code, and it looks like you're forcing the use of `asyncio` there. But uvicorn does not always run on `asyncio` — it may run on `asyncio` or `uvloop` (see [Settings - Implementation](https://www.uvicorn.org/settings/#implementation)).

I'm wondering, though — doesn't Faust support using whichever event loop is currently configured? I believe it should simply call `asyncio.get_event_loop()` when it needs to access it, instead of requiring to pass the loop upon instanciation. Perhaps this would be worth discussing on the Faust repo?

databasedav:
I'm not forcing the use of `asyncio` since `Config.setup_event_loop` sets the event loop policy to the one of your choosing. I haven't changed any behavior, just wrapped `run` a in class for a tiny bit of convenience really.

And yes, I believe `Faust` does do this and I can't figure out why (doesn't look like `Faust` creates a new event loop anywhere; it may be because `Uvicorn` does?) but in my testing simply doing
```python
import asyncio
import uvloop
from faust import App
import uvicorn

faust_app = App(..., loop=asyncio.get_event_loop())

...

if __name__ == '__main__':
    uvicorn.run(...)
```
leads to `RuntimeError: Task ... running at ... got Future <Future pending> attached to a different loop` errors, while extracting the event loop from the `Uvicorn` instance like I do in the docs does not.

florimondmanca:
Oh, yes, my bad, I didn’t realize uvloop is just a different loop that still allows using asyncio everywhere else.

lovelydinosaur:
Closing this in favor of https://github.com/encode/uvicorn/pull/455 which I think allows you the same functionality without us having to introduce much more API. Does that work okay for you?

databasedav:
This actually doesn't work because there's no way to extract the event loop prior to `server.run()` since the loop is created then. I'm not sure what you meant by "setup the loop yourself" here https://github.com/encode/uvicorn/pull/442#issuecomment-544928506 since the `loop` arg for `Config` takes a string for the type of loop but not the loop object itself. But I agree that there's no need to introduce more API then needed so I suggest one of the following.

- Allow `Config`'s `loop` arg to take an event loop object so we actually can setup the loop ourselves and pass it to whatever we want (which is the issue pointed out in https://github.com/encode/uvicorn/pull/442)
- scrap the `Uvicorn` class but keep my additions to `Server` from this PR, i.e. this stuff:
```python
class Server:
    def __init__(self, config):
        self.config = config
        self.server_state = ServerState()

        self.started = False
        self.should_exit = False
        self.force_exit = False
        self.last_notified = 0
        # begin stuff
        self.loop = None
        self.event_loop_setup = False
    
    def setup_event_loop(self):
        self.event_loop_setup = True

    def get_event_loop(self):
        if not self.event_loop_setup:
            self.setup_event_loop()
        if not self.loop:
            self.loop = asyncio.get_event_loop()
        return self.loop

    def run(self, sockets=None, shutdown_servers=True):
        loop = self.get_event_loop()
        # end stuff
```

I think the former may make more sense but is inconsistent with the current API. Please let me know what you think. Thanks!

lightoyou:
ReOpen it please

euri10:
> ReOpen it please

why ?

lightoyou:
Is it' working ? I try to use uvicorn (fastapi) and faust maybe. I need an example. Cause it's not working for me.

euri10:
something like this should, feel free to ask precise questions or join the chat to get help, it's not working is not helpful

```
config = Config(app=fastapi_app, loop=loop_instance)
server.run()
```

lightoyou:
Yeah i join the chat to get more help :)

PR #402, opened 2019-08-09
Add switch to disable the handling of signals

Tamles:
Sometime, you don't want uvicorn to handle SIGTERM signals. For example when you want to run uvicorn with asyncio.create_task. This change doesn't modify the default behavior of uvicorn.


lovelydinosaur:
We've dropped `shutdown_servers` now, since it didn't fit in with the rest of the public API.
I'd be okay with the functionality change this PR adds, but let's instead use the standard `Config` for it, and add a `config.handled_signals` which should default to `(signal.SIGINT, signal.SIGTERM)` but can be overridden.

PR #406, opened 2019-08-14
Raise uvloop version to 0.*

jlaine:
uvloop 0.13.0 was released a few hours ago. Although it's not *needed* for uvicorn, it's nice for projects which do require the updated uvloop (for instance for better UDP performance) in the same virtualenv.

An alternative would be to loosen the versioning scheme.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ requirements = [
     "h11==0.8.*",
     "websockets==7.*",
     "httptools==0.0.13 ;" + env_marker,
-    "uvloop==0.12.* ;" + env_marker,
+    "uvloop==0.* ;" + env_marker,
 ]
 
 


jlaine:
I'm not sure whether the CI breakage was actually caused by the PR or not

lovelydinosaur:
Want to merge master in, and see where we're at?

jlaine:
Somehow this broke Python 3.7-dev's tests

lovelydinosaur:
Lovely 😐

lovelydinosaur:
Will try to dig tomoz 🙃

jlaine:
I had a look and it looks like the failing test is not the root cause. If I print `caplog.records` at the very start of `test_request_logging` it already contains the warning about a pending task being destroyed.

Adding `caplog.records.clear()` at the top of `test_request_logging` works around the problem but it would be nice to know why we have this destroyed pending task.

UPDATE: the failure originates in the WSGI middleware tests, most likely the ones which trigger an exception

lovelydinosaur:
> the WSGI middleware tests, most likely the ones which trigger an exception

That might be a good call - yup.

lovelydinosaur:
> UPDATE: the failure originates in the WSGI middleware tests, most likely the ones which trigger an exception

Do you have a good way of isolating and reproducing that? I'd like to have a dig into resolving it.

jlaine:
I can reproduce the issue locally (Debian / Python 3.7 / uvloop 0.13.0) by running the full test suite: the wsgi middleware tests pass, but they produce a warning which trips up the `test_request_logging` test, which fails in the same way as:

https://travis-ci.org/encode/uvicorn/jobs/574436943#L300

EDIT: it's not 100% deterministic, sometimes the tests pass. When it fails the error looks like:

```
>       assert '"GET {} HTTP/1.1" 200'.format(path) in caplog.records[0].message
E       assert '"GET / HTTP/1.1" 200' in 'Task was destroyed but it is pending!\ntask: <Task pending coro=<WSGIResponder.sender() done, defined at /home/...orn/uvicorn/middleware/wsgi.py:100> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f097c0820a8>()]>>'
E        +  where '"GET / HTTP/1.1" 200' = <built-in method format of str object at 0x7f097ee6d468>('/')
E        +    where <built-in method format of str object at 0x7f097ee6d468> = '"GET {} HTTP/1.1" 200'.format
E        +  and   'Task was destroyed but it is pending!\ntask: <Task pending coro=<WSGIResponder.sender() done, defined at /home/...orn/uvicorn/middleware/wsgi.py:100> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f097c0820a8>()]>>' = <LogRecord: asyncio, 40, /home/..corn/env/lib/python3.7/site-packages/pluggy/callers.py, 17...rn/uvicorn/middleware/wsgi.py:100> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f097c0820a8>()]>>">.message
```
The culprit seems to be `test_wsgi_exception`
```

**Comment by jlaine** [comment]
Yay, #418 indeed fixed CI!

File: setup.py [hunk 1]
```diff
@@ -45,7 +45,7 @@ def get_packages(package):
     "h11==0.8.*",
     "websockets==7.*",
     "httptools==0.0.13 ;" + env_marker,
-    "uvloop==0.12.* ;" + env_marker,
```
lovelydinosaur, line 48: Let's go to `0.*` instead.

lovelydinosaur (changes requested):
I think we ought to switch to `0.*` instead.
I can't really see uvloop breaking their existing (very very minimal) API contract, without a 1.x version bump, so that seems like the best tradeoff to me, here.

jlaine:
It might be nice to have a Travis cron job to run CI periodically regardless of whether there are new commits to uvicorn, to catch an unexpected regression?

https://docs.travis-ci.com/user/cron-jobs/

lovelydinosaur:
Interesting possibility, yeah.
Not sure how we'd work that into a notification of some kind, tho? (So that it's visible when it starts failing.)

jlaine:
uvloop does periodic CI runs using a cron, you can see the builds marked as "CRON":

https://travis-ci.org/MagicStack/uvloop/builds

Not sure if @1st1 receives notifications when the build fails?

lovelydinosaur (approved):

lovelydinosaur:
Let's get this in.
Thanks!

PR #407, opened 2019-08-16
Run Travis against Python 3.7 release

jlaine:


jlaine:
OK, well this bombed on Python 3.8, so I'll change the PR to only switch to Python 3.7 final.

jlaine:
Any thoughts on this @tomchristie ? I don't think there's a valid reason to run our tests against a dev snapshot of Python 3.7

lovelydinosaur:
Thanks, yup!

PR #409, opened 2019-08-19
Parse WSGI ThreadPoolExecuter worker amount via parameter

tomnaumann:
I recently recognized that the ThreadPoolExecuter worker count for the WSGI-Middleware is hard set to 10, which means even though the worker_connections are set to e.g. 1000 the ThreadPoolExecuter limits my application to 10 sub threads per worker.
By parsing a new parameter into the config and furthermore to the WSGI-Middleware we can gain some more flexibility here, that every developer can decide how high the limit should be by himself.
This could become interesting for heavy I/O bounded functions.

This could look something like this ..

    from uvicorn.workers import UvicornWorker

    class MyAwesomeUvicornWorker(UvicornWorker):
        CONFIG_KWARGS = {"thread_pool_worker": 50}

.. and then call ..

    gunicorn -k MyAwesomeUvicornWorker

_Please correct me if I said something stupid or wrong :D_

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -89,6 +89,7 @@ class Config:
         workers=1,
         proxy_headers=False,
         root_path="",
+        thread_pool_worker=10,
         limit_concurrency=None,
         limit_max_requests=None,
         timeout_keep_alive=5,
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Some thoughts:

* We should automatically tie it in with gunicorn's [threads](https://docs.gunicorn.org/en/stable/settings.html#threads) config.
* I think we should place it just below "workers" in the config section.
* I'd ideally like this to raise error if it's used on anything except a wsgi interface.
* Needs adding to the documentation.

Kludex:
This can easily be done. I'm not sure if we want to do it, tho.

Considering that it's an ancient PR, and I haven't seen people interested on this, I'll close it.

PR #416, opened 2019-08-30
Fix exception on TCP liveness check

paulhodson:
The following would cause an exception to be raised:
  nc -z -v -w 3 <host> <port>

### DIFF
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -137,7 +137,8 @@ class HttpToolsProtocol(asyncio.Protocol):
         if self.cycle and not self.cycle.response_complete:
             self.cycle.disconnected = True
         self.message_event.set()
-        self.flow.resume_writing()
+        if self.flow is not None:
+            self.flow.resume_writing()
 
     def eof_received(self):
         pass


uvicorn/protocols/http/httptools_impl.py
```diff
@@ -137,7 +137,8 @@ def connection_lost(self, exc):
         if self.cycle and not self.cycle.response_complete:
             self.cycle.disconnected = True
         self.message_event.set()
-        self.flow.resume_writing()
+        if self.flow:
+            self.flow.resume_writing()
```
lovelydinosaur, line 141: I'm a bit confused what this resolves. It looks like it could only fail if `self.flow` is None, but it's not clear how that could possibly occur, since surely `connection_made` is always called prior to `connection lost`?

What's the traceback of the error that occurs?

lovelydinosaur: I'm not neccessarily against this (although I think we should use `if self.flow is not None`) since it'd be more strictly correct from a type-checked point of view. I'd just like to understand the issue better first.

paulhodson: Here is the traceback on invocation of the above command:
```
ERROR: Exception in callback <bound method HttpToolsProtocol.connection_lost of <uvicorn.protocols.http.httptools_impl.HttpToolsProtocol object at 0x7fe275424710>>
handle: <Handle HttpToolsProtocol.connection_lost>
  File "uvloop/cbhandles.pyx", line 68, in uvloop.loop.Handle._run
  File "/usr/local/lib/python3.6/site-packages/uvicorn/protocols/http/httptools_impl.py", line 140, in connection_lost
    self.flow.resume_writing()
AttributeError: 'NoneType' object has no attribute 'resume_writing'
```

The `nc` call is not sending any traffic, just checking that the port is open.

lovelydinosaur: Okay, tho let's change the branch statement to this:

```python
if self.flow is not None:
```

lovelydinosaur (changes requested):
Happy to have this in, with one minor change as described.

paulhodson:
@tomchristie Can you please re-review? I probably shouldn't have force-pushed an update addressing the comment but, having done so, there may not have been an automated notification. Thanks, and sorry for the trouble.

gvbgduh (approved):

lovelydinosaur:
Have made the same change in the `h11` implementation too.

PR #418, opened 2019-09-04
If the WSGI task blows up, allow the sender task to shut down

jlaine:
This should hopefully fix the CI failures which were observed in PRs #399 and #406.

I have run the test suite 100 times and am no longer observing any failures.

Still, I'm not 100% confident I'm doing the right thing, so comments are welcome.


lovelydinosaur:
This looks really sensible yup, let's get it in!

PR #419, opened 2019-09-04
Run Travis against Python 3.8-dev

jlaine:


jlaine:
uvloop fails to build for Python 3.8-dev, looks like a Cython version thing. I'll close for now and we can revisit this at the next uvloop release.

PR #420, opened 2019-09-04
Update 'websockets' dependency to 8.x

### LINKED ISSUE #412: Update 'websockets' to 8.x
I realize that https://github.com/encode/uvicorn/pull/396 was done because the tests appear to be failing with `websockets>=8.0`, but that release also fixed a bunch of DeprecationWarnings that we've been seeing in [httpx builds on 3.8](https://travis-ci.org/encode/httpx/jobs/574949180).

Since it's only a `DeprecationWarning`, and only on 3.8 ([3.7 and below is fine](https://travis-ci.org/encode/httpx/jobs/574949179)), I definitely would agree that this is a low priority issue.

I dug into the tests to try and figure out what exactly was going wrong, but the `AttributeError`s I looked at looks like the main issue is two missing functions in [websockets_impl.py](https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/websockets/websockets_impl.py)

`15 AttributeError: 'Server' object has no attribute 'is_serving'` --> https://github.com/aaugustin/websockets/blob/8.0.2/src/websockets/server.py#L642-L668
` 9 AttributeError: 'WebSocketProtocol' object has no attribute 'transfer_data_task'` --> https://github.com/aaugustin/websockets/blob/8.0.2/src/websockets/protocol.py#L276-L313

lovelydinosaur:
Closes #412

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
     - if [ "$TRAVIS_OS_NAME" = "windows" ]; then
         choco install python3;
         export PATH=/c/Python37:/c/Python37/Scripts:/c/Python38:/c/Python38/Scripts:$PATH;
-        python -m pip install -U click h11 wsproto==0.13.* websockets==7.*;
+        python -m pip install -U click h11 wsproto==0.13.* websockets==8.*;
         python -m pip install -U autoflake black codecov isort pytest pytest-cov requests;
       elif [ "$TRAVIS_PYTHON_VERSION" = "pypy3.5" ]; then
         pip install -U click h11 wsproto==0.13.*;
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,6 @@ matrix:
       python: "pypy3.5"
     # The main tests run on Linux for a all relevant Python versions.
     - os: linux
-      python: "3.5"
-    - os: linux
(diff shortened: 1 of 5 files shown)


lovelydinosaur:
Okay, there's a call to be made here. `websockets 8.x` drops support for Python 3.5.

We need to figure out if we're okay with that.

There is an alternative, which would be to make it possible to install just the base `uvicorn` package without dependencies, and have our tests omit `websockets` for 3.5, but that's all a bit fiddly, and it's also a big jump (I'd kinda been planning something a bit similar to that for the 1.0 release.)

lovelydinosaur:
I think we're gonna have to accept this, update to a 0.9 release, and document that if folks want Python 3.5 support, then they need to install 0.8.

3.5 EOL is [only a year away now](https://devguide.python.org/#status-of-python-branches), so that doesn't seem unreasonable. (Tho I know there's some install environments, eg. raspbian, glitch, that are still defaulting to 3.5 right now)

StephenBrown2:
Woo! Thanks for this. I thought it would be far more involved than that. Now we just need a release (when it's time).

lovelydinosaur:
Done. 0.9

PR #423, opened 2019-09-13
use IOCP in windows

abersheeran:
https://docs.python.org/3/library/asyncio-policy.html#asyncio.WindowsProactorEventLoopPolicy

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -32,7 +32,7 @@ equivalent keyword arguments, eg. `uvicorn.run(app, port=5000, debug=True, acces
 
 ## Implementation
 
-* `--loop <str>` - Set the event loop implementation. The uvloop implementation provides greater performance, but is not compatible with Windows or PyPy. **Options:** *'auto', 'asyncio', 'uvloop'.* **Default:** *'auto'*.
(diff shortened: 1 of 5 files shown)


gvbgduh (changes requested):
It might be not a bad idea to bring another implementation of the event loop. But in this implementation, it implicitly replaces the asyncio event loop. It may confuse the user it's generally better to keep things explicit.
So for this particular case, it would be better to create a separate module, like `uvloop`, so it would be possible to specify it with the `--loop` param running the uvicorn server.

abersheeran:
I agree it. I will make changes later and resubmit.

gvbgduh:
@abersheeran looks really good to me. Let's run it with @tomchristie what he thinks.

lovelydinosaur:
Liking it, yup!

abersheeran:
So, why not merge the PR? What else do I need to modify?

gvbgduh:
hey @abersheeran, I'm going to merge it today, but if you could run `./scripts/lint` and then push again it would save some time.

gvbgduh (approved):

gvbgduh:
Actually, never mind!

abersheeran:
I am sorry that I didn't notice lint, I will use it next time. 😅

gvbgduh:
all good, thanks for the PR!

PR #424, opened 2019-09-13
Bump tested PyPy3 to 3.6

johnthagen:
Travis recently added PyPy3-3.6 to the available Python languages: https://docs.travis-ci.com/user/languages/python/#pypy-support

As of writing this PR,

> `"pypy3"  # currently Python 3.6.1,  PyPy 7.1.1-beta0`

Since the docs say that `uvicorn` requires Python 3.6+, I presume this would be preferable.

The Travis config comments say:

```
    # The pypy3 job runs a fully pure Python setup, which triggers a part of the
    # auto-detect tests that the other tests don't. We cannot test websockets though.
```

Does bumping to 3.6 allow PyPy3 to test websockets? Can anything else be enabled by bumping to PyPy3-3.6?


scripts/test
```diff
@@ -15,7 +15,8 @@ ${PREFIX}coverage html
 ${PREFIX}autoflake --recursive uvicorn tests
 if [ "${PYTHON_VERSION}" = '3.7' ]; then
     echo "Skipping 'black' on 3.7. See issue https://github.com/ambv/black/issues/494"
-elif [ "${PYTHON_VERSION}" = '3.6' ]; then
+# black triggers known bug in PyPy3-3.6-7.1.1: https://bitbucket.org/pypy/pypy/issues/2985
+elif [ "${PYTHON_VERSION}" = '3.6' ] && [ "${TRAVIS_PYTHON_VERSION}" != 'pypy3' ]; then
```
lovelydinosaur: Since we've dropped 3.5, and black runs fine on 3.7 now, I think we can simplify all of this block right down to just `if [ "${TRAVIS_PYTHON_VERSION}" != 'pypy3' ]; then ...skip black...; else ... ; fi`

johnthagen:
@tomchristie I have resolved your suggestions

lovelydinosaur:
Thanks!

PR #433, opened 2019-09-22
fix: unix socket support multi workers

### LINKED ISSUE #428: UNIX domain socket raises 503 when workers are used
Is it Ok when a combination of `--workers ` and `--uds` raises 503 error via nginx? For example:
```
--workers 4 --uds /var/run/uvicorn-test.sock
```

it works well when `--workers` is omitted (1 worker). Please note that `gunicorn` works well via Unix socket and multiple workers.

Thanks!

Hanaasagi:
- resolve #428 .
- fix `None` remote addr in access log.

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -271,10 +271,18 @@ def run(app, **kwargs):
         sock = config.bind_socket()
         supervisor = StatReload(config)
         supervisor.run(server.run, sockets=[sock])
+        sock.close()
     elif config.workers > 1:
-        sock = config.bind_socket()
         supervisor = Multiprocess(config)
-        supervisor.run(server.run, sockets=[sock])
+        if config.uds is not None:
+            sock = config.bind_unix_socket()
+            supervisor.run(server.run, sockets=[sock])
+            sock.close()
+            os.remove(config.uds)
+        else:
+            sock = config.bind_socket()
+            supervisor.run(server.run, sockets=[sock])
+            sock.close()
     else:
         server.run()
 
--- a/uvicorn/protocols/utils.py
+++ b/uvicorn/protocols/utils.py
@@ -8,11 +8,11 @@ def get_remote_addr(transport):
         family = socket_info.family
         if family in (socket.AF_INET, socket.AF_INET6):
             return (str(info[0]), int(info[1]))
-        return None
+        return "-"
     info = transport.get_extra_info("peername")
     if info is not None and isinstance(info, (list, tuple)) and len(info) == 2:
         return (str(info[0]), int(info[1]))
-    return None
+    return "-"
 
 
 def get_local_addr(transport):
--- a/tests/protocols/test_utils.py
+++ b/tests/protocols/test_utils.py
@@ -40,7 +40,7 @@ def test_get_local_addr_with_socket():
 
 def test_get_remote_addr_with_socket():
     transport = MockTransport({"socket": MockSocket(family=socket.AF_IPX)})
-    assert get_remote_addr(transport) == None
+    assert get_remote_addr(transport) == "-"
 
     transport = MockTransport(
         {"socket": MockSocket(family=socket.AF_INET6, peername=["::1", 123])}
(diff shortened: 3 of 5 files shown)


uvicorn/main.py
```diff
@@ -271,10 +271,18 @@ def run(app, **kwargs):
...
+        sock.close()
     elif config.workers > 1:
-        sock = config.bind_socket()
         supervisor = Multiprocess(config)
-        supervisor.run(server.run, sockets=[sock])
+        if config.uds is not None:
+            sock = config.bind_unix_socket()
+            supervisor.run(server.run, sockets=[sock])
+            sock.close()
+            os.remove(config.uds)
```
lovelydinosaur, line 281: Curious: Why do we remove the UDS socket here?

uvicorn/protocols/utils.py
```diff
@@ -8,11 +8,11 @@ def get_remote_addr(transport):
         family = socket_info.family
         if family in (socket.AF_INET, socket.AF_INET6):
             return (str(info[0]), int(info[1]))
-        return None
+        return "-"
     info = transport.get_extra_info("peername")
     if info is not None and isinstance(info, (list, tuple)) and len(info) == 2:
         return (str(info[0]), int(info[1]))
-    return None
```
lovelydinosaur, line 15: Is there a reason why we're changing the `None` case here?

uvicorn/main.py
Hanaasagi: From `man 7 Unix`

> Binding  to  a  socket with a filename creates a socket in the filesystem that must be deleted by the caller when it is no longer needed (using unlink(2)).  The usual UNIX close-behind semantics apply; the socket can be unlinked at any time and will be finally removed from the filesystem when the last reference to it is closed.

I think there is another problem. If our process crashes, `os.remove()` will not be called and we’ll have a dangling socket. It seems that we need to check the socket file before `bind()`.


----
Why haven’t encountered this problem when use one worker and unix socket mode?

Because we use the `create_unix_server`

https://github.com/encode/uvicorn/blob/e0a77de6baf436d182868fa280cd5e2f0d66501c/uvicorn/main.py#L363-L373

`create_unix_server` will unlink the socket file before it starts.

https://github.com/python/cpython/blob/0d3fe8ae4961bf551e7d5e42559e2ede1a08fd7c/Lib/asyncio/unix_events.py#L283-L293

uvicorn/protocols/utils.py
Hanaasagi: `None` was displayed in access log.

```
2019-10-08 15:47:52,870 - INFO - None - "GET / HTTP/1.1" 200
2019-10-08 15:48:00,546 - INFO - None - "GET / HTTP/1.1" 200
2019-10-08 15:48:01,285 - INFO - None - "GET / HTTP/1.1" 200
```

Or we can fix here

```diff
diff --git a/uvicorn/protocols/http/httptools_impl.py b/uvicorn/protocols/http/httptools_impl.py
index 2e1dd90..8b414b2 100644
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -436,7 +436,7 @@ class RequestResponseCycle:
             if self.access_log:
                 self.logger.info(
                     '%s - "%s %s HTTP/%s" %d',
-                    self.scope["client"],
+                    self.scope["client"] or "-",
                     self.scope["method"],
                     get_path_with_query_string(self.scope),
                     self.scope["http_version"],
```

Any good idea, thanks?

jcwilson:
I just bumped into this today, as well. I can use TCP for now, but would love to help out here if I can.

gisthere:
really needed

Hanaasagi:
Emm, I will fix the conflicts recently.

Roang-zero1:
Both the issue I had #586 and the Issue #722 could also be solved by using the fix proposed in this pull request.
By moving the check to config.bind_socket as proposed in #722 any further additions to watcher types would also be agnostic to the type of socket used.
I am willing to work on refactoring as needed.

Kludex:
I'll close this PR, as #1009 addressed the issue that motivated it.

PR #436, opened 2019-09-24
fix: handle gunicorn max_requests config

### LINKED ISSUE #426: --limit-max-requests not working
Hi! I'm trying to figure out why my workers are not restarting as expected when using the `--limit-max-requests` flag. I ran `uvicorn` in debug mode and noticed that the `self.server_state.total_requests` count is not increasing (stays at 0 after each request) so `self.server_state.total_requests >= self.config.limit_max_requests` never returns `True`. 
When looking into where the `total_requests` was used, I noticed that the `protocols.http.[auto/h11/httptools]` were never getting called. I tried forcing the `--http h11` and `--http httptools` parameters, without any change in behavior. Any help would be appreciated!

Hanaasagi:
It will resolve #426 .

### DIFF
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -25,6 +25,7 @@ class UvicornWorker(Worker):
             "timeout_keep_alive": self.cfg.keepalive,
             "timeout_notify": self.timeout,
             "callback_notify": self.callback_notify,
+            "limit_max_requests": self.max_requests,
         }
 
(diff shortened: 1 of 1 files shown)


Hanaasagi:
Hey @tomchristie ,may I get a review :)

gvbgduh:
Awesome investigation @Hanaasagi! I think I can merge this one.

Issue #389, opened 2019-07-17
Gunicorn access logging is broken


nwalsh1995:
Hi all,

I originally found out about this bug which I was affected by here: https://github.com/tiangolo/uvicorn-gunicorn-starlette-docker/issues/1

# Description
When using Gunicorn deployed with `--access-logfile` set and using the `-k uvicorn.workers.UvicornWorker` option, no access logs are sent to the access log file. Note that access logs do get sent to error logger (`gunicorn.error`) but they are supposed to go to access logger (`gunicorn.access`)

# The Problem in Code
The issue is due to the following in `httptools_impl.py`:

https://github.com/encode/uvicorn/blob/15c4e9cfe2d996bd73ed6b542e44160f620b9519/uvicorn/protocols/http/httptools_impl.py#L436-L444

The problem is `self.logger` references a gunicorn logger: https://github.com/benoitc/gunicorn/blob/dc7b5d5c4876b49f86ea2460698a335d0f5ef7c9/gunicorn/glogging.py#L166 and the methods for logging there are only for the error logger (see the definition for the `info` method). 

To use the access logger it would have to be `self.logger.access_log.info(...)` instead. I've confirmed this by making those changes manually and running the same gunicorn config and it fixes it. I believe hardcoding that piece will break the logging done by running `uvicorn` directly so I have not attached a PR. I am still looking into it, but if anyone has a good fix please go ahead.

lovelydinosaur:
I assume to properly address this, `uvicorn` would need to be using two independently configured logger instances in the first place, in the same way as `gunicorn`, right?

So instead of `self.logger.info()`, we'd use `self.access_logs.info()` and `self.error_logs.info()`.

nwalsh1995:
Hi @tomchristie, yes, that is the same approach I was thinking and allows uvicorn to match gunicorn for easier compatibility.

kireet:
hello, is there an ETA for this fix? Also will the gunicorn [access log format](http://docs.gunicorn.org/en/stable/settings.html#access-log-format) setting be honored?

lovelydinosaur:
Okay - I've taken a pass at this: https://github.com/encode/uvicorn/pull/462

lovelydinosaur:
Okay, we still need to deal with the access log format, but we've got properly seperated access and error logs in master now.

Stigjb:
[...]

PR #440, opened 2019-09-26
Differentiate access_logger and error_logger

euri10:
Attempt at solving https://github.com/encode/uvicorn/issues/389 which bites me too
Honestly I'm not too sure about it and got a little bit overwhelmed by it, so I'd appreciate guidance on this, happy to rework it if necessary


LMolr:
Hello, is this under review?
This would fix #389 , which makes uvicorn unappealing for a production deployment.
Thank you !

uvicorn/lifespan/on.py
```diff
@@ -49,10 +50,10 @@ def __init__(self, config):
                 return
             if self.config.lifespan == "auto":
                 msg = "ASGI 'lifespan' protocol appears unsupported."
-                self.logger.info(msg)
+                self.access_logger.info(msg)
```
lovelydinosaur: I think we should use `error_logger.info(...)` for this one. (It's kinda a soft error)

uvicorn/middleware/message_logger.py
```diff
@@ -39,23 +40,27 @@ def __init__(self, app):
...
             log_text = "%s - ASGI [%d] Received %s"
-            self.logger.debug(log_text, client_addr, task_counter, logged_message)
+            self.access_logger.debug(
+                log_text, client_addr, task_counter, logged_message
+            )
             await send(message)
 
         log_text = "%s - ASGI [%d] Started"
-        self.logger.debug(log_text, client_addr, task_counter)
+        self.access_logger.debug(log_text, client_addr, task_counter)
```
lovelydinosaur, line 57: I *think* we probably want to reserve the access logger strictly for `startup/shutdown/request-response`, and put this kind of debug logging on the error logger. (?)

uvicorn/supervisors/statreload.py
```diff
@@ -30,9 +30,9 @@ def handle_fds(target, fd_stdin, **kwargs):
 
     def run(self, target, *args, **kwargs):
         pid = os.getpid()
-        logger = self.config.logger_instance
+        access_logger = self.config.access_logger_instance
 
-        logger.info("Started reloader process [{}]".format(pid))
+        access_logger.info("Started reloader process [{}]".format(pid))
```
lovelydinosaur, line 35: It'd be really useful to know which logger gunicorn uses for this kind of messaging.

lovelydinosaur:
Great stuff! There's a few cases where it's not super obvious which of the two cases we'd want to use - any obvious points of reference here (eg based on what gunicorn does?)

LMolr:
From a monitoring perspective it would be handy to have a uniform access log with just the http hits.
It would make the access log easy to parse e.g. in Splunk/ELK.
I don't know what the gunicorn rationale is..

lovelydinosaur:
> From a monitoring perspective it would be handy to have a uniform access log with just the http hits.

I'd be okay with that, yup. Seems pretty reasonable.

uvicorn/lifespan/on.py
euri10: yep I was wrong

uvicorn/middleware/message_logger.py
euri10: not sure I understand what you mean by`startup/shutdown/request-response`, is it in the context of this module or generally speaking ?.

Currently there are currently 5 message in this module, 4 are on access

1. `"%s - ASGI [%d] Sent %s"`
2. `"%s - ASGI [%d] Received %s"`
3. `"%s - ASGI [%d] Started"`
4. `"%s - ASGI [%d] Completed"`

the last one is on error_logger `"%s - ASGI [%d] Raised exception"`

From your comment I'm not sure if
A. you want all of them to be on the error_logger because it's meant for debugging logging
B. something else

uvicorn/supervisors/statreload.py
euri10: ok I dig a little
Gunicorn deals with reloading in [here](https://github.com/benoitc/gunicorn/blob/235f06c3aff23f72edd17f2054e29cb49102002e/gunicorn/workers/base.py#L120-L122):
```
        if self.cfg.reload:
            def changed(fname):
                self.log.info("Worker reloading: %s modified", fname)
```
Now `self.log` in this context is <gunicorn.glogging.Logger object at 0x7f137990bac8> which has in it's [dunder init ](https://github.com/benoitc/gunicorn/blob/235f06c3aff23f72edd17f2054e29cb49102002e/gunicorn/glogging.py#L185-L195) the 2 gunicorn loggers:
```
    def __init__(self, cfg):
        self.error_log = logging.getLogger("gunicorn.error")
        self.error_log.propagate = False
        self.access_log = logging.getLogger("gunicorn.access")
```

when it calls ` self.log.info("Worker reloading: %s modified", fname)` it spits it to the `error_logger` as you can see [here](https://github.com/benoitc/gunicorn/blob/235f06c3aff23f72edd17f2054e29cb49102002e/gunicorn/glogging.py#L261):

```
    def info(self, msg, *args, **kwargs):
        self.error_log.info(msg, *args, **kwargs)
```

More generally, it seems to me that everything goes to the `error_logger` in gunicorn except ONCE [here](https://github.com/benoitc/gunicorn/blob/235f06c3aff23f72edd17f2054e29cb49102002e/gunicorn/glogging.py#L327-L346)

euri10:
ok solved what I could @tomchristie , I have one question above, and I explained a little bit what gunicorn does

uvicorn/supervisors/statreload.py
euri10: > It'd be really useful to know which logger gunicorn uses for this kind of messaging.

so the answer is clearly error_logger in case of gunicorn

lovelydinosaur:
Thanks - this has been really helpful!
I've taken a crack at https://github.com/encode/uvicorn/pull/462 which adds some additional functionality wrt. configuration.

euri10:
amzing, closing this for #462 then

PR #442, opened 2019-09-30
Extracting event loop reference to synchronise with Faust

aorumbayev:
This PR is an attempt to add support for Kafka streaming library called Faust. 
The changes introduced are originally implemented by @gitavi, the idea is to have access to loop reference available from Uvicorn instance. Using that, it is then possible to run both Faust's consumers and producers within the same event loop with FastAPI using Uvicorn. 

Here are the main issues on corresponding Faust and FastAPI repositories:
• https://github.com/tiangolo/fastapi/issues/408
• https://github.com/robinhood/faust/issues/399

If there is a different approach on how to achieve it without exposing the loop, or a better way to refactor edit - any comments are welcome! 

Later on if this is merged and approved, I can also help contributing to writing a few documentation notes on integration with Faust and possibly other Kafka clients (I have also tested it with aiokafka!)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -135,6 +135,21 @@ if __name__ == "__main__":
     uvicorn.run(app, host="127.0.0.1", port=5000, log_level="info")
 ```
 
+#### Extracting Event Loop to Provide to other Applications
+
+```python
+from fastapi import FastAPI
+from uvicorn import Uvicorn
+from faust import App
+
+app = FastAPI()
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
See https://github.com/encode/uvicorn/pull/455 for using the lower level API. You can setup the loop yourself and use `await server.serve()` for this functionality.

mura28info:
This PR not yet available in the release branch?

PR #445, opened 2019-10-03
REL-0.10.0

gvbgduh:


lovelydinosaur:
Grand - thanks.

I've had a policy of merging to master, and then having releases bump the version number, rather than merging to a release branch, and then later merging that to master.

It might not be a bad policy, so *possibly* worth considering across all encode projects.
In the meantime I'd like us to keep a standard approach across the board, so for now let's stick with what we're doing elsewhere, and merge change into master. 👍

gvbgduh:
Yeah, thanks for the clarification. I was a bit confused as I had to push some final adjustments and wasn't sure what's the best way to do it as I'm generally hesitant to push straight to master.
I think we can still do the new branch with the version update after merging this one to get back on the track.

PR #446, opened 2019-10-05
Fix for https://github.com/encode/uvicorn/issues/183

RobertoPrevato:
Hi,
When running programmatically an application with uvicorn, as described in the existing issue #183; any event loop the user configured outside of uvicorn is ignored. This is a problem for example, when an http client is configured as part of the business logic / data access layers when preparing the application used by uvicorn. Currently this doesn't work because dependent coroutines are awaited by a different event loop.

This situation makes debugging impossible when running an application programmatically.

**Proposed solution**:
Support `"manual"` configuration setting for the event loop. If manual, no event loop is configured inside `config.py`, because it means the user already configured the event loop.

```python
import uvicorn
from server import app

uvicorn.run(app, host='127.0.0.1', port=44777, log_level='debug', loop='manual')
```

**Note:** I also took into consideration to check if the `loop` is already an instance of event loop, but I understand this option is not viable because uvicorn wants to support different frameworks (asyncio, trio, curio) - although I am not sure about this.

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -214,7 +214,7 @@ class Config:
         self.loaded = True
 
     def setup_event_loop(self):
-        if self.loop == 'manual':
+        if self.loop == "manual":
             # do not configure an event loop, as the user already configured it
             return
         loop_setup = import_from_string(LOOP_SETUPS[self.loop])
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -214,6 +214,9 @@ class Config:
         self.loaded = True
 
     def setup_event_loop(self):
+        if self.loop == 'manual':
+            # do not configure an event loop, as the user already configured it
+            return
         loop_setup = import_from_string(LOOP_SETUPS[self.loop])
         loop_setup()
 


RobertoPrevato:
Not sure why `black` is complaining about the format under Python 3.6, while not complaining for the other builds Python 3.7, pypy3.5?

lovelydinosaur:
There's two alternate routes on to this. Either...

> This is a problem for example, when an http client is configured as part of the business logic / data access layers when preparing the application used by uvicorn. 

1. Have this setup be performed in the application startup, rather than prior to running uvicorn.

...or...

2. Drop down to a lower API level - instantiate a server instance and call `serve` once you're running within the event loop. Really we need to do some extra documentation for how to use the lower level programmatic API.

I generally think we ought to be pushing for (1) and ahve some better docs about how users can hook tasks into the setup phase. Really we don't want the gritty bits of "setup an event loop" to be exposed at all to application developers.

lovelydinosaur:
I'll close this off at the moment, since I don't think it's something we neccessarily want to expose in this way, but I *do* think we need clearer docs for both (1) and (2) above.

RobertoPrevato:
Thanks for the explanation, @tomchristie! I will go for the second option. :tada:

PR #449, opened 2019-10-09
Reenabled black check in Python 3.7

### LINKED ISSUE #448: Reenabled black check in Python 3.7
https://github.com/psf/black/issues/494 is fixed.

Hanaasagi:
It will resolve #448 . uvicorn drops Python 3.5 since version 0.8.6. But pypy3.5 is still in our test matrix. We need to upgrade pypy3.5 to pypy3.6.


lovelydinosaur:
Fantastic, thanks!

Hanaasagi:
PR #424 is about upgrading pypy version.

PR #439, opened 2019-09-26
Add __main__ module, enabling 'python -m uvicorn'.

jaraco:
Fixes #438

PR #454, opened 2019-10-19
#334: add __main__.py to allow running of uvicorn via python's `-m` flag

### LINKED ISSUE #334: Invoke from command line via python -m uvicorn
Would this be a good idea? It is as simple as creating a `uvicorn/__main__.py` with:

```python
if __name__ == "__main__":
    from .main import main
    main()
```

retnikt:
Resolves #334 (even though it's closed)

### DIFF
--- a/uvicorn/__main__.py
+++ b/uvicorn/__main__.py
@@ -1,0 +1,9 @@
+import uvicorn.main
+
+uvicorn.main.main.main(prog_name="uvicorn")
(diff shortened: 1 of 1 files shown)


Hanaasagi (commented):
I think there is already a PR(https://github.com/encode/uvicorn/pull/439) doing the same thing.

lovelydinosaur:
Closed in #439, and released as 0.9.1.

Thanks!

PR #457, opened 2019-10-22
Drop shutdown_servers flag

lovelydinosaur:
Drops the `.run(shutdown_servers=False)` flag.

It's not obvious why this API is available. We don't use it anyplace, it's not documented anyplace, and I don't see that it's obviously a piece of API that we ought to expose.


gvbgduh (approved):
Yeah, looks much cleaner indeed

PR #470, opened 2019-10-29
Support --env-file

lovelydinosaur:
Requires https://github.com/theskumar/python-dotenv

```shell
$ pip install python-dotenv
$ uvicorn example:app --env-file .env
INFO:	Loading environment from '.env'
INFO:	Started server process [73666]
INFO:	Uvicorn running on http://0.0.0.0:5000 (Press CTRL+C to quit)
...
```

Having the envfile loading available as part of the server run command, rather than in the codebase itself is cleaner, and means that it can eg. be omitted from the `Procfile`, but included if running `uvicorn` locally.

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -180,6 +180,7 @@ class Config:
 
         if env_file is not None:
             from dotenv import load_dotenv
+
             logger.info("Loading environment from '%s'", env_file)
             load_dotenv(dotenv_path=env_file)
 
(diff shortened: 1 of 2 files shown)


sylvoslee:
Does not work on Windows

euri10:
> Does not work on Windows

A very useful read suggestion for the next time you want to report something: http://www.catb.org/~esr/faqs/smart-questions.html

PR #474, opened 2019-10-31
Add 'trace' log level

lovelydinosaur:
Adds a new log level, TRACE, which is a level below DEBUG.

Here's why I like this...

**INFO** Output for stuff such as Process startup, Database connected, Access logs.
*Typically suitable for production.*

Screenshot 2019-10-31 at 15 04 08

**DEBUG** Output for stuff such as SQL on database accesses, Outgoing HTTP requests, Cache lookups, SMTP requests, Background tasks.
*Really useful during development, without presenting too much info. Everything you'd want to see to get a good idea of what I/O the system is performing on each request.*

Screenshot 2019-10-31 at 15 03 06

**TRACE** ASGI messages & other low level stuff
*For when you really need to dig into something obscure, or understand exactly whats happening under the hood.*

Screenshot 2019-10-31 at 15 07 28

I can see this fitting in really well as a rule of thumb for other components, too. Eg. I'd like to see `httpx` issuing one-line-per-request at DEBUG level, or comprehensive logging of state changes at TRACE level. Looking at a web stack as a whole it feels like a natural split to me.

### DIFF
--- a/tests/middleware/test_message_logger.py
+++ b/tests/middleware/test_message_logger.py
@@ -1,9 +1,9 @@
-import logging
-
 import pytest
 
 from tests.client import TestClient
 from uvicorn.middleware.message_logger import MessageLoggerMiddleware
+
+TRACE_LOG_LEVEL = 5
 
 
 def test_message_logger(caplog):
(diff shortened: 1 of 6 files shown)


lovelydinosaur:
Somewhat related: https://github.com/encode/databases/pull/156

gvbgduh:
That's the great idea

PR #479, opened 2019-11-09
--reload-dir now defaults to CWD, instead of sys.path

lovelydinosaur:
Previously `--reload-dir` was defaulting to `sys.path`.

This seems excessive to me - we probably don't need to default to checking for reloads of python modules located anywhere other than under the current working directory.

I think we should just change the default to `os.getcwd()`.

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -178,7 +178,7 @@ class Config:
         self.configure_logging()
 
         if reload_dirs is None:
-            self.reload_dirs = sys.path
+            self.reload_dirs = [os.getcwd()]
         else:
             self.reload_dirs = reload_dirs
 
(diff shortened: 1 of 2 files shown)


wshayes:
It's --reload that is changed rather than --reload-dir. Thanks for making this change @tomchristie. The title/description is the only thing that needs to be updated.

lovelydinosaur:
Okay, you've confused me there... 🤣

It's actually the `--reload-dir` default that has changed in this PR, not `--reload`.
The title/description seem correct to me, no?

wshayes:
The --reload-dir option changed, but the effect is that --reload now defaults to CWD right? Or have I mis-read the code and intention? Do you need to have --reload and --reload-dirs (no path set here) to have this work or can you just enter --reload and have it default to CWD?

lovelydinosaur:
Just enter `--reload` on it's own, yeah.  (It's a flag, it doesn't take any arguments itself)

PR #498, opened 2019-11-23
Don't serve requests until startup complete

zhammer:
one potential solution to https://github.com/encode/starlette/issues/733, where uvicorn _does_ process requests before the lifespan startup event is complete

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -370,6 +370,11 @@ class Server:
         )
 
     async def startup(self, sockets=None):
+        await self.lifespan.startup()
+        if self.lifespan.should_exit:
+            self.should_exit = True
+            return
+
         config = self.config
 
         create_protocol = functools.partial(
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Seems like a very sensible change, yup.

One thing worth checking is what the behavior is when the socket fails to bind - really we'd need to make sure that any shutdown tasks get run.

Looks like there's some code linting to apply here.

zhammer:
@tomchristie thanks for taking a look at this. this was a quick draft as a fix for #499 

i'll take a closer look when i have some time!

zhammer:
@tomchristie re:
> One thing worth checking is what the behavior is when the socket fails to bind - really we'd need to make sure that any shutdown tasks get run.

are you thinking something like this? https://github.com/zhammer/uvicorn/commit/0da316a70ade9a04178a250887465db75e0d5572

could add a log saying something like: "server start failed, sending lifespan.shutdown event to app before exiting". also not sure what would be ideal behavior on shutdown exception here.

zhammer:
let me know what i can do to get this more production ready!

aviramha:
Any update on this?

lovelydinosaur:
Righty, I've had a chance to test this all locally and just double check that it all works as expected.
Looks good, great stuff!

zhammer:
Awesome :) thanks tom!

Issue #497, opened 2019-11-23
should_use_colors inside debugger


euri10:
the colorized formatter in pycharm debugger doesnt work (it displays everything in black) because for the `AccessFormatter` we have
```
    def should_use_colors(self):
        return sys.stdout.isatty()
```
that returns False and same for the `DefaultFormatter`, where
```
    def should_use_colors(self):
        return sys.stderr.isatty()
``` 
that returns the same False

In that particular case, the process is ran with [pydevd](https://github.com/fabioz/PyDev.Debugger).
I'm not familar with its internals, but a quick look makes me think it redirects stderr/stdout 
https://github.com/fabioz/PyDev.Debugger/search?q=isatty&type=Code 

do you think the `should_use_colors` test could be improved or that I should raise an issue there ?

lovelydinosaur:
> do you think the should_use_colors test could be improved

I'm not aware that there's anything smarter we can do there?

What do we think about adding a `--log-colors`/`--no-log-colors` command line flag, that explicitly forces the colors on/off, rather than relying on auto-detection?

Any pull requests for that would be very welcome.

lovelydinosaur:
Closed via #502

PR #502, opened 2019-11-30
Add `--use-colors` / `--no-use-colors` flags.

euri10:
A potential attempt at https://github.com/encode/uvicorn/issues/497

I wish I could have found a way to add tests but I have no idea as to what to test, I can see if I pass the `use_colors` flag to False  that it's working but that's weak

True
True

False
https://imgur.com/pyHzfpL


uvicorn/config.py
```diff
@@ -116,9 +149,10 @@ def __init__(
         ws="auto",
         lifespan="auto",
         env_file=None,
-        log_config=LOGGING_CONFIG,
+        log_config=LOGGING_CONFIG_COLORS,
         log_level=None,
         access_log=True,
+        use_colors=True,
```
lovelydinosaur: We should default this to `None`

uvicorn/config.py
```diff
@@ -57,12 +57,12 @@
 SSL_PROTOCOL_VERSION = getattr(ssl, "PROTOCOL_TLS", ssl.PROTOCOL_SSLv23)
 
 
-LOGGING_CONFIG = {
+LOGGING_CONFIG_COLORS = {
```
lovelydinosaur: Let's just stick with a single `LOGGING_CONFIG`.

We can pass an argument `"use_colors": None,` in the config to the `DefaultFormatter`, to indicate that it should be autodetermined.

If use_colors=[True|False] is pass to the config, then we should update *just that bit* in the config dictionary.

uvicorn/config.py
```diff
@@ -215,10 +250,15 @@ def __reduce__(self):
             logging.Logger.__reduce__ = __reduce__
 
         if self.log_config is not None:
+            if not self.use_colors:
+                self.log_config = LOGGING_CONFIG_NO_COLORS
```
lovelydinosaur: Instead of switching the entire config here, lets use...

```python
if self.use_colors in (True, False):
    self.log_config["formatters"]["default"]["use_colors"] = self.use_colors
```

We should only do this within the `if isinstance(self.log_config, dict):` case. If someone is using the file config then it's up to them to choose the correct settings there.

lovelydinosaur (changes requested):
Great stuff!

I've got a bit of a requested tweak in how in think we should handle this... let me know if I've not explained something clearly enough here!

uvicorn/config.py
euri10: done

uvicorn/config.py
euri10: done

uvicorn/config.py
euri10: added the same logic for access too

euri10:
> Great stuff!
> 
> I've got a bit of a requested tweak in how in think we should handle this... let me know if I've not explained something clearly enough here!

hopefully I got this right ;)

uvicorn/config.py
```diff
@@ -216,9 +219,18 @@ def __reduce__(self):
...
+                        "use_colors"
+                    ] = self.use_colors
+                    self.log_config["formatters"]["access"][
+                        "use_colors"
+                    ] = self.use_colors
                 logging.config.dictConfig(self.log_config)
             else:
                 logging.config.fileConfig(self.log_config)
+        else:
+            pass
```
lovelydinosaur: Doesn't look like we need the `else: pass` block here.

uvicorn/logging.py
```diff
@@ -27,8 +27,8 @@ class ColourizedFormatter(logging.Formatter):
         ),
     }
 
-    def __init__(self, fmt=None, datefmt=None, style="%"):
-        self.use_colors = self.should_use_colors()
+    def __init__(self, fmt=None, datefmt=None, style="%", use_colors=None):
+        self.use_colors = use_colors
```
lovelydinosaur: We'd want to do...

```python
if use_colors in (True, False):
    self.use_colors = use_colours
else:
    self.use_colors = sys.stdout.isatty()

**Review by lovelydinosaur — CHANGES_REQUESTED** [review]
Yes, great stuff!
Just a coupla minor comments to address.

**Comment by lovelydinosaur** [comment]
Fantastic!
Last bit is the docs...

* The homepage and deployment pages both include the output of `uvicorn --help` which will need to be updated.
* The settings page includes a "Logging" section which is probably where we want to add your flag docs into.

**Comment by euri10** [comment]
> Fantastic!
> Last bit is the docs...
> 
> * The homepage and deployment pages both include the output of `uvicorn --help` which will need to be updated.
> * The settings page includes a "Logging" section which is probably where we want to add your flag docs into.

yes sorry I forgot the docs, will do that asap

**Comment by euri10** [comment]
just one thing, I modified the cli flag too as it was wrong (it was `"--use-colors/--use-colors"` and I realized it pasting the --help output), I have one question about the default I set to None, but I'm not too sure it's valid in case of a bool

also I'm not sure but think the huge diff in the `uvicorn --help` output comes from the fact it has not been updated in a long time, and for the `index.md` it seems previous version had not the 2 spaces before each flag which explains the huge diff

**Comment by euri10** [comment]
added some examples of custom logging config for https://github.com/encode/uvicorn/issues/491
I realized during this docs aren't tested, might be worth it

edit: commit message wanted to link 491 not 409

**Comment by lovelydinosaur** [comment]
> added some examples of custom logging config

Okay - Let's drop those out of this pull request. We *may* want a PR for them, but that's a seperate issue, and I'd prefer to review it in isolation to the rest of this work.

**Comment by euri10** [comment]
> > added some examples of custom logging config
> 
> Okay - Let's drop those out of this pull request. We _may_ want a PR for them, but that's a seperate issue, and I'd prefer to review it in isolation to the rest of this work.

ok I'm no git pro but I think this might be ok, I did
 git branch logging_examples
git reset --keep HEAD~1
git push -f origin 

so now colors branch should be in the state it was before adding the examples,

File: docs/deployment.md [hunk 6]
```diff
@@ -35,25 +35,35 @@ Options:
...
-  --proxy-headers                 Use X-Forwarded-Proto, X-Forwarded-For,
-                                  X-Forwarded-Port to populate remote address
-                                  info.
+                                  application interface.  [default: auto]
+  --env-file PATH                 Environment configuration file.
+  --log-config PATH               Logging configuration file.
+  --log-level [critical|error|warning|info|debug|trace]
+                                  Log level. [default: info]
+  --access-log / --no-access-log  Enable/Disable access log.
+  --use-colors / --use-colors     Enable/Disable colorized logging.
```
lovelydinosaur: Looks like this should be `--use-colors / --no-use-colors`

docs/index.md
```diff
@@ -67,67 +67,67 @@ The uvicorn command line tool is the easiest way to run your application...
...
+  --lifespan [auto|on|off]        Lifespan implementation.  [default: auto]
+  --interface [auto|asgi3|asgi2|wsgi]
+                                  Select ASGI3, ASGI2, or WSGI as the
+                                  application interface.  [default: auto]
+  --env-file PATH                 Environment configuration file.
+  --log-config PATH               Logging configuration file.
+  --log-level [critical|error|warning|info|debug|trace]
+                                  Log level. [default: info]
+  --access-log / --no-access-log  Enable/Disable access log.
+  --use-colors / --use-colors     Enable/Disable colorized logging.
```
lovelydinosaur: Presumably `--use-colors / --no-use-colors`, right?

docs/settings.md
```diff
@@ -30,6 +30,7 @@ equivalent keyword arguments, eg. `uvicorn.run("example:app", port=5000, reload=
 * `--log-config <path>` - Logging configuration file.
 * `--log-level <str>` - Set the log level. **Options:** *'critical', 'error', 'warning', 'info', 'debug', 'trace'.* **Default:** *'info'*.
 * `--no-access-log` - Disable access log only, without changing log level.
+* `--use-colors` - Enable colorized formatting of the log records.
```
lovelydinosaur: We should change this doc around so that:

* We include both the "on" and "off" flag variants.
* It's clear that we'll autodetect if the flag isn't used.

lovelydinosaur:
> "I'm no git pro but I think this might be ok, I did"

Looks okay to me, yup.

As it happens you *don't actually need to do those smarts* - we have a policy of squashing changes down into a single-commit-per-pull-request on merges, so we don't get any of the messy "getting towards the goal" individual commits, but instead just have a single "here's the PR functionality" commit in the history.

You'd be absolutely fine just making a new commit that removed the work we didn't want in here.

docs/deployment.md
euri10: ok I was sure I pasted it....now I'm sure

docs/index.md
euri10: same...

docs/settings.md
euri10: changed, let me know if this is better

lovelydinosaur:
Thanks for this!

euri10:
I think I somewhat f*** up this PR by fogetting to add the new flag in the main() @tomchristie 

tthe below patch should fix it I think, not sure how to proceed from here

```
Index: uvicorn/main.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- uvicorn/main.py	(revision abd4b006c8a443bc80201409187ab3a662ac7c7e)
+++ uvicorn/main.py	(date 1576008880785)
@@ -257,6 +257,7 @@
     ssl_ca_certs: str,
     ssl_ciphers: str,
     headers: typing.List[str],
+    use_colors: bool,
 ):
     sys.path.insert(0, ".")
 
@@ -292,6 +293,7 @@
         "ssl_ca_certs": ssl_ca_certs,
         "ssl_ciphers": ssl_ciphers,
         "headers": list([header.split(":") for header in headers]),
+        "use_colors": use_colors,
     }
     run(**kwargs)
````

lovelydinosaur:
> the below patch should fix it I think

Looks right to me too. Wanna issue a PR for that?

euri10:
Will do for sure tomorrow!

Le mar. 10 déc. 2019 à 10:48 PM, Tom Christie <notifications@github.com> a
écrit :

euri10:
ok done in https://github.com/encode/uvicorn/pull/520

Issue #320, opened 2019-03-08
TypeError: __init__() when run "uvicorn app:App"


adriangallegos:
I'm working on Mac Os Sierra 10.12.6, python 3.7.2 and uvicorn via pip3 0.5.1.
When I run the example uvicorn app:App get the following error:
Traceback (most recent call last):
  File "/usr/local/bin/uvicorn", line 11, in <module>
    load_entry_point('uvicorn==0.5.1', 'console_scripts', 'uvicorn')()
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2793, in load_entry_point
    return ep.load()
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2411, in load
    return self.resolve()
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2417, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/local/lib/python3.7/site-packages/uvicorn/__init__.py", line 2, in <module>
    from uvicorn.main import Server, main, run
  File "/usr/local/lib/python3.7/site-packages/uvicorn/main.py", line 212, in <module>
    ssl_ciphers: str,
  File "/usr/local/lib/python3.7/site-packages/click/decorators.py", line 170, in decorator
    _param_memo(f, OptionClass(param_decls, **attrs))
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 1460, in __init__
    Parameter.__init__(self, param_decls, type=type, **attrs)
TypeError: __init__() got an unexpected keyword argument 'hidden'

Thank you

florimondmanca:
@adriangallegos Not a maintainer, but I think providing a minimum example application and full command listing allowing to reproduce this error would help people to help you. :-)

adriangallegos:
Thank you @florimondmanca 

I tried with the example:

class App():
    def __init__(self, scope):
        assert scope['type'] == 'http'
        self.scope = scope
    async def __call__(self, receive, send):
        await send({
            'type': 'http.response.start',
            'status': 200,
            'headers': [
                [b'content-type', b'text/plain'],
            ]
        })
        await send({
            'type': 'http.response.body',
            'body': b'Hello, world!',
        })

[...]

PR #508, opened 2019-12-03
Upgrade h11 dependency to 0.9

kmichel2:
Upgrading is not required but I don't see any reason not to upgrade (I might be missing something).

All dependencies were pinned in [#324](https://github.com/encode/uvicorn/pull/324) with the main intent of pinning the click version to fix an issue: #320.

Local tests are passing successfully.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ env_marker = (
 
 requirements = [
     "click==7.*",
-    "h11==0.8.*",
+    "h11==0.9.*",
     "websockets==8.*",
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Yup, looks good. 👍

jlowin:
Just a note - this pin is not compatible with `httpx` (which pins `h11==0.8.*`). Perhaps pinning to `h11 >= 0.8, < 0.10` would achieve the goal more gracefully?

lovelydinosaur:
@jlowin - Thanks, that's a really good point.

PR #510, opened 2019-12-04
Handle signals

euri10:
attempt to replace https://github.com/encode/uvicorn/pull/402

I added 2 new tests (`test_run_signal` and `test_run_signal_multi_threads`)  that make use of the flag and therefore don't need to use the `CustomServer` class that override signal handling, I guess it's a test only workaround.
```
    class CustomServer(Server):
            pass
```

~~I would need / love guidance in how to properly handle one of the tests here, namely `test_run_signal_multi_threads`, because I have currenlty an issue:~~

~~- `test_run_signal_multi_threads` in essence spawns 10 threads, run uvicorn in them with  each time a new loop, the test works fine independently however it needs an `import asyncio` to set loops. And **this single import break about 80 other tests** while on its own the test passes fine and we can have 10 uvicorn servers running in 10 threads with 10 loops...well~~

~~So I left the PR in a state where all tests pass but this one because of the missing import. 
Hope it makes sense~~

ok I found the fix:  because of the tests I added, the `_ASGIAdapter`  had a closed loop it tried to use and therefore failed, so a simple test on it and all tests pass

### DIFF
--- a/tests/client.py
+++ b/tests/client.py
@@ -99,7 +99,11 @@ class _ASGIAdapter(requests.adapters.HTTPAdapter):
         response_started = False
         raw_kwargs = {"body": io.BytesIO()}
 
-        loop = asyncio.get_event_loop()
+        if not asyncio.get_event_loop().is_closed():
+            loop = asyncio.get_event_loop()
+        else:
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
 
         try:
             loop.run_until_complete(self.app(scope, receive, send))
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,8 +1,10 @@
+import asyncio
 import threading
 import time
-import asyncio
+import typing
 
 import requests
+from requests.adapters import HTTPAdapter
 
 from uvicorn.config import Config
 from uvicorn.main import Server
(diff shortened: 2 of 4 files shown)


euri10:
add a click option in main ?

uvicorn/config.py
```diff
@@ -172,6 +174,7 @@ def __init__(
         self.ssl_ciphers = ssl_ciphers
         self.headers = headers if headers else []  # type: List[str]
         self.encoded_headers = None  # type: List[Tuple[bytes, bytes]]
+        self.handled_signals = handled_signals
```
lovelydinosaur: Let's use `handled_signals or []` here, so that it's *always* a list.
That'll simplify the `install_signal_handlers` implementation slightly.

tests/client.py
```diff
@@ -99,7 +99,11 @@ def send(self, request, *args, **kwargs):
         response_started = False
         raw_kwargs = {"body": io.BytesIO()}
 
-        loop = asyncio.get_event_loop()
+        if not asyncio.get_event_loop().is_closed():
+            loop = asyncio.get_event_loop()
+        else:
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
```
lovelydinosaur, line 106: Is this change neccessary?

tests/test_main.py
```diff
@@ -80,3 +83,82 @@ def install_signal_handlers(self):
...
+    ]
+    start_threads(*workers)
+    time.sleep(1)
+    for x in range(threads_count):
+        port = 10000 + x
+        s = requests.Session()
+        s.mount(f"http://127.0.0.1:{port}", HTTPAdapter(max_retries=1))
+        response = s.get(f"http://127.0.0.1:{port}")
+        assert response.status_code == 204
+    join_t(*workers)
```
lovelydinosaur, line 204: I actually don't think we need to add these tests. We already have coverage of the code that's being used - that's enough for me here.

euri10: fair enough

however I wanted to say that the main use case for this PR, if I understanded it correctly what people wanted, I mean using `handled_signals=None` is for people who may want a finer control over that Config part, so that they could run uvicorn in one thread and something else in another.

that test is effectively far-fetched in the sense it runs 10 servers in 10 threads, but it just demonstrates how it could be done,

skipping it removes the need for modifying the `tests/client.py` above since one-side effect of the test is that it as to close the loop!

tests/client.py
euri10: yes if we keep the  `test_run_signal_multi_threads`.

if this isn't modified here's what happens:
================================================================ 80 failed, 92 passed, 12 warnings in 9.79s ================================================================

this happens on all tests after this one, because the `loop = asyncio.get_event_loop(`) I replaced above is closed now so you get a `E   RuntimeError: Event loop is closed` for the 80 failed tests after 

should we skip this test then this is not required but my personal guess is that it's a good example

uvicorn/config.py
euri10: done

uvicorn/config.py
```diff
@@ -139,6 +140,7 @@ def __init__(
         ssl_ca_certs=None,
         ssl_ciphers="TLSv1",
         headers=None,
+        handled_signals=(signal.SIGINT, signal.SIGTERM),
```
florimondmanca: How about `exit_signals`?

euri10: done

devxpy:
Thanks for the work @euri10! Anything left to do here?

([Shameless plug](https://stackoverflow.com/questions/60182619/run-multiple-asgi-apps-in-same-thread-with-uvicorn))

Issue #491, opened 2019-11-17
Add example of logging config file


matrixise:
In the doc, there is the description of `--log-config <path>` for the configuration of a config file. I think an example of config file could be really awesome for the new comers.

> [!Important]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/uvicorn/issues/491">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/491/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/uvicorn/issues/491/pledge.svg">
</picture>
</a>

lovelydinosaur:
Agreed yup. As a starting point we could just link to this: https://docs.python.org/3/library/logging.config.html#logging-config-fileformat

Tho ideally a proper example that's relevant to uvicorn would be ace.

Any pull requests here would be most welcome (even if it's just linkig to the Python docs to start with.)

ex-nerd:
It would also be worth mentioning that by enabling the custom log, other uvicorn loggers are likely to become disabled.  See #511

NightMachinery:
I found this helpful:
```
import sys
import uvicorn
from uvicorn.config import LOGGING_CONFIG

def main():
    root_path = ''
    if len(sys.argv) >= 2:
        root_path = sys.argv[1]
    ##
    # %(name)s : uvicorn, uvicorn.error, ... . Not insightful at all.
    LOGGING_CONFIG["formatters"]["access"]["fmt"] = '%(asctime)s %(levelprefix)s %(client_addr)s - "%(request_line)s" %(status_code)s'
    LOGGING_CONFIG["formatters"]["default"]["fmt"] = "%(asctime)s %(levelprefix)s %(message)s"

    date_fmt = "%Y-%m-%d %H:%M:%S"
    LOGGING_CONFIG["formatters"]["default"]["datefmt"] = date_fmt
    LOGGING_CONFIG["formatters"]["access"]["datefmt"] = date_fmt
    ##
    uvicorn.run("brishgarden.garden:app", host="127.0.0.1", port=7230, log_level="info", proxy_headers=True, root_path=root_path)

```

ievgennaida:
[...]

PR #515, opened 2019-12-09
Logging custom docs

euri10:
attempt at solving https://github.com/encode/uvicorn/issues/491

It adds a simple example of an app with custom logging

I added `markdown-include` so that examples are code, it could be a good idea to add to the tests CI that part of the docs


lovelydinosaur:
Great work getting started on this!

It'd be great if we could make this a bit more minimal. In particular:

* It'd be helpful to just give an example logging config file that *replicates* the existing default log config.
* It'd be nice if the only things that changed in the PR were (1) Adding an example log config file. and (2) linking to it from the settings docs.

I'd prefer it if we could drop all the other surrounding context.

euri10:
ok I made the logging_example.yml replicate the default config the only difference being it's a yaml file and not a dict
I kept markdown-include addition but I can remove it, it just makes things simpler imho in the case we put a yaml file as a config, any change in it will be automatically reflected in the docs, but I understand if you'd prefer not to.

Next slash in the PR could be to just  copy/paste the LOGGING_CONFIG dictionnary into a file and link it in the settings page, that;d more minimal, let me know what you prefer :)

docs/examples.md
```diff
@@ -0,0 +1,17 @@
...
+You can pass it loading a dict, using the `--log-config` flag.
+It sets 2 loggers:
+1. `uvicorn.error` whose formatter is the `uvicorn.logging.DefaultFormatter`. 
+2. `uvicorn.access` whose formatter is the `uvicorn.logging.AccessFormatter`. 
+
+Both formatters will output a colorized automatically if a tty is detected.
+
+If you used the `--use-colors / --no-use-colors` then the output will / won't be colorized.
+
+```yaml hl_lines="7 11 38 39 40 41"
```
lovelydinosaur: I'd rather than any example is just included directly, than using this style.

docs/examples.md
```diff
@@ -0,0 +1,17 @@
+# Examples
+
+## Custom logging
+
+The below `logging_example.yaml` is a yaml representation of the default logging configuration.
+You can pass it loading a dict, using the `--log-config` flag.
```
lovelydinosaur: Just checking... Does this definately work?
I haven't checked out if Python's log file format also supports using the yaml style?

euri10: this is definitely an ambiguous formulation, python logging supports dict / file (ini style). A lot of time people use yaml and load it to dict with pyyaml for instance, that's why I said "representation"

docs/examples.md
euri10: ok so you prefer to remove markdown-include and just copy-paste examples right ?

lovelydinosaur: Yup

docs/examples.md
lovelydinosaur: Right, but it wouldn't directly work with our `--log-config` flag right?

(Tho we *could* potentially change things here, so that we *expect* a YAML style logging config file.)

euri10: correct, one would have to do something like
```
    with open('logging_example.yaml', 'r') as stream:
        config = yaml.load(stream, yaml.FullLoader)
    uvicorn.run("main:app", reload=True, log_config=config, log_level="trace")
```

docs/examples.md
euri10: done

Issue #476, opened 2019-11-04
Add missing --version flag on cli


hvdklauw:
Would make figuring out which version we have a lot easier.

lovelydinosaur:
Yup, absolutely agreed.
Would certainly accept a pull request implementing this.

lovelydinosaur:
I guess ideally we want something that allows us to print out the Python version, and our dependency versions, to help with filing comprehensive issue reports. Wondering if there's any other Python projects in particular that do a good job of this?

hvdklauw:
flake8 lists it's version, it's dependencies versions and the python version used.

lovelydinosaur:
Good call, yup...

```shell
$ venv/bin/flake8 --version
3.7.9 (mccabe: 0.6.1, pycodestyle: 2.5.0, pyflakes: 2.1.1) CPython 3.7.1 on Darwin
```

Also similar...

```shell
$ python -m requests.help
{
  "chardet": {
    "version": "3.0.4"
  },
  "cryptography": {
    "version": ""
  },
  "idna": {
    "version": "2.8"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.7.1"
  },
  "platform": {
    "release": "18.6.0",
    "system": "Darwin"
  },
  "pyOpenSSL": {
    "openssl_version": "",
    "version": null
  },
  "requests": {
    "version": "2.22.0"
  },
  "system_ssl": {
    "version": "1010009f"
  },
  "urllib3": {
    "version": "1.25.6"
  },
  "using_pyopenssl": false
}
```

florimondmanca:
We have a `--version` flag now:

```console
$ uvicorn --version
Running uvicorn 0.12.0 with CPython 3.8.2 on Darwin
```

So, closing! (Probably we meant to close this via #518 but didn't include a `Fixes` magic keyword there.)

PR #518, opened 2019-12-10
Add `--version` flag.

euri10:
attempts at https://github.com/encode/uvicorn/issues/476

~~there are many ways to add the functionality, I choose not to touch to the only existing command (main) and add a new entrypoint `uvicorn.version`
the alternative would be to add a `click.option("--version"....)` and inside the main function do a if version then echo the version else run as usual~~

```
(venv) ➜  uvicorn git:(version) ✗ uvicorn.version 
Running uvicorn 0.10.8 with CPython 3.8.0 on Linux
```

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -2,6 +2,7 @@ import asyncio
 import functools
 import logging
 import os
+import platform
 import signal
 import socket
 import ssl
@@ -12,6 +13,7 @@ from email.utils import formatdate
 
 import click
 
+import uvicorn
 from uvicorn.config import (
     HTTP_PROTOCOLS,
     INTERFACES,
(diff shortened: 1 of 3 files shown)


uvicorn/main.py
```diff
@@ -225,6 +242,9 @@
     multiple=True,
     help="Specify custom default HTTP response headers as a Name:Value pair",
 )
+@click.option(
+    "--version", is_flag=True, callback=print_version, expose_value=False, is_eager=True
```
lovelydinosaur: Looks great, yup! I think we just want to add something like `help="Display the uvicorn version and exit."` here.

lovelydinosaur:
Lovely, thanks!

PR #519, opened 2019-12-11
Add --reload-extension option to specify which files are watched.

wapiflapi:
This pull request ads the option `--reload-extension` to specify which extensions should be watched. The default behavior doesn't change and is still only watching `.py` files.

This is useful in projects using non-python files that change during development (for example graphql `schema.sdl` files, or configuration files.) 

I updated the tests so that combinations of extensions and file changes are tested (as well as the default behaviors.) I had to add a `time.sleep()` to the tests because we are actually testing if the system detects the time change and on my machine the tests were failing often due to the the `Path.touch()` being too fast and close to each other.

I also updated the documentation to the best of my ability but not being super familiar with the project I might have forgotten something, please point it out and I will fix it!

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -36,6 +36,8 @@ Options:
   --reload                        Enable auto-reload.
   --reload-dir TEXT               Set reload directories explicitly, instead
                                   of using the current working directory.
+  --reload-extension TEXT         Set reloaded file extensions explicitly,
+                                  instead of only watching .py.
   --workers INTEGER               Number of worker processes. Defaults to the
                                   $WEB_CONCURRENCY environment variable if
                                   available. Not valid with --reload.
--- a/docs/index.md
+++ b/docs/index.md
@@ -78,6 +78,8 @@ Options:
   --reload                        Enable auto-reload.
   --reload-dir TEXT               Set reload directories explicitly, instead
                                   of using the current working directory.
+  --reload-extension TEXT         Set reloaded file extensions explicitly,
+                                  instead of only watching .py.
   --workers INTEGER               Number of worker processes. Defaults to the
                                   $WEB_CONCURRENCY environment variable if
                                   available. Not valid with --reload.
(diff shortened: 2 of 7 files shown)


lovelydinosaur:
I'd like us to aim for less configuration, rather than more.
I'd suggest that instead of being able to choose which extensions to watch, we change the behavior so that we just watch *all* files in the reload directory. Sound reasonable?

wapiflapi:
> I'd like us to aim for less configuration, rather than more.
> I'd suggest that instead of being able to choose which extensions to watch, we change the behavior so that we just watch _all_ files in the reload directory. Sound reasonable?

Sounds good. I'll do a PR for that.

wapiflapi:
@tomchristie thinking about it some more, are we concerned about changing the behavior for existing users? I'm thinking most people have it on the current directory which is probably the top level of their project which might include stuff such as:
  - logfiles
  - sqlite (or other testing / dummy) databases
  - `.git` folder 
  - probably more stuff

All of the above might change more or less frequently during development without requiring a server reload which might be bothersome for users that do not expect it.

All of this can be fixed by the user by specifying the directories reload watches, but it will require users to do something. 

If you want I can open an issue to recap the discussion, but to be honest I don't want to make this a big thing (and eg. start talking about following `.gitignore` for reload).

rafalp:
I would take this step further and say that we should watch all non-hidden files and directories within `reload-dir`.

There is an unknown number of file extensions that app may depend on for configuration, and trying to maintain list of supported ones will result in "My project parses .XYZ files on startup, can you add that to file watcher?" issues/PRs.

Next, the reason for omitting `.hidden` files/dirs by default, is because those are usually used to store dev data (such as tools configuration) or caches. For example, `.git` directory can easily contain hundreds of thousands of files for larger projects, but those files are effectively a database about project's code. Checkouts and commit hooks will likely cause filesystem changes in app's sources, triggering reload anyway, but at less expense. For same reason reloader should also ignore python cache files.

PR #521, opened 2019-12-11
Watch all files for reload not only `.py` files.

wapiflapi:
Follow up to https://github.com/encode/uvicorn/pull/519

This pull request **changes** the default behavior so that all files in the reload directories are watched and not only `.py` files. (See discussion on #519 as to why we want this instead of specifying which extensions to watch.)

This is useful in projects using non-python files that change during development (for example graphql schema.sdl files, or configuration files.)

The downside is that people using the default configuration that now watches all files in the current working directory which might include files that change often and should not cause a reload. Such files could include:
  - `.git`: committing and other git operations could cause reloads.
  - logfiles: If during development logs are kept in the current working directory they will cause infinite reloads of the server.
  - data files, test databases: it is not uncommon during development to use sqlite or json files to store state between runs, those might reasonably be in the current working directory.
  - IDE configuration, backup, cache, etc.

All of the above can be solved by specifying `--reload-dir` but that requires the user to do something they didn't have to before.

I'm not sure what the **correct** solution is in this case, maybe following `.gitignore` but that seems awfully out of scope for uvicorn.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -19,7 +19,7 @@ equivalent keyword arguments, eg. `uvicorn.run("example:app", port=5000, reload=
 ## Development
 
 * `--reload` - Enable auto-reload.
-* `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default all directories in `sys.path` will be watched.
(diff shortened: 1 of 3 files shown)


euri10:
I'm going to close this, seems like it has been forgotten and implemented, see https://www.uvicorn.org/settings/#development

PR #531, opened 2019-12-19
Output correct port if port=0, refs #530

### LINKED ISSUE #530: Correctly output selected port if port=0
If you run Uvicorn with `port=0`, it selects an unused port (as [it should](https://www.lifewire.com/port-0-in-tcp-and-udp-818145)). But... it logs the incorrect URL:

```
$ datasette --memory -p 0
Serve! files=() (immutables=()) on port 0
INFO:     Started server process [28673]
INFO:     Uvicorn running on http://127.0.0.1:0 (Press CTRL+C to quit)
```

It should output the selected port in that message instead.

simonw:
Should fix #530

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -428,6 +428,9 @@ class Server:
             except OSError as exc:
                 logger.error(exc)
                 sys.exit(1)
+            port = config.port
+            if port == 0:
+                port = server.sockets[0].getsockname()[1]
             protocol_name = "https" if config.ssl else "http"
             message = "Uvicorn running on %s://%s:%d (Press CTRL+C to quit)"
             color_message = (
@@ -439,7 +442,7 @@ class Server:
                 message,
                 protocol_name,
                 config.host,
-                config.port,
+                port,
                 extra={"color_message": color_message},
             )
             self.servers = [server]


simonw:
Weird: https://travis-ci.org/encode/uvicorn/jobs/627147183

Test fails but only under Python 3.6:
```
tests/supervisors/test_statreload.py .F                                  [100%]

=================================== FAILURES ===================================

______________________________ test_should_reload ______________________________

tmpdir = local('/tmp/pytest-of-travis/pytest-0/test_should_reload0')
    @pytest.mark.skipif(
        sys.platform.startswith("win"),
        reason="Skipping reload test on Windows, due to low mtime resolution.",
    )
    def test_should_reload(tmpdir):
        update_file = Path(os.path.join(str(tmpdir), "example.py"))
        update_file.touch()

        working_dir = os.getcwd()

        os.chdir(str(tmpdir))

        try:
            config = Config(app=None, reload=True)
            reloader = StatReload(config, target=run, sockets=[])
            reloader.startup()
    
            assert not reloader.should_restart()
            update_file.touch()
>           assert reloader.should_restart()
E           assert False
E            +  where False = <bound method StatReload.should_restart of <uvicorn.supervisors.statreload.StatReload object at 0x7f60701cdb00>>()
E            +    where <bound method StatReload.should_restart of <uvicorn.supervisors.statreload.StatReload object at 0x7f60701cdb00>> = <uvicorn.supervisors.statreload.StatReload object at 0x7f60701cdb00>.should_restart
```

uvicorn/main.py
```diff
@@ -428,6 +428,9 @@ def run(self, sockets=None):
             except OSError as exc:
                 logger.error(exc)
                 sys.exit(1)
+            port = config.port
+            if port == 0:
+                port = server.sockets[0].getsockname()[1]
```
simonw, line 458: Will `server.sockets` only ever have a single item in the list here? I think so but I'm not certain.

lovelydinosaur:
Thanks @simonw.

> Test fails but only under Python 3.6

Righty, that's actually a flaky test right there, since it's dependant on the time resolution and how quickly the `.touch` takes effect.

Let me take a look at resolving that, then once that's in you can merge master back into this.

lovelydinosaur:
Righty, worth pulling master into this now, which ought to resolve the issue.

simonw:
That fixed it!

PR #532, opened 2019-12-19
More resiliant reload test

lovelydinosaur:
Sleep briefly to before performing the file `touch`, to ensure that `should_reload` correctly detects an updated file.

### DIFF
--- a/tests/supervisors/test_statreload.py
+++ b/tests/supervisors/test_statreload.py
@@ -1,6 +1,6 @@
 import os
 import signal
-import sys
+import time
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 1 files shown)


euri10 (commented):
shouldn't the wait be after the touch ?

lovelydinosaur:
Nope. We want to ensure that the timestamp on the file has changed.

euri10:
> Nope. We want to ensure that the timestamp on the file has changed.

obviously, and this happens when you modify / touch the file doesnt it ?

lovelydinosaur:
So, we're checking the timestamp in the initial `should_reload()`, then waiting a bit, then touching the timestamp, then calling `should_reload()` again which sees that it's been updated since *later* than during the last check.

Issue #529, opened 2019-12-18
uvicorn reload fails on Windows and Python 3.8


f0ff886f:
I know this was referred to in #477 and #509 but it still isn't solved (making a new issue as I hijacked a closed one before).

Using uvicorn checked out 996aa2d, running with or without `--loop asyncio` changes nothing (that code was removed):

```
ERROR:    Accept failed on a socket
socket: <asyncio.TransportSocket fd=764, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('127.0.0.1', 8000)>
Traceback (most recent call last):
  File "C:\Users\f\scoop\apps\python\current\lib\asyncio\proactor_events.py", line 817, in loop
    f = self._proactor.accept(sock)
  File "C:\Users\f\scoop\apps\python\current\lib\asyncio\windows_events.py", line 545, in accept
    self._register_with_iocp(listener)
  File "C:\Users\f\scoop\apps\python\current\lib\asyncio\windows_events.py", line 714, in _register_with_iocp
    _overlapped.CreateIoCompletionPort(obj.fileno(), self._iocp, 0, 0)
OSError: [WinError 87] The parameter is incorrect
```

```
Python 3.8.0 (tags/v3.8.0:fa919fd, Oct 14 2019, 19:37:50) [MSC v.1916 64 bit (AMD64)] on win32
```

f0ff886f:
This does not repro on py37, reloading works fine there.

```
Python 3.7.5 (tags/v3.7.5:5c02a39a0b, Oct 15 2019, 00:11:34) [MSC v.1916 64 bit (AMD64)] on win32
```

lovelydinosaur:
Right, so the problem here is to do with IOCP.

In Python 3.8 the default loop policy changes on Windows to use IOCP by default. We probably want to *explcitly* turn this off (and remove our `loop="iocp"` support completly, since it appears to not work.)

Does this only occur when using multiple workers / reload, or does it always occur on Python 3.8 with Windows?

euri10:
~I can trigger the same on windows and 3.7.5~
passing `--loop="asyncio"` is fine on 3.7.5
passing `--loop="asyncio"` on 3.8.0 the bug still here

f0ff886f:
>Does this only occur when using multiple workers / reload, or does it always occur on Python 3.8 with Windows?

[...]

PR #533, opened 2019-12-20
Simplified subprocess by enabling socket pickling

euri10:
while working on trying to understand https://github.com/encode/uvicorn/issues/529 I realized this windows specificity can be handled more simply
it doenst solve the issue mentionned though


lovelydinosaur:
Lovely change, yup. Thanks.

Issue #306, opened 2019-02-23
Add websocket max size


divad:
Uvicorn does not appear to support setting a web socket max message size. 

Without such an option an attacker can overwhelm the memory of an application running within Uvicorn (a denial of service). 

Other frameworks such as aiohttp, tornado and Sanic have max message size options. Hypercorn also has a max message size:

https://gitlab.com/pgjones/hypercorn/blob/master/hypercorn/config.py#L55

It is discussed/documented here: https://pgjones.gitlab.io/hypercorn/dos_mitigations.html#large-websocket-message

Cheers!

lovelydinosaur:
It would be worth looking at the wsproto and websockets packages here. Do either of them include a limit by default? (We support either)

lovelydinosaur:
But either way, yes we should definitely have this as a configuration setting.

cgarciae:
Hey, any updates on this?

lovelydinosaur:
@cgarciae Nope - we'd use this ticket if there was any progress.
If anyone wants to take a crack at a pull request for it I'd happily help review & get it in.

pgjones:
> It would be worth looking at the wsproto

I don't think wsproto has something for this (may have forgotten), but I do this in Hypercorn - firstly there is a [config setting](https://github.com/pgjones/hypercorn/blob/master/hypercorn/protocol/ws_stream.py#L147) then it is checked in the [message buffer](https://github.com/pgjones/hypercorn/blob/master/hypercorn/protocol/ws_stream.py#L121). Also see [Some commentary](https://pgjones.gitlab.io/hypercorn/dos_mitigations.html#large-websocket-message)

selimb:
Looks like #538 fixes this, but has been collecting dust? Needing to switch to a different web server just to have control over that option is unfortunate :(

Many thanks for your time spent on this project.

untereiner:
Hello,

I was reading this issue and I am wondering about two size parameters. 
I think there are two sizes who comes into play: 
- the max WebSocket message payload size: the maximum size in bytes allowed for a complete WebSocket message (with multiple frames) which I think is discussed by this issue
- the max WebSocket frame payload size: the maximum size in bytes allowed for a single WebSocket frame.

I don't see the second in wsproto.
[...]

PR #538, opened 2019-12-20
Defaults ws max_size on server to 16MB

euri10:
Rough attempt at solving https://github.com/encode/uvicorn/issues/306

There are a few points I'm not sure:

1. seems like this can be implemented only on `WebSocketProtocol` but the truth is I didn't find anything relevant for `WSProtocol`
2. In `class WebSocketCommonProtocol(asyncio.Protocol):` the default max_size is 2**20 (1MB) so I wrote a new test that sends 1 more byte than this default, using a default config on the server side that uses a 16MB default (following [hypercorn default](https://gitlab.com/pgjones/hypercorn/blob/master/hypercorn/config.py#L77)), would you want to see more tests, which ones  (send 1 bytes more that default config of 16MB > 1009 , anything else ?_)


lovelydinosaur:
Okay looks good. Shall we call this `ws_max_size` and make sure to push it all the way through to being a configurable setting? (Including the cli)

euri10:
Will do it asap, kind of swamped with work now.

euri10:
OK @tomchristie I changed the name and pushed it all the way up so that it's configurable,
there are 2 things I'd like feedback on before advancing:

1. is the 16MB default good for you ?
2. since the option is configurable only for the `WebSocketProtocol` (I didnt find anything relevant for the `WSProtocol`) should we add a warning / sys.exit(1) if someone uses the `--ws-max-size` with the `--ws=wsproto` ? I added a note in the docs in case

deepankarm:
@euri10 Is there any reason this wasn't merged?

euri10:
Yes we need a reviewer's approval to merge something.

jjehl:
Maybe have you find a reviewer ? This limitation of 1MB could be a real problem when using uvicorn.  And the patch appear to be simple : 
`super().__init__(ws_handler=self.ws_handler, ws_server=self.ws_server, max_size=self.config.websocket_max_message_size)`

castaway2000:
I am also finding this to be a major hindrance in some of the work i am doing. is there a way we can merge this in?

darthlegit:
eagerly awaiting this.

PR #542, opened 2019-12-29
color codes in windows console not escaped

### LINKED ISSUE #815: color codes in windows console
### Describe the bug

Color/format codes in console are literally printed.

### To Reproduce

Use a simple fastapi setup like your readme example.

```Python
from fastapi import FastAPI

app = FastAPI()


@app.get("/")
def read_root():
    return {"Hello": "World"}
```

### Expected behavior

I figured out that this can be fixed by using `os.system("cls")` on top of your `main.py`

### Screenshots

grafik

Expected with `os.system("cls")`:

grafik

### Environment

- OS: Windows 10
- FastAPI Version 0.45.0
- Python version 3.8

### Additional context

I do not think adding a raw `os.system("cls")` should be the right solution for a library. However, I do not know another way to fix the issue.
On Ubuntu18.04 everything works as expected. Could be a problem with my windows machine.
If this is a problem with my local machine, I am sorry.
Also I know that I should post this on the uvicorn repo instead. But I figured that you should mention this as long as the bug exists.

euri10:
Fixes https://github.com/tiangolo/fastapi/issues/815 that should have been reported upstream

There are many ways to handle the case obviously, I choose to use click.clear() since we use already click.style and because it already performs the os check and issues the right command for that.

### DIFF
--- a/uvicorn/logging.py
+++ b/uvicorn/logging.py
@@ -32,6 +32,7 @@ class ColourizedFormatter(logging.Formatter):
             self.use_colors = use_colors
         else:
             self.use_colors = sys.stdout.isatty()
+        click.clear()
         super().__init__(fmt=fmt, datefmt=datefmt, style=style)
 
     def color_level_name(self, level_name, level_no):


blep:
I can confirm that calling `click.clear()` just before creating the FastAPI app fix the log colorization issue on Windows with `cmd.exe` console. I added the call just before creating the FastApi app.

tiangolo:
In theory, adding [Colorama](https://pypi.org/project/colorama/) to the dependencies "should" fix all these color issues.

If Click finds that `colorama` is installed, it will use it and pass all the color stuff through it, and it does special tricks to handle colors in Windows.

Here are the relevant Click docs: http://click.palletsprojects.com/en/7.x/utils/#ansi-colors

I haven't checked (I don't have a Windows at hand), but maybe that's something worth trying too...

fghso:
Hi, I can also confirm that calling `click.clear()` in `logging.py` fix the issue on Windows with Powershell. The only drawback is that, as the function name implies, it also clears the terminal window.

I tried @tiangolo's suggestion of installing `Colorama` but unfortunaly it didn't work out of the box as expected. To make it work you have to explicitly call `colorama.init()`. For example, it works if you replace the call to `click.clear()` in this PR with `colorama.init()`.

However, the fact that [Click's documentation](https://click.palletsprojects.com/en/7.x/utils/#ansi-colors) says "Do not call colorama.init()!" led me to dig a little deeper into the issue. I found out that both [`click.echo()`](https://github.com/pallets/click/blob/97de2c97275f52adc5c6ef8bb8aa993ce997fe7e/src/click/utils.py#L171) and [`colorama.init()`](https://github.com/tartley/colorama/blob/master/colorama/initialise.py#L23) create wrappers around `sys.stdout` and `sys.stderr` (`click.echo()` uses [`colorama.AnsiToWin32()`](https://github.com/tartley/colorama/blob/3d8d48a95de10be25b161c914f274ec6c41d3129/colorama/ansitowin32.py#L64) under the hood to create an ansi wrapped stream), so they can interfere with each other. That's probably the reason behind the instruction in Click's docs.

In this case, a possible solution to avoid calling `colorama.init()` is to emit the logs through `click.echo()` (note that Colorama must be installed anyway). That could be done by creating a specific log handler and configuring uvicorn accordingly. Something like this:

#### In [`logging.py`](https://github.com/encode/uvicorn/blob/master/uvicorn/logging.py):
```python
import io

class ClickEchoHandler(logging.StreamHandler):

    def __init__(self, **kwargs):
        self.echo_args = kwargs
        self.terminator = ""
        super().__init__()
        
    def emit(self, record):
        self.setStream(io.StringIO())
        super().emit(record)
        click.echo(self.stream.getvalue(), **self.echo_args)
```

#### In [`config.py`](https://github.com/encode/uvicorn/blob/master/uvicorn/config.py#L73):
```python
"handlers": {
    "default": {
        "formatter": "default",
        "()": "uvicorn.logging.ClickEchoHandler",
        "err": True,
    },
    "access": {
        "formatter": "access",
        "()": "uvicorn.logging.ClickEchoHandler",
    },
}
```
I tested this solution locally on my Windows machine and it worked very well.

Issue #544, opened 2020-01-04
UvicornWorker ignores backlog configuration of Gunicorn


aviramha:
When stress testing my application I found out that UvicornWorker ignores the backlog configuration of Gunicorn.
My guess it happens on create_server (where it should pass the backlog argument, but doesn't)
To reproduce:
`gunicorn app:app -k uvicorn.workers.UvicornWorker --backlog 100000 --log-level=debug`

Then run `ss -a | grep gunicorn`
Send-Q value will be 100 (default backlog value) instead of the given.
I'll submit a PR.

aviramha:
Fixed in #545 .
Thanks @euri10 for debugging with me and helping :)

PR #545, opened 2020-01-04
Support backlog configuration

aviramha:
Addresses #544 
Adds a new field for configuration - backlog.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -69,6 +69,8 @@ Options:
   --limit-concurrency INTEGER     Maximum number of concurrent connections or
                                   tasks to allow, before issuing HTTP 503
                                   responses.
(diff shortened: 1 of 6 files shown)


lovelydinosaur:
Seems reasonable. We should also bump up the defult for this value up to `2048`, to match `gunicorn`.

aviramha:
> Seems reasonable. We should also bump up the defult for this value up to `2048`, to match `gunicorn`.

Done!

euri10:
An addition to the docs would be cool as well

aviramha:
> An addition to the docs would be cool as well

Done :)

lovelydinosaur:
Thanks!

aviramha:
You're welcome, my pleasure supporting this wonderful project :)

PR #546, opened 2020-01-07
incorrect parameter passed

### LINKED ISSUE #523: access log show nothing when using file config
h11_impl.py line 361 self.access_logger = logger should be  self.access_logger = access_logger

Loknar:
Fixes #523

Causes logger mixup when h11 is used, like on windows, logs that should to to uvicorn.access go to uvicorn.error instead

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -360,7 +360,7 @@ class RequestResponseCycle:
         self.transport = transport
         self.flow = flow
         self.logger = logger
-        self.access_logger = logger
+        self.access_logger = access_logger
         self.access_log = access_log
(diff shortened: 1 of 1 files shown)


Loknar:
Same as https://github.com/encode/uvicorn/pull/587

unsure what changing `propagate` in `"uvicorn.access": {"handlers": ["access"], "level": "INFO", "propagate": False},` to `True` achieves, but if it is necessary feel free to close this pull and accept the other one

edit: having propagate default to True is somewhat explained here: https://github.com/encode/uvicorn/issues/523#issuecomment-598522664

ClericPy:
So late to see this pr......

euri10:
closing this as it was fixed part of #645 the access_logger was indeed on the wrong logger.

Issue #528, opened 2019-12-18
Add support for --reload to monitor additional file types.


steven-sprouse-sndk:
The "reload" process currently only monitors ".py" files in various directories. I have a changes that will pass in a list of additional "reload_suffixes" that the process will monitor. This allows the service to monitor data files in addition to code files.

Any feedback on whether this is useful to others?

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/uvicorn/issues/528">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/528/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/uvicorn/issues/528/pledge.svg">
</picture>
</a>

hanneskuettner:
@rafalp could we reopen this now that #659 has reverted the feature of #646?
We could use this to flush out the intended API that we want to see implemented. 

We should do something like @euri10 suggested in https://github.com/encode/uvicorn/pull/659#issuecomment-621084234.

`reload_watch`: a glob pattern for watched files / file types that is handled with `fnmatch` which defaults to `*.py` for backwards compatibility
`reload_ignore`: a global pattern for ignored files , which defaults to empty

The logic should be imho: iterate all files that are matched by `reload_watch` and exclude all that are matched by `reload_ignore`.

PR #550, opened 2020-01-12
Add statreload config option reload_types

WillDaSilva:
Solves #528 
Helps #102 

`reload_types` is a config option that allows the specification of multiple file suffixes to be monitored.

### DIFF
--- a/tests/supervisors/test_statreload.py
+++ b/tests/supervisors/test_statreload.py
@@ -27,6 +27,9 @@ def test_statreload():
 
 
 def test_should_reload(tmpdir):
+    """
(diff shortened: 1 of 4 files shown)


rafalp:
Not sure if I like an additional option here. We could simply watch everything that is not:

- `.dotted`: because it's dev tools configs and/or caches
- `.pyc`: because changes in those are triggered by changes in `.py`

We've also merged an alternative strategy for watching changes that uses file system events (aka the watchdog strategy) that will also need updating.

rafalp:
Thank you for taking the time to open PR! While it wasn't merged, it was helpful as a reference for me <3

PR #556, opened 2020-01-20
reload option only with IPv4 host fix

### LINKED ISSUE #553: reload option works only with IPv4 host
I can not specify --reload option with IPv6 only --host.
`uvicorn --reload --host="::" example:app`

The server binds only to IPv4 host.

### LINKED ISSUE #791: Built address of server in logs is wrong in case of a ipv6 host
Prompted by the below issue, we are missing the brackets around the host in case of a ipv6 host, the below should be 
`http://[::]:8000`

```
❯ uvicorn app:app --host="::"
INFO:     Started server process [15903]
INFO:     Uvicorn running on http://:::8000 (Press CTRL+C to quit)
INFO:     ::1:35892 - "GET / HTTP/1.1" 200 OK
INFO:     ::1:35892 - "GET /favicon.ico HTTP/1.1" 200 OK
```

_Originally posted by @euri10 in https://github.com/encode/uvicorn/issues/782#issuecomment-700578964_

euri10:
fixes https://github.com/encode/uvicorn/issues/553

correct the url printed in the logs if ipv6 is used, using brackets arout the ip as recommanded in https://www.ietf.org/rfc/rfc2732.txt

so fixes https://github.com/encode/uvicorn/issues/791 as well


euri10:
anyone knows if the ipv6 link could be clickable or if that's a terminal issue ? most likely the terminal but you never know


Imgur

Issue #567, opened 2020-02-07
incompatible httptools 0.1.0


alfredo-ardito:
ERROR: uvicorn 0.11.2 has requirement httptools==0.0.13; sys_platform != "win32" and sys_platform != "cygwin" and platform_python_implementation != "pypy", but you'll have httptools 0.1.0 which is incompatible.

lovelydinosaur:
Are you saying "httptools 0.1.0 is available, you folks should update the dependancy to that now", or something else?

Yup would be happy to accept any pull requests upgrading to the latest version.

* Do we know if there are any API differences between 0.0.13 and 0.1.0?
* If so, do any of those API differences affect us?

One simple place to start with this would be issuing a pull request that updates our `setup.py`, and seeing if any of the tests fail with the latest version.

alfredo-ardito:
I'll check the differences between 0.0.13 and 0.1.0 and let you know.

PR #569, opened 2020-02-09
Update httptools compatibility

alippai:
https://github.com/encode/uvicorn/issues/567


alippai:
In https://github.com/MagicStack/httptools/issues/52 uvicorn is asked to change it's code so it doesn't rely on httptools' private packages.

alippai:
@tomchristie thanks!

PR #580, opened 2020-02-22
Document proxy-headers middleware wildcard feature

aogier:
Hello! I just found myself wondering if there was a wildcard for proxy-headers `--forwarded-allow-ips` and to figure it out, I had to find it in the code. I think it could be documented so here it is.

What do you think?

Thank you for uvicorn, ciao!


lovelydinosaur:
👍

PR #591, opened 2020-03-17
Add search for a trusted host in ProxyHeadersMiddleware

### LINKED ISSUE #590: Proxy Middleware grabs the last IP in `x-forwarded-for`, not the first
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For says the first IP in is a client IP, but ProxyHeadersMiddleware grabs the last IP:
https://github.com/encode/uvicorn/blob/9d9f8820a8155e36dcb5e4d4023f470e51aa4e03/uvicorn/middleware/proxy_headers.py#L41

b0g3r:
Fixes #590 

If we have `*` in trusted proxies we should return the first host in the `X-Forwarded-For` as the client address.

In other cases, we should iterate over `X-Forwarded-For` hosts backward and return the first host which isn't included in trusted proxies

### DIFF
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -52,7 +52,7 @@ async def test_proxy_headers_trusted_hosts(trusted_hosts, response_text):
             ["127.0.0.1", "10.0.2.1", "192.168.0.2"],
             "Remote: https://1.2.3.4:0",
         ),
-        # order isn't matter
+        # order doesn't matter
         (
             ["10.0.2.1", "192.168.0.2", "127.0.0.1"],
             "Remote: https://1.2.3.4:0",
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -13,34 +13,79 @@ async def app(scope, receive, send):
     await response(scope, receive, send)
 
 
-app = ProxyHeadersMiddleware(app, trusted_hosts="*")
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("trusted_hosts", "response_text"),
+    [
+        # always trust
+        ("*", "Remote: https://1.2.3.4:0"),
+        # trusted proxy
+        ("127.0.0.1", "Remote: https://1.2.3.4:0"),
+        (["127.0.0.1"], "Remote: https://1.2.3.4:0"),
+        # trusted proxy list
+        (["127.0.0.1", "10.0.0.1"], "Remote: https://1.2.3.4:0"),
+        ("127.0.0.1, 10.0.0.1", "Remote: https://1.2.3.4:0"),
+        # request from untrusted proxy
+        ("192.168.0.1", "Remote: http://127.0.0.1:123"),
+    ],
+)
+async def test_proxy_headers_trusted_hosts(trusted_hosts, response_text):
+    app_with_middleware = ProxyHeadersMiddleware(app, trusted_hosts=trusted_hosts)
+    async with httpx.AsyncClient(
+        app=app_with_middleware, base_url="http://testserver"
+    ) as client:
+        headers = {"X-Forwarded-Proto": "https", "X-Forwarded-For": "1.2.3.4"}
+        response = await client.get("/", headers=headers)
+
+    assert response.status_code == 200
+    assert response.text == response_text
 
 
 @pytest.mark.asyncio
-async def test_proxy_headers():
-    async with httpx.AsyncClient(app=app, base_url="http://testserver") as client:
-        headers = {"X-Forwarded-Proto": "https", "X-Forwarded-For": "1.2.3.4"}
+@pytest.mark.parametrize(
+    ("trusted_hosts", "response_text"),
+    [
+        # always trust
+        ("*", "Remote: https://1.2.3.4:0"),
+        # all proxies are trusted
+        (
+            ["127.0.0.1", "10.0.2.1", "192.168.0.2"],
+            "Remote: https://1.2.3.4:0",
+        ),
+        # order isn't matter
+        (
+            ["10.0.2.1", "192.168.0.2", "127.0.0.1"],
+            "Remote: https://1.2.3.4:0",
+        ),
+        # should set first untrusted as remote address
+        (["192.168.0.2", "127.0.0.1"], "Remote: https://10.0.2.1:0"),
+    ],
+)
+async def test_proxy_headers_multiple_proxies(trusted_hosts, response_text):
+    app_with_middleware = ProxyHeadersMiddleware(app, trusted_hosts=trusted_hosts)
+    async with httpx.AsyncClient(
+        app=app_with_middleware, base_url="http://testserver"
+    ) as client:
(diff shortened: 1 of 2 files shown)


b0g3r:
I found a discussion about this feature: https://github.com/encode/uvicorn/commit/723d3dc225cc2658634bdbfa29570ceedd6d40ed#r36079732

So, I'll try to find examples from other servers with this feature and make comparison

uvicorn/middleware/proxy_headers.py
```diff
@@ -14,11 +14,19 @@ class ProxyHeadersMiddleware:
     def __init__(self, app, trusted_hosts="127.0.0.1"):
         self.app = app
         if isinstance(trusted_hosts, str):
-            self.trusted_hosts = [item.strip() for item in trusted_hosts.split(",")]
+            self.trusted_hosts = {item.strip() for item in trusted_hosts.split(",")}
         else:
-            self.trusted_hosts = trusted_hosts
+            self.trusted_hosts = set(trusted_hosts)
```
b0g3r, line 20: For speedup `str not in collection` operations

uvicorn/middleware/proxy_headers.py
```diff
@@ -14,11 +14,21 @@ class ProxyHeadersMiddleware:
...
         self.always_trust = "*" in self.trusted_hosts
 
+    def get_trusted_client_host(
+        self, x_forwarded_for_hosts
+    ):  # type: (List[str]) -> str
+        if self.always_trust:
+            return x_forwarded_for_hosts[0]
+
+        for host in reversed(x_forwarded_for_hosts):
```
iamlordaubrey, line 28: Hi @b0g3r,

Doesn't adding `reversed()` here negate the idea of returning the original client ip address in the x_forwarded_for list (which should be the very first ip address)?

iamlordaubrey (commented):
From my understanding, the idea here is to return the actual client ip address. This as it is returns the last ip address (which could be a proxy address)

uvicorn/middleware/proxy_headers.py
b0g3r, line 29: It's just a security feature: we should return only the last non-trusted host.

Trivial case:
 - Proxy-server A
 - Proxy-server B
 - Client C
 - We trust both of proxies (`trusted_hosts = A, B`) or trust every proxy (`trusted_hosts = *`)

So in this scenario, a header will be `X-Forwarded-For: C, B, A` and we can securely extract C as a trusted client address.

Unfortunately, IRL we have a situation when we have a chain of proxies, and we can trust only a few of them, e.g:
 - Proxy-server A
 - Proxy-server B
 - Client C
 - We trust only A (`trusted_hosts = A`)

In this case, non-trusted proxy-server B can replace the real client address, so we should return the first non-trusted host and it will be `B`. Because the chain is actually reverted, we should check it also with `reversed`.

tl;dr: In a simple case, we can trust all of the proxies with `trusted_hosts = *` and it will return real client address (the first in X-Forwarded-For)

iamlordaubrey:
> From my understanding, the idea here is to return the actual client ip address. This as it is returns the last ip address (which could be a proxy address)

Ok, I think I get it now. Thanks

rlittlefield:
This change is working well for me. The old way of specifying N proxies worked better for my use case because google cloud load balancers + kubernetes services always adds two hosts to the header, and I don't get to know what those are in advance. I can get this to work with your changes. Thank you!

iamlordaubrey:
@tomchristie please can someone take a look at this PR? Sorry for the inconvenience 🙏

ghost (approved):

ghost:
Is there any chance to release this any time soon? 
Slow API (https://pypi.org/project/slowapi/) can not work without this change.

uvicorn/middleware/proxy_headers.py
```diff
@@ -14,11 +14,21 @@ class ProxyHeadersMiddleware:
...
+            if host not in self.trusted_hosts:
+                return host
```
lovelydinosaur, line 31: Do we need a fall through case so there's no way we can end up returning `None` here?

euri10: I fail to see how it would be possible

euri10 (commented):
lgtm, this would be a great addition, is there something we can do @tomchristie to alleviate your concerns on the above point ?

uvicorn/middleware/proxy_headers.py
dp-rufus: It would happen if all the IPs in the chain were explicitly listed in `trusted_hosts` - but this would only occur if `trusted_hosts` were incorrectly set and/or the first reverse proxy in the chain failed to set `x-forwarded-for` to the client's IP (not our fault).

In this case, is there any other justifiable return value than `None`? I can't think of one

dp-rufus:
@euri10 I'm not sure a fall-through case is necessary above - `None` might be the most sensible return value under the (unlikely) circumstances that would trigger this behaviour.

@tomchristie could you take another look at this PR please? I've just been bitten by this problem in the wild and it would be good to get this merged. Thanks

euri10:
updated it in https://github.com/encode/uvicorn/compare/master...euri10:b0g3r-uvicorn-590?expand=1 to solve conflicts because of the "new" test suite (original PR was done previous to the test suite being "revamped")

b0g3r:
@euri10 Thank you for your comment!

I rebased this PR onto master and fix related issues (`127.0.0.1` instead of `testclient` and others.) The last piece to merge is code review, right?

tests/middleware/test_proxy_headers.py
```diff
@@ -13,34 +13,79 @@ async def app(scope, receive, send):
...
+    ("trusted_hosts", "response_text"),
+    [
+        # always trust
+        ("*", "Remote: https://1.2.3.4:0"),
+        # all proxies are trusted
+        (
+            ["127.0.0.1", "10.0.2.1", "192.168.0.2"],
+            "Remote: https://1.2.3.4:0",
+        ),
+        # order isn't matter
```
euri10:
```suggestion
        # order doesn't matter
```

euri10 (approved):
just correct this small typo in the comment and we gtg I think, this is a great addition thanks @b0g3r

uvicorn/middleware/proxy_headers.py
```diff
@@ -38,7 +49,10 @@ async def __call__(self, scope, receive, send):
                     # X-Forwarded-For header. We've lost the connecting client's port
                     # information by now, so only include the host.
                     x_forwarded_for = headers[b"x-forwarded-for"].decode("latin1")
-                    host = x_forwarded_for.split(",")[-1].strip()
+                    x_forwarded_for_hosts = [
+                        item.strip() for item in x_forwarded_for.split(",")
+                    ]
+                    host = self.get_trusted_client_host(x_forwarded_for_hosts)
```
zffocussss, line 55: what if the x-forwarded-for format is IP1:PORT1,IP2:PORT2,which is common in the L7 load balance? the port here is always set zero which is not reasonable.

PR #592, opened 2020-03-17
Add support for `forwarded-allow-ips` to gunicorn worker

### LINKED ISSUE #589: Uvicorn via gunicorn worker doesn't respect `--forwarded-allow-ips`
I use uvicorn in docker as uvicorn-worker for gunicorn for my fastapi app. My application needs to know the real client IP of each request, so I use proxy-server with the `X-Forwarded-For` header.

Gunicorn has a special option to change proxy-ip to real-ip, so I running gunicorn like this:
```
gunicorn \
  ppm_telegram_bot.api:app \
  --forwarded-allow-ips="*" 
  --worker-class=uvicorn.workers.UvicornWorker \
  --bind=0.0.0.0:$PORT
```

Because I'm in a container, my WSGI/ASGI server receives requests not from the localhost, but from the docker network.

But uvicorn-worker doesn't respect gunicorn's `forwarded-allow-ips`, so in `ProxyHeadersMiddleware.trusted_hosts` I receive default `127.0.0.1` and proxy-ip instead of real-ip.
https://github.com/encode/uvicorn/blob/9d9f8820a8155e36dcb5e4d4023f470e51aa4e03/uvicorn/middleware/proxy_headers.py#L14-L17

It looks like uvicorn-worker can forward this information to config via `config_kwargs`: https://github.com/encode/uvicorn/blob/9d9f8820a8155e36dcb5e4d4023f470e51aa4e03/uvicorn/workers.py#L28-L35

I could do PR with this change, if required 🙌

b0g3r:
Fixes #589

### DIFF
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -2,6 +2,7 @@ import asyncio
 import logging
 
 from gunicorn.workers.base import Worker
+
 from uvicorn.config import Config
 from uvicorn.main import Server
 
@@ -32,6 +33,7 @@ class UvicornWorker(Worker):
             "timeout_notify": self.timeout,
             "callback_notify": self.callback_notify,
             "limit_max_requests": self.max_requests,
+            "forwarded_allow_ips": self.cfg.forwarded_allow_ips,
         }
 
         if self.cfg.is_ssl:


florimondmanca (approved):
Nice one, thanks @b0g3r!

Irfan-wani:
the flag is still not respected. I tried with environment variable too, but none of them works

Kludex:
> the flag is still not respected. I tried with environment variable too, but none of them works

I can't reproduce. Seems to be working just fine.

PR #595, opened 2020-03-19
Fix missed %s in H11 protocol logging

### LINKED ISSUE #594: Typo in protocol logging.
Line 137 of /protocols/http/h11_impl.py has a typo:
`self.logger.log(TRACE_LOG_LEVEL, "%Connection lost", prefix)` should be `self.logger.log(TRACE_LOG_LEVEL, "%sConnection lost", prefix)`. This causes an exception to be thrown when logging at trace level.

b0g3r:
Closes #594


lovelydinosaur:
Thanks!

PR #597, opened 2020-03-21
Added asgi version and spec_version to scope

### LINKED ISSUE #578: HTTP Scope missing fields
https://asgi.readthedocs.io/en/latest/specs/www.html#connection-scope

>The connection scope contains:
    ...
    asgi["version"] (Unicode string) – Version of the ASGI spec.
    asgi["spec_version"] (Unicode string) – Version of the ASGI HTTP spec this server understands; one of "2.0" or "2.1". Optional; if missing assume 2.0.

I sent a request through uvicorn and noticed these fields are missing. Additionally searching the code for "spec_version" does not return any hits so I think these fields are not being handled. Potentially these were added in a later ASGI version that is not supported or I may be misunderstanding what the ASGI docs mean.

euri10:
attempt at solving #578 

Fixes #578


uvicorn/protocols/http/h11_impl.py
```diff
@@ -188,6 +188,7 @@ def handle_events(self):
                 raw_path, _, query_string = event.target.partition(b"?")
                 self.scope = {
                     "type": "http",
+                    "asgi": {"version": "2.1", "spec_version": "2.1"},
```
florimondmanca: It seems that `scope["asgi"]["version"]` actually refers to either ASGI 2 or ASGI 3, i.e. the [ASGI spec](https://github.com/django/asgiref/blob/master/specs/asgi.rst):

> The key `scope["asgi"]` will also be present as a dictionary containing a `scope["asgi"]["version"]` key that corresponds to the ASGI version the server implements. If missing, the version should default to "2.0".

(OTOH `spec_version` refers to the ASGI [HTTP/WebSocket spec](https://github.com/django/asgiref/blob/master/specs/www.rst).)

We support both versions, and which one is used depends on the app callable interface - that information is available as `config.interface`.

So adding this to `Config`…

```python
class Config:
    @property
    def asgi_version(self) -> str:
        return {"asgi2": "2.0", "asgi3": "3.0"}[self.interface]
```

We should be able to use:

```suggestion
                    "asgi": {"version": config.asgi_version, "spec_version": "2.1"},
```

florimondmanca (commented):
Great start! Did a bit of digging around `scope["asgi"]["version"]`, my notes below.

uvicorn/protocols/http/h11_impl.py
euri10: done

euri10:
@florimondmanca should I write a test for it @florimondmanca ? seems a little bit useless to me but if you want

florimondmanca:
Yes ideally let's add a test to prevent regressions. Surely there's somewhere we test the content of `scope`?

tests/test_config.py
```diff
@@ -66,3 +66,29 @@ def test_ssl_config(certfile_and_keyfile):
     config.load()
 
     assert config.is_ssl is True
+
+
+async def asgi3app(scope, receive, send):
+    pass
```
florimondmanca: Seems like we can reuse the `asgi_app` defined at the top of this file

tests/test_config.py
```diff
@@ -66,3 +66,29 @@ def test_ssl_config(certfile_and_keyfile):
...
+
+
+def asgi2app(scope):
+    async def asgi(receive, send):
+        raise RuntimeError("Something went wrong")
+
+    return asgi
```
florimondmanca: Should we move this alongside `asgi_app` and `wsgi_app`, as `asgi2_app`?

tests/test_config.py
```diff
@@ -66,3 +66,29 @@ def test_ssl_config(certfile_and_keyfile):
...
+]
+
+
+@pytest.mark.parametrize(
+    "asgi2or3_app, expected_interface, expected_scope", asgi_scope_data
+)
+def test_interface_config(asgi2or3_app, expected_interface, expected_scope):
+    config = Config(app=asgi2or3_app)
+    config.load()
+    assert config.interface == expected_interface
```
florimondmanca: This seems to duplicate some of the testing we already do elsewhere here… Maybe we can repurpose this as a unit test for the new `.asgi_version` attribute?

```suggestion
@pytest.mark.parametrize(
    "app, asgi_version", [(asgi_app, "3.0"), (asgi2_app, "2.0")]
)
def test_asgi_version(app, asgi_version):
    config = Config(app=app)
    config.load()
    assert config.asgi_version == asgi_Version
```

florimondmanca (commented):
Forgot to submit this one :-)

tests/test_config.py
euri10: ok I was making changes while you were commenting so I incorporated the test changed you want and still kept the protocol test for scopes

tests/test_config.py
euri10: ok

tests/test_config.py
euri10: that said I dont think this duplicates anything, or i failed to see any mention of interface testing, this could have been in the auto test too but there wasnt anything, but maybe I didnt search hard enough, always an option

euri10:
I put the latest master changes, which caught some small stuff with flake8 !

I think it should be ok now, would be great if you have time to take a look at it @florimondmanca

florimondmanca (approved):
Yup, looks good!

Issue #551, opened 2020-01-13
[FEATURE] Make it possible to restrict keys in scope.headers log entries


LasseGravesen:
I had an issue pop up when I updated uvicorn to a newer version, specifically it seems that scope.headers got added to the log output, and there appears to be no way to filter what goes into that log.

Here is an example log:
```
{
	"message": "255.255.255.255:37326 - \"POST /api/v1/count HTTP/1.1\" 200",
	"timestamp": "2020-01-13T12:33:44.384008",
	"levelno": 20,
	"levelname": "INFO",
	"name": "uvicorn.access",
	"status_code": 200,
	"scope": {
		"type": "http",
		"http_version": "1.1",
		"server": ["0.0.0.0", 8080],
		"client": ["255.255.255.255", 37326],
		"scheme": "http",
		"method": "POST",
		"root_path": "",
		"path": "/api/v1/count",
		"raw_path": "b'/api/v1/count'",
		"query_string": "b''",
		"headers": [["b'host'", "b'lb.api.production.company.io:54776'"], ["b'sentry-trace'", "b'00-27422174e7f14c9aac3572a42d34006f-8f059e2531e39fb5-00'"], ["b'user-agent'", "b'test (version 2019.12.1)'"], ["b'accept-encoding'", "b'gzip, deflate'"], ["b'accept'", "b'*/*'"], ["b'content-type'", "b'application/json'"], ["b'authorization'", "b'Bearer fake-bearer-token'"], ["b'content-length'", "b'137'"], ["b'x-forwarded-port'", "b'54776'"], ["b'x-forwarded-for'", "b'255.255.255.255'"], ["b'connection'", "b'close'"]],
		"fastapi_astack": "<contextlib.AsyncExitStack object at 0x7f62ec631090>",
		"app": "<fastapi.applications.FastAPI object at 0x7f62eeda7710>",
		"state": {
		},
		"router": "<fastapi.routing.APIRouter object at 0x7f62eeda7bd0>",
		"endpoint": "<function count at 0x7f62ee6bc560>",
		"path_params": {}
	}
}
```

The problem are headers like `["b'authorization'", "b'Bearer fake-bearer-token'"]` which can give away sensitive information like API Keys or valid Bearer tokens.

I would like a feature that gives me the option to exclude specific headers that may include sensitive information that I do not want to be logged.
The current way I do this is by creating a custom formatter that drops the scope.headers field entirely.

lovelydinosaur:
[...]

PR #599, opened 2020-03-21
WIP: ability to blur some headers values in access logs to avoid leaking sensitive info

euri10:
rough draft for #551 

It adds a config kwarg `exclude_scope_keys` that takes a list of bytes

so if you pass for instance `exclude_scope_keys=[b"authorization"]` you're getting access logs with the headers values whose key match authorization replaced by `****`, no bearer tokens leaked.

I just wrote one test and there are some further checks that should be made, but I just wanted to see if in current state that would be something to work more on.

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -200,9 +200,7 @@ def test_request_logging2(path, protocol_cls, caplog):
     caplog.set_level(logging.INFO, logger="uvicorn.access")
     logging.getLogger("uvicorn.access").propagate = True
 
-    app = Response(
-        "Hello, world", media_type="text/plain",
-    )
+    app = Response("Hello, world", media_type="text/plain",)
 
     protocol = get_connected_protocol(app, protocol_cls, log_config=None)
     protocol.data_received(get_request_with_query_string)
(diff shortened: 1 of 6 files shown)


uvicorn/protocols/http/h11_impl.py
```diff
@@ -360,8 +364,9 @@ def __init__(
         self.transport = transport
         self.flow = flow
         self.logger = logger
-        self.access_logger = logger
```
euri10, line 363: this is not part of the PR but it triggered a bug in my new test so i figured i'd fix it

tests/protocols/test_http.py
```diff
@@ -1,14 +1,13 @@
...
 import logging
 
-import h11
 import pytest
 
 from tests.response import Response
 from uvicorn.config import Config
 from uvicorn.main import ServerState
 from uvicorn.protocols.http.h11_impl import H11Protocol
-from uvicorn.protocols.websockets.wsproto_impl import WSProtocol
```
euri10, line 11: those 2 imports removal serve no purpose in the PR too, they are just not used

PR #612, opened 2020-04-06
Access logger variable fixed, should_upgrade never used

euri10:
Small fix in h11 implementation, one variable is never used, the other was assigned incorrectly

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -200,7 +200,6 @@ class H11Protocol(asyncio.Protocol):
                     "headers": self.headers,
                 }
 
-                should_upgrade = False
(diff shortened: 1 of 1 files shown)


euri10:
not sure why I closed this, still valid I think

uvicorn/protocols/http/h11_impl.py
```diff
@@ -200,7 +200,6 @@ def handle_events(self):
                     "headers": self.headers,
                 }
 
-                should_upgrade = False
```
euri10, line 203: this is never used

euri10: already picked by flake8 on master

euri10:
dealt in #645

PR #617, opened 2020-04-08
GitHub Actions CI workflow

gvbgduh:


gvbgduh:
@tomchristie @rafalp here's a PR for a CI workflow, I wonder if you could give some feedback on how it looks.

rafalp:
What do you think about splitting this into multiple separate actions, so failing tests don't  prohibit pylint/black from running too? I've found such approach faster to iterate against as "fixed test, missed pylint elsewhere" doesn't happen that much.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+name: CI
+
+on:
+  push:
+    branches:
+      - '*'
+      - '!master'
+  pull_request:
+    branches: [ master ]
```
florimondmanca, line 12: It looks like all jobs ran twice, once due to the "push" event and one due to the "pull_request" event. Should this be…?

```suggestion
on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - master
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,68 @@
...
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        python-version: [3.6, 3.7, 3.8, pypy3]
+        exclude:
+        - os: windows-latest
+          python-version: 3.6
+        - os: windows-latest
+          python-version: pypy3
+
+    steps:
```
florimondmanca, line 29: I think CI should reuse the `scripts/` as much as possible. Otherwise there's a lot of duplication and risk of discrepencies between CI and local testing.

If we need to add stuff to `scripts/install`/`scripts/test` to make them more OS-independent (very likely) and CI-friendly, then IMO we should, and then use them here, like we do in the current `.travis.yml`.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+on:
+  push:
+    branches:
+      - '*'
+      - '!master'
+  pull_request:
+    branches: [ master ]
+
+jobs:
```
florimondmanca, line 14: One thing we don't have currently here to allow for what @rafalp suggests is `scripts/check`, i.e. a "run linters only" script. We could run it in a separate task from `scripts/test`, so that we'd know earlier on which of checks or actual tests have failed.

One quite good example of this approach is HTTPX's [`.travis.yml`](https://github.com/encode/httpx/blob/master/.travis.yml). I also use this extensively in my projects, eg see [aiometer `azure-pipelines.yml`](https://github.com/florimondmanca/aiometer/blob/master/ci/azure-pipelines.yml).

Eventually the pipeline could probably look like this:

```yaml
jobs:
  check:
    steps:
    # Runs scripts/install, then scripts/check
  test:
    strategy: ...
    steps:
    # Runs scripts/install, then scripts/test
```

florimondmanca (commented):
Great that this passes! :tada:

My two cents…

.github/workflows/ci.yaml
florimondmanca: (I realize `scripts/install` isn't currently used in `.travis.yml`, but I think we can very much make that work.)

.github/workflows/ci.yaml
gvbgduh: It actually replicates the current travis behaviour, tho I agree it might be redundant.
I'm also trying to cover commits to other branches as it's a nice feedback, while it's WIP.
But the proposed approach I see a problem that it'll skip other branchs.
I think it might be
```on:
  push:
    branches:
      - '!master'
  pull_request:
    branches: [ master ]
```
But i'll check it

.github/workflows/ci.yaml
gvbgduh: yes, jobs are available as well, but i'm not sure the matrix {os|python-verion} can be reused at a higher level, so we can duplicate it actually save steps with installed environments as artifacts and pass it further.

.github/workflows/ci.yaml
gvbgduh: yes, it's convenient to have it one place, but scripts have one disadvantage is they concatenate logs from all steps, which is less convenient for failures in the middle.
And there can be quite a bit of them, `mypy`, `flake8`, `autoflake`, `black`, `coverage`, `bandit(?)` etc.
Per step granularity allows seeing the status and logs within the step.

gvbgduh: it's also what makes `tox` convenient, as keeps steps the same locally and in CI.
tho github actions have a local runner - https://github.com/nektos/act

lovelydinosaur:
Superseeded by #649. Apologies that I never found the time to review this.

The style in #649 is the way I'd like to see us tying together the `/scripts/...` together with the CI workflow. Tho there are some things this PR does a better job on. (eg. #649 doesn't yet include a windows build in the matrix, which is worth a follow up PR, amongst other things.)

Issue #549, opened 2020-01-10
[QUESTION]How to specify application path when running uvicorn outside of the application directory.


yezyilomo:
I thought there would be an option like `--chdir` in Gunicorn but I didn't find one on the documentation, at first I thought that's what `--root-path` option is used for, turns out I was wrong.

fsecada01:
Would also like to know if this is a thing.  This would make deployments much easier.

euri10:
Using PYTHONPATH maybe, if I understood correctly what you want?

fsecada01:
No, something more like `--chdir backend\src` for deploying uvicorn

anandandy139:
I am also looking for something like this. Could you help ?

lovelydinosaur:
I think we could probably accept a pull request adding that, yup.

Sub0X:
Following this question, is it possible to also have an app-dir kwarg for uvicorn.Config?

PR #619, opened 2020-04-09
Add --app-dir option for running uvicorn from any location

yezyilomo:
This PR is an attempt to solve #549 
I chose `--app-dir` to be the name of the option, please let me know if it makes sense, I've also updated the documentation to include `--app-dir` option, please check if I have added it to the right place.


docs/settings.md
```diff
@@ -8,6 +8,7 @@ equivalent keyword arguments, eg. `uvicorn.run("example:app", port=5000, reload=
 ## Application
 
 * `APP` - The ASGI application to run, in the format `"<module>:<attribute>"`.
+* `--app-dir <str>` - Specifty application directory.
```
florimondmanca: This place in the docs looks sensible to me. :-)

uvicorn/main.py
```diff
@@ -256,6 +256,9 @@ def print_version(ctx, param, value):
     is_eager=True,
     help="Display the uvicorn version and exit.",
 )
+@click.option(
+    "--app-dir", "app_dir", default=".", help="Specifty application directory."
```
florimondmanca: I think we can add `show_default=True`, so that `.` shows as a default when running `$ uvicorn --help`:

```suggestion
    "--app-dir", "app_dir", default=".", show_default=True, help="Specifty application directory."
```

docs/settings.md
florimondmanca: Maybe the description could be a bit more detailed, maybe something like this?

```suggestion
* `--app-dir <str>` - Look for `APP` in the specified directory. Defaults to `.`.
```

florimondmanca (commented):
My two cents on the `--app-dir` vs `--chdir` naming…

The gunicorn description of [`--chdir`](https://docs.gunicorn.org/en/stable/settings.html#chdir) is:

> Chdir to specified directory before apps loading.

This is not *accurately* what we do here — we don't `cd` or really change the working directory. Instead, we allow Python to find the `APP` in the specified directory by adding it to `sys.path`.

So `--app-dir` is _probably_ okay IMO?

uvicorn/main.py
yezyilomo: Should we update help to "Look for `APP` in the specified directory. Defaults to `.`."?

anein:
Waiting the approval of this meaningful PR. 😭 

//cc @florimondmanca

yezyilomo:
> Waiting the approval of this meaningful PR. :sob:
> 
> //cc @florimondmanca

& //cc @tomchristie

docs/deployment.md
```diff
@@ -87,6 +87,7 @@ Options:
                                   [default: TLSv1]
   --header TEXT                   Specify custom default HTTP response headers
                                   as a Name:Value pair
+  --app-dir TEXT                  Look for APP in the specified directory. Defaults to '.'
```
lovelydinosaur: What do we think about changing this to "Defaults to the current working directory" ?

lovelydinosaur:
Yup, I like this, thanks! One query about the phrasing of the help text... should we maybe prefer "Defaults to the current working directory."?

docs/deployment.md
yezyilomo: Yes "Defaults to the current working directory" looks good.

lovelydinosaur (approved):
Fab, thanks!

PR #618, opened 2020-04-09
Cleanup fixes for watchdog reload

florimondmanca:
Cleanup for #609 

- `.run()` method on `BaseReloader` was duplicated. (Could have been detected by flake8, but it's not part of the toolchain here. Has it been removed?)
- `setup.py` is broken due to a typo on the `extras_require` declaration:

```console
    ERROR: Command errored out with exit status 1:
     command: /Users/florimond.manca/Developer/encode-projects/uvicorn/venv/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/Users/florimond.manca/Developer/encode-projects/uvicorn/setup.py'"'"'; __file__='"'"'/Users/florimond.manca/Developer/encode-projects/uvicorn/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info
         cwd: /Users/florimond.manca/Developer/encode-projects/uvicorn/
    Complete output (1 lines):
    error in uvicorn setup command: 'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers.
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

PR #620, opened 2020-04-09
Improve shutdown robustness when using `--reload` or multiprocessing

### LINKED ISSUE #364: docker-compose graceful shutdown of uvicorn --reload

euri10:
rework of https://github.com/encode/uvicorn/pull/522 because master moved too much and my brain couldn't rebase.

Fixes #364


uvicorn/supervisors/multiprocess.py
```diff
@@ -23,11 +23,32 @@ def __init__(self, config, target, sockets):
...
+                    if os.WIFEXITED(status):
+                        logger.debug(
+                            "process returning status exited via the exit() system call"
+                        )
+                    elif os.WIFSIGNALED(status):
+                        logger.debug(
+                            "process returning status was terminated by a signal"
+                        )
+                    elif os.WIFSTOPPED(status):
+                        logger.debug("process returning status was stopped")
```
florimondmanca: Do we _need_ all this debug information? I think we should keep this as lightweight as possible.

uvicorn/supervisors/multiprocess.py
```diff
@@ -23,11 +23,32 @@ def __init__(self, config, target, sockets):
         self.processes = []
         self.should_exit = threading.Event()
         self.pid = os.getpid()
+        self.child_pids = []
```
florimondmanca: Getting back to https://github.com/encode/uvicorn/pull/522#discussion_r406355579 …

I haven't tried the Docker test yet, but I'm not sure I understand _why_ using a separate attribute to store PIDs of `self.processes` is necessary for this to work.

In my mind, `child_pids` is equivalent to `[p.pid for p in self.processes]`. Isn't that guaranteed/can the `.pid` of a process object change in its lifespan?

uvicorn/supervisors/basereload.py
```diff
@@ -22,22 +22,37 @@ def __init__(self, config, target, sockets):
...
     def signal_handler(self, sig, frame):
         """
         A signal handler that is registered with the parent process.
         """
+        logger.info(f"Handling signal: {sig}")
+        for child_pid in self.child_pids:
+            logger.debug(f"Attempt at killing child PID {child_pid}")
+            try:
+                os.kill(child_pid, signal.SIGINT)
+                (pid, status) = os.waitpid(child_pid, 0)
```
florimondmanca: Any thoughts on my suggestion to add a reusable util in `uvicorn/subprocess.py`? My rationale is that this is fairly "sensitive" logic, it shouldn't be duplicated.

```python
def stop_subprocess(pid):
    os.kill(pid, signal.SIGINT)
    os.waitpid(pid, 0)
```

uvicorn/supervisors/multiprocess.py
euri10: while I agree for the multiprocess part,  I fail to see how to tackle this on the reload part

florimondmanca: The reload part runs a single process - and it seems to be stored on `self.process`. We don't even need a `for` loop, only to `stop_subprocess(self.process.pid)`, correct? Or am I missing something?

euri10: you're right indeed, dunno why I've been stuck with that in mind, just had to setup the same `self.process = []` for the basereload so that the `test_statreload` and `test_watchdogreload` pass
it would fail otherwise since self.process would be non existant

uvicorn/supervisors/basereload.py
euri10: yep did it !

setup.py
```diff
@@ -50,7 +50,7 @@ def get_packages(package):
     "uvloop>=0.14.0 ;" + env_marker,
 ]
 
-extras_require = {"watchdogreload": ["watchdog>0.10,<0.11"]},
+extras_require = {"watchdogreload": ["watchdog>0.10,<0.11"]}
```
florimondmanca: Feel free to update from `master`, #618 got in earlier today. :-)

uvicorn/supervisors/basereload.py
```diff
@@ -20,21 +20,21 @@ def __init__(self, config, target, sockets):
...
     def signal_handler(self, sig, frame):
         """
         A signal handler that is registered with the parent process.
         """
-        self.should_exit.set()
+        for process in self.processes:
+            try:
+                shutdown_subprocess(process.pid)
+            except Exception as exc:
+                logger.error(f"Could not stop child process {process.pid}: {exc}")
```
florimondmanca: The `self.processes` attribute is not necessary IMO, one alternative is to check for `self.process` being set:

```suggestion
        if hasattr(self, 'process'):
            try:
                shutdown_subprocess(self.process.pid)
            except Exception as exc:
                logger.error(f"Could not stop child process {self.process.pid}: {exc}")
```

or (cleaner), add `self.process = None` in the constructor, and check for `if self.process is not None: ...` instead.

WDYT?

uvicorn/supervisors/watchdogreload.py
```diff
@@ -1,10 +1,11 @@
 import logging
 from os import path
 
-from uvicorn.supervisors.basereload import BaseReload
 from watchdog.events import PatternMatchingEventHandler
 from watchdog.observers import Observer
 
+from uvicorn.supervisors.basereload import BaseReload
```
florimondmanca, line 7: Agreed, this change is commanded by `scripts/lint`.

uvicorn/workers.py
```diff
@@ -2,6 +2,7 @@
 import logging
 
 from gunicorn.workers.base import Worker
+
```
florimondmanca: `scripts/lint` would remove this, I don't think we should leave this in.

```suggestion
```

tests/supervisors/test_multiprocess.py
```diff
@@ -19,3 +23,14 @@ def test_multiprocess_run():
     supervisor = Multiprocess(config, target=run, sockets=[])
     supervisor.signal_handler(sig=signal.SIGINT, frame=None)
     supervisor.run()
+
+
+@pytest.mark.skipif(
+    sys.platform.startswith("win"), reason="Skipping os.kill on windows"
+)
+def test_multiprocess_run2():
```
florimondmanca: Maybe a more descriptive name for this test…?

```suggestion
def test_multiprocess_cleanup_children():
    """
    Verify that stopping the multiprocess supervisor
    cleanly shuts down all child subprocesses.
    """
```

Also, given that `run` is a no-op in this test module, does this actually test anything, given that the subprocess will exit immediately (so `.shutdown()` will basically have nothing to do)?

setup.py
euri10: done

uvicorn/supervisors/basereload.py
euri10: yep

tests/supervisors/test_multiprocess.py
euri10: yep removed it

tests/supervisors/test_multiprocess.py
```diff
@@ -1,5 +1,7 @@
 import signal
 
+import pytest
```
florimondmanca: Unused import now it seems? :)

euri10: I blame flake8 not merged :grin:

uvicorn/supervisors/basereload.py
```diff
@@ -36,6 +43,11 @@ def run(self):
                 self.restart()
         self.shutdown()
 
+        while not self.should_exit.wait(0.25):
+            if self.should_restart():
+                self.restart()
+        self.shutdown()
+
```
florimondmanca: Probably an artifact from the merge…

```suggestion
```

uvicorn/workers.py
florimondmanca: Right?

florimondmanca (changes requested):
It seems like the changes to `multiprocess.py` went away…?

uvicorn/supervisors/basereload.py
euri10: ok _

uvicorn/workers.py
euri10: well scripts/lint doesnt remove it but I removed it >_

euri10:
thanks for your patience on this @florimondmanca 

> It seems like the changes to `multiprocess.py` went away…?

not sure, seems to me they are here, it's becoming hot in here, thanks for your patience !

uvicorn/supervisors/basereload.py
```diff
@@ -36,6 +43,7 @@ def run(self):
                 self.restart()
         self.shutdown()
 
+
```
florimondmanca: Last nit 🙏  (don't know if this would make CI fail, probably not as `flake8` isn't here yet)

florimondmanca (commented):
You're right about multiprocess, I had just marked the file as "viewed" but it's there 😅

uvicorn/supervisors/basereload.py
euri10: it was failing black, I should have tested it ....

florimondmanca (approved):
LGTM! :tada:

florimondmanca:
I went ahead and updated the PR title to something that I think is more descriptive of these changes.

Great work on this @euri10!

PR #621, opened 2020-04-09
Add flake8 to linters

florimondmanca:
This PR is a proposal to add `flake8` to the uvicorn toolchain.

HTTPX and other projects have shown me that having a linter (and not only a formatting fixer) is invualuable for catching subtle bugs and ensuring code quality as a whole. Eg #618 could have been avoided with `flake8` installed.

In any case, happy to pull down if that's not something we want — the original set of check tools added via #290 sticked to black, autoflake and isort, because that's what is in Starlette. (But I'd advocate for adding `flake8` to Starlette, too.)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,6 @@
 
 import os
 import re
-import sys
-import platform
 
 from setuptools import setup
 
@@ -50,7 +48,7 @@ requirements = [
     "uvloop>=0.14.0 ;" + env_marker,
 ]
 
-extras_require = {"watchdogreload": ["watchdog>0.10,<0.11"]},
+extras_require = {"watchdogreload": ["watchdog>0.10,<0.11"]}
 
 
 setup(
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,10 +27,10 @@ install:
         choco install python3;
         export PATH=/c/Python37:/c/Python37/Scripts:/c/Python38:/c/Python38/Scripts:$PATH;
         python -m pip install -U click h11 wsproto==0.13.* websockets==8.*;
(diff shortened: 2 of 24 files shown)


setup.py
```diff
@@ -50,7 +48,7 @@ def get_packages(package):
     "uvloop>=0.14.0 ;" + env_marker,
 ]
 
-extras_require = {"watchdogreload": ["watchdog>0.10,<0.11"]},
+extras_require = {"watchdogreload": ["watchdog>0.10,<0.11"]}
```
florimondmanca: These changes are necessary to make flake8 pass, we should wait for #618.

lovelydinosaur:
Looks great actually yes.

Can you talk me through the bits in `setup.cfg`?
I'd *probably* prefer to avoid another top-level config file if possible, so perhaps we could inline those or something?

florimondmanca:
> Can you talk me through the bits in setup.cfg?
> I'd probably prefer to avoid another top-level config file if possible, so perhaps we could inline those or something?

Indeed, it looks like it should be possible to pass this config as command line arguments to `flake8`.

florimondmanca:
Updated the PR. :-)

One other thing that'd be very nice too would be a `scripts/check`, and making `scripts/install` CI-friendly and platform-accommodating, like we have in HTTPX. We can use them straight in CI, but do the platform switching in the scripts. That way, those scripts work locally on those platforms too… But anyway, that's something else.

tiangolo (approved):
I love `flake8`! :nerd_face: :heart:

florimondmanca:
Thanks all! 🎉

PR #624, opened 2020-04-11
Watchgod

euri10:
~~- [ ] add a config option for the reloader ? `auto|watchdog|watchgod|statsreload` or we want to chose one or the other ? currently I import watchGod then watchDog then the default, but of course this is a draft.~~

- [x] fix test on windows

- [x] write a CustomWatcher that follows @rafalp [comment1](https://github.com/encode/uvicorn/pull/519#issuecomment-609060238) / [comment2](https://github.com/encode/uvicorn/pull/550#issuecomment-610969440) ie follow all non hidden files in `reload-dir` if I got it ?


rafalp:
> add a config option for the reloader ? auto|watchdog||statsreload

My vote is to don't do this. Provide working default out of the box (`statsreload` cos it has no dependencies) but allow users who need it to opt in to more performant option via installing new dependency. If we agree that `watchgod` is better than `watchdog` (which it appears to be the case), then we should just replace one with another to keep number of options limited and simple for the users, docs and maintainers sake. :)

euri10:
> > add a config option for the reloader ? auto|watchdog||statsreload
> 
> My vote is to don't do this. 

Agree on not doing it too!

uvicorn/supervisors/basereload.py
```diff
@@ -45,6 +45,8 @@ def run(self):
 
     def startup(self):
         message = "Started reloader process [{}]".format(str(self.pid))
+        if self.__class__.__name__ == "WatchGodReload":
+            message += " powered by watchgod"
```
rafalp:
```suggestion
            message += " using watchgod"
```

tests/supervisors/test_watchgodreload.py
```diff
@@ -13,27 +15,30 @@ def run(sockets):
...
     reloader.signal_handler(sig=signal.SIGINT, frame=None)
     reloader.run()
 
 
-def test_should_reload(tmpdir):
-    update_file = Path(os.path.join(str(tmpdir), "example.py"))
+@pytest.mark.parametrize(
+    "should_reload, file", [(True, "example.py"), (False, ".dotted")]
+)
+def test_should_reload(tmpdir, should_reload, file):
```
rafalp: Please name scenarios which parameters test, or better, create two separate tests:

- `test_should_reload_is_true_when_python_file_is_changed`
- `test_should_reload_is_false_when_dot_file_is_changed`

Parametrization obfuscates the intention between assertions and leads to bad growth when more parametrized scenarios are bolted on on top of old parameters set, further hiding the intention behind tests.

uvicorn/supervisors/watchgodreload.py
```diff
@@ -0,0 +1,63 @@
...
+        self._ignored = tuple(re.compile(r) for r in self.ignored)
+        super().__init__(root_path)
+
+    def should_watch_file(self, entry):
+        return not any(r.search(entry.name) for r in self._ignored)
+
+
+class WatchGodReload(BaseReload):
+    def __init__(self, config, target, sockets):
+
```
rafalp:
```suggestion
```

uvicorn/supervisors/watchgodreload.py
```diff
@@ -0,0 +1,63 @@
...
+from watchgod import DefaultWatcher
+
+from uvicorn.supervisors.basereload import BaseReload
+
+logger = logging.getLogger("uvicorn.error")
+
+
+class CustomWatcher(DefaultWatcher):
+
```
rafalp:
```suggestion
```

uvicorn/supervisors/basereload.py
rafalp: Not a fan of peeking into child class type. How about message template being class attribute? It could default to `"Started reloader process [{}]"` but `WatchGodReload` could define custom.

rafalp (approved):

rafalp:
Smooth work here, approved! 👍

rafalp:
Could you also update the docs and changelog? We mention watchdog in three places.

PR #512, opened 2019-12-04
Fixes bug where --log-config disables uvicorn loggers

petermorrowdev:
Fixes #511 

### Example

**`logging_config.ini`**
```
[loggers]
keys=root

[handlers]
keys=h

[formatters]
keys=f

[logger_root]
level=INFO
handlers=h

[handler_h]
class=StreamHandler
level=INFO
formatter=f
args=(sys.stderr,)

[formatter_f]
format=%(asctime)s %(name)s %(levelname)-4s %(message)s
```

**`fastapi_scratch.py`**
```
import logging
from fastapi import FastAPI

logger = logging.getLogger(__name__)
app = FastAPI()

@app.on_event("startup")
async def startup():
    logger.warning('Raising an exception on startup')
    raise Exception("Nope!")
```

### Current Behavior on `master`

```
uvicorn scratch_fastapi:app --log-config=logging_config.ini
```
```
2019-12-04 11:56:43,681 scratch_fastapi WARNING Raising an exception on startup
```

### Behavior with this PR

```
uvicorn scratch_fastapi:app --log-config=logging_config.ini
```
```
2019-12-04 12:00:05,646 uvicorn.error INFO Started server process [7523]
2019-12-04 12:00:05,651 uvicorn.error INFO Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
2019-12-04 12:00:05,652 uvicorn.error INFO Waiting for application startup.
2019-12-04 12:00:05,652 scratch_fastapi WARNING Raising an exception on startup
2019-12-04 12:00:05,654 uvicorn.error ERROR Traceback (most recent call last):
  File "/Users/peter/.virtualenvs/JR/lib/python3.7/site-packages/starlette/routing.py", line 473, in __call__
    await self.startup()
  File "/Users/peter/.virtualenvs/JR/lib/python3.7/site-packages/starlette/routing.py", line 457, in startup
    await handler()
  File "./scratch_fastapi.py", line 10, in startup
    raise Exception("Nope!")
Exception: Nope!

2019-12-04 12:00:05,655 uvicorn.error ERROR Application startup failed. Exiting.
```

PR #626, opened 2020-04-11
Config coverage increase

euri10:
Little steps towards https://github.com/encode/uvicorn/issues/102

Added tests to increase config.py coverage (a little bit of main.py too), went from 81 to 99%.

Incidentally when I added the ini log config file in that commit (aff30ae3708603c70f96570c5c4a9ea909915276) it broke the http protocol logs tests [here](https://github.com/encode/uvicorn/blob/a839da2564e5168c6e99c3b22819950f26f7d178/tests/protocols/test_http.py#L171)  and it looks the fix from #512 is necessary or no logs are sent indeed.

Also trying to understand the specifics of <3.7 for loggers it appears to me the check is not necessary anymore (last commit 32b2c727f31d9429427838b9afc2b587d9f28f9d)


euri10:
tests pass fine locally, dunno what travis is currently doing, all tests are failing and logs are not possible to fetch

euri10:
need to get why pypy timeout

lovelydinosaur (approved):
Seems okay, yup.

I'd generally suggest that we try to keep PRs split up wherever ossible (eg. the change in the `disable_existing_loggers=False` seems unrelated to the coverage, as does the `sys.version_info < (3,7)` case. Having said that I don't think we should let those block this from progressing.

Looks like there's some conflicts that'll need resolving, but otherwise, great - let's move on with this! 😃

euri10:
> Seems okay, yup.
> 
> I'd generally suggest that we try to keep PRs split up wherever ossible (eg. the change in the `disable_existing_loggers=False` seems unrelated to the coverage, as does the `sys.version_info < (3,7)` case. Having said that I don't think we should let those block this from progressing.
> 
> Looks like there's some conflicts that'll need resolving, but otherwise, great - let's move on with this!

ok I rebased it and all tests pass

I'm sorry for not separating them and will keep them seperated next time.

just one thing I'd like you to confirm,not sure you saw it, it's the change here in uvicorn/config.py

just asking because you seemed not ok about it in https://github.com/encode/uvicorn/issues/511

lovelydinosaur:
> just one thing I'd like you to confirm, not sure you saw it, it's the change here in uvicorn/config.py

Well, *ideally* keep it separate from this PR.

I'm not sure, I'm not convinced about it, but we also get a lot of feedback around the behaviour there, so... 🤷‍♂️

PR #629, opened 2020-04-12
Flake8 config file for project

djbaldey:
I found that the Flake8 parameters are hard-coded only in tests, but not used in the developer IDE. Therefore, I offer this universal option for convenience. Now you may be run command "flake8" without parameters in the root directory for checking a code or configure your IDE for automation with it.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,15 @@
+[flake8]
+ignore =
+    # Conflict W504 with old style W503. One of these things.
+    W503,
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
I'm not against us making improvements here, but we need to aim for consistency across all the encode projects, so if we're going to change something in this area, let's discuss it on https://gitter.im/encode/community and consider what we'd like across the board.

PR #631, opened 2020-04-12
Add test for uvicorn worker

euri10:
I added a test that runs an asgi app with gunicorn with the uvicorn worker.

It increases slightly coverage but I think the main benefit is to be able to test rapidly issues people may have using both servers in conjunction: I always had the feeling looking at those issues that I needed to "manually" set things up, the test provides a layout that may prove (or not :grin: ) useful


lovelydinosaur:
Looks good to me - bit uncertain about what shape the tests are in since they still don't seem to be getting reported properly from travis.

euri10:
> Looks good to me - bit uncertain about what shape the tests are in since they still don't seem to be getting reported properly from travis.

last one is ok, https://travis-ci.org/github/euri10/uvicorn/builds/674294964

PR #637, opened 2020-04-14
ReadMe updated for uvloop if trying to build in a docker image

cemtu:
If users trying to install in docker image they will get this error:

`ERROR: Failed building wheel for uvloop`

We can solve this problem by adding `RUN apk add build-base` in docker file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,12 @@ Install using `pip`:
 $ pip install uvicorn
 ```
 
+If trying to install in a docker image add this to `Dockerfile`:
+```
+RUN apk add build-base
+```
+
+
 Create an application, in `example.py`:
 
 ```python


lovelydinosaur:
This doesn't feel like something we ought to be having to add into the README.
Are you able to explain what the line means, and why it's needed?
Can anyone confirm this?

euri10:
if I get it correctly it implies that installing uvicorn in an alpine container (guessing it from the `RUN apk add build-base`) will fail unless you're adding the `build-base` package to the image because of the uvloop compilation, which in a sense is correct but rather very specific to that type of container and has nothing to do with uvicorn in my opinion.

I think most people building images would know that and/or a simple stackoverflow search will lead to the missing packages required to build the uvloop so that uvicorn can be installed.

Issue #632, opened 2020-04-12
Multiprocess supervisor hangs indefinitely if limit-max-requests hit and workers are dead


euri10:
run `uvicorn app:app --workers 2 --limit-max-requests 1 --log-level=debug` and send 2 requests.

both worker servers will be closed after each subsequent request, but the supervisor parent will hang indefinitely and ctrl+c must be hit to close it.

```
/home/lotso/PycharmProjects/uvicorn/venv/bin/python -m uvicorn app:app --workers 2 --limit-max-requests 1 --log-level=debug
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started parent process [48076]
INFO:     Started server process [48093]
INFO:     Waiting for application startup.
INFO:     Started server process [48092]
INFO:     Waiting for application startup.
WARNING:  Executing <Task pending name='Task-1' coro=<Server.serve() running at ./uvicorn/main.py:399> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7fb7a8b471c0>()] created at /home/lotso/.pyenv/versions/3.8.1/lib/python3.8/asyncio/locks.py:306> cb=[run_until_complete.<locals>.<lambda>()] created at ./uvicorn/main.py:382> took 0.159 seconds
WARNING:  Executing <Task pending name='Task-1' coro=<Server.serve() running at ./uvicorn/main.py:399> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f06035181c0>()] created at /home/lotso/.pyenv/versions/3.8.1/lib/python3.8/asyncio/locks.py:306> cb=[run_until_complete.<locals>.<lambda>()] created at ./uvicorn/main.py:382> took 0.159 seconds
INFO:     Application startup complete.
INFO:     Application startup complete.
INFO:     127.0.0.1:45616 - "GET / HTTP/1.1" 200 OK
INFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [48092]
INFO:     127.0.0.1:45618 - "GET / HTTP/1.1" 200 OK
INFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [48093]
```

Kludex:
Given that `Multiprocess` doesn't monitor the processes it creates, this is quite normal.

[...]

PR #641, opened 2020-04-18
Add check to multiprocess supervisor 

euri10:
should solve #632 

it adds a check to see if workers are still alive (limit_max_request will kill them) annd in that case it sets the should_exit event thread


uvicorn/supervisors/multiprocess.py
```diff
@@ -39,9 +40,14 @@ def signal_handler(self, sig, frame):
 
     def run(self):
         self.startup()
-        self.should_exit.wait()
+        while not self.should_exit.wait(0.25):
+            if not self.workers_alive():
+                self.should_exit.set()
         self.shutdown()
 
+    def workers_alive(self):
```
rafalp, line 48: Is it possible for this function to return negative number, eg. when old processes haven't finished shutdown but new ones have already spawned?

euri10: I dont see a situation where it could happen, the max number of workers is the one when it's initialized and set to `self.workers_number`, after that it can only decrease, 

for this number to be negative there would be the need for new workers to be spawned, but in our case it only happens once at start, if it dies it dies

rafalp (approved):

uvicorn/supervisors/multiprocess.py
```diff
@@ -16,13 +16,14 @@
 
 
 class Multiprocess:
-    def __init__(self, config, target, sockets):
+    def __init__(self, config, target, sockets, workers_number):
```
lovelydinosaur, line 19: It doesn't look like we should need `workers_number` here.
In startup we're using `self.config.workers` - can we not rely on that instead? (Or am I missing something?)

lovelydinosaur (changes requested):

euri10:
will reopen if necessary, I feel like the #853 might be a better take at it, not many people seem interested in graceful shutdown anyway it seems,  :grapes:

PR #647, opened 2020-04-22
Fix to get correct client ip behind proxy

sinaso:
According to Wikipedia [X-Forwarded-For](https://en.wikipedia.org/wiki/X-Forwarded-For):

> The general format of the field is:
>
> X-Forwarded-For: client, proxy1, proxy2[2]
> where the value is a comma+space separated list of IP addresses, the left-most being the original client, and each successive proxy that passed the request adding the IP address where it received the request from.

Therefore, in order to get the client address we need the leftmost ip, i.e first ip in the list.


sinaso:
I might make sense to go through the list forwarded-for ips, (from last to first) until find an untrusted "ip" and report that as the client.

With the current setup, (ie. using last ip) uvicorn always returns the ip of the proxy server not the client.

PR #648, opened 2020-04-24
Removed watchdogreload.py that was wrongly added part of last commit

euri10:
Refs #626


lovelydinosaur (approved):
I've updated this PR description with a reference to #626.
Otherwise yup, let's fix this.

PR #649, opened 2020-04-24
GitHub actions TestSuite

lovelydinosaur:
Looking into setting up GitHub actions for the TestSuite.

https://github.com/encode/uvicorn/actions

Closes #617


lovelydinosaur:
I'm going to pull this in unilaterally - Since Travis reporting of tests has been so flaky I *really* want us to get in a first pass on this.

Superseeds #617 which is also a great piece of work towards this, but doesn't quite tie in the "Scripts to rule them all" pattern with the CI in the way I'd like us to approach it.

Some *incremental* pieces of work that we can *follow up with* from this PR on...

* Add a windows build to the matrix.
* Pull the linting into a separate step that runs before the tests.
* Switch from "ubuntu-latest" to a fixed version for better reproducibility.
* Frozen requirements for better reproducibility.

Each of those needs a little discussion, and we'll want to aim for following the same style across all the encode async stack.

PR #652, opened 2020-04-28
Fix docs with correct asgi type

### LINKED ISSUE #651: Error at https://www.uvicorn.org/#the-asgi-interface
At the 'HTTP Messages' block, there is an error at sample:

It says: 

`await send({
    'type': 'http.request.start',
    'status': 200,
    'headers': [
    ]
})`

but it should be:

`await send({
    'status': 200,
    'headers': [
    ]
})`

Until then, it cause an error: RuntimeError: Expected ASGI message 'http.response.start', but got 'http.request.start'.

euri10:
fix #651

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -234,14 +234,14 @@ The instance coroutine communicates back to the server by sending messages to th
 
 ```python
 await send({
-    'type': 'http.request.start',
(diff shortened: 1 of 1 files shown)


lovelydinosaur (approved):
Fab, thanks!

PR #653, opened 2020-04-28
Automate releases

lovelydinosaur:
Hi @encode/maintainers,

So, this PR is the start of rolling out GitHub Action based releases across the encode repos.
The great thing here is that we'll be able to create releases directly through GitHub, and anyone on the maintainer team will be able to do so.

I've turned on a bunch of settings that'll help us enforce a really careful, deliberate flow...

Screenshot 2020-04-28 at 15 29 45

So:

* Everything *has* to come through a PR.
* All PRs *have* to have an approval review, including those from admins.
* Tests have to pass.
* Branches have to be up to date, which is nice since we then we only have to run tests on PRs, not also against the final "merge to master".

We'll have to see how this goes. We can make adjustments as needed, but seems like a great start.

I'm going to also issue a release PR alongside this, and we can use that to run a release, and check that there's nothing missing in the flow here.

I'll need a review from someone before I'm able to merge this.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,23 @@
+---
+name: Publish
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  publish:
+    name: "Publish release"
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - uses: "actions/checkout@v2"
+      - uses: "actions/setup-python@v1"
+        with:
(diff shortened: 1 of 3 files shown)


yeraydiazdiaz (approved):
The bash script LGTM, never used GH actions though 🙂

rafalp (approved):

tiangolo:
Awesome! :rocket: :tada:

PR #655, opened 2020-04-28
Test those docs snippets

euri10:
draft on testing examples / snippets in docs.

it essentially boils down to:

1. remove the snippets from the markdown
2. put them into a .py in `docs/src`
3. include the .py file in the md with help of the newly added `markdown-include`
3. adds a test on the snippets in `tests`

after doing it I found it's not that useful since the docs has a low number of examples so not sure it's worth it.
that sid examples could be expanded, I think in particular of logging where there seem to be lots of recurrent questions.

side-note: what would be amazing though is a markdown extension like shpinx-click to take care of the `uvicorn --help` in docs automatically but that's another story...

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -36,8 +36,8 @@ $ pip install uvicorn
 
 Create an application, in `example.py`:
 
-```python
-{!./src/index.py!}
+```python hl_lines="4"
+{!./src/index/example.py!}
 ```
 
 Run the server:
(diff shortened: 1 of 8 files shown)


lovelydinosaur:
Okay. I think we should probably pass on this. Not convinced by the cost/benefit trade-off.

lovelydinosaur:
Thanks for taking an exploratory dig into it! 😃

Issue #656, opened 2020-04-28
Reload Behavior: Documentation != Code, Unnecessary Reloads


kalzoo:
Hey! I upgraded to 0.11.4 shortly after release today, and it has a couple of quirks.

1. The documentation [here](https://github.com/encode/uvicorn/compare/0.11.3...0.11.4#diff-1819b1daaccb3d358620ade9c67e9118R22) says "python file changes" but the code says ["all non-dotfile changes"](https://github.com/encode/uvicorn/compare/0.11.3...0.11.4#diff-b0da863c7164698a2ef0fa805e4a9197R40).
2. That behavior, while from the test cases seems to be intended to roll up things like `.graphql` files, also unfortunately rolls up `.pyc` files, meaning every restart is a double restart:

```
WARNING:  Detected file change in 'app/main.py'. Reloading...
INFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [87024]
INFO:     Started server process [87080]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
WARNING:  Detected file change in 'app/__pycache__/main.cpython-37.pyc'. Reloading...
INFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [87080]
INFO:     Started server process [87093]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
```

It might be better to use [Path.glob](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) so users can specify file extensions and paths more explicitly than with `os.walk`, but it's published already so maybe as another flag?

3. A minor point, but worth noting in the docs: `--reload_dir` on the CLI becomes `reload_dirs=['my_dir']` in code: `uvicorn.run('app.main:app', host="0.0.0.0", port=8000, reload=True, reload_dirs=['app'])`

Thanks for making this great library!

euri10:
Can you please paste the output just above what you pasted just to see what reloader it uses?

euri10:
ok it's on statreloader, will send a PR asap

lovelydinosaur:
Thanks for reporting this! 0.11.5 is now published... https://github.com/encode/uvicorn/runs/629310851?check_suite_focus=true

PR #657, opened 2020-04-28
Avoid reload on .pyc files

euri10:
fix for #656

### DIFF
--- a/uvicorn/supervisors/statreload.py
+++ b/uvicorn/supervisors/statreload.py
@@ -9,12 +9,15 @@ logger = logging.getLogger("uvicorn.error")
 
 
 class StatReload(BaseReload):
-    ignored_file_regexes = r'\.py[cod]$', r'\.___jb_...___$', r'\.sw.$', '~$'
+    ignored_file_regexes = r"\.py[cod]$", r"\.___jb_...___$", r"\.sw.$", "~$"
+
     def __init__(self, config, target, sockets):
         super().__init__(config, target, sockets)
         self.reloader_name = "statreload"
         self.mtimes = {}
-        self._ignored_file_regexes = tuple(re.compile(r) for r in self.ignored_file_regexes)
+        self._ignored_file_regexes = tuple(
+            re.compile(r) for r in self.ignored_file_regexes
+        )
 
     def should_restart(self):
         for filename in self.iter_files():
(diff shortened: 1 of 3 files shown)


uvicorn/supervisors/statreload.py
```diff
@@ -8,10 +9,15 @@
 
 
 class StatReload(BaseReload):
+    ignored_file_regexes = r"\.py[cod]$", r"\.___jb_...___$", r"\.sw.$", "~$"
```
lovelydinosaur, line 12: It’s not clear to what all of these are?

It might be worth having a look into what gunicorn or Django devserver do for reload behavious, since this seems a bit ad hoc at the moment?

kalzoo:
Thanks for the quick PR, @euri10. As noted in #656, letting the user specify a glob or list of globs instead of a list of directories would let them filter out other files as well, so you wouldn't have to hard code them here and then possibly change the default behavior in the future. Just a thought.

lovelydinosaur:
Personally I’m thinking that we probably ought to do something like have a pattern/glob whitelisting, with “*.py” as the default.

uvicorn/supervisors/statreload.py
euri10: from a quick look at it it seems gunicorn's reloader will only react to `.py` files:

```
    def get_files(self):
        fnames = [
            COMPILED_EXT_RE.sub('py', module.__file__)
            for module in tuple(sys.modules.values())
            if getattr(module, '__file__', None)
        ]


        fnames.extend(self._extra_files)


        return fnames
```

it get all files to watch by looking at `sys.modules.values()` and uses the regex
[`r'py[co]$'` ](https://github.com/benoitc/gunicorn/blob/5425af8941e1d43c57b5b2fb3fbd7f728688ce01/gunicorn/reloader.py#L14) to transform all the matched extensions (pyc, py and pyo, which is almost the 1st ignored regex above by the way) to `.py`

the above regexes are taken from what watchgod does in its DefaultWatcher:

1. `r"\.py[cod]$"` wil ignore all `.pyc` and `.pyo` and `.pyd`

1. `r"\.___jb_...___$"` honestly I dont know

1. `r"\.sw.$"` this should be swap files, handy for vim users

1. `"~$"` same category as above I suppose, vim backup files

erewok: Drive-by comment: `"~$"` represents backup files produced by emacs.

lovelydinosaur:
Superseeded by #659

PR #659, opened 2020-04-29
Revert "Reload when other files than py are changes, small test maintenance"

lovelydinosaur:
Reverts encode/uvicorn#646

I think that just reverting this and issuing a release is probably the best first step. We can then follow up later with a different tack. (Eg. perhaps allow a filename pattern to be passed as config).

I think we want to get this and #658 in, and then issue a re-release right?


euri10 (approved):

euri10:
agreed, the question is whitelist of blacklist, or a mix of both >)_

Issue #328, opened 2019-03-14
uvicorn.run with logger & reload=True crashed


tchiotludo:
If I pass a custom logger to `uvicorn.run` with `reload=True`
```python

logging.basicConfig(level=logging.DEBUG)

if __name__ == '__main__':
    uvicorn.run("dcp.api.main:app",
                logger=logging.getLogger("uvicorn"),
                reload=True)
```

Uvicorn crash with : 
```
INFO:uvicorn:Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:uvicorn:Started reloader process [13940]
Traceback (most recent call last):
  File "~/bin/server", line 41, in <module>
    reload=True)
  File "~/venv/lib/python3.6/site-packages/uvicorn/main.py", line 262, in run
    supervisor.run(server.run, sockets=[socket])
  File "~/venv/lib/python3.6/site-packages/uvicorn/supervisors/statreload.py", line 34, in run
    process.start()
  File "/usr/lib/python3.6/multiprocessing/process.py", line 105, in start
    self._popen = self._Popen(self)
  File "/usr/lib/python3.6/multiprocessing/context.py", line 284, in _Popen
    return Popen(process_obj)
  File "/usr/lib/python3.6/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/usr/lib/python3.6/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
  File "/usr/lib/python3.6/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/usr/lib/python3.6/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
TypeError: can't pickle _thread.RLock objects
```

Changing with Reload = False and the server start properly

Hanaasagi:
@tchiotludo 
In reload mode, uvicorn will fork a child process to run server. And parent process will check status to stop child and spawn new child to implement reloading.

https://github.com/encode/uvicorn/blob/b2b5410a72f9ef9870dc4443f0fc3a659916e509/uvicorn/supervisors/statreload.py#L23-L46


To make it easier to understand, Supposed you have following code

```Python
import logging
import multiprocessing

logging.basicConfig(level=logging.DEBUG)


class T:
    def __init__(self, logger):
        self.logger = logger

    def run(self):
[...]

PR #661, opened 2020-04-30
Removed python 3.6 specific logging leftover

euri10:
was introduced following #328 

I think it's not necessary as it's solving a use case that doesnt exist anymore ie passing a `logger` kwargs as described in the bug report
```
    uvicorn.run("dcp.api.main:app",
                logger=logging.getLogger("uvicorn"),
                reload=True)
```


lovelydinosaur (approved):
Agreed, thanks!

PR #665, opened 2020-05-03
Allow .json or .yaml --log-config files

jcwilson:
This adds support for `.json` and `.yaml` config files when running `uvicorn` from the CLI.
This allows users to define how they wish to deal with existing loggers (#511, #512)
by providing`disable_existing_loggers` in their config file (the last item described in [this section](https://docs.python.org/3/library/logging.config.html#dictionary-schema-details)).

Furthermore, it addresses the desire to allow users to replicate the default hard-coded
`LOGGING_CONFIG` in their own configs and tweak it as necessary, something that is not currently
possible for clients that wish to run their apps from the CLI.


setup.py
```diff
@@ -48,7 +48,7 @@ def get_packages(package):
     "uvloop>=0.14.0 ;" + env_marker,
 ]
 
-extras_require = {"watchgodreload": ["watchgod>=0.6,<0.7"]}
+extras_require = {"watchgodreload": ["watchgod>=0.6,<0.7"], "yamllogconfig": ["PyYAML>=5.1"]}
```
jcwilson: This PR also adds items to the `extras_require` list: #298. There's a fair amount of discussion there trying to come up with the best names for the optional features. As this will become part of the public interface and troublesome to update once it's live, it might make sense to omit this line for now until the best name for this feature is agreed upon. This wouldn't conflict with the changes in #298, but there they appear to be trying to name the extras as categories as opposed to features to enable.

docs/settings.md
```diff
@@ -33,10 +33,17 @@ $ pip install uvicorn[watchgodreload]
 
 ## Logging
 
-* `--log-config <path>` - Logging configuration file.
+* `--log-config <path>` - Logging configuration file. **Options:** *`dictConfig()` formats: .json, .yaml*. Any other format will be processed with `fileConfig()`. Set the `formatters.default.use_colors` and `formatters.access.use_colors` values to override the auto-detected behavior.
```
euri10, line 36: are there any other format than ini that can be processed by `fileConfig()` ?

requirements.txt
```diff
@@ -3,6 +3,7 @@ h11
 
 # Optional
 httptools
+PyYAML>=5.1
```
euri10: why not the latest ?

setup.py
euri10: yep I'd prefer @tomchristie to decide on this, not sure we should pile on extras_require for every little tweak, even if we did it already with the reloader

tests/test_config.py
```diff
@@ -66,3 +89,64 @@ def test_ssl_config(certfile_and_keyfile):
     config.load()
 
     assert config.is_ssl is True
+
+
+@pytest.mark.parametrize(
+    "use_colors, expected",
+    [(None, None), (True, True), (False, False), ("invalid", False)],
+)
```
euri10, line 119: can we use ids kwargs on this parametrized test to get a sense of what it means ?

euri10 (changes requested):
> This allows users to define how they wish to deal with existing loggers (#511, #512)
by providingdisable_existing_loggers in their config file (the last item described in this section.

I may be off on this but I thought that it was not possible to add that key to an ini file and that the only option if you want that behaviour was to pass it as a kwarg to fileConfig(), I don't think this PR solves anything from #511 but I admit I'm not sure about it, if it does it's cool.

otherwise LGTM minus few changes and @tomchristie thoughts on the `extra_requires` thing !

tests/test_config.py
jcwilson: sure thing. i'll give that a shot

requirements.txt
jcwilson: My reasoning was that this was the least restrictive modern version of the library. I would expect that the latest version would be installed in the absence of other version specifications from other required libraries on the system.

docs/settings.md
jcwilson: There is only one file format that `fileConfig()` supports and that's the `configparser` [format](https://docs.python.org/3/library/logging.config.html#logging-config-fileformat). It's not precisely `.ini`, but `.ini` is the closest well-known format that we can use to give a hint as to what one should use if going the `fileConfig()` route. These changes don't require the `.ini` extension, it's just the fallback if `.json` and `.yaml` are not detected.

I would expect anyone wishing to use `fileConfig()` would be acquainted well-enough with it's format that seeing `.ini` wouldn't cause any confusion.

jcwilson:
> I may be off on this but I thought that it was not possible to add that key to an ini file and that the only option if you want that behaviour was to pass it as a kwarg to fileConfig(), I don't think this PR solves anything from #511 but I admit I'm not sure about it, if it does it's cool.

You're correct that it's not possible to disable existing loggers with a setting in the `fileConfig()` file, but one can with `dictConfig()`. So anyone wishing to disable existing loggers would switch to a json or yaml logging config and get the desired behavior without having to invoke uvicorn programmatically.

tests/test_config.py
euri10: excellent !

requirements.txt
jcwilson: should i update this to the latest?

setup.py
jcwilson: I backed this change out

jcwilson:
@tomchristie @euri10 Can we merge this? I've backed out the changes that related to the "extras" decision. As it's implemented now, it expects the user to install a `yaml` library and there are code comments explaining how to address the situation if it's missing.

docs/settings.md
```diff
@@ -33,10 +33,17 @@ $ pip install uvicorn[watchgodreload]
...
+    * If you wish to use a YAML file for your logging config, you will want to specify the optional `yamllogconfig` dependency to ensure YAML support:
+
+    ```
+    $ pip install uvicorn[yamllogconfig]
+    ```
+
```
euri10: sorry to get back on this that late :) time flies, work is insane,

we can remove this part now package install is merged

requirements.txt
euri10: lgtm

euri10 (changes requested):
Some minor changes to be made I think, my fault because it's been a while since you posted this, I hope you dont mind, let me know if you want me to take this on

1. the `pip install uvicorn[yamllogconfig]` is obsolete in settings.md
2. we would need a section in the README.md about this, after the `python-dotenv` part this should do it  
3. add pyyaml in the setup.py so that we have it when using `uvicorn[standard]`

otherwise this is great stuff

euri10:
I pushed https://github.com/euri10/uvicorn/tree/jcwilson-master based on your work and making the changes so that it doesnt clash with latest master here @jcwilson 
Cant push into your PR unfortunately or there are some git commands I'm missing so feel free to take the necessary commits !

jcwilson:
Sorry for not getting back to you earlier last week. I think I've made all the changes you've requested. Let me know if there's anything else and I should be able to address it tomorrow. Thanks!

euri10 (approved):

euri10:
thanks for your contribution @jcwilson this is a great addition :100: !

PR #666, opened 2020-05-05
Use optional package installs

euri10:
continuation of https://github.com/encode/uvicorn/pull/298/

from last comments there this does:

 `pip install uvicorn` # A minimal install
- click
- h11

`pip install uvicorn[standard]` # An install complete with extras and platform dependencies
- uvloop (cpython only)
- httptools (cpython only)
- websockets
- colorama (windows only, would close #542)
- watchgod for the `--reload` option
- python-dotenv for the `--env-file` option
- pyyaml for the proposed PR #665

that leaves wsproto aside as @tomchristie last comment suggests.


rafalp (dismissed):
My biggest gripe with this change is that as a person only just coming to uvicorn, I have no idea what  "fast deps" are, what they give me, what "possibles" I am to gain or lose.

We'll have to expand readme's intro and describe what I get when I do `pip install uvicorn` vs what I'll get when I do `pip install uvicorn[smth]`.

euri10:
I tried to be more explicit on what those 2 installation options mean @rafalp , you're right that was cryptic.
Not convinced by the term "goodies" on the reloader thing but I couldnt come up with something better on top of my mind, happy to change the wording

mojimi:
How can I help to move this forward?

README.md
```diff
@@ -33,6 +33,26 @@ Install using `pip`:
...
 
+This will install uvicorn with minimal (pure Python) dependencies.
+
+```shell
+$ pip install uvicorn[standard]
+```
+
+This will install uvicorn with "fast" dependencies (where possible) and other "goodies".
+
+Fast meaning in that context that:
```
erewok: Minor comment, but I'd probably phrase this slightly differently:

`In this context, "fast" means the following:` or

`"Fast" here means the following:`

erewok (approved):

lovelydinosaur (commented):
Looking ace yup! I think we might want some minor tweaks in the docs languages. (Eg maybe "optional extras" over "goodies", and something like "Cython-based" rather than "fast"?)

README.md
```diff
@@ -39,15 +39,15 @@ This will install uvicorn with minimal (pure Python) dependencies.
...
 
-This will install uvicorn with "fast" dependencies (where possible) and other "goodies".
+This will install uvicorn with "Cython-based" dependencies (where possible) and other "optional extras".
 
-In this context, "fast" means the following:
+In this context, "Cython-based" means the following:
 
 - the event loop `uvloop` will be installed and used if possible.
 - the http protocol will be handled by `httptools` if possible.
 - the websocket protocol will be handled by `websockets` (should you want to use `wsproto` you'd need to install it manually) if possible.
```
lovelydinosaur: I guess the "websockets" one ought to go under optional extras really. (The "uvloop" and "httptools" cases here are kinda different in that they're not adding extra behaviour, so they're purely due to performance-something-something)

euri10: yes this is indeeed the case !

euri10: amended

Issue #668, opened 2020-05-08
multiprocess graceful shutdown in docker and non-docker discrepency


euri10:
while working on #620 I fixed the issue one could have using several workers in a container ie when you send a sigterm to the container then the subprocess were not killed, or to be more precise they were but not in a graceful manner and after waiting the default docker timeout of 10 or 15s.

while that indeed fixes the issue that existed in containers, that introduced a regression in "normal" usage, ie subprocess are killed too abruptly and the lifespan shutdown event is skipped.

with current master uvicorn version, and the following app:

```
from starlette.applications import Starlette
from starlette.responses import JSONResponse
from starlette.routing import Route


async def homepage(request):
    return JSONResponse({'hello': 'world'})

routes = [
    Route("/", endpoint=homepage)
]


async def some_startup_task():
    print('start')


async def some_shutdown_task():
    print('end')


app = Starlette(
    routes=routes,
    on_startup=[some_startup_task],
    on_shutdown=[some_shutdown_task]
)
```

we have in a container the following logs when running `app:app --workers 2 --log-level=trace`, everything's normal

```
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started parent process [1]
INFO:     Started server process [8]
INFO:     Waiting for application startup.
TRACE:    ASGI [1] Started scope={'type': 'lifespan'}
TRACE:    ASGI [1] Receive {'type': 'lifespan.startup'}
TRACE:    ASGI [1] Send {'type': 'lifespan.startup.complete'}
INFO:     Application startup complete.
INFO:     Started server process [9]
INFO:     Waiting for application startup.
TRACE:    ASGI [1] Started scope={'type': 'lifespan'}
TRACE:    ASGI [1] Receive {'type': 'lifespan.startup'}
TRACE:    ASGI [1] Send {'type': 'lifespan.startup.complete'}
INFO:     Application startup complete.
DEBUG:    pid: 1
DEBUG:    sig: 15
DEBUG:    <SpawnProcess name='SpawnProcess-1' pid=8 parent=1 started>
INFO:     Shutting down
INFO:     Waiting for application shutdown.
TRACE:    ASGI [1] Receive {'type': 'lifespan.shutdown'}
TRACE:    ASGI [1] Send {'type': 'lifespan.shutdown.complete'}
TRACE:    ASGI [1] Completed
INFO:     Application shutdown complete.
INFO:     Finished server process [8]
start
end
[...]

PR #636, opened 2020-04-14
Corrected --proxy-headers client ip/host when using a unix socket

euri10:
should fix #634

PR #671, opened 2020-05-09
SIGTERM on multiple workers fails

euri10:
Fix for #668 

it essentially reverts #620 and applies what the python documentation recommends ie terminate or close a process before join

To summarize, before #620 processes in docker containers were not killed gracefully, so it fixed that, but it introduced a bug described in #668 where for non-docker containers the shutdown was too abrupt and children lifespan events were skipped.

I think this PR deals nicely with both issues, it's rather boring to test things inside and out container, so I hope this is ok


uvicorn/supervisors/basereload.py
```diff
@@ -71,6 +64,7 @@ def restart(self):
         self.process.start()
 
     def shutdown(self):
+        self.process.terminate()
```
rafalp, line 67: Does this mean my worker can shutdown in middle of transaction happening in, say, Starlette's [backgorund task](https://www.starlette.io/background/)?

euri10: not sure on this one, 
one thing also it that I miread the doc, it says to terminate before join in case of a pool, we are not in this case....(https://docs.python.org/3/library/multiprocessing.html?highlight=join#multiprocessing.pool.Pool.join)

kind of stuck, dunno how to solve this, this is annoying to get the 20s timeout reached in a container

aviramha: @rafalp I think it wont shutdown in middle of transaction.
Uvicorn handles sigterm, and sends shutdown event to Starlette. 
If it does, it's a Starlette issue (maybe on shutdown Starlette should wait for all background tasks..)

euri10:
ok after putting loggers basically on every line it appear to me that the mutiprocess can hang indefinitely not only on docker but also normally.

docker sends a SIGTERM signal and not a SIGINT

now if your NOT in a container and `kill -15 parentPID` then the supervisor will correctly capture the signal, enter ` def signal_handler(self, sig, frame):` that sets the event, then enter the shutdown method.
```
def shutdown(self):
        for process in self.processes:
            process.join()
```
now here is the deal, if the signal is a SIGTERM, it will hang indefinitely on `process.join()`

so I think in that case only we need to terminate() the process, maybe there would be a nicer way that handles that case

In case of one worker a SIGTERM is nicely dealt with the 
https://github.com/encode/uvicorn/blob/a839da2564e5168c6e99c3b22819950f26f7d178/uvicorn/main.py#L571

lovelydinosaur:
Similar comment wrt. the Pull Request name on this as in https://github.com/encode/uvicorn/pull/636 😄

Issue #630, opened 2020-04-12
Uvicorn adds a logging handler to the root logger


Ronserruya:
When using the default config, unicorn adds an handler to the root logger:
https://github.com/encode/uvicorn/blob/master/uvicorn/config.py#L86

Since all loggers in python are propagated back to the root logger, this means that all logs from all loggers will reach the handler Uvloop adds

When I create a custom logger in my app all my log messages are printed twice, once from my handlers, and then from the uvicorn handler on the root logger.

Since uvicorn always use "uvicorn.*" loggers, I was able to create a new "unicorn" logger and move the handler to it on app startup

```python
def disable_uvicorn_root_logger():
    """ Uvicorn adds a default handler to the root logger, so all logs messages are duplicated"""
    uvicorn_logger = logging.getLogger('uvicorn')
    uvicorn_logger.addHandler(logging.root.handlers[0])
    logging.root.handlers = []
```

I'm not sure if its the intended functionality, it was kinda unexpected for me.

If you wish to fix it, It can be easily fixed by changing "" to "uvicorn" in 
https://github.com/encode/uvicorn/blob/master/uvicorn/config.py#L86

But doing so will break the existing functionality for users that currently just use logging without any config and it works for them.

xeor:
This bit me as well in https://github.com/Delgan/loguru/issues/247 trying to get all logs handled by loguru. It started with some logs from the `aioredis` that didnt want to get intercepted by loguru.

pawamoy:
I posted a solution to intercept everything (gunicorn/uvicorn/httpx/etc.) [here](https://github.com/benoitc/gunicorn/issues/1572#issuecomment-638391953), hope this helps the next ones who want to do the same thing :slightly_smiling_face:

euri10:
closed by #674

PR #674, opened 2020-05-15
Don't override the root logger

stchris:
Overriding the root logger is probably not intended here. 

This is just a PR for the fix proposed in https://github.com/encode/uvicorn/issues/630


lovelydinosaur (approved):
👍

lovelydinosaur:
Thanks!

PR #676, opened 2020-05-16
Fix typo.

Asday:
Noticed the issue [here](https://www.uvicorn.org/deployment/).  I assume this file is the source of truth, so I'm hoping this fixes it.


lovelydinosaur (approved):
Thanks!

PR #692, opened 2020-06-09
Removing iocp mentions from documentation.

Bloodielie:


euri10 (approved):

euri10:
Good catch @Bloodielie  thanks

PR #696, opened 2020-06-10
ENH: Allow click lib accept environment variables

### LINKED ISSUE #693: Include a environment variable to change arg commands
Hello,

I'm testing the FASTAPI (https://fastapi.tiangolo.com/) and there's a nice environment variable when running with gunicorn (https://docs.gunicorn.org/en/stable/settings.html#settings) `GUNICORN_CMD_ARGS`, it include command line args using enviroment variables

this is very usefull when using docker compose, or k8s. without rebuild the container i can change how it starts (more debug messages instead of only critical)

could be possible include a enviroment at uvicorn? `UVICORN_CMD_ARGS`

the gunicorn_cmd_args implementation can be saw here:

https://github.com/benoitc/gunicorn/blob/ee685e197b3f7cf899dc7d6e0688ff169e9d10df/gunicorn/app/base.py#L171

https://github.com/benoitc/gunicorn/blob/6aab4decde5735fc77daf4fecaf9ef3632189f62/gunicorn/config.py#L79

thanks!

rspadim:
This allow environment variables by click lib:
references: at gitlabs issue https://github.com/encode/uvicorn/issues/693

Edit by @Kludex : closes #693

### DIFF
--- a/uvicorn/__main__.py
+++ b/uvicorn/__main__.py
@@ -1,6 +1,4 @@
 import uvicorn
 
 if __name__ == "__main__":
-    uvicorn.main(
-        auto_envvar_prefix='UVICORN'
-    )
+    uvicorn.main(auto_envvar_prefix="UVICORN")
--- a/uvicorn/__main__.py
+++ b/uvicorn/__main__.py
@@ -1,4 +1,6 @@
 import uvicorn
 
 if __name__ == "__main__":
-    uvicorn.main(auto_envvar_prefix='UVICORN')
+    uvicorn.main(
+        auto_envvar_prefix='UVICORN'
+    )
--- a/uvicorn/__main__.py
+++ b/uvicorn/__main__.py
@@ -1,4 +1,4 @@
 import uvicorn
 
 if __name__ == "__main__":
-    uvicorn.main()
+    uvicorn.main(auto_envvar_prefix='UVICORN')


rspadim:
@euri10 could you check this? thanks!

NargiT:
up ! this is very nice, would be nice to document current behaviour vs expected !

Kludex:
I've asked your review to see if you have something against this, jfyk. @euri10 

But anyway, we need to document it, I guess...

euri10:
nothing against, I just dont see what issue it solves, maybe some examples we'd put in the docs would help clarify it ?

Kludex:
@rspadim Do you have any good examples we can include in the documentation? :kissing:

rushilsrivastava:
@Kludex I think that @rspadim's original issue (#693) offers some good insight into why this is useful. This mimics the behavior of Gunicorn's `GUNICORN_CMD_ARGS` and allows for on the fly change of Uvicorn settings via k8s or docker-compose.

rspadim:
> @Kludex I think that @rspadim's original issue (#693) offers some good insight into why this is useful. This mimics the behavior of Gunicorn's `GUNICORN_CMD_ARGS` and allows for on the fly change of Uvicorn settings via k8s or docker-compose.

sorry folks, i was a bit off, yes that's the point k8s / docker like gunicorn

Kludex:
@rspadim Would you like to rebase your branch? 

We also probably want to add a test for this behavior and write a note in the `Settings` section on our documention. If you are willing, I'll glad accept, otherwise I'll try to do it later on.

rushilsrivastava:
@Kludex If @rspadim doesn't have time, I would be happy to implement this in a different PR.

Kludex:
@rushilsrivastava Feel free to do so. We need: 
- add a test for this behavior;
- and write a note in the Settings section on our docs, showing this possibility and succinctly showing a use case.
Thanks! 

If you can fork @rspadim 's branch and create a branch on top of it, so we don't lose his commit, it'll be cool. :)

Kludex:
We need to twitch the entry point on the `setup.py` for this to work. The solution here only works with `python -m uvicorn`, not with `uvicorn` directly. 

EDIT: I've found a better solution on https://github.com/pallets/click/issues/1379#issuecomment-525329399

Kludex:
Closing this in favor of #1114.
I've cherry-picked those commits and change the implementation, so you're still the original author @rspadim .

Thank you very much for your contribution! :)
I'll take the lead now.

Issue #596, opened 2020-03-21
Handshaking may not be completed yet at `shutdown` in wsproto impl


MisLink:
Traceback:
```
LocalProtocolError: Event CloseConnection(code=1012, reason=None) cannot be sent during the handshake
  File "gunicorn/arbiter.py", line 583, in spawn_worker
    worker.init_process()
  File "uvicorn/workers.py", line 57, in init_process
    super(UvicornWorker, self).init_process()
  File "gunicorn/workers/base.py", line 140, in init_process
    self.run()
  File "uvicorn/workers.py", line 66, in run
    loop.run_until_complete(server.serve(sockets=self.sockets))
  File "uvloop/loop.pyx", line 1456, in uvloop.loop.Loop.run_until_complete
  File "uvicorn/main.py", line 403, in serve
    await self.shutdown(sockets=sockets)
  File "uvicorn/main.py", line 539, in shutdown
    connection.shutdown()
  File "uvicorn/protocols/websockets/wsproto_impl.py", line 115, in shutdown
    output = self.conn.send(wsproto.events.CloseConnection(code=1012))
  File "__init__.py", line 61, in send
    data += self.handshake.send(event)
  File "wsproto/handshake.py", line 101, in send
    "Event {} cannot be sent during the handshake".format(event)
```

euri10:
Could you please send the command you used?
Does it happen without gunicorn?
Do you have a minimal example of the code that triggers it?

MisLink:
@euri10 Yes, it happens every time we send TERM to gunicorn master process in our production environment. I checked the source code:
https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/websockets/wsproto_impl.py#L113-L117
and I concluded by myself that there is no check for completion of the handshake.

Sorry for the lack of info, I'll make a minimal example without gunicorn.

MisLink:
The minimal app:
```python
import asyncio


async def app(scope, receive, send):
    assert scope["type"] == "websocket"
    event = await receive()
    if event["type"] == "websocket.connect":
        await asyncio.sleep(10)
        await send({"type": "websocket.accept"})
```

1. start it with uvicorn and wsproto: `uvicorn --ws wsproto app:app` and write down the process id;
2. connect it with a websocket client, I use [websocat](https://github.com/vi/websocat): `websocat ws://127.0.0.1:8000`;
3. kill the server before the sleep finished: `kill <pid>`, then you will see the same traceback.

[...]

PR #704, opened 2020-06-19
Fix failures when ASGI app rejects a connection during handshake

### LINKED ISSUE #244: TypeError when connection is closed in connect()
I have the same problem then #185 but I think, I can reproduce it.

When you close a connection in the connect() method of a WebsocketConsumer, it works fine with daphne but uvicorn raises a TypeError 

```
  File "/home/ossi/src/channels-examples/.venv/lib/python3.7/site-packages/uvicorn/protocols/websockets/websockets_impl.py", line 140, in run_asgi
    result = await asgi(self.asgi_receive, self.asgi_send)
  File "/home/ossi/src/channels-examples/.venv/lib/python3.7/site-packages/channels/sessions.py", line 179, in __call__
    return await self.inner(receive, self.send)
  File "/home/ossi/src/channels-examples/.venv/lib/python3.7/site-packages/channels/middleware.py", line 41, in coroutine_call
    await inner_instance(receive, send)
  File "/home/ossi/src/channels-examples/.venv/lib/python3.7/site-packages/channels/consumer.py", line 59, in __call__
    [receive, self.channel_receive], self.dispatch
  File "/home/ossi/src/channels-examples/.venv/lib/python3.7/site-packages/channels/utils.py", line 59, in await_many_dispatch
    await task
  File "/home/ossi/src/channels-examples/.venv/lib/python3.7/site-packages/channels/utils.py", line 51, in await_many_dispatch
    result = task.result()
  File "/home/ossi/src/channels-examples/.venv/lib/python3.7/site-packages/uvicorn/protocols/websockets/websockets_impl.py", line 222, in asgi_receive
    data = await self.recv()
  File "/home/ossi/src/channels-examples/.venv/lib/python3.7/site-packages/websockets/protocol.py", line 419, in recv
    return_when=asyncio.FIRST_COMPLETED,
  File "/usr/lib/python3.7/asyncio/tasks.py", line 361, in wait
    fs = {ensure_future(f, loop=loop) for f in set(fs)}
  File "/usr/lib/python3.7/asyncio/tasks.py", line 361, in <setcomp>
    fs = {ensure_future(f, loop=loop) for f in set(fs)}
  File "/usr/lib/python3.7/asyncio/tasks.py", line 592, in ensure_future
    raise TypeError('An asyncio.Future, a coroutine or an awaitable is '
TypeError: An asyncio.Future, a coroutine or an awaitable is required
```

I created this branch on the django-channels-example repository: https://github.com/ostcar/channels-examples/tree/uvicorn-test

All I changed was, that the websocket connection is always closed: https://github.com/andrewgodwin/channels-examples/compare/master...ostcar:uvicorn-test

I think the problem is, that you call `websockets.WebSocketServerProtocol.recv()` before the websocket connection is open so `websockets.WebSocketServerProtocol.connection_open()` was not called. In this case the attribute `transfer_data_task` is still None. So when `websockets.WebSocketServerProtocol.recv()` calls 

```python
yield from asyncio.wait(
                    [pop_message_waiter, self.transfer_data_task],
                    loop=self.loop,
                    return_when=asyncio.FIRST_COMPLETED,
                )
```
it tries to await `None` what causes the exception.

I don't know, if this is a but in the websockets package or if you should check `self.closed` before calling `self.recv()`

adamhooper:
Fixes #244. The unit test already existed, but it logged exceptions that
weren't being noticed.

* websockets_impl: `transfer_data_task` is unset if the handshake fails, so we can't call recv().
* wsproto_impl: `h11.Reject` is invalid during handshake; send `RejectConnection` instead.

### DIFF
--- a/uvicorn/protocols/websockets/websockets_impl.py
+++ b/uvicorn/protocols/websockets/websockets_impl.py
@@ -216,6 +216,7 @@ class WebSocketProtocol(websockets.WebSocketServerProtocol):
 
             elif message_type == "websocket.close":
                 code = message.get("code", 1000)
+                self.close_code = code  # for WebSocketServerProtocol
                 await self.close(code)
                 self.closed_event.set()
 
@@ -236,6 +237,14 @@ class WebSocketProtocol(websockets.WebSocketServerProtocol):
             return {"type": "websocket.connect"}
 
         await self.handshake_completed_event.wait()
+
+        if self.closed_event.is_set():
+            # If the client disconnected: WebSocketServerProtocol set self.close_code.
+            # If the handshake failed or the app closed before handshake completion,
+            # use 1006 Abnormal Closure.
+            code = getattr(self, "close_code", 1006)
+            return {"type": "websocket.disconnect", "code": code}
+
         try:
             data = await self.recv()
         except websockets.ConnectionClosed as exc:
--- a/uvicorn/protocols/websockets/wsproto_impl.py
+++ b/uvicorn/protocols/websockets/wsproto_impl.py
@@ -255,10 +255,8 @@ class WSProtocol(asyncio.Protocol):
                 )
                 self.handshake_complete = True
                 self.close_sent = True
-                msg = h11.Response(status_code=403, headers=[])
+                msg = events.RejectConnection(status_code=403, headers=[])
                 output = self.conn.send(msg)
-                msg = h11.EndOfMessage()
-                output += self.conn.send(msg)
                 self.transport.write(output)
                 self.transport.close()
 


adamhooper:
I could use some help on this [unit-test failure](https://github.com/encode/uvicorn/pull/704/checks?check_run_id=788643507#step:5:63). I've tested 10 times in the venv on localhost and there are no failures.

I also wonder: do other tests have un-awaited tasks? Is my only mistake the use of `caplog`? `pytest -v -s` indeed prints such a backtrace in `tests/protocols/test_websocket.py::test_send_before_handshake[WSProtocol]` on my machine.

matteing:
Any updates? This is a blocker for us in prod.

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -230,6 +229,12 @@ async def asgi_receive(self):
             return {"type": "websocket.connect"}
 
         await self.handshake_completed_event.wait()
+        if not hasattr(self, "transfer_data_task"):
+            # If the handshake completed and there's no data-transfer task,
+            # that means the handshake failed.
+            assert self.closed_event.is_set()
+            return {"type": "websocket.disconnect", "code": self.initial_response[0]}
```
florimondmanca: Rather than relying on an implementation detail of the `websockets` library, what do we think of this…?

```suggestion
        if self.closed_event.is_set():
            # Handshake has failed.
            return {"type": "websocket.disconnect", "code": self.initial_response[0]}
```

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -50,7 +50,6 @@ def __init__(self, config, server_state, _loop=None):
         self.initial_response = None
         self.connect_sent = False
         self.accepted_subprotocol = None
-        self.transfer_data_task = None
```
florimondmanca: Based on my other comment, it's likely this `transfer_data_task` attribute is unrelated to resolving this bug (or at least, there are other ways to tell if the handshake has failed). So not sure if we should just keep this line for the sake of reducing the diff footprint, or go ahead and remove this since we're obviously not using it anywhere else.

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -255,10 +255,8 @@ async def send(self, message):
                 )
                 self.handshake_complete = True
                 self.close_sent = True
-                msg = h11.Response(status_code=403, headers=[])
+                msg = events.RejectConnection(status_code=403, headers=[])
```
florimondmanca, line 258: Yup, looks good to me — results in the desired behavior, same as the `websockets` case.

florimondmanca (commented):
Nice one, did some local testing against this and this looks pretty good. Left a suggestion to simplify the implementation in the `websockets_impl` case!

florimondmanca:
@matteing Mind giving a look at the Windows error we're seeing…? Occurs in the test modified here, apparently there's a coroutine that's not awaited on Windows but I can't tell which coroutine that is exactly, i.e. is it due to the test server, or due to the ASGI application…? I don't have a Windows machine so debugging this is pretty iffy, and I don't want to rush things in case this could be a hint that there's something deeper we need to take care of here.

```console
task: <Task pending name='Task-168' coro=<IocpProactor.accept.<locals>.accept_coro() running at c:\hostedtoolcache\windows\python\3.8.5\x64\lib\asyncio\windows_events.py:562> wait_for=<_OverlappedFuture cancelled>>
```

adamhooper:
@florimondmanca we're all having trouble reproducing this bug on Windows. But it seems to me _this error is currently on Windows, and it has been for ages_?

It looks to me like this patch merely _reveals_ the bug -- it doesn't cause it.

adamhooper:
(My suggesion being: 1. Make unit tests pass _somehow_, so we can deploy this and finally fix #244; and then 2. Investigate the un-awaited coroutine on Windows, which is a different bug that already exists and shouldn't delay the fix to #244....)

euri10:
my 2 cents on the CI issue, I didn't follow the issue at stake here...

we dropped iocp support for windows a while ago and now we force uvicorn to run the SelectorEventLoop, which is not the default loop on windows > 3.8 iirc

here the issue (I think) is that the CI uses the default windows loop, so I'd just force it in the `run_server` that should fix the CI and be closer to what we're doing anyway in the windows case

adamhooper:
Thanks, @euri10! That made the test pass

@florimondmanca I added a new function, `uvicorn.loops.asyncio._new_event_loop()`. Is that okay? Is this pull request [gulp] mergeable?

adamhooper:
ping? Is there anything I can do to help this along?

florimondmanca:
Taking a look. :)

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -236,6 +235,10 @@ async def asgi_receive(self):
             return {"type": "websocket.connect"}
 
         await self.handshake_completed_event.wait()
+        if self.closed_event.is_set():
+            # Handshake has failed.
+            return {"type": "websocket.disconnect", "code": self.initial_response[0]}
```
florimondmanca: That `code` is tricky… We actually need to consider two situations: A/ the handshake failed / got rejected, in that case `.initial_response` is available and we can use it like here, and B/ the connection was closed normally by the client sending a `websocket.close` event, and in that case `.initial_response` isn't available and we need to send the close code that was given by the client.

florimondmanca:
@adamhooper Sadly it seems the entire suite of websocket tests seems off. The `run_server()` helper is potentially plain broken, because it creates an starts tasks on a loop created in the main thread, but runs the loop in a different thread, which `asyncio` simply doesn't allow. So eventually, we need to switch all of those tests to using actual `async/await` code instead, with proper `asyncio` tasks for running the server in the background.

So I'd say — forget about the test suite for now. Let's have it work on manual tests. If that works, let's ship it. :shipit:

For now I was able to confirm that your `closed_event.is_set()` check makes the bug in the `websockets` case go away. Then I looked at the logs in the test cases and saw the one about `h11.Response`, and switching to `RejectConnection` resolves it.

So… How about a _new_ pull request, that makes _those two changes only_?

As noted in https://github.com/encode/uvicorn/pull/704#discussion_r538750016 we'll need to introduce a new `.close_code` variable, set it when receiving the `websocket.close` event, and use that instead of `.initial_response` to derive the close code. I can provide some more pointers on the PR if necessary but hopefully you get the idea?

florimondmanca:
Here's the diff I got to, if that's helpful:

```diff
diff --git a/uvicorn/protocols/websockets/websockets_impl.py b/uvicorn/protocols/websockets/websockets_impl.py
index 407674c..9c26306 100644
--- a/uvicorn/protocols/websockets/websockets_impl.py
+++ b/uvicorn/protocols/websockets/websockets_impl.py
@@ -49,6 +49,7 @@ class WebSocketProtocol(websockets.WebSocketServerProtocol):
         self.closed_event = asyncio.Event()
         self.initial_response = None
         self.connect_sent = False
+        self.close_code = None
         self.accepted_subprotocol = None
         self.transfer_data_task = None
 
@@ -216,6 +217,7 @@ class WebSocketProtocol(websockets.WebSocketServerProtocol):
 
             elif message_type == "websocket.close":
                 code = message.get("code", 1000)
+                self.close_code = code
                 await self.close(code)
                 self.closed_event.set()
 
@@ -236,6 +238,18 @@ class WebSocketProtocol(websockets.WebSocketServerProtocol):
             return {"type": "websocket.connect"}
 
         await self.handshake_completed_event.wait()
+
+        if self.closed_event.is_set():
+            if self.initial_response is None:
+                # Client has disconnected.
+                assert self.close_code is not None
+                code = self.close_code
+            else:
+                # Handshake has failed, or the connection was rejected.
+                code = self.initial_response[0]
+
+            return {"type": "websocket.disconnect", "code": code}
+
         try:
             data = await self.recv()
         except websockets.ConnectionClosed as exc:
diff --git a/uvicorn/protocols/websockets/wsproto_impl.py b/uvicorn/protocols/websockets/wsproto_impl.py
index d712963..456af13 100644
--- a/uvicorn/protocols/websockets/wsproto_impl.py
+++ b/uvicorn/protocols/websockets/wsproto_impl.py
@@ -255,10 +255,8 @@ class WSProtocol(asyncio.Protocol):
                 )
                 self.handshake_complete = True
                 self.close_sent = True
-                msg = h11.Response(status_code=403, headers=[])
+                msg = events.RejectConnection(status_code=403, headers=[])
                 output = self.conn.send(msg)
-                msg = h11.EndOfMessage()
-                output += self.conn.send(msg)
                 self.transport.write(output)
                 self.transport.close()

```

adamhooper:
@florimondmanca I'm a bit confused as to what my role is here. I assume I am to copy/paste your diff into the pull request, reset to master....

... and here's the irony: out of curiosity, I added my `caplog` hack from my original commit, just to see what would happen. And your patch caused a big stack trace:

```
INFO     uvicorn.error:websockets_impl.py:194 ('127.0.0.1', 53910) - "WebSocket /" 403
ERROR    asyncio:base_events.py:1738 Exception in callback _SelectorSocketTransport._call_connection_lost(None)
handle: <Handle _SelectorSocketTransport._call_connection_lost(None)>
  File "/usr/lib64/python3.9/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/lib64/python3.9/asyncio/selector_events.py", line 978, in _call_connection_lost
    super()._call_connection_lost(exc)
  File "/usr/lib64/python3.9/asyncio/selector_events.py", line 736, in _call_connection_lost
    self._protocol.connection_lost(exc)
  File "/home/adam/src/upstream/uvicorn/uvicorn/protocols/websockets/websockets_impl.py", line 72, in connection_lost
    super().connection_lost(exc)
  File "/home/adam/src/upstream/uvicorn/venv/lib64/python3.9/site-packages/websockets/protocol.py", line 1367, in connection_lost
    self.abort_pings()
  File "/home/adam/src/upstream/uvicorn/venv/lib64/python3.9/site-packages/websockets/protocol.py", line 1307, in abort_pings
    exc = self.connection_closed_exc()
  File "/home/adam/src/upstream/uvicorn/venv/lib64/python3.9/site-packages/websockets/protocol.py", line 779, in connection_closed_exc
    exception = ConnectionClosedError(self.close_code, self.close_reason)
  File "/home/adam/src/upstream/uvicorn/venv/lib64/python3.9/site-packages/websockets/exceptions.py", line 134, in __init__
    super().__init__(code, reason)
  File "/home/adam/src/upstream/uvicorn/venv/lib64/python3.9/site-packages/websockets/exceptions.py", line 121, in __init__
    super().__init__(format_close(code, reason))
  File "/home/adam/src/upstream/uvicorn/venv/lib64/python3.9/site-packages/websockets/exceptions.py", line 93, in format_close
    if 3000 <= code < 4000:
TypeError: '<=' not supported between instances of 'int' and 'NoneType'
```

The fix is to nix the `self.close_code = None`. (`WebSocketServerProtocol` uses `hasattr()`, not `is None`.) So I committed that.

Also, I changed the logic so the `websocket.disconnect` event (which I guess is part of ASGI spec?) has code=1006 in this case. As I understand it, `code` here is meant to be a valid WebSocket close code, and 403 isn't one. ref: https://asgi.readthedocs.io/en/latest/specs/www.html#disconnect-receive-event-ws

All that to say: I've force-pushed a very brief diff. I also recommend adding caplog to _every_ test ... but I agree with you that we should merge this post-haste to tackle my thousands of spurious stack traces per day :).

florimondmanca:
@adamhooper Thanks for following up, sorry if my comments looked patronizing — didn't mean to have it come across that way.

Your approach for `.close_code` sounds good to me!

So is this PR fixing the two main issues that are overflowing your logs? I don't have a workstation handy just now, but I can do an extra-checkup, then I guess we can go ahead and release this? :-)

adamhooper:
@florimondmanca Not patronizing at all! I'm all eagerness here.

Yes, I heartily expect this PR to fix #244, the spurious log messages in `websockets_impl`.

I don't have experience with the parallel bug in `wsproto_impl`, but my `caplog` change exposed it. It looks like that might be #596 and/or #811. My change makes tests pass, so I _assume_ it's correct.

florimondmanca (approved):
Was able to confirm this fixes the following two issues:

* When running with `--ws websockets` and the ASGI app rejects the connection, a `TypeError` is raised -> Fixed
* When running with `--ws wsproto` and the ASGI app rejects the connection, a `LocalProtocolError` is raised -> Fixed

florimondmanca:
@adamhooper Thanks! Seems like we can go ahead and get this in.

We'll get to fixing the test suite later — I'll log an issue about that to track it.

euri10:
sorry to be late in the game but this feels like a wrong fix to me at least in the websocket case, if you read the comment here https://github.com/aaugustin/websockets/pull/670#issuecomment-530368604

a simpler and cleaner imho fix would look like this :
```
        try:
            await self.ensure_open()
            data = await self.recv()
        except websockets.ConnectionClosed as exc:
            return {"type": "websocket.disconnect", "code": exc.code}
```

kylebebak:
@euri10 

We can't call `WebSocketCommonProtocol.ensure_open` before calling `WebSocketCommonProtocol.recv`, because there may be unread frames in the server's read queue that were sent by the client **before** the client sent the close frame.

`WebSocketCommonProtocol.ensure_open` raises an exception and prevents the server from reading these unread frames. Neither the client, nor application code using uvicorn, has any way of knowing that frames successfully sent by the client to the server's read queue, before any close frame was sent, will never be read by the server.

The spec is clear on this: https://datatracker.ietf.org/doc/html/rfc6455#section-1.4

> After sending a control frame indicating the connection should be closed, a peer does not send any further data; after receiving a control frame indicating the connection should be closed, a peer discards any further data received.

The current behavior in uvicorn is rather "a peer discards any further data received, and any data previously received"

-----

More detail in https://github.com/encode/uvicorn/issues/1230

Also, https://github.com/aaugustin/websockets/pull/670 was eventually closed without any changes being made; in other words, there was no bug in `websockets`, and the maintainer agrees that calling `await self.ensure_open()` in `recv` is incorrect

Issue #527, opened 2019-12-13
Honor gunicorn access log format


ipmb:
When running under gunicorn, [`--access-logformat`](http://docs.gunicorn.org/en/stable/settings.html#access-log-format) isn't honored.

There was some discussion of this in #389, but I didn't see an open issue for it.

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/uvicorn/issues/527">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/527/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/uvicorn/issues/527/pledge.svg">
</picture>
</a>

danieltahara:
This is a pretty big loss for observability for us, and means we have to implement something out of band (and therefore less efficient). Any word on if it will be worked on?

loudsquelch:
Any update on this?

asgoel:
Any update on this one?

lovelydinosaur:
Hey folks - "Any update on this" is almost always not helpful. If there was an update on it, it'd be here on the ticket.
If anyone wants to put some work into progressing it that'd be welcome.

asgoel:
@tomchristie I would be willing to take a stab. i've taken a look at the code and it seems like the `UvicornWorker` is fundamentally written in a different way than the rest of the gunicorn workers that come bundled by default. 

We would basically need to thread through the `Logger` instance from gunicorn and call the `access` method when the request is complete in `httptools_impl`/`h11_impl`. We'd also need to record request time which the app currently does not do. 

Does this approach sound somewhat correct to you?

asgoel:
@tomchristie just following up to see if my above comment made sense.

insurandoag:
We have the same problem at the moment and are about to write a custom logging engine for this, because we must be able to alter the uvicorn log format when running it from gunicorn

KlausGlueckert:
We have the same problem at the moment and are about to write a custom logging engine for this, because we must be able to alter the uvicorn log format when running it from gunicorn. Any update on this if there will be a fix? In addition, ideally it should also be configurable in a logging.conf:

```
[formatter_AccessFormatter]
[...]

PR #718, opened 2020-07-21
add response headers to scope

mindrunner:
While implementing a custom logger formatter to support gunicorn logging syntax (https://docs.gunicorn.org/en/stable/settings.html#logging), we had some crucial information missing. One of which, we could solve by injecting the response headers to the scope object. This might be a first step towards 
https://github.com/encode/uvicorn/issues/527

### DIFF
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -443,6 +443,7 @@ class RequestResponseCycle:
 
             status_code = message["status"]
             headers = self.default_headers + list(message.get("headers", []))
+            self.scope["response_headers"] = headers
 
             if self.access_log:
                 self.access_logger.info(


marco-loche-work:
Please merge it!!!!

cosimomeli (approved):

gianpaolocaprara (approved):

marco-loche-work (approved):

euri10:
I don't think the spec has that scope key, am I missing something ?

mindrunner:
For me it worked in the specific context I had to use it. However, I am not really into the gunicon/uvicorn code, so I am not entirely sure about the consequences of this change.

florimondmanca:
Hello!

> I don't think the spec has that scope key, am I missing something ?

Yup, the [ASGI spec](https://asgi.readthedocs.io/en/latest/) doesn't define putting response headers into the scope. If that's something you'd be ready to champion in the ASGI repo, then please feel free to. But until then I'd advise us against implementing ASGI-level behavior into Uvicorn that's not defined in the spec — otherwise we break the purpose of ASGI as a standard that enables interoperability between apps and servers. :confused:

I'll close this for now, not to say that the original use case isn't appropriate for discussion, but I don't think this is the way we'll be going forward with. If there's any specific use case you can expose for discussion then we can still try and find an alternative solution. :-)

PR #721, opened 2020-07-23
return 'connection: close' header in response

### LINKED ISSUE #720: connection: close header not returned in response
Hello, the [W3 standard](https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html) seems to say that a server which receives a request with a `connection: close` header should also return this header to the client. Specifically I'm talking about the first line in `8.1.2.1 Negotiation` quoted here:
> An HTTP/1.1 server MAY assume that a HTTP/1.1 client intends to maintain a persistent connection unless a Connection header including the connection-token "close" was sent in the request. If the server chooses to close the connection immediately after sending the response, it SHOULD send a Connection header including the connection-token close.

The wording is maybe a little ambiguous but to me it definitely seems to imply that if the server closes the connection (wether due to a `connection: close` header in the request or by it's own choice) it should return a `connection: close` header. I ran into an issue with the asks library were this was causing an issue: https://github.com/theelous3/asks/issues/115.

RmStorm:
According to https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html a
server should return the 'connection: close' header in the response when
a request has that header. Resolves #720

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -12,6 +12,21 @@ async def app(scope, receive, send):
 
 
 @pytest.mark.asyncio
+async def test_return_close_header():
+    config = Config(app=app, host="localhost", loop="asyncio", limit_max_requests=1)
+    async with run_server(config):
+        async with httpx.AsyncClient() as client:
+            response = await client.get(
+                "http://127.0.0.1:8000", headers={"connection": "close"}
+            )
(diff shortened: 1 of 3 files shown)


euri10 (approved):
thanks @RmStorm !

Issue #727, opened 2020-07-28
ValueError in version 0.11.7


felipe3dfx:
there are a bug in the last version.

![image](https://user-images.githubusercontent.com/4070833/88726744-e7095400-d0f3-11ea-8c7e-bbe8f5bfa115.png)

Packages:

acme | 1.1.0
-- | --
aioredis | 1.3.1
asgiref | 3.2.10
async-timeout | 3.0.1
attrs | 19.3.0
autobahn | 20.7.1
automat | 20.2.0
ax3-mercadopago | 0.2.6
ax3-model-extras | 1.2.5
ax3-redactor | 1.0.11
backcall | 0.2.0
blinker | 1.4
cachetools | 4.1.1
certbot | 0.40.0
certifi | 2020.6.20
cffi | 1.14.1
channels | 2.4.0
channels-redis | 3.0.1
chardet | 3.0.4
click | 7.1.2
colorama | 0.4.3
command-not-found | 0.3
configargparse | 0.13.0
configobj | 5.0.6
constantly | 15.1.0
cryptography | 3.0
cssselect | 1.1.0
cssutils | 1.0.2
daphne | 2.5.0
dbus-python | 1.2.16
decorator | 4.4.2
defusedxml | 0.6.0
distro | 1.4.0
distro-info | 0.23ubuntu1
django | 3.0.8
django-admin-sortable | 2.2.3
django-appconf | 1.0.4
django-axes | 5.4.1
django-compressor | 2.4
django-extensions | 3.0.3
django-hashid-field | 3.1.3
django-ipware | 3.0.0
django-redis | 4.12.1
django-widget-tweaks | 1.4.8
entrypoints | 0.3
fail2ban | 0.11.1
future | 0.18.2
gdal | 3.0.4
gunicorn | 20.0.4
h11 | 0.9.0
hashids | 1.3.1
hiredis | 1.1.0
httplib2 | 0.14.0
httptools | 0.1.1
huey | 2.2.0
hyperlink | 19.0.0
idna | 2.10
incremental | 17.5.0
iotop | 0.6
ipython | 7.16.1
ipython-genutils | 0.2.0
jedi | 0.17.2
josepy | 1.2.0
keyring | 18.0.1
language-selector | 0.1
launchpadlib | 1.10.13
lazr.restfulclient | 0.14.2
lazr.uri | 1.0.3
linecache2 | 1.0.0
lxml | 4.5.2
mock | 3.0.5
msgpack | 1.0.0
[...]

Issue #634, opened 2020-04-14
--proxy-headers with supervisor get client ip failed


m0nksy:
#### uvicorn version
`Running uvicorn 0.11.3 with CPython 3.8.2 on Linux` 
#### supervisor config
```
[fcgi-program:uvicorn]
socket=tcp://0.0.0.0:5000    
directory=/var/www/webapp/    
command=/usr/local/python3/bin/uvicorn --fd 0 --proxy-headers main:app    
numprocs=2    
process_name=uvicorn-%(process_num)d    
stdout_logfile_maxbytes=0    
stdout_logfile=/var/log/uvicorn.log    
```

### fastapi 
version 0.52.0

```
# cat /var/www/webapp/main.py

#!/usr/bin/env python
# -*- coding: utf-8 -*-
from fastapi import FastAPI, Request

app = FastAPI(redoc_url=None)


@app.get('/')
def index(request: Request):
    return 'hello {}!'.format(request.client.host)

```

### run on supervisor
supervisord version 4.1.0
```
supervisord -n
curl  http://127.0.0.1:5000

```
why client.host is None?

euri10:
confirmed, the client lookup returns None if the socket is a unix socket which is the case here

euri10:
closed via https://github.com/encode/uvicorn/pull/636

PR #729, opened 2020-07-29
Get the request.client under case uvicorn is run with a fd or a unix socket

euri10:
should solve #727 and still give the correct info that https://github.com/encode/uvicorn/issues/634 wanted

### DIFF
--- a/uvicorn/protocols/utils.py
+++ b/uvicorn/protocols/utils.py
@@ -3,9 +3,6 @@ import urllib
 
 def get_remote_addr(transport):
     socket_info = transport.get_extra_info("socket")
-    peer_info = transport.get_extra_info("peername")
-    print(socket_info)
-    print(peer_info)
     if socket_info is not None:
         try:
             info = socket_info.getpeername()
--- a/uvicorn/protocols/utils.py
+++ b/uvicorn/protocols/utils.py
@@ -17,7 +17,7 @@ def get_remote_addr(transport):
 
         if family in (socket.AF_INET, socket.AF_INET6):
             return (str(info[0]), int(info[1]))
-        elif family is socket.AF_UNIX:
+        elif hasattr(socket, "AF_UNIX") and family is socket.AF_UNIX:
             if isinstance(info, tuple):
                 # fd case
                 # <uvloop.PseudoSocket fd=13, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, laddr=('127.0.0.1', 8000), raddr=('127.0.0.1', 38634)>
--- a/uvicorn/protocols/utils.py
+++ b/uvicorn/protocols/utils.py
@@ -1,32 +1,20 @@
-import socket
 import urllib
 
 
 def get_remote_addr(transport):
     socket_info = transport.get_extra_info("socket")
+    peer_info = transport.get_extra_info("peername")
+    print(socket_info)
+    print(peer_info)
     if socket_info is not None:
         try:
             info = socket_info.getpeername()
+            return (str(info[0]), int(info[1])) if isinstance(info, tuple) else None
         except OSError:
             # This case appears to inconsistently occur with uvloop
             # bound to a unix domain socket.
-            family = None
-            info = None
-        else:
-            family = socket_info.family
+            return None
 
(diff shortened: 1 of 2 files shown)


uvicorn/protocols/utils.py
```diff
@@ -20,9 +15,17 @@ def get_remote_addr(transport):
...
-
+        elif hasattr(socket, "AF_UNIX") and family is socket.AF_UNIX:
+            if isinstance(info, tuple):
+                # fd case
+                # <uvloop.PseudoSocket fd=13, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, laddr=('127.0.0.1', 8000), raddr=('127.0.0.1', 38634)>
+                return (str(info[0]), int(info[1]))
+            else:
+                # unix socket case
+                # <uvloop.PseudoSocket fd=21, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, laddr=/tmp/gunicorn.sock>
+                return None
```
florimondmanca: I think I'd do something like this, to simplify:

```python
HAS_AF_UNIX = hasattr(socket, "AF_UNIX")

def get_remote_addr(transport):
    ...

        if family in (socket.AF_INET, socket.AF_INET6) or (
            HAS_AF_UNIX and family == socket.AF_UNIX and isinstance(family, tuple)
        ):

        return None
```

florimondmanca (commented):
My understanding is that in some cases the `socket_info.getpeername()` may return a string containing the path to the Unix socket, and we need to ignore that case, correct?

I'm curious - in what cases, when bound to a Unix socket, can it return a (host, port) tuple? Isn't binding to a Unix socket not done via a host/port by definition? Had we confirmed that #636 _did fix_ #634?

florimondmanca:
We might want to simply revert #636. It's not clear to me that it did fix anything, given how according to https://docs.python.org/3/library/socket.html there is no host/port information available when binding to a Unix socket. I think #634 was a non-issue ("yes there's no host/port because your server is not bound to a TCP socket").

euri10:
> My understanding is that in some cases the `socket_info.getpeername()` may return a string containing the path to the Unix socket, and we need to ignore that case, correct?

correct

> I'm curious - in what cases, when bound to a Unix socket, can it return a (host, port) tuple? Isn't binding to a Unix socket not done via a host/port by definition? Had we confirmed that #636 _did fix_ #634?

if you run uvicorn with `supervisord -n` using

```
[supervisord]

[fcgi-program:uvicorn]
socket=tcp://localhost:8000
command=venv/bin/uvicorn --fd 0 app:app
process_name=uvicorn-%(process_num)d
stdout_logfile=/dev/stdout
stdout_logfile_maxbytes=0
```

then you end up having `socket_info` to what I left in comments ` <uvloop.PseudoSocket fd=13, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, laddr=('127.0.0.1', 8000), raddr=('127.0.0.1', 38634)>`

florimondmanca:
I see, so in that case Supervisord's [FastCGI support](http://supervisord.org/configuration.html#fcgi-program-x-section-settings) makes the TCP socket available at `fd=0`, which _is_ a Unix socket, _and_ for some reason the original TCP host:port info is made available on the Unix socket object too? (I'm definitely no Unix networking wizard, so trying to get my head around this…)

uvicorn/protocols/utils.py
lovelydinosaur: Could we just change the whole block to...

```python
return (str(info[0]), int(info[1])) if isinstance(info, tuple) else None
```

bukowa:
I merged this patch into integration test with nginx - it works https://github.com/bukforks/uvicorn/tree/django_nginx
https://github.com/encode/uvicorn/pull/734

uvicorn/protocols/utils.py
euri10: I did it that way, way simpler,
Just note that I kept in `get_remote_addr` the try except introduced it seems in #495

florimondmanca (approved):

florimondmanca:
I guess we can merge this and get it along for 0.11.8. 👍 #731

PR #730, opened 2020-07-29
Fix crash when --interface is wsgi

mpaolini:
This is a regression introduced in
ae0fd316f03dbef926e40216024dfb934417d48d (#597)

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -202,7 +202,7 @@ class Config:
 
     @property
     def asgi_version(self) -> str:
-        return {"asgi2": "2.0", "asgi3": "3.0"}[self.interface]
+        return {"asgi2": "2.0", "asgi3": "3.0", "wsgi": "2.0"}[self.interface]
 
     @property
     def is_ssl(self) -> bool:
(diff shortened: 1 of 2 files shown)


uvicorn/config.py
```diff
@@ -202,7 +202,7 @@ def __init__(
 
     @property
     def asgi_version(self) -> str:
-        return {"asgi2": "2.0", "asgi3": "3.0"}[self.interface]
+        return {"asgi2": "2.0", "asgi3": "3.0", "wsgi": "2.0"}[self.interface]
```
florimondmanca: The `WSGIMiddleware` uses the ASGI3 callable interface, so I _think_ we should use `"wsgi": "3.0"`.

(That said I think we can acknowledge that this is fuzzy, because "I'm serving a WSGI app under the hood" isn't something that's planned by the ASGI spec. This is merely a convenience feature provided by Uvicorn.)

mpaolini: done!

florimondmanca (approved):
Sounds good to me. As I said it's a fuzzy area not covered by the spec, but obviously there's a bug worth resolving and we need to provide an `asgi_version` in all cases, so… Thanks!

PR #739, opened 2020-08-01
Align most of scripts of httpx here

euri10:
WIP, kind of massive but in the end not too many changes to be made


euri10:
should be good for review, not that I set the coverage pass at 90% here
and commented mypy check, I think that would be good for another PR once this one done

euri10:
I forgot many things before a proper review

euri10:
should be good for a review @tomchristie or @florimondmanca since you are very familiar with those ! :)

I mostly copied the scripts from httpx and adapted

note that I commented mypy usage (another pending PR building on that one hopefully incoming after)
note that I arbitrarily fixed the coverage at 80 (so that we can grind back to 100 !!)

requirements_windows.txt
```diff
@@ -5,6 +5,10 @@ h11
 websockets==8.*
 wsproto==0.13.*
 
+# Packaging
+twine
+wheel
+
```
florimondmanca, line 11: Probably out of scope of this PR but - is there a reason this Windows-specific requirements file exists, instead of using a single requirements file with environment markers such as `some-dep; platform="win32"`?

scripts/check
```diff
@@ -0,0 +1,14 @@
...
+if [ -d 'venv' ] ; then
+    export PREFIX="venv/bin/"
+fi
+export SOURCE_FILES="uvicorn tests"
+
+set -x
+
+${PREFIX}black --check --diff --target-version=py36 $SOURCE_FILES
+${PREFIX}flake8 $SOURCE_FILES
+#${PREFIX}mypy $SOURCE_FILES
```
florimondmanca: I suppose we can remove this commented line and add it if/when mypy gets added to the repo.

scripts/clean
```diff
@@ -0,0 +1,14 @@
...
+fi
+if [ -d 'site' ] ; then
+    rm -r site
+fi
+if [ -d 'htmlcov' ] ; then
+    rm -r htmlcov
+fi
+if [ -d 'uvicorn.egg-info' ] ; then
+    rm -r uvicorn.egg-info
+fi
```
florimondmanca: In my experience this script has been made obsolete since switching to the "publish via GH actions" workflow. I don't think we absolutely need it. WDYT?

setup.cfg
```diff
@@ -0,0 +1,22 @@
...
+[tool:isort]
+profile = black
+combine_as_imports = True
+known_first_party = uvicorn,tests
+known_third_party = click,does_not_exist,gunicorn,h11,httptools,pytest,requests,setuptools,urllib3,uvloop,watchgod,websockets,wsproto
+
+[tool:pytest]
+addopts = --cov=uvicorn --cov=tests -rxXs
+markers =
+  copied_from(source, changes=None): mark test as copied from somewhere else, along with a description of changes made to accodomate e.g. our test setup
```
florimondmanca: It seems we don't need this marker. :-) (And we should remove it from HTTPX - it's an old leftover.)

tests/middleware/test_trace_logging.py
```diff
@@ -13,10 +13,10 @@
     "disable_existing_loggers": False,
     "formatters": {
         "test_formatter_default": {
-            "format": "[TEST_DEFAULT] %(levelname)-9s %(name)s - %(lineno)d - %(message)s"
+            "format": "[TEST_DEFAULT] %(levelname)-9s %(name)s - %(lineno)d - %(message)s"  # noqa: E501
```
florimondmanca, line 16: Thoughts on ignoring E501 (line too long) by default, ie putting it in setup.cfg? Something I've been thinking of doing in HTTPCore and HTTPX too...

florimondmanca (commented):
Nice chunk of work!

requirements_windows.txt
euri10: this is where I'm not confortable and mix probably things,

I originally thought once https://github.com/encode/uvicorn/pull/666 land that we could avoid having separate `requirements.txt` for win and linux , that's probably what you imply, the use of specific os markers inside the setup.py ?

but now I htink about it we may still need both requirements files because they're just used by the CI and specifically the install script

again, I may be totally off, happy to learn somehting today !

scripts/check
euri10: done

setup.cfg
euri10: done

scripts/clean
euri10: yup !  in the trash

tests/middleware/test_trace_logging.py
euri10: I'd rather not have it by default so that we are explicit about tolerating it on some rare occasions where the line is long enough but it wouldnt make sense to split it, but I guess it's really a matter of taste and I'm happy to put it by default

requirements_windows.txt
euri10: oula ok yes now I get it, didn't remember ou could do that in the requirements.txt too

euri10: will do another PR on top of this one and we can have the install script back were it was

tests/middleware/test_trace_logging.py
florimondmanca: Sure, not sure either, so let's keep things like they are for now. :)

florimondmanca (approved):

PR #686, opened 2020-05-28
ignore OSError on killing old process when reloading

trim21:
close #684

Issue #684, opened 2020-05-27
autoreload got PermissionError after any exception on windows


trim21:
```
INFO:     Application startup complete.
INFO:     <Server sockets=[<socket.socket fd=592, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('127.0.0.1', 8000)>]> is serving
WARNING:  Detected file change in 'app\fast.py'. Reloading...
2020-05-28 03:23:58.519 | DEBUG    | app.log:setup_logger:24 - setup logger
2020-05-28 03:23:58.748 | DEBUG    | app.middlewares.sentry:setup_sentry:14 - setup sentry
WARNING:  Executing <Task finished coro=<Server.serve() done, defined at C:\Users\Trim21\.venv\pol\lib\site-packages\uvicorn\main.py:384> exception=NameError("name 'a' is not defined") created at C:\Users\Trim21\scoop\apps\python37\3.7.7\lib\asyncio\base_events.py:566> took 0.782 seconds
Process SpawnProcess-2:
Traceback (most recent call last):
  File "C:\Users\Trim21\scoop\apps\python37\3.7.7\lib\multiprocessing\process.py", line 297, in _bootstrap
    self.run()
  File "C:\Users\Trim21\scoop\apps\python37\3.7.7\lib\multiprocessing\process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\Trim21\.venv\pol\lib\site-packages\uvicorn\subprocess.py", line 62, in subprocess_started
    target(sockets=sockets)
  File "C:\Users\Trim21\.venv\pol\lib\site-packages\uvicorn\main.py", line 382, in run
    loop.run_until_complete(self.serve(sockets=sockets))
  File "C:\Users\Trim21\scoop\apps\python37\3.7.7\lib\asyncio\base_events.py", line 587, in run_until_complete
    return future.result()
  File "C:\Users\Trim21\.venv\pol\lib\site-packages\uvicorn\main.py", line 389, in serve
    config.load()
  File "C:\Users\Trim21\.venv\pol\lib\site-packages\uvicorn\config.py", line 288, in load
    self.loaded_app = import_from_string(self.app)
  File "C:\Users\Trim21\.venv\pol\lib\site-packages\uvicorn\importer.py", line 20, in import_from_string
    module = importlib.import_module(module_str)
[...]

PR #743, opened 2020-08-04
Reloader yields PermissionError on windows after a syntax error

euri10:
Attempt at solving #684 
It replaces  #686 which seems weaker
see https://docs.python.org/3/library/os.html#os.kill

### DIFF
--- a/tests/supervisors/test_statreload.py
+++ b/tests/supervisors/test_statreload.py
@@ -45,3 +45,28 @@ def test_should_reload(tmpdir):
         reloader.shutdown()
     finally:
         os.chdir(working_dir)
+
+
+def test_should_reload_after_exception(tmpdir):
+    update_file = Path(os.path.join(str(tmpdir), "example.py"))
+    update_file.touch()
+
+    working_dir = os.getcwd()
+    os.chdir(str(tmpdir))
+    try:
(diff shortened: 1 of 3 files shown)


abersheeran:
Great solution, hope to merge as soon as possible~

euri10:
> Great solution, hope to merge as soon as possible~

would you test it please ? I have no windows at hand unfortunately, and I would be interested in knowing if there are process left after a reload or if the signal kills them cleanly ?
would be amazing if someone could take a look at that

abersheeran:
Unfortunately, after trying it on Windows 10, I found that the restart can be successful the first time, but I cannot restart the second time. It got stuck there and there was no response.

There should be some questions that I don't know. I will check the Windows signal again.

abersheeran:
I found it. `TerminateProcess` should be used on Windows.

euri10:
https://maruel.ca/post/python_windows_signal/

abersheeran:
Maybe you can take a look at https://github.com/encode/uvicorn/pull/744? 

By the way, it's early morning in China, and I need to sleep. Maybe the next reply will be a long time later. good night.

euri10:
not the correct solution it seems

PR #751, opened 2020-08-14
Mypy lifespan

euri10:
In the process of typing uvicorn, this does the lifespan module

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,11 +6,11 @@ max-line-length = 88
 disallow_untyped_defs = True
 ignore_missing_imports = True
 follow_imports = silent
+files = uvicorn/lifespan,tests/test_lifespan.py
 
 [mypy-tests.*]
 disallow_untyped_defs = False
 check_untyped_defs = True
-follow_imports = silent
 
 [tool:isort]
 profile = black
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,10 +5,12 @@ max-line-length = 88
 [mypy]
 disallow_untyped_defs = True
 ignore_missing_imports = True
+follow_imports = silent
 
 [mypy-tests.*]
 disallow_untyped_defs = False
 check_untyped_defs = True
+follow_imports = silent
 
 [tool:isort]
 profile = black
--- a/setup.py
+++ b/setup.py
@@ -41,11 +41,12 @@ env_marker_cpython = (
 )
 
 env_marker_win = "sys_platform == 'win32'"
-
+env_marker_below_38 = "python_version < '3.8'"
 
 minimal_requirements = [
     "click==7.*",
     "h11>=0.8,<0.10",
+    "typing-extensions;" + env_marker_below_38
 ]
 
 extra_requirements = [
--- a/uvicorn/_types.py
+++ b/uvicorn/_types.py
@@ -27,15 +27,15 @@ else:
     from typing import Literal
     from typing import TypedDict
 
-from uvloop.loop import TCPTransport
 from websockets import Subprotocol
 
 if TYPE_CHECKING:  # pragma: no cover
+    from uvloop.loop import TCPTransport
+
     from uvicorn.protocols.http.h11_impl import H11Protocol
     from uvicorn.protocols.http.httptools_impl import HttpToolsProtocol
     from uvicorn.protocols.websockets.websockets_impl import WebSocketProtocol
     from uvicorn.protocols.websockets.wsproto_impl import WSProtocol
-
 
 # useless until there is a possibility to add extra keys
 # https://github.com/python/mypy/issues/4617
@@ -190,4 +190,4 @@ AutoWebSocketsProtocolType = Type[Union["WebSocketProtocol", "WSProtocol"]]
 
 StrPath = Union[str, PathLike]
 
-TransportType = Union[TCPTransport, Transport]
+TransportType = Union["TCPTransport", Transport]
--- a/uvicorn/_types.py
+++ b/uvicorn/_types.py
@@ -1,5 +1,5 @@
 import sys
-from typing import List, Optional, Sequence, Tuple, Union
+from typing import List, Optional, Tuple, Union
 
 if sys.version_info < (3, 8):
     from typing_extensions import Literal, TypedDict
@@ -17,6 +17,7 @@ class HTTPReceiveRequest(TypedDict):
     """
     https://asgi.readthedocs.io/en/latest/specs/www.html#request-receive-event
     """
+
     type: Literal["http.request"]
     body: Optional[bytes]
     more_body: Optional[bool]
@@ -26,6 +27,7 @@ class HTTPReceiveDisconnect(TypedDict):
     """
     https://asgi.readthedocs.io/en/latest/specs/www.html#disconnect-receive-event
     """
+
     type: Literal["http.disconnect"]
 
 
@@ -38,6 +40,7 @@ class HTTPSendResponseStart(TypedDict):
     """
     https://asgi.readthedocs.io/en/latest/specs/www.html#response-start-send-event
     """
+
     type: Literal["http.response.start"]
     status: int
     headers: Optional[List[Tuple[bytes, bytes]]]
@@ -47,6 +50,7 @@ class HTTPSendResponseBody(TypedDict):
     """
     https://asgi.readthedocs.io/en/latest/specs/www.html#response-body-send-event
     """
+
     type: Literal["http.response.body"]
     body: Optional[bytes]
     more_body: Optional[bool]
@@ -61,6 +65,7 @@ class WSReceiveConnect(TypedDict):
     """
     https://asgi.readthedocs.io/en/latest/specs/www.html#connect-receive-event
     """
+
     type: Literal["websocket.connect"]
 
 
@@ -68,6 +73,7 @@ class WSReceive(TypedDict):
     """
     https://asgi.readthedocs.io/en/latest/specs/www.html#receive-receive-event
     """
+
     type: Literal["websocket.receive"]
     bytes: Optional[bytes]
     text: Optional[str]
@@ -78,6 +84,7 @@ class WSReceiveDisconnect(TypedDict):
     yeah I know this link looks off but it is the one !
     https://asgi.readthedocs.io/en/latest/specs/www.html#id2
     """
+
     type: Literal["websocket.disconnect"]
     code: int
 
@@ -91,6 +98,7 @@ class WSSendAccept(TypedDict):
     """
     https://asgi.readthedocs.io/en/latest/specs/www.html#accept-send-event
     """
+
     type: Literal["websocket.accept"]
     subprotocol: Optional[Subprotocol]
     headers: Optional[List[Tuple[bytes, bytes]]]
@@ -100,16 +108,17 @@ class WSSend(TypedDict):
     """
     https://asgi.readthedocs.io/en/latest/specs/www.html#send-send-event
     """
+
     type: Literal["websocket.send"]
     bytes: Optional[bytes]
     text: Optional[str]
-
 
 
 class WSSendClose(TypedDict):
     """
     https://asgi.readthedocs.io/en/latest/specs/www.html#close-send-event
     """
+
     type: Literal["websocket.close"]
     code: int
 
--- a/uvicorn/_types.py
+++ b/uvicorn/_types.py
@@ -6,82 +6,26 @@ if sys.version_info < (3, 8):
 else:
     from typing import Literal, TypedDict
 
-# SCOPES
 
-
-class ASGIDict(TypedDict):
+class ASGISpecInfo(TypedDict):
     version: str
     spec_version: Optional[Union[Literal["2.0"], Literal["2.1"]]]
 
 
 class LifespanScope(TypedDict):
-    """
-    https://asgi.readthedocs.io/en/latest/specs/lifespan.html#scope
-    """
-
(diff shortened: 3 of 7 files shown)


setup.cfg
```diff
@@ -5,10 +5,12 @@ max-line-length = 88
 [mypy]
 disallow_untyped_defs = True
 ignore_missing_imports = True
+follow_imports = silent
```
florimondmanca, line 8: Doesn't mypy allow a `files` option here? (Maybe it doesn't, but I think it does.) It would then be picked up by IDEs too. :)

uvicorn/_types.py
```diff
@@ -0,0 +1,193 @@
...
+    Sequence,
+    Tuple,
+    Type,
+    Union,
+)
+
+if sys.version_info < (3, 8):
+    if TYPE_CHECKING:
+        from typing_extensions import Literal, TypedDict
+    else:  # due to different mypy warnings raised during CI for python 3.7 and 3.8
```
florimondmanca: This looks a bit off. Are you able to make the topmost try/except pass mypy by adding `# type: ignore` for any of the imports?

uvicorn/_types.py
```diff
@@ -0,0 +1,193 @@
...
+
+Sockets = Optional[List[socket.socket]]
+
+AutoHTTPProtocolType = Type[Union["H11Protocol", "HttpToolsProtocol"]]
+
+AutoWebSocketsProtocolType = Type[Union["WebSocketProtocol", "WSProtocol"]]
+
+StrPath = Union[str, PathLike]
+
+TransportType = Union["TCPTransport", Transport]
```
florimondmanca: Should we strip down these types to only what's necessary for annotating lifespan?

euri10: done

uvicorn/_types.py
euri10: I dont understand what you mean :)

setup.cfg
euri10: done

uvicorn/_types.py
euri10: well maybe like this, not sure that's what you meant !

euri10:
ok I had it wrong on first pass, that's why reviews are so good !

this now uses typed lifespan messages in place of the http and ws protocol message before, and using those revealed 

1. that we didn't have a proper lifespan scope, it was missing the asgi spec, I corrected it
2. there is no use that is made of the "lifespan.shutdown.failed" in the `async def send`. Is that an omission ? Readin the spec we should deal with it, not sure how yet, any ideas ?

uvicorn/lifespan/on.py
```diff
@@ -33,18 +37,21 @@ async def startup(self):
...
         self.logger.info("Application shutdown complete.")
 
-    async def main(self):
+    async def main(self) -> None:
         try:
             app = self.config.loaded_app
-            scope = {"type": "lifespan"}
+            scope: LifespanScope = {
+                "type": "lifespan",
+                "asgi": {"version": self.config.asgi_version, "spec_version": "2.0"},
```
florimondmanca, line 53: Interesting! Should we submit this as a separate pull request, so that this focuses strictly on adding type hints?

florimondmanca:
> there is no use that is made of the "lifespan.shutdown.failed" in the async def send. Is that an omission ? Readin the spec we should deal with it, not sure how yet, any ideas ?

True: https://asgi.readthedocs.io/en/latest/specs/lifespan.html#shutdown-failed-send-event

Probably let's track this as an issue? Interesting that this hasn't come up before - maybe application frameworks (eg Starlette) don't send this either?

uvicorn/lifespan/on.py
euri10: while I completely agree if I leave the scope as is and just add the type then mypy will fail with
`uvicorn/lifespan/on.py:51: error: Key 'asgi' missing for TypedDict "LifespanScope"`

so I PR the full dict in a separate one then once merged we rebase this one and add types ?

euri10: ok added #754 !

florimondmanca (commented):
Still need to think about this — the amount of verbosity in these types vs the benefit they bring doesn't seem clear yet…

In the meantime, do you know how come CI passes on 3.6 and 3.7, even though we haven’t added `typing_extensions` to Uvicorn’s dependencies? There’s probably a dep/dev dep that installs it for us. Should we add it explicitly?

euri10:
> In the meantime, do you know how come CI passes on 3.6 and 3.7, even though we haven’t added `typing_extensions` to Uvicorn’s dependencies? There’s probably a dep/dev dep that installs it for us. Should we add it explicitly?

seems like pip does it ! this might be in the mypy setup for those versions ?

edit: well it's a dependency of mypy in fact, whatever the python version

```
❯ mkdir typing_extensions_36
❯ cd typing_extensions_36
❯ asdf local python 3.6.9
❯ which python
/home/lotso/.asdf/shims/python
❯ pip install mypy
Collecting mypy
  Downloading https://files.pythonhosted.org/packages/14/9d/914bb0c71292ffd7e1c9728b9366f9ea34c874f5ca12ac6005a400e4d130/mypy-0.782-cp36-cp36m-manylinux1_x86_64.whl (20.8MB)
    100% |████████████████████████████████| 20.8MB 1.8MB/s
Collecting typing-extensions>=3.7.4 (from mypy)
  Using cached https://files.pythonhosted.org/packages/0c/0e/3f026d0645d699e7320b59952146d56ad7c374e9cd72cd16e7c74e657a0f/typing_extensions-3.7.4.2-py3-none-any.whl
Collecting typed-ast<1.5.0,>=1.4.0 (from mypy)
  Using cached https://files.pythonhosted.org/packages/90/ed/5459080d95eb87a02fe860d447197be63b6e2b5e9ff73c2b0a85622994f4/typed_ast-1.4.1-cp36-cp36m-manylinux1_x86_64.whl
Collecting mypy-extensions<0.5.0,>=0.4.3 (from mypy)
  Using cached https://files.pythonhosted.org/packages/5c/eb/975c7c080f3223a5cdaff09612f3a5221e4ba534f7039db34c35d95fa6a5/mypy_extensions-0.4.3-py2.py3-none-any.whl
Installing collected packages: typing-extensions, typed-ast, mypy-extensions, mypy
Successfully installed mypy-0.782 mypy-extensions-0.4.3 typed-ast-1.4.1 typing-extensions-3.7.4.2
You are using pip version 18.1, however version 20.2.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

florimondmanca:
> Collecting typing-extensions>=3.7.4 (from mypy)

Yup. But then it means that running Uvicorn will fail if the user doesn't have mypy installed! :-) So we ought to add it explicitly ourselves as well, I believe (with a proper `; python_version < '3.8'` environment marker).

euri10:
> > Collecting typing-extensions>=3.7.4 (from mypy)
> 
> Yup. But then it means that running Uvicorn will fail if the user doesn't have mypy installed! :-) So we ought to add it explicitly ourselves as well, I believe (with a proper `; python_version < '3.8'` environment marker).

those eagle eyes, I didn't saw this ! :eagle:

euri10:
> Still need to think about this — the amount of verbosity in these types vs the benefit they bring doesn't seem clear yet…

yes you're right, maybe we can merge this one and subsequent attempts on other modules to a `mypy_attempt` branch then once more has been merged get a clearer picture whether or not it's a benefit or not

uvicorn/_types.py
```diff
@@ -0,0 +1,87 @@
...
+
+class LifeSpanReceiveShutdown(TypedDict):
+    """
+    https://asgi.readthedocs.io/en/latest/specs/lifespan.html#shutdown-receive-event
+    """
+
+    type: Literal["lifespan.shutdown"]
+
+
+LifespanReceiveMessage = Union[LifespanReceiveStartup, LifeSpanReceiveShutdown]
```
florimondmanca: Could all of this be simplified like this…? (The type hints in `lifespan/off.py`/`lifespan/on.py` only use `LifespanReceiveMessage` and `LifespanSendMessage`, so we don't _actually need_ individual declarations for each individual message type.)

```python
class LifespanReceiveMessage(TypedDict):
    type: Literal["lifespan.startup", "lifespan.shutdown"]
```



(Handy trick here - `Literal` accepts multiple literal values, which mypy interprets as a `Union` of those literals.)

And then, similarly for send messages:

```python
class LifespanSendMessage(TypedDict):
    type: Literal[
        "lifespan.startup.complete",
        "lifespan.startup.failed",
        "lifespan.shutdown.complete",
    ]
```

(I know we'd need an optional `message` attribute for `startup.shutdown.failed`, but since we don't handle it anyway this is the form we need.)

uvicorn/_types.py
```diff
@@ -0,0 +1,87 @@
...
+
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal, TypedDict
+else:
+    from typing import Literal, TypedDict
+
+# SCOPES
+
+
+class ASGIDict(TypedDict):
```
florimondmanca: Maybe this alternative naming…? (We know it's a `dict`, so we might want to indeed answer "what about ASGI does it contain?".)

```suggestion
class ASGISpecInfo(TypedDict):
```

uvicorn/_types.py
```diff
@@ -0,0 +1,87 @@
+import sys
+from typing import Optional, Union
+
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal, TypedDict
+else:
+    from typing import Literal, TypedDict
+
+# SCOPES
```
florimondmanca: I'm not sure we need these comments - it's quite readable what each item corresponds to, imo.

setup.py
```diff
@@ -41,11 +41,12 @@ def get_packages(package):
 )
 
 env_marker_win = "sys_platform == 'win32'"
-
+env_marker_below_38 = "python_version < '3.8'"
```
florimondmanca, line 44: I find this style a bit odd, but at least it's consistent with what we already have. (I tend to rather see the markers inline, eg `"typing-extensions; python_version < '3.8'"`.)

florimondmanca (commented):
Looking good! :+1: Left some ideas to simplify the types and reduce their LOC footprint.

florimondmanca:
Wild idea: I don't know if we could eventually package these types up as an `asgi-typing` package, so that these type hints can be reused across ASGI-related projects… Since the spec is pretty stable now, it should be possible to define these once and for all. Could be an extra-thin package like [trio-typing](https://github.com/python-trio/trio-typing), for example.

uvicorn/_types.py
euri10: > (Handy trick here - Literal accepts multiple literal values, which mypy interprets as a Union of those literals.)

super nice, I learnt something today.

> (I know we'd need an optional message attribute for startup.shutdown.failed, but since we don't handle it anyway this is the form we need.)

I left both the message and the `lifespan.shutdown.failed` since we need message anyway for the `lifespan.startup.failed`

uvicorn/_types.py
euri10: done yes

uvicorn/_types.py
euri10: yep, my rough way to mentally keep track on where I am, shouldnt be here

setup.py
euri10: originally this was because we grouped windows / cython markers, I have no preference

florimondmanca (approved):
Lovely! :tada: I think this is ready for merging. :100:

uvicorn/lifespan/on.py
```diff
@@ -1,24 +1,28 @@
...
+    def __init__(self, config: Config) -> None:
         if not config.loaded:
             config.load()
 
         self.config = config
         self.logger = logging.getLogger("uvicorn.error")
         self.startup_event = asyncio.Event()
         self.shutdown_event = asyncio.Event()
-        self.receive_queue = asyncio.Queue()
+        self.receive_queue: Queue = asyncio.Queue()
```
JayH5: asyncio.Queue is generic, but only in the typestubs. So if you quote it you can do:
```suggestion
        self.receive_queue: "Queue[LifespanReceiveMessage]" = asyncio.Queue()
```

uvicorn/_types.py
```diff
@@ -0,0 +1,31 @@
...
+
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal, TypedDict
+else:
+    from typing import Literal, TypedDict
+
+
+class ASGISpecInfo(TypedDict):
+    version: str
+    spec_version: Optional[Union[Literal["2.0"], Literal["2.1"]]]
```
JayH5:
```suggestion
    spec_version: Optional[Literal["2.0", "2.1"]]
```

uvicorn/lifespan/on.py
JayH5: Or possibly to avoid the extra/slightly redundant input:
```suggestion
        self.receive_queue: "asyncio.Queue[LifespanReceiveMessage]" = asyncio.Queue()
```

JayH5 (approved):
Small notes but I'm also +1 on this with/without suggestions.

uvicorn/lifespan/on.py
euri10: done !

uvicorn/_types.py
euri10: and done

florimondmanca (approved):

PR #754, opened 2020-08-15
Add `asgi` field to the lifespan scope

### LINKED ISSUE #753: Lifespan scope is missing asgi dict
the lifespan scope (https://asgi.readthedocs.io/en/latest/specs/lifespan.html#scope) is missing the asgi dict

prompted by

_Originally posted by @florimondmanca in https://github.com/encode/uvicorn/pull/751_

euri10:
fix #753 
note that spec version is 2.0 according to https://asgi.readthedocs.io/en/latest/specs/lifespan.html#version-history

### DIFF
--- a/tests/test_lifespan.py
+++ b/tests/test_lifespan.py
@@ -144,3 +144,40 @@ def test_lifespan_with_failed_startup(mode, raise_exception):
 
     loop = asyncio.new_event_loop()
     loop.run_until_complete(test())
+
+
+@pytest.mark.parametrize("mode", ("auto", "on"))
+def test_lifespan_scope_asgi3app(mode):
+    async def asgi3app(scope, receive, send):
+        assert scope == {"version": "3.0", "spec_version": "2.0"}
+
+    async def test():
+        config = Config(app=asgi3app, lifespan=mode)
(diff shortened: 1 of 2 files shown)


euri10:
should we add tests for this ?

florimondmanca:
> should we add tests for this ?

Ideally, yup! 😄 I think there are already some "unit" tests related to the `asgi_version`, we can take inspiration from them to check that uvicorn passes this info when doing the lifespan dance.

euri10:
mmm we test scope and asgi version here on the protocol level, and only for the http protocol this said (this could be done for the ws one too)
https://github.com/encode/uvicorn/blob/fbce393fd34038a589bba02e759b0f051f52ad84/tests/protocols/test_http.py#L690-L702

one simple test could be in `test_lifespan.py` with somethings like this ?
or should we mock the lifespan protocol the way we moch the http one in tests and test the scope rightness there ?

```
        nonlocal startup_complete, shutdown_complete
+++ assert scope.get("asgi") +++
        message = await receive()
```

euri10:
ok added some tests for both cases , asgi 2 or 3

florimondmanca (approved):
Very nice!

PR #755, opened 2020-08-15
Implemented lifespan.shutdown.failed

### LINKED ISSUE #752: lifespan.shutdown.failed message is not implemented
there is no use that is made of the "lifespan.shutdown.failed" in the `async def send`. Is that an omission ? Readin the spec we should deal with it, not sure how yet, any ideas ?

_Originally posted by @euri10 in https://github.com/encode/uvicorn/pull/751#issuecomment-674366666_

euri10:
fixes #752 

one improvment in another PR could be to  add tests to see if the server sends correctly the reason for failed startup / shutdown, but out of scope for now


euri10:
this is an old PR but since we are dealing with tests right now hopefully this will be easier to review :)

florimondmanca (approved):
Superb! :)

Issue #679, opened 2020-05-19
Uvicorn >=0.11.4 does not send shutdown event when using --reload


ghost:
Hello, I have a problem. After update Uvicorn from 0.11.3 to 0.11.5, it does not send shutdown event.

Source code:
```python
from starlette.applications import Starlette

def startup():
    print('Ready to go')

def shutdown():
    print('Shutdown!')

app = Starlette(debug=True, on_startup=[startup], on_shutdown=[shutdown])
```

Uvicorn 0.11.3:
```
[uvicorn-ERV875EY]─[~/p/uvicorn]─> uvicorn --version
Running uvicorn 0.11.3 with CPython 3.8.2 on Darwin

[uvicorn-ERV875EY]─[~/p/uvicorn]─> uvicorn --reload --log-level trace main:app
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started reloader process [47131]
INFO:     Started server process [47134]
INFO:     Waiting for application startup.
TRACE:    ASGI [1] Started scope={'type': 'lifespan'}
TRACE:    ASGI [1] Receive {'type': 'lifespan.startup'}
Ready to go
TRACE:    ASGI [1] Send {'type': 'lifespan.startup.complete'}
INFO:     Application startup complete.
^CINFO:     Shutting down
INFO:     Waiting for application shutdown.
TRACE:    ASGI [1] Receive {'type': 'lifespan.shutdown'}
Shutdown!
TRACE:    ASGI [1] Send {'type': 'lifespan.shutdown.complete'}
TRACE:    ASGI [1] Completed
INFO:     Application shutdown complete.
INFO:     Finished server process [47134]
INFO:     Stopping reloader process [47131]
```

Uvicorn 0.11.4:
```
[uvicorn-ERV875EY]─[~/p/uvicorn]─> uvicorn --version
Running uvicorn 0.11.4 with CPython 3.8.2 on Darwin

[uvicorn-ERV875EY]─[~/p/uvicorn]─> uvicorn --reload --log-level trace main:app
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started reloader process [47903] using statreload
INFO:     Started server process [47905]
INFO:     Waiting for application startup.
TRACE:    ASGI [1] Started scope={'type': 'lifespan'}
TRACE:    ASGI [1] Receive {'type': 'lifespan.startup'}
Ready to go
[...]

PR #756, opened 2020-08-15
Revert "Improve shutdown robustness when using `--reload` or multiprocessing #620"

euri10:
Fix for #679 as discussed 

This reverts commit fdb89e9c


euri10:
what is still obscure to me is why no tests showed the regression, maybe we should take the opportunity of this revert to add some, thoughts ?

florimondmanca (approved):

PR #759, opened 2020-08-18
Test assertions are broken in websocket server tests

stefanw:
This PR is showing that assertion failures in the server part of websocket tests do not affect the test outcomes.


lovelydinosaur:
Oh, fun!

euri10:
Yeah I wasn't even aware of this,
after some digging this exists : https://github.com/bjoluc/pytest-reraise

should we use this ?

Kludex:
I've checked this. Yes, it's weird. The thing is that we do assertions outside the thread, so I'd say this is fine, as the `pytest.raises` we have there are a way of not printing the exceptions (and only that...).

PR #758, opened 2020-08-17
Add test to simulate lost connection on server

stefanw:
Test for issue #757.

Shows the effects of websocket server sending data on a connection that has been closed by transport without receiving a close frame.

PR #762, opened 2020-08-20
Deal gently with ConnectionClosedError fron websockets

### LINKED ISSUE #757: ConnectionClosedError raised from websockets library
### Checklist

### Describe the bug

I'm running a Django project with channels and uvicorn (as a gunicorn worker).

Occasionally, when the server tries to send a message and the client has disconnected improperly, the following exception is raised from the `websockets` library:

```
...
  File "channels/consumer.py", line 81, in send
    await self.base_send(message)
  File "channels/sessions.py", line 236, in send
    return await self.real_send(message)
  File "uvicorn/protocols/websockets/websockets_impl.py", line 212, in asgi_send
    await self.send(data)
  File "websockets/protocol.py", line 555, in send
    await self.ensure_open()
  File "websockets/protocol.py", line 812, in ensure_open
    raise self.connection_closed_exc()
ConnectionClosedError
```
(full traceback under debugging below)

### Expected behavior

As a user of channels, I expect not to be required to catch these lower-level exceptions in my message sending code (right? I'm a bit unsure about this). I would expect uvicorn to handle this exception from `websockets`.

### Actual behavior

I have to guard my websocket sending code against exceptions thrown by an implementation detail of the underlying ASGI server. I'm running the daphne ASGI server in development, so this exception handling is unexpected.

### Debugging material

<details>
<summary>Full traceback</summary>

```
CancelledError: null
  File "websockets/protocol.py", line 827, in transfer_data
    message = await self.read_message()
  File "websockets/protocol.py", line 895, in read_message
    frame = await self.read_data_frame(max_size=self.max_size)
  File "websockets/protocol.py", line 971, in read_data_frame
    frame = await self.read_frame(max_size)
  File "websockets/protocol.py", line 1051, in read_frame
    extensions=self.extensions,
  File "websockets/framing.py", line 105, in read
    data = await reader(2)
  File "asyncio/streams.py", line 674, in readexactly
    yield from self._wait_for_data('readexactly')
  File "asyncio/streams.py", line 464, in _wait_for_data
    yield from self._waiter

ConnectionClosedError: code = 1006 (connection closed abnormally [internal]), no reason
  File "uvicorn/protocols/websockets/websockets_impl.py", line 154, in run_asgi
    result = await self.app(self.scope, self.asgi_receive, self.asgi_send)
  File "uvicorn/middleware/proxy_headers.py", line 45, in __call__
  File "uvicorn/middleware/asgi2.py", line 7, in __call__
    await instance(receive, send)
  File "channels/sessions.py", line 183, in __call__
    return await self.inner(receive, self.send)
  File "channels/middleware.py", line 41, in coroutine_call
    await inner_instance(receive, send)
  File "channels/consumer.py", line 59, in __call__
    [receive, self.channel_receive], self.dispatch
  File "channels/utils.py", line 51, in await_many_dispatch
    await dispatch(result)
  File "channels/consumer.py", line 73, in dispatch
    await handler(message)
  File ".../consumers.py", line 63, in userlist
    'user': event['user']
  File "channels/generic/websocket.py", line 273, in send_json
    await super().send(text_data=await self.encode_json(content), close=close)
  File "channels/generic/websocket.py", line 211, in send
    await super().send({"type": "websocket.send", "text": text_data})
  File "channels/consumer.py", line 81, in send
    await self.base_send(message)
  File "channels/sessions.py", line 236, in send
    return await self.real_send(message)
  File "uvicorn/protocols/websockets/websockets_impl.py", line 212, in asgi_send
    await self.send(data)
  File "websockets/protocol.py", line 555, in send
    await self.ensure_open()
  File "websockets/protocol.py", line 812, in ensure_open
    raise self.connection_closed_exc()
```

</details>

### Environment

- Running uvicorn 0.11.8 with CPython 3.6.6 on Linux
- Running as a gunicorn worker via `uvicorn.workers.UvicornWorker`

euri10:
based on test nicely provided in https://github.com/encode/uvicorn/pull/758 
using pytest-reraise to avoid the bug shown in https://github.com/encode/uvicorn/pull/759
fix #757 

Would really appreciate some feedback on this,  not sure I got this right

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -466,3 +466,34 @@ def test_subprotocols(protocol_cls, subprotocol):
         accepted_subprotocol = loop.run_until_complete(get_subprotocol(url))
         assert accepted_subprotocol == subprotocol
         loop.close()
+
+
+@pytest.mark.parametrize("protocol_cls", WS_PROTOCOLS)
+def test_server_lost_connection(protocol_cls):
+    class App(WebSocketResponse):
+        async def websocket_connect(self, message):
+            await self.send({"type": "websocket.accept"})
+            await self.receive()
+
+            # Simulate a lost connection
+            # without receiving a close frame
+            self.send.__self__.connection_lost(None)
+
+            with pytest.raises(Exception) as exc:
+                await self.send({"type": "websocket.send", "text": "123"})
+
+            assert exc.value.code == 1006
+
+    async def websocket_session(url):
+        async with websockets.connect(url) as websocket:
+            await websocket.ping()
+            await websocket.send("abc")
+            # Delay exiting context manager
+            # to avoid sending a close frame before server attempts send
+            await asyncio.sleep(1)
+
+    if protocol_cls is WebSocketProtocol:
+        with run_server(App, protocol_cls=protocol_cls) as url:
+            loop = asyncio.new_event_loop()
+            loop.run_until_complete(websocket_session(url))
+            loop.close()
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -469,7 +469,7 @@ def test_subprotocols(protocol_cls, subprotocol):
 
 
 @pytest.mark.parametrize("protocol_cls", WS_PROTOCOLS)
-def test_server_lost_connection(protocol_cls):
+def test_server_lost_connection(protocol_cls, reraise):
     class App(WebSocketResponse):
         async def websocket_connect(self, message):
             await self.send({"type": "websocket.accept"})
@@ -479,10 +479,8 @@ def test_server_lost_connection(protocol_cls):
             # without receiving a close frame
             self.send.__self__.connection_lost(None)
 
-            with pytest.raises(Exception) as exc:
+            with reraise:
                 await self.send({"type": "websocket.send", "text": "123"})
-
-            assert exc.value.code == 1006
 
     async def websocket_session(url):
         async with websockets.connect(url) as websocket:
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -490,8 +490,7 @@ def test_server_lost_connection(protocol_cls, reraise):
             # to avoid sending a close frame before server attempts send
             await asyncio.sleep(1)
 
-    if protocol_cls is WebSocketProtocol:
-        with run_server(App, protocol_cls=protocol_cls) as url:
-            loop = asyncio.new_event_loop()
-            loop.run_until_complete(websocket_session(url))
-            loop.close()
+    with run_server(App, protocol_cls=protocol_cls) as url:
+        loop = asyncio.new_event_loop()
+        loop.run_until_complete(websocket_session(url))
+        loop.close()
--- a/uvicorn/protocols/websockets/websockets_impl.py
+++ b/uvicorn/protocols/websockets/websockets_impl.py
@@ -4,6 +4,7 @@ import logging
 from urllib.parse import unquote
 
 import websockets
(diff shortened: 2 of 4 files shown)


tests/protocols/test_websocket.py
```diff
@@ -466,3 +466,32 @@ async def get_subprotocol(url):
...
+
+    async def websocket_session(url):
+        async with websockets.connect(url) as websocket:
+            await websocket.ping()
+            await websocket.send("abc")
+            # Delay exiting context manager
+            # to avoid sending a close frame before server attempts send
+            await asyncio.sleep(1)
+
+    if protocol_cls is WebSocketProtocol:
```
stefanw: I would let this run on both implementations, neither should raise exceptions.

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -212,8 +213,11 @@ async def asgi_send(self, message):
                 bytes_data = message.get("bytes")
                 text_data = message.get("text")
                 data = text_data if bytes_data is None else bytes_data
-                await self.send(data)
-
+                try:
+                    await self.send(data)
+                except ConnectionClosedError as exc:
```
stefanw, line 218: Catching the exception is certainly one way to deal with this. Just want to mention that the `wsproto` implementation checks for `self.transport.is_closing()`:

https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/websockets/wsproto_impl.py#L278

Maybe this is more agnostic to the underlying library. On the other hand it's circumventing the library and checks transport conditions directly. Unsure.

tests/protocols/test_websocket.py
```diff
@@ -466,3 +466,32 @@ async def get_subprotocol(url):
...
+@pytest.mark.parametrize("protocol_cls", WS_PROTOCOLS)
+def test_server_lost_connection(protocol_cls, reraise):
+    class App(WebSocketResponse):
+        async def websocket_connect(self, message):
+            await self.send({"type": "websocket.accept"})
+            await self.receive()
+
+            # Simulate a lost connection
+            # without receiving a close frame
+            self.send.__self__.connection_lost(None)
```
stefanw, line 480: I'm still wondering if there's a better way to access the underlying protocol here.

tests/protocols/test_websocket.py
euri10: yep, didn't see that >_ fixed

uvicorn/protocols/websockets/websockets_impl.py
euri10: yes, I dont like it too, I didn't see something satisfying enough from the faq here too.
https://websockets.readthedocs.io/en/stable/faq.html#what-does-connectionclosederror-code-1006-mean

tests/protocols/test_websocket.py
euri10: yes, you're right.
add to that the fact that once we call the protocol here, right after in the `asgi_send` mehtod at `elif not self.closed_event.is_set():` we have that event not set and the State that is `State.CLOSED` which is dubious

euri10: fail_connection ?

stefanw: I used `connection_lost` because it was used in [`test_early_disconnect`](https://github.com/encode/uvicorn/blob/ae0fd316f03dbef926e40216024dfb934417d48d/tests/protocols/test_http.py#L487) in the http tests. It's the method called on protocol from [transport.close()](https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseTransport.close), so seems appropriate. I was just wondering if there's a nicer way to access it instead of going through the method's `__self__`.

PR #776, opened 2020-09-14
adding parameter --ssl-password to decrypt ssl key file

### LINKED ISSUE #581: Load certificates with custom function
Is it possible to point Uvicorn to a custom function to load an SSL certificate and keyfile, rather than passing in the filepaths? Or alternatively, point to a function to pass into the password argument when SSLContext.load_cert_chain is called? Thanks.

remster85:
Add parameter `--ssl-password` to be able to decrypt the ssl key
If provided, this parameter will be passed to load_cert_chain as password.

The password argument may be a function to call to get the password for decrypting the private key. It will only be called if the private key is encrypted and a password is necessary. It will be called with no arguments, and it should return a string, bytes, or bytearray. If the return value is a string it will be encoded as UTF-8 before using it to decrypt the key. Alternatively a string, bytes, or bytearray value may be supplied directly as the password argument. It will be ignored if the private key is not encrypted and no password is needed.

https://docs.python.org/3.3/library/ssl.html#ssl.SSLContext.load_cert_chain

This issue required a pull request https://github.com/encode/uvicorn/issues/581

Fixes https://github.com/encode/uvicorn/issues/581

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -79,6 +79,7 @@ Options:
                                   5]
   --ssl-keyfile TEXT              SSL key file
   --ssl-certfile TEXT             SSL certificate file
+  --ssl-password TEXT             SSL key file password
   --ssl-version INTEGER           SSL version to use (see stdlib ssl module's)
                                   [default: 2]
   --ssl-cert-reqs INTEGER         Whether client certificate is required (see
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -106,7 +106,7 @@ def create_ssl_context(
 ):
     ctx = ssl.SSLContext(ssl_version)
     if password:
-        getpassword = lambda _: password
+        def getpassword() : return password
         ctx.load_cert_chain(certfile, keyfile, getpassword)
     else:
         ctx.load_cert_chain(certfile, keyfile)
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -106,7 +106,10 @@ def create_ssl_context(
 ):
     ctx = ssl.SSLContext(ssl_version)
     if password:
-        def getpassword() : return password
+
+        def getpassword():
+            return password
+
         ctx.load_cert_chain(certfile, keyfile, getpassword)
     else:
         ctx.load_cert_chain(certfile, keyfile)
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -101,9 +101,15 @@ LOGGING_CONFIG = {
 logger = logging.getLogger("uvicorn.error")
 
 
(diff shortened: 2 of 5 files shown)


remster85:
Let's go! Could you please merge?

docs/deployment.md
```diff
@@ -79,6 +79,7 @@ Options:
                                   5]
   --ssl-keyfile TEXT              SSL key file
   --ssl-certfile TEXT             SSL certificate file
+  --ssl-password TEXT             SSL key file password
```
florimondmanca, line 82: Thoughts on naming this `--ssl-keyfile-password`? Both for extra explicitness that this is related to `--ssl-keyfile`, as well as visually hinting that "this is a more advanced option, since its name is longer" :-)

```suggestion
  --ssl-keyfile-password TEXT             SSL key file password
```

uvicorn/config.py
```diff
@@ -101,9 +101,18 @@
...
     ctx = ssl.SSLContext(ssl_version)
-    ctx.load_cert_chain(certfile, keyfile)
+    if password:
+
+        def getpassword():
+            return password
+
+        ctx.load_cert_chain(certfile, keyfile, getpassword)
+    else:
+        ctx.load_cert_chain(certfile, keyfile)
```
florimondmanca, line 115: (Nit) Dunno if we could go for a more compact form - hopefully this is readable enough? Or can we _not_ pass `None` as the `password` argument? (The Python docs aren't very explicit about that.)

```suggestion
    password = (lambda: password) if password else None
    ctx.load_cert_chain(certfile, keyfile, password)
```

florimondmanca (commented):
Sounds like an acceptable feature to add, yup. I left a naming suggestion which I think we should address before merging this.

Also, looks like we should add a test to our test suite, at least to verify that the config ingestion logic works well?

Eg we've got fixtures for a certfile and a keyfile in `conftest.py`. So we can probably add an `ENCRYPTED_PRIVATE_KEY` that's an encrypted version of `PRIVATE_KEY`, and then make sure our config is able to load that properly into the `ssl_context`, by updating `test_config.py`?

florimondmanca (changes requested):
Marking as "Changes requested" since I don't think this should be merged as-is for now, at least not without a basic test to prevent future regressions. :-)

remster85:
Thanks @florimondmanca for the review , allow me some time to get back to you.

euri10:
thanks a lot @remster85, I built on your PR in #808 and added the tests etc using trustme.

remster85:
Thanks so much @euri10 !

Issue #344, opened 2019-04-04
Sporatic errors with nginx connection pipelining


nhumrich:
There seems to be some weird issue with pipelining and or connection reuse and nginx. I have nginx sitting in front of my uvicorn/starlette app, and on patch/post requests I occasionally get 502's from nginx (about 30-40% of the time). The nginx logs say:

```
upstream prematurely closed connection while reading response header from upstream
```

all over the place

and my app logs say:
```
WARNING: Invalid HTTP request received.
WARNING: Invalid HTTP request received.
```
with not much more information.

I forced connection reuse to not work by adding the default header `Connection: close`, which forces nginx to kill the connection. Performance drops significantly, but at least I don't get 502's. 

```python
uvicorn.run(app, host='0.0.0.0', port=8445, headers=[('Server', 'custom'), ('Connection', 'close')], proxy_headers=True)
```

potentially relevant: the 502 doesn't seem to ever happen on a GET

lovelydinosaur:
Okay - so I'd start by trying each of the `h11` and `httptools` implementations, and see if replicates in both cases. You could also dig into if you can replicate with hypercorn or daphne.

That'll give us a first pass onto what's going on.

Uvicorn will respond like that if it gets a malformed request. So another route into figuring out the issue would be to change the source code so that it logs out the raw byte streams that are being sent, and examine how they look in the broken cases.

nhumrich:
Great idea. How do I swap between h11 and httptools?

lovelydinosaur:
So. `--http h11` or `--http httptools`.

Def. also recommend trying to replicate on Daphne or Hypercorn following that too.

Let's narrow it down as much as poss as a first step.

nhumrich:
The issue only happens with httptools.
I could not reproduce with h11 on uvicorn. 
I could not reproduce using hypercorn.
I wasn't able to get daphne working.

Hope that helps.

leobuskin:
I can confirm this issue, can see such behavior from uvicorn==0.4.3 (didn't test earlier). Seems related to websockets, going to investigate deeper next week. @nhumrich are you using ws?

nhumrich:
@leobuskin no, I am not using websockets at all

gvbgduh:
[...]

PR #778, opened 2020-09-16
Handle `on_url` being called multiple times on the same request

malthejorgensen:
Refs #344 

In some cases -- specifically when there's TCP packet fragmentation -- the `on_url` callback in httptools can be called multiple times in a single request. In that case the URL has to be built from the separate calls.

This PR moves the per-request state reset to the `on_message_begin`-callback and moves the URL parsing to `on_headers_complete` where we know that all of the URL has been received.

### DIFF
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -116,7 +116,7 @@ class HttpToolsProtocol(asyncio.Protocol):
         self.pipeline = []
 
         # Per-request state
-        self.url = None
+        self.url = b""
         self.scope = None
         self.headers = None
         self.expect_100_continue = False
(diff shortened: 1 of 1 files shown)


florimondmanca:
@malthejorgensen Thanks for submitting this PR! I'm tempted to close this _for now_ for housekeeping purposes, since others have reported that this might not solve the problem fully — there might be a deeper root cause.

I've posted a strategy to digging deeper into the issue here: https://github.com/encode/uvicorn/issues/344#issuecomment-743745260

In the meantime I believe you can keep running off the branch on your fork, right?

FWIW this "it's possible that the URL be fragmanted into multiple TCP packets leading to multiple calls to `.on_url()`" idea _sounds_ legit in its own right. But until we get a confirmation I'd rather we focus on a shared investigation in #344, and perhaps come back to this one we're sure TCP fragmentation of URLs is what's _causing_ the issue you're seeing.

Thank you. :-)

PR #781, opened 2020-09-17
Add watchdog to gunicorn worker

jadbin:
It is similar to [watchdog](https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/gtornado.py#L55) and [heartbeat](https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/gtornado.py#L63) in tornado worker implemented by gunicorn.

1.  **Exit appropriately when worker should not be alive**

If we set `reload = True` in gunicorn configuration, gunicorn could restart workers automatically when code changes. The reloader set ``self.alive = False`` and the worker should exit then.  It is one of the most useful features when we developing web applications. However, the original UvicornWorker could not set `server.should_exit = True` when  gunicorn reloader set `self.alive = False` . So the worker could not exit appropriately and fail to reboot.

2. **Shutting down when parent changed**

The original UvicornWorker will keep alive when parent changed, e.g. kill -9 the gunicorn master process.

### DIFF
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import os
 
 from gunicorn.workers.base import Worker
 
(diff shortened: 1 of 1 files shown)


jadbin:
The worker should keep the signal handlers defined by gunicorn to response the kill signal from master process.
https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L628

euri10:
interesting use case ! I'm going to close this for the following reason :

One of the weak points in uvicorn where there's no testing whatsoever made on the gunicorn workers and it's kind of hard to pull the trigger and change that part without any tests.

I'd be in favor of looking at that once we got our ass covered on that front.

I added some tests long time ago and those were reverted, just to mention some work has been done and should you be interested in taking it back I could use some some `git log` magic to point you at it.

PR #790, opened 2020-09-28
Move package editable install to requirements.txt

florimondmanca:
Hey team!

Quick tweak to the `requirements.txt` so that it's treated as the "all dev dependencies" file. Basically drops the duplicate requirements in favor of using those specified in `setup.py`, by installing `-e .[standard]`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,15 +1,4 @@
-# Minimal
-click
-h11
-
-# Optional
-httptools; sys_platform != 'win32'
-uvloop>=0.14.0; sys_platform != 'win32'
-websockets==8.*
-wsproto==0.15.*
-watchgod>=0.6,<0.7
-python_dotenv==0.13.*
-PyYAML>=5.1
+-e .[standard]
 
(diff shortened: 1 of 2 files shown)


tiangolo (approved):
Nice! I didn't know you can do that :nerd_face: :rocket:

euri10:
yes but this fails the CI because wsproto as agreed in #666 (for the real comment it's one issue above in https://github.com/encode/uvicorn/pull/298#issuecomment-549308242) is left at user's discretion and we need it for the tests

requirements.txt
```diff
@@ -1,15 +1,4 @@
-# Minimal
-click
-h11
-
-# Optional
-httptools; sys_platform != 'win32'
-uvloop>=0.14.0; sys_platform != 'win32'
-websockets==8.*
-wsproto==0.15.*
```
euri10, line 9: we need wsproto explcitely for the tests and we dont have it in the setup.py so cant really remvove that one

PR #796, opened 2020-10-03
Note the need to configure trusted "ips" when using unix sockets

### LINKED ISSUE #713: Docs: clarify how to use proxy-headers with unix sockets
It would be great if it could be documented that when using a unix socket the client address which `ProxyHeadersMiddleware` gets will be `None` and that you therefore need to pass `*` as the trusted hosts to get the proxy headers to work.

Even better would be for the proxy headers to "just work" when using a unix socket, in the same way that `ProxyHeadersMiddleware` trusts localhost by default, though I suspect that may be more involved.

PeterJCLaw:
It is unfortunately non-obvious that when using an unix socket, clients connecting to that socket are not trusted as a source of headers for proxying to the underlying application.

Fixes https://github.com/encode/uvicorn/issues/713


euri10 (approved):

euri10:
thanks @PeterJCLaw

PR #799, opened 2020-10-05
Support .yml log config files

mrakitin:
Here is a tiny follow up on https://github.com/encode/uvicorn/pull/665. As the `.yml` is another widely-used extension, it would be great to support both `.yaml` and `.yml` formats.

Useful post https://stackoverflow.com/a/22812835/4143531helped to eliminate the second call of `.endswith()`.


tests/test_config.py
```diff
@@ -165,6 +165,23 @@ def test_log_config_yaml(
     mocked_logging_config_module.dictConfig.assert_called_once_with(logging_config)
 
 
+def test_log_config_yml(
```
euri10: haven't looked at it too closely but it seems this is a full repeat of `test_log_config_yaml` so maybe a better option would be to use a 
`@pytest.mark.parametrize("config_filename", ["log_config.yml", "log_config.yaml"])`

uvicorn/config.py
```diff
@@ -236,7 +236,7 @@ def configure_logging(self):
                 with open(self.log_config) as file:
                     loaded_config = json.load(file)
                     logging.config.dictConfig(loaded_config)
-            elif self.log_config.endswith(".yaml"):
+            elif self.log_config.endswith((".yaml", ".yml")):
```
euri10, line 239: TIL endswith could take a tuple, thanks :+1:

euri10 (changes requested):
Looks good to me pending a small change in tests, thanks

tests/test_config.py
mrakitin: Thanks for the hint, @euri10, I was too lazy and indeed just copied the existing test. Now it's parametrized.

.gitignore
```diff
@@ -1,5 +1,5 @@
 .cache
-.coverage
+.coverage*
```
euri10: Can you remove that please, let's keep the changes tightly scoped

euri10 (changes requested):

euri10:
you can use the `scripts/lint` to solve your black issue

.gitignore
mrakitin: In my case, it was producing a file called `.coverage.<hostname>.77907.207781`, which I accidentally committed. Don't we want to ignore such cases?

euri10:
It's not relevant to the current PR nor to the repo itself, you can use a
global gitignore if you feel like there's something that should be ignored,

Should we go this way, we should ignore vscode config stuff, jetbrains ones
etc etc.. it's endless.

euri10 (approved):

euri10:
thanks @mrakitin  !

mrakitin:
Thanks @euri10 for your review & merging :)

mrakitin:
> It's not relevant to the current PR nor to the repo itself, you can use a global gitignore if you feel like there's something that should be ignored, Should we go this way, we should ignore vscode config stuff, jetbrains ones etc etc.. it's endless.

I like the proposed approach. However, I am seeing a slight difference between the editors' config files/dirs (`.idea/`, `.vscode/`, etc.) and the coverage file(s). The editor configs are brought by a user, and indeed, the list can be endless (however, that can be covered by existing configs such as from our [cookiecutter's config](https://github.com/NSLS-II/scientific-python-cookiecutter/blob/master/%7B%7B%20cookiecutter.repo_name%20%7D%7D/.gitignore)). However, the coverage files are generated by the `pytest` run on a repo. In other words, such files are generated by the project's infrastructure, not something that users bring in, so it'd be the right place to control it on the repo side.

But it's your project, and I am happy to follow "the terms and conditions".

euri10:
It's not my project !!!
If you're annoyed by the .coverage in the ignore then just open another pr
and we deal with just that.
There won't be any consequences for this particular change but down the
road it proved useful to have tightly scoped pr in case of regressions,
it's way easier to revert a small pr than some parts of a big one dealing
with unrelated changes.
I'm not against this change by any means,just not here 😁

mrakitin:
Got it, thanks!

PR #800, opened 2020-10-05
Add support for proposed "lifespan.shutdown.notice" ASGI extension

### LINKED ISSUE #451: How do I cancel a streaming response on shutdown?

```python
import asyncio
from datetime import datetime

import uvicorn

class App:

    def __call__(self, scope):
            return self.lifespan
        elif scope['type'] == 'http':
            return self.http

        message = await receive()
        print('startup')

        message = await receive()
        print('shutdown')
        
        """Handle http messages"""

        message = await receive()

        while message['more_body']:
            message = await receive()

        await send({
            'status': 200,
            'headers': [
            ]
        })

        while True:
            await send({
                'type': 'http.response.body',
                'more_body': True
            })

            try:
                    print('disconnect')
                    break
            except asyncio.TimeoutError:
                print('timeout')
                receive_task.cancel()

if __name__ == '__main__':
    uvicorn.run(App(), port=9009)
```

justindujardin:
I'm working with Uvicorn and an indefinitely looping endpoint using [Server-Sent Events](https://www.w3.org/TR/eventsource/) to push unidirectional data to browser clients without client polling or WebSockets.

When Uvicorn shuts down, my response is unable to break out of its loop because the spec does not fire "shutdown" signals until after all connections have been closed. 

I opened a proposed change to the ASGI spec here: https://github.com/django/asgiref/pull/199

This PR is code to support the proposed spec changes, and could be used to fix #451, but has to remain a draft until (if) the changes are adopted.

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -549,9 +549,6 @@ class Server:
         return False
 
     async def shutdown(self, sockets=None):
-        # Send the lifespan shutdown before event, allowing looping requests to exit
-        await self.lifespan.shutdown_notice()
-
         logger.info("Shutting down")
 
         # Stop accepting new connections.
@@ -561,6 +558,9 @@ class Server:
             sock.close()
         for server in self.servers:
             await server.wait_closed()
+
+        # Send the lifespan shutdown before event, allowing looping requests to exit
+        await self.lifespan.shutdown_notice()
 
         # Request shutdown on all existing connections.
         for connection in list(self.server_state.connections):
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -559,6 +559,9 @@ class Server:
         for server in self.servers:
             await server.wait_closed()
 
+        # Send the lifespan shutdown before event, allowing looping requests to exit
+        await self.lifespan.before_shutdown()
+
         # Request shutdown on all existing connections.
         for connection in list(self.server_state.connections):
             connection.shutdown()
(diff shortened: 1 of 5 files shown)


tiangolo (approved):
Awesome! :nerd_face: :rocket: 

This LGTM :tada: 

Let's wait for the conversation in the ASGI spec, but I think this is great.

uvicorn/main.py
```diff
@@ -549,6 +549,9 @@ async def on_tick(self, counter) -> bool:
         return False
 
     async def shutdown(self, sockets=None):
+        # Send the lifespan shutdown before event, allowing looping requests to exit
+        await self.lifespan.shutdown_notice()
```
sm-Fifteen: Isn't it race-y to close long-running requests here, before the server stopped receiving new ones? That still leaves a time window (however brief) where a new connection could be opened, thus preventing the server from closing again. I know the proposed ASGI lifespan RFC says that's when the signal is meant to be emitted, but is there a reason for that?

justindujardin: The intent was to send the event as early as possible when doing shutdown. You're right that doing it before we stop accepting connections creates a race, so it should happen after that.

I adjusted the implementation and the RFC language to reflect moving when the event is triggered. 

Thanks!

justindujardin:
It's been a month, so I think the ASGI spec modification for this PR is unlikely to be merged.

@tomchristie Can you suggest an agreeable change to support this use-case? I'm not very familiar with Uvicorn, but I'm willing to do the implementation + tests if you can help me figure out how it should look.

PR #803, opened 2020-10-06
Fix reload with ipv6 host

### LINKED ISSUE #553: reload option works only with IPv4 host

### LINKED ISSUE #791: Built address of server in logs is wrong in case of a ipv6 host

euri10:
fixes #553

correct the url printed in the logs if ipv6 is used, using brackets arout the ip as recommanded in https://www.ietf.org/rfc/rfc2732.txt

so fixes #791 as well


uvicorn/config.py
```diff
@@ -112,6 +112,20 @@ def create_ssl_context(certfile, keyfile, ssl_version, cert_reqs, ca_certs, ciph
     return ctx
 
 
+def ipvXmessage(version=None):
```
cdeler: I have some concerns about the function.

First of all, it's a kind of a public function (at least the name doesn't start with `_`) and from my point of view it means that user can (technically) call the function. And from my point of view the naming is not pythonic, I'd change the signature to 

```suggestion
def _get_server_start_message(host_ip_version: int = 4) -> Tuple[str, str]:
```

uvicorn/config.py
```diff
@@ -112,6 +112,20 @@ def create_ssl_context(certfile, keyfile, ssl_version, cert_reqs, ca_certs, ciph
...
+    if version == 6:
+        ip_repr = "%s://[%s]:%d"
+    elif version == 4:
```
cdeler: Having missed `version` argument, we'll get a `NameError` below, as `ip_repr` won't be defined.

I'd replace this `elif` by else:
```suggestion
    else:
```

(moreover I'd introduce an enum there, but it's might be an over-engineering)

euri10: Let me know what you think @cdeler 
I didn't find enums for the 4 or 6 version in the std lib ipaddress but I may have searched badly so I created one.
The only thing missing here is that the ipv6 link is not clickable but I guess it's a shell issue, will check that separetly

cdeler: It's fine, it's what I meant about the enum.

uvicorn/config.py
```diff
@@ -112,6 +113,27 @@ def create_ssl_context(certfile, keyfile, ssl_version, cert_reqs, ca_certs, ciph
...
+    IPv6 = "IPv6"
+
+
+def _get_server_start_message(
+    host_ip_version: _IPKind = _IPKind.IPv4,
+) -> Tuple[str, str]:
+    if host_ip_version == _IPKind.IPv6:
+        ip_repr = "%s://[%s]:%d"
+    elif host_ip_version == _IPKind.IPv4:
+        ip_repr = "%s://%s:%d"
```
cdeler:
```suggestion
    if host_ip_version is _IPKind.IPv6:
        ip_repr = "%s://[%s]:%d"
    else:
        ip_repr = "%s://%s:%d"
```

cdeler (approved):
Everything is fine apart from one remark. If you think it's a reasonable suggestion you can commit it (but I'm ok without it)

Approved :fire:

Issue #806, opened 2020-10-07
Support custom SSL Context


samypr100:
### Checklist

- [x] There are no similar issues or pull requests for this yet.
- [ ] I discussed this idea on the [community chat](https://gitter.im/encode/community) and feedback is positive.

### Is your feature related to a problem? Please describe.

I would like to pass an existing [SSL Context](https://docs.python.org/3/library/ssl.html#ssl.SSLContext) to `uvicorn.run()`. For example, I have a certificate that needs a password to load. Typically I would do that by setting up a context like so:
```python
ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
# ... customize context even more
ssl_context.load_cert_chain(ssl_crt_path, keyfile=ssl_key_path, password=ssl_key_password)
```
The current options are limited to these kinds of advance scenarios and I'd like to avoid keep adding/requesting `--ssl-xyz` options for each of those scenarios. I know I can decrypt the key before loading it into python, but I'm limited on the environment I need to deploy on since I'm given the encrypted key and the password via a secret.

## Describe the solution you would like.

Adding the ability to pass a `ssl_context` to `uvicorn.run` in python code that supersedes any of the `ssl_*` settings if provided.

Example changes in `uvicorn/config.py`:

```python
@property
def is_ssl(self) -> bool:
    return bool(self.ssl_keyfile or self.ssl_certfile)

@property
def is_ssl_context(self) -> bool:
    return isinstance(self.ssl_context, ssl.SSLContext)

# ...

if self.is_ssl and not self.is_ssl_context:
    self.ssl = create_ssl_context(
        keyfile=self.ssl_keyfile,
        certfile=self.ssl_certfile,
        ssl_version=self.ssl_version,
        cert_reqs=self.ssl_cert_reqs,
        ca_certs=self.ssl_ca_certs,
        ciphers=self.ssl_ciphers,
    )
elif self.is_ssl_context:
    self.ssl = self.ssl_context
else:
    self.ssl = None

# ...
```

## Describe alternatives you considered

Searched source code to see if there was a way to pass a custom context to no avail.

## Additional context

Since ssl context is createe via python, it would not quite be supported via command line. Unless we want to get fancy. I can attempt to do a PR if permitted. Thanks!

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
[...]

PR #807, opened 2020-10-07
Refactor ssl tooling using trustme

euri10:
This removes hardcoded values for certificates / private keys and uses fixtures based on the trustme library

I think it's worthy addition as a preliminary work towards better tests for https://github.com/encode/uvicorn/pull/776 and maybe in anticipation of a resolution of https://github.com/encode/uvicorn/issues/806

It also adds a test that shows how to use a combined key and certificate (`test_run_chain`)


tests/supervisors/test_watchgodreload.py
```diff
@@ -11,7 +11,7 @@ def run(sockets):
     pass
 
 
-def test_watchgodreload(certfile_and_keyfile):
+def test_watchgodreload(tls_ca_certificate_pem_path, tls_ca_certificate_private_key_path):
```
euri10: have to say I'm surprised I had to modify this since those fixtures are not used in the test, I think we shoud remove them, letting it that way just in case I missed something

cdeler (approved):
Everything is fine :rocket:

euri10:
much appreciated @cdeler  !

PR #808, opened 2020-10-07
Adding ability to decrypt ssl key file

euri10:
Should be merge after https://github.com/encode/uvicorn/pull/807 as it builds on it

Built on https://github.com/encode/uvicorn/pull/776 , added tests and changed what @florimondmanca  wanted

The encrypted key fixture is stolen from https://github.com/urllib3/urllib3/blob/master/dummyserver/server.py#L217-L226 as advised by trio wizards !


euri10:
Now #807 is merged this one is rather easy !

tests/conftest.py
```diff
@@ -1,5 +1,7 @@
 import pytest
 import trustme
+from cryptography.hazmat.backends import default_backend
```
cdeler, line 3: I'm sorry, but I missed where this dependency is specified?

May be I'm wrong, but using a dependency in import section, you should specify it somewhere, e.g. in either `requirements.txt` or  `setup.py`

cdeler (approved):
Everything is fine, apart one comment (which you can fix if you agree with my remark)

tests/conftest.py
euri10: you're absolutely right, but I think trustme installs it by default, will check to be 100% sure

euri10: yep we got covered by it:

https://github.com/python-trio/trustme/blob/a09f8e2ccc2ff321c4aee7870077bffb998d73f0/setup.py#L17-L23

cdeler: If you asked me, I'd add it to the requirements because you directly import this library in the project source.

The logic is, having omitted `cryptography` for now, you create an implicit dependency on `trustme` code (a-la "`trustme` provides us with `cryptography`).

If it wasn't "test" code, I'd say that it's mandatory to add `cryptography` to the `uvicorn` dependencies, but may be it's acceptible for test code

euri10: OK I added it, I agree with you, better be explicit, you changed my mind !

cdeler: Awesome! Feel free to merge it 🚀

euri10:
Fixes https://github.com/encode/uvicorn/issues/581

PR #805, opened 2020-10-06
Update .gitignore to ignore .coverage.*

mrakitin:
This is a scoped PR to update the `.gitignore` file with the extra files generated by `pytest`. This was discussed within https://github.com/encode/uvicorn/pull/799 with @euri10. Not sure why `pytest` generates both the `.coverage` and `.coverage.<hostname>.<some float number>` files.

PR #809, opened 2020-10-07
Drop pytest-cov and uses vanilla coverage

euri10:
Prompted by https://github.com/encode/uvicorn/pull/805

I personally dont see the advantage of using pytest-cov, this PR drops it and uses vanilla coverage.

It solves the creation of those coverage.hostname.xxxx left by pytest-cov that the aforementioned PR wanted to ignore.

Any preferences ?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -19,4 +19,4 @@ known_first_party = uvicorn,tests
 known_third_party = click,does_not_exist,gunicorn,h11,httptools,pytest,requests,setuptools,urllib3,uvloop,watchgod,websockets,wsproto,yaml
 
 [tool:pytest]
-addopts = --cov=uvicorn --cov=tests -rxXs
+addopts = -rxXs
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,3 +20,7 @@ known_third_party = click,does_not_exist,gunicorn,h11,httptools,pytest,requests,
 
 [tool:pytest]
 addopts = -rxXs
+
+[coverage:run]
+omit=venv/*
+source=uvicorn,tests
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,5 +22,5 @@ known_third_party = click,does_not_exist,gunicorn,h11,httptools,pytest,requests,
 addopts = -rxXs
 
 [coverage:run]
-omit=venv/*
-source=uvicorn,tests
+omit = venv/*
+include = uvicorn/*, tests/*
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,7 +14,6 @@ codecov
 flake8
 isort
 pytest
-pytest-cov
 pytest-mock
 requests
 mypy
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,13 +10,13 @@ wheel
 # Testing
 autoflake
 black
-codecov
 flake8
 isort
 pytest
 pytest-mock
 requests
 mypy
+coverage
 
 # Documentation
 mkdocs
--- a/scripts/test
+++ b/scripts/test
@@ -4,7 +4,6 @@ export PREFIX=""
 if [ -d 'venv' ] ; then
     export PREFIX="venv/bin/"
 fi
-export SOURCE_FILES="uvicorn tests"
 
 set -ex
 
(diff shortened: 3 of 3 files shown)


setup.cfg
```diff
@@ -19,4 +19,8 @@ known_first_party = uvicorn,tests
 known_third_party = click,does_not_exist,gunicorn,h11,httptools,pytest,requests,setuptools,urllib3,uvloop,watchgod,websockets,wsproto,yaml
 
 [tool:pytest]
-addopts = --cov=uvicorn --cov=tests -rxXs
+addopts = -rxXs
+
+[coverage:run]
+omit=venv/*
+source=uvicorn,tests
```
cdeler: I don't want to be so pedantic, but can you add a newline at the end of the file?

cdeler:
I wonder if it solves the problem with tests debugging under PyCharm. If so it's might be reasonable to copy it to `httpcore` and `httpx` (`pytest-cov` is causing [breakpoints to be skipped](https://pytest-cov.readthedocs.io/en/latest/debuggers.html))

Approved 👍 
_with a small comment_

cdeler (approved):

setup.cfg
JayH5: From experience, `include` produces coverage reports with nicer path information (if you ever use something like Codecov).
```suggestion
include = uvicorn/*, tests/*
```

JayH5 (commented):
I also prefer vanilla coverage over pytest-cov, but don't really have a _strong_ argument for why. 

I guess my two main reasons for preferring it are:
1. pytest-cov implicitly adds a few config options to coverage that are sometimes surprising, like with the multiprocess thing you pointed out
2. I'd rather learn to use Coverage, a generic (and very well-maintained) tool that is not pytest-specific, than pytest-cov. Using vanilla Coverage with pytest is not generally difficult/complicated.

But I don't think these are super strong reasons so it's perhaps a discussion for the larger encode org to have.

cdeler:
@JayH5 the reason might be the problem with debugging in IDE I mentioned. For example I'm executing each test with `--no-cov` which drives me crazy (sorry I cannot copy the link from prev comment as I'm writing from phone)

I meant that I want to be able to run tests under debugging without cov enabled

florimondmanca:
I think we're using this alternative "no Pytest-cov" approach in the HTTPCore repo...? There's a separate `scripts/coverage` script that checks for coverage, and I don't remember that `scripts/test` had to change much... Or maybe we actually do use Pytest-cov, not sure 😅 But might be worth checking how we're doing it there!

setup.cfg
euri10: done

euri10: yes indeed, done

euri10:
> @JayH5 the reason might be the problem with debugging in IDE I mentioned. For example I'm executing each test with `--no-cov` which drives me crazy (sorry I cannot copy the link from prev comment as I'm writing from phone)
> 
> I meant that I want to be able to run tests under debugging without cov enabled

yep I have a pytest template with `--no-cov` for every projet using pytest-cov and it's super annoying indeed.

euri10:
> I think we're using this alternative "no Pytest-cov" approach in the HTTPCore repo...? There's a separate `scripts/coverage` script that checks for coverage, and I don't remember that `scripts/test` had to change much... Or maybe we actually do use Pytest-cov, not sure  But might be worth checking how we're doing it there!

nope you are using pytest-cov and coverage report, the latter being obviously useless since pytest-cov take care of it

graingert:
`coverage run -m pytest` changes how the sys.path exists when running the tests - eg you'll be testing the source files rather than the installed package

euri10:
> `coverage run -m pytest` changes how the sys.path exists when running the tests - eg you'll be testing the source files rather than the installed package

care to expand on this @graingert ?
from pytest-cov docs I can indeed see a reference to sys.path

> Consistent pytest behavior. If you run coverage run -m pytest you will have slightly different sys.path (CWD will be in it, unlike when running pytest).

I fail to see however how that translates into testing sources files vs installed package, I see no installation process here.

graingert:
@euri10 ah it seems you're not installing uvicorn as part of the tests. That's usually desirable because the source files and the distributed package often have subtle differences

euri10:
yes indeed, we may want to change that.

the biggest grip I had on pytest-cov was that I had to put a `--no-cov` flag everytime I wanted to debug in pycharm, if not set breakpoints were simply skipped.

PR #813, opened 2020-10-11
Added cli suport for headers containing colon

romulocollopy:
Some http headers need to have colon in its values, such as:
```
Content-Security-Policy: default-src 'self'; script-src https://example.com;
```
Passing this value though the client used to raise an error, because this
header would be parsed as:
```
["Content-Security-Policy", "default-src 'self'; script-src https", "//example.com;"]
```
And could no be unpached as key, value pair.

This commit limits the spliting on colon to its first occuence in the string,
enabling us to pass a value containing semicolon in the cli:

```
$ uvicorn main:app --header Content-Security-Policy:default-src *:8000
```

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,0 +1,30 @@
+import importlib
+from unittest import mock
+
+from click.testing import CliRunner
+
+from uvicorn.main import main as cli
+
+HEADERS = "Content-Security-Policy:default-src 'self'; script-src https://example.com"
+main = importlib.import_module("uvicorn.main")
+
+
+def test_cli_headers():
+    runner = CliRunner()
+
+    with mock.patch.object(main, "run") as mock_run:
+        result = runner.invoke(cli, ["tests.test_cli:App", "--header", HEADERS])
+
+    assert result.output == ""
+    assert result.exit_code == 0
+    mock_run.assert_called_once()
+    assert mock_run.call_args[1]["headers"] == [
+        [
+            "Content-Security-Policy",
+            "default-src 'self'; script-src https://example.com",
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change Log
 
+## 0.12.2 - 2020-10-11
+
+### Fixed
+- Accepts headers values containing colon
+
```
euri10: we'll make a PR for the changelog, can you remove that please ?

euri10 (changes requested):
Good catch !
Can you please also run the `scripts/lint` so that it formats everything well ?
thanks

romulocollopy:
@euri10 changes applied!
I was considering to put a .strip() here: https://github.com/encode/uvicorn/blob/master/uvicorn/config.py#L278

Tell me what you think and I can open another PR by this week (I didn't find where to test it yet)

euri10:
> @euri10 changes applied!
> I was considering to put a .strip() here: https://github.com/encode/uvicorn/blob/master/uvicorn/config.py#L278
> 
> Tell me what you think and I can open another PR by this week (I didn't find where to test it yet)

I dont think that's necessary, I just looked briefly but it seems to me we get key / value in Config from maiin so the strip is not necessary since it will have happened already with this fix, but I can mix up things .)

If that's not the case then writing a failing test will make it obvious.

euri10 (approved):

euri10:
thanks @romulocollopy !

romulocollopy:
> thanks @romulocollopy !

my pleasure =)

PR #817, opened 2020-10-14
Fixes watchgod with common prefixes

Roang-zero1:
In the current watchgod implementations directories with similar names and same prefix will remove all except the shortest one.

Example:
``` bash
uvicorn --reload --reload-dir "app" --reload-dir "app_extension" --proxy-headers --host 0.0.0.0 limes.asgi:application
```
With this command line the app_extension directory will be removed form the watch_dirs_set

With this pull request the logic is changed so it uses pathlib (python 3.4+) to recognize common ancestry instead of string compare.

### DIFF
--- a/uvicorn/supervisors/watchgodreload.py
+++ b/uvicorn/supervisors/watchgodreload.py
@@ -1,6 +1,6 @@
 import logging
 import re
-from os import path
+from pathlib import Path
 
 from watchgod import DefaultWatcher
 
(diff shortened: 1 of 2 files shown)


uvicorn/supervisors/watchgodreload.py
```diff
@@ -43,10 +43,9 @@ def __init__(self, config, target, sockets):
                 if compare_dir is watch_dir:
                     continue
 
-                if watch_dir.startswith(compare_dir) and len(watch_dir) > len(
-                    compare_dir
-                ):
+                if compare_dir in watch_dir.parents:
```
euri10, line 46: this is much cleaner indeed and will works up to the top of the tree

euri10 (commented):
LGTM, this is better indeed,
One addition maybe, could you write a small test, won't block on it though so up to you.

Roang-zero1:
Sure, let me look into it.

Roang-zero1:
I've added the test and also replaced the `os.path` with `pathlib` for all tests in this group.

euri10 (approved):
excellent, thanks a lot @Roang-zero1 :+1:

PR #818, opened 2020-10-14
Support disabling default Server and Date headers

### LINKED ISSUE #688: Disable Server Header
Hi, 

I'm looking for a  way to prevent the `server:uvicorn` header from being returned in every response. 
I found https://github.com/encode/uvicorn/issues/243 which is similar but the solution doesn't address this particular problem. 

@davidism [asked this question](https://github.com/encode/uvicorn/pull/321#issuecomment-471183072) but @Jitsusama's suggestion of using "middleware to remove it at the last possible second" doesn't work because the default header is added after all the middlewares & routers have run. 

@antirais's [suggestion](https://github.com/encode/uvicorn/issues/243#issuecomment-470603894) of monkey-patching does work but it's very hacky and I'm looking for a cleaner solution.

Thank you

stiiin:
As described in the individual commit messages, this pull request adds configuration flags to control the default Server and Date headers. With these flags, the Server header can be entirely removed rather than overwritten, and the Date header can be suppressed in (the extremely rare) case that HTTP/1.1 requires an origin server to do so.

Fixes #688

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -68,6 +68,8 @@ Options:
                                   populate remote address info.
   --server-header / --no-server-header
                                   Enable/Disable default Server header.
+  --date-header / --no-date-header
+                                  Enable/Disable default Date header.
   --forwarded-allow-ips TEXT      Comma seperated list of IPs to trust with
                                   proxy headers. Defaults to the
                                   $FORWARDED_ALLOW_IPS environment variable if
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -66,6 +66,8 @@ Options:
                                   Enable/Disable X-Forwarded-Proto,
                                   X-Forwarded-For, X-Forwarded-Port to
                                   populate remote address info.
+  --server-header / --no-server-header
+                                  Enable/Disable default Server header.
   --forwarded-allow-ips TEXT      Comma seperated list of IPs to trust with
                                   proxy headers. Defaults to the
                                   $FORWARDED_ALLOW_IPS environment variable if
--- a/tests/test_default_headers.py
+++ b/tests/test_default_headers.py
@@ -83,3 +83,17 @@ async def test_add_additional_header():
                 and response.headers["server"] == "uvicorn"
                 and response.headers["date"]
             )
+
+
+@pytest.mark.asyncio
+async def test_disable_default_date_header():
+    config = Config(
+        app=app,
+        loop="asyncio",
+        limit_max_requests=1,
+        date_header=False,
+    )
+    async with run_server(config):
+        async with httpx.AsyncClient() as client:
+            response = await client.get("http://127.0.0.1:8000")
+            assert "date" not in response.headers
--- a/tests/test_default_headers.py
+++ b/tests/test_default_headers.py
@@ -37,6 +37,20 @@ async def test_override_server_header():
 
 
 @pytest.mark.asyncio
+async def test_disable_default_server_header():
+    config = Config(
+        app=app,
+        loop="asyncio",
+        limit_max_requests=1,
+        server_header=False,
+    )
+    async with run_server(config):
+        async with httpx.AsyncClient() as client:
+            response = await client.get("http://127.0.0.1:8000")
+            assert "server" not in response.headers
+
+
+@pytest.mark.asyncio
 async def test_override_server_header_multiple_times():
     config = Config(
         app=app,
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -182,6 +182,12 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     help="Enable/Disable default Server header.",
 )
 @click.option(
+    "--date-header/--no-date-header",
+    is_flag=True,
+    default=True,
+    help="Enable/Disable default Date header.",
+)
+@click.option(
     "--forwarded-allow-ips",
     type=str,
     default=None,
@@ -317,6 +323,7 @@ def main(
     access_log: bool,
     proxy_headers: bool,
     server_header: bool,
+    date_header: bool,
     forwarded_allow_ips: str,
     root_path: str,
     limit_concurrency: int,
@@ -359,6 +366,7 @@ def main(
         "workers": workers,
         "proxy_headers": proxy_headers,
         "server_header": server_header,
+        "date_header": date_header,
         "forwarded_allow_ips": forwarded_allow_ips,
         "root_path": root_path,
         "limit_concurrency": limit_concurrency,
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -176,6 +176,12 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     "populate remote address info.",
 )
 @click.option(
+    "--server-header/--no-server-header",
+    is_flag=True,
+    default=True,
+    help="Enable/Disable default Server header.",
+)
+@click.option(
     "--forwarded-allow-ips",
     type=str,
     default=None,
@@ -310,6 +316,7 @@ def main(
     log_level: str,
     access_log: bool,
     proxy_headers: bool,
+    server_header: bool,
     forwarded_allow_ips: str,
     root_path: str,
     limit_concurrency: int,
@@ -351,6 +358,7 @@ def main(
         "reload_delay": reload_delay,
         "workers": workers,
         "proxy_headers": proxy_headers,
+        "server_header": server_header,
         "forwarded_allow_ips": forwarded_allow_ips,
         "root_path": root_path,
         "limit_concurrency": limit_concurrency,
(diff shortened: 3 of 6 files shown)


euri10:
for reference
https://gitter.im/encode/community?at=5f8726a399e1ab4dd1fc7b64

euri10 (commented):
Hi @stiiin sorry for getting back this late,
could you please if you still fancy doing it rebase this PR and adapt your tests to the new "layout" we setup for tests, reading current tests this should be fairly easy, if you dont have time let me know I'll pursue this PR.

euri10:
also if you can find again the relevant part of the RFC that would be awesome to keep it written here for future reference, thanks !

stiiin:
Hi @euri10! I rebased my changes, but without running tests. Let's see if the workflow likes it.

> also if you can find again the relevant part of the RFC that would be awesome to keep it written here for future reference, thanks !

I already mentioned a section of the RFC in the commit message for the Date header change. Is that what you were looking for, or was it something else?

euri10:
the CI fails, you can use `./scripts/lint`

you also can use `./scripts/check` and `./scripts/test` locally also, it would be preferable as now Github forces me to approve workflow runs for 1st time contributor which is a pain in the **** :)

and thanks for the rfc, ~~but I cant find the link to the server one~~

date:  https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.1.2
server: https://datatracker.ietf.org/doc/html/rfc7231#section-7.4.2

tests/test_default_headers.py
```diff
@@ -36,6 +36,20 @@ async def test_override_server_header():
...
+    config = Config(
+        app=App,
+        loop="asyncio",
+        limit_max_requests=1,
+        default_server_header=False,
+    )
+    async with run_server(config):
+        async with httpx.AsyncClient() as client:
+            response = await client.get("http://127.0.0.1:8000")
+            assert "server" not in response.headers
```
euri10, line 50: shouldn't it be `assert "uvicorn" not in response.headers` ?

Kludex: The value is "uvicorn", but the key is "server".

euri10: mmm my bad :facepalm:

stiiin:
Sorry for the failed run, I should've just looked around a bit more to find the linter and test run script.

After fixing the linter issues, it seems I messed up the rebase in a way that removed the behaviour for the Date header, so I re-implemented that. Please review those changes, maybe the naming or the style could be better.

I've changed the commit message about the Server option so that it refers to 7.4.2 from the same RFC, where the wording implies that the Server header is optional.

docs/deployment.md
```diff
@@ -66,6 +66,10 @@ Options:
                                   Enable/Disable X-Forwarded-Proto,
                                   X-Forwarded-For, X-Forwarded-Port to
                                   populate remote address info.
+  --default-server-header / --no-default-server-header
```
euri10: maybe a better name would be `enable-server-header / -- disable-server-header` ? when I read `default-server-header` I would like to see about an option to set it when here it's just a switch.
same for date ?
thoughts ?

stiiin: Naming is tough ;)

I chose these names because they reflect the behaviour as I understand it (but my understanding may be imperfect): if the headers are not provided by the application, uvicorn has default values for these headers, unless you disable that default through this new feature. Users may read `--disable-server-header` as "uvicorn will never send a Server header".

Do `--disable-default-server-header` and `--disable-default-date-header` make more sense?

euri10: > Do --disable-default-server-header and --disable-default-date-header make more sense?

I think that's a good compromise between the 2, @Kludex any preference ? 
I'd prefer having more eyes on this : as you said naming is tough and one can see different things for the same variable name ; so the more eyes will have the better, until everyone disagree :)
let's wait a little bit on this.

Kludex: I'm more in favor of following the pattern already in place. I mean, the option above is `--proxy-headers / --no-proxy-headers`, so I guess `--server-header / --no-server-header` makes more sense. What do you guys think?

stiiin: Fair point. Disregard my `--disable-*` suggestion, I agree that having a predictable interface is more important than having a concise one.

euri10: ok let's go for that last suggestion !

euri10 (approved):
approved with the last naming agreed on ! I'll do the merge once done
thanks a lot for this @stiiin !

uvicorn/main.py
```diff
@@ -175,6 +175,20 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     help="Enable/Disable X-Forwarded-Proto, X-Forwarded-For, X-Forwarded-Port to "
     "populate remote address info.",
 )
+@click.option(
+    "--server-header/--no-server-header",
+    "default_server_header",
```
euri10: for consistency with other flags we have can u use `server_header` and `server_date` ?
```suggestion

```

uvicorn/main.py
```diff
@@ -175,6 +175,20 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
...
+    is_flag=True,
+    default=True,
+    help="Enable/Disable default Server header.",
+)
+@click.option(
+    "--date-header/--no-date-header",
+    "default_date_header",
```
euri10:
```suggestion

```

uvicorn/main.py
```diff
@@ -310,6 +324,8 @@ def main(
     log_level: str,
     access_log: bool,
     proxy_headers: bool,
+    default_server_header: bool,
```
euri10:
```suggestion
    server_header: bool,
```

uvicorn/main.py
```diff
@@ -310,6 +324,8 @@ def main(
...
+    default_date_header: bool,
```
euri10:
```suggestion
    date_header: bool,
```

uvicorn/main.py
```diff
@@ -351,6 +367,8 @@ def main(
         "reload_delay": reload_delay,
         "workers": workers,
         "proxy_headers": proxy_headers,
+        "default_server_header": default_server_header,
```
euri10:
```suggestion
        "server_header": server_header,
```

uvicorn/main.py
```diff
@@ -351,6 +367,8 @@ def main(
...
+        "default_date_header": default_date_header,
```
euri10:
```suggestion
        "date_header": date_header,
```

euri10 (changes requested):

stiiin:
You're right, I should've acted on my _other_ gut feeling ;)

If you don't mind, I'll amend my changes rather than using Github's commit feature. At least I know what's going to happen if I do it through git's CLI.

uvicorn/main.py
stiiin: Resolved as suggested, thanks!

tests/test_default_headers.py
```diff
@@ -36,6 +36,20 @@ async def test_override_server_header():
             )
 
 
+@pytest.mark.asyncio
+async def test_disable_default_server_header():
```
Kludex, line 40:
```suggestion
async def test_disable_server_header():
```

tests/test_default_headers.py
```diff
@@ -69,3 +83,17 @@ async def test_add_additional_header():
                 and response.headers["server"] == "uvicorn"
                 and response.headers["date"]
             )
+
+
+@pytest.mark.asyncio
+async def test_disable_default_date_header():
```
Kludex, line 89:
```suggestion
async def test_disable_date_header():
```

uvicorn/main.py
```diff
@@ -175,6 +175,18 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     help="Enable/Disable X-Forwarded-Proto, X-Forwarded-For, X-Forwarded-Port to "
     "populate remote address info.",
 )
+@click.option(
+    "--server-header/--no-server-header",
+    is_flag=True,
+    default=True,
+    help="Enable/Disable default Server header.",
```
Kludex, line 182:
```suggestion
    help="Enable/Disable server header.",
```

uvicorn/main.py
```diff
@@ -175,6 +175,18 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
...
+)
+@click.option(
+    "--date-header/--no-date-header",
+    is_flag=True,
+    default=True,
+    help="Enable/Disable default Date header.",
```
Kludex, line 188:
```suggestion
    help="Enable/Disable date header.",
```

euri10 (approved):

euri10:
thanks a lot @stiiin !

Kludex (commented):
I forgot to press enter on the review. :sweat_smile: 

I'm going to create a small PR later.

uvicorn/main.py
stiiin: I think the distinction in the help text would still be somewhat helpful. You can always issue a `Server` header from the application code, right?

Kludex: Fair point. :kissing: :+1:

PR #820, opened 2020-10-16
Feature: Change reload to be configurable with glob patterns

### LINKED ISSUE #966: Ignore changes to any top-level directory starting with a "."
### Checklist

I want to build a service that accepts file uploads. During development these uploads are saved in a project-level directory. Currently every upload will trigger `uvicorn` to reload:

```
INFO:     127.0.0.1:55846 - "POST /api/v1/upload HTTP/1.1" 200 OK
WARNING:  WatchGodReload detected file change in '['PROJECT_PATH/.media/3fa85f64-5717-4562-b3fc-2c963f66afa6/cropped.png']'. Reloading...
```

The watchgodreload watcher should implement `should_watch_directory` and [apply the same logic as here](https://github.com/encode/uvicorn/blob/master/uvicorn/supervisors/watchgodreload.py#L24).

### LINKED ISSUE #984: Support --ignore or --reload-file option

My application's file structure is as follows:

```
cloud-app git:(dev) ✗ tree -L 1
.
├── Dockerfile
├── README.md
├── frontend
├── gatekeeper
├── iam
├── main.py
├── manage.py
├── requirements.txt
├── templates
└── util.py
```

`frontend` is a React SPA, `gatekeeper` and `iam` belong to a Django application that we mount as an ASGI sub-application alongside our fastapi ASGI app which is defined in `main.py`.

Currently I run uvicorn as follow:

```
uvicorn main:app --reload --host 0.0.0.0 --port 8000 --reload-dir iam --reload-dir gatekeeper
```

I would like to see one of the following (or both!) flags supported:

* `--ignore-dir frontend`: would enable me to ignore changes to frontend source code that are triggering a large number of reloads of my uvicorn app, despite those changes being irrelevant;
* `--reload-file main.py`: would enable me to whitelist _specific files_ to watch.

I believe WatchDog [can be configured](https://pythonhosted.org/watchdog/api.html) to ignore/exclude files and dirs based on pattern matching.

## Refs

https://pythonhosted.org/watchdog/api.html#watchdog.events.RegexMatchingEventHandler.ignore_regexes
https://pythonhosted.org/watchdog/api.html#watchdog.events.PatternMatchingEventHandler.ignore_patterns

Roang-zero1:
With this change it is possible to control which files will trigger a reload of uvicorn.

Features:

- New Option `reload-include` is a repeatable list of unix style glob patterns [pathlib.Path.glob](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) that will be included in reload watch.
- New Option `reload-exclude` is a repeatable list of unix style glob patterns [pathlib.Path.glob](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) that will be excluded in reload watch.
- Common Watchgod and Startreload configuration:
  - Both reloaders get their directories from the new configurations
- Improved Watchgod implementation:
  - Use `should_watch_dir` to reduce setup overhead
  - Use `should_watch_file` with `pathlib` to verify included paths
  - Only watch .py files by default
- Improved Startreload implementation:
  - Use pathlib glob to search for .py files
- Updated documentation
- Updated tests

Edit by @Kludex:
- Closes #984 
- Closes #966

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -40,7 +40,7 @@ Options:
   --reload-delay FLOAT            Delay between previous and next check if
                                   application needs to be. Defaults to 0.25s.
                                   [default: 0.25]
-  --reload-include TEXT           Set glob patterns to exclude files in
+  --reload-include TEXT           Set glob patterns to include files in
                                   reload watch.
   --reload-exclude TEXT           Set glob patterns to exclude files from
                                   reload watch.
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -31,39 +31,50 @@ Usage: uvicorn [OPTIONS] APP
 Options:
   --host TEXT                     Bind socket to this host.  [default:
                                   127.0.0.1]
+
   --port INTEGER                  Bind socket to this port.  [default: 8000]
   --uds TEXT                      Bind to a UNIX domain socket.
   --fd INTEGER                    Bind to socket from this file descriptor.
   --reload                        Enable auto-reload.
   --reload-dir PATH               Set reload directories explicitly, instead
                                   of using the current working directory.
+
+  --reload-include TEXT           Set glob patterns to include while watching
+                                  for files.
+
+  --reload-exclude TEXT           Set glob patterns to exclude while watching
+                                  for files.
+
   --reload-delay FLOAT            Delay between previous and next check if
                                   application needs to be. Defaults to 0.25s.
                                   [default: 0.25]
-  --reload-include TEXT           Set glob patterns to include files in
-                                  reload watch.
-  --reload-exclude TEXT           Set glob patterns to exclude files from
-                                  reload watch.
+
   --workers INTEGER               Number of worker processes. Defaults to the
                                   $WEB_CONCURRENCY environment variable if
                                   available, or 1. Not valid with --reload.
+
   --loop [auto|asyncio|uvloop]    Event loop implementation.  [default: auto]
   --http [auto|h11|httptools]     HTTP protocol implementation.  [default:
                                   auto]
+
   --ws [auto|none|websockets|wsproto]
                                   WebSocket protocol implementation.
                                   [default: auto]
+
   --ws-max-size INTEGER           WebSocket max size message in bytes
                                   [default: 16777216]
-  --ws-ping-interval FLOAT        WebSocket ping interval  [default: 20.0]
-  --ws-ping-timeout FLOAT         WebSocket ping timeout  [default: 20.0]
+
+  --ws-ping-interval FLOAT        WebSocket ping interval  [default: 20]
+  --ws-ping-timeout FLOAT         WebSocket ping timeout  [default: 20]
   --lifespan [auto|on|off]        Lifespan implementation.  [default: auto]
   --interface [auto|asgi3|asgi2|wsgi]
                                   Select ASGI3, ASGI2, or WSGI as the
                                   application interface.  [default: auto]
+
   --env-file PATH                 Environment configuration file.
   --log-config PATH               Logging configuration file. Supported
                                   formats: .ini, .json, .yaml.
+
   --log-level [critical|error|warning|info|debug|trace]
                                   Log level. [default: info]
   --access-log / --no-access-log  Enable/Disable access log.
@@ -72,6 +83,7 @@ Options:
                                   Enable/Disable X-Forwarded-Proto,
                                   X-Forwarded-For, X-Forwarded-Port to
                                   populate remote address info.
+
   --server-header / --no-server-header
                                   Enable/Disable default Server header.
   --date-header / --no-date-header
@@ -80,36 +92,48 @@ Options:
                                   proxy headers. Defaults to the
                                   $FORWARDED_ALLOW_IPS environment variable if
                                   available, or '127.0.0.1'.
+
   --root-path TEXT                Set the ASGI 'root_path' for applications
                                   submounted below a given URL path.
+
   --limit-concurrency INTEGER     Maximum number of concurrent connections or
                                   tasks to allow, before issuing HTTP 503
                                   responses.
+
   --backlog INTEGER               Maximum number of connections to hold in
                                   backlog
+
   --limit-max-requests INTEGER    Maximum number of requests to service before
                                   terminating the process.
+
   --timeout-keep-alive INTEGER    Close Keep-Alive connections if no new data
                                   is received within this timeout.  [default:
                                   5]
+
   --ssl-keyfile TEXT              SSL key file
   --ssl-certfile TEXT             SSL certificate file
   --ssl-keyfile-password TEXT     SSL keyfile password
   --ssl-version INTEGER           SSL version to use (see stdlib ssl module's)
                                   [default: 17]
+
   --ssl-cert-reqs INTEGER         Whether client certificate is required (see
                                   stdlib ssl module's)  [default: 0]
+
   --ssl-ca-certs TEXT             CA certificates file
   --ssl-ciphers TEXT              Ciphers to use (see stdlib ssl module's)
                                   [default: TLSv1]
+
   --header TEXT                   Specify custom default HTTP response headers
                                   as a Name:Value pair
+
   --version                       Display the uvicorn version and exit.
   --app-dir TEXT                  Look for APP in the specified directory, by
                                   adding this to the PYTHONPATH. Defaults to
                                   the current working directory.  [default: .]
+
   --factory                       Treat APP as an application factory, i.e. a
                                   () -> <ASGI app> callable.  [default: False]
+
   --help                          Show this message and exit.
 ```
 
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -38,9 +38,12 @@ Options:
   --reload-dir PATH               Set reload directories explicitly, instead
                                   of using the current working directory.
   --reload-include TEXT           Set glob patterns to include while watching
-                                  for files.
+                                  for files. Includes '*.py' by default, which
+                                  can be overridden in reload-excludes.
   --reload-exclude TEXT           Set glob patterns to exclude while watching
-                                  for files.
+                                  for files. Includes '.*, .py[cod], .sw.*,
+                                  ~*' by default, which can be overridden in
+                                  reload-excludes.
   --reload-delay FLOAT            Delay between previous and next check if
                                   application needs to be. Defaults to 0.25s.
                                   [default: 0.25]
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -40,6 +40,10 @@ Options:
   --reload-delay FLOAT            Delay between previous and next check if
                                   application needs to be. Defaults to 0.25s.
                                   [default: 0.25]
+  --reload-include TEXT           Set glob patterns to exclude files in
+                                  reload watch.
+  --reload-exclude TEXT           Set glob patterns to exclude files from
+                                  reload watch.
   --workers INTEGER               Number of worker processes. Defaults to the
                                   $WEB_CONCURRENCY environment variable if
                                   available, or 1. Not valid with --reload.
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -31,50 +31,39 @@ Usage: uvicorn [OPTIONS] APP
 Options:
   --host TEXT                     Bind socket to this host.  [default:
                                   127.0.0.1]
-
   --port INTEGER                  Bind socket to this port.  [default: 8000]
   --uds TEXT                      Bind to a UNIX domain socket.
   --fd INTEGER                    Bind to socket from this file descriptor.
   --reload                        Enable auto-reload.
   --reload-dir PATH               Set reload directories explicitly, instead
                                   of using the current working directory.
-
   --reload-include TEXT           Set glob patterns to include while watching
                                   for files.
-
   --reload-exclude TEXT           Set glob patterns to exclude while watching
                                   for files.
-
   --reload-delay FLOAT            Delay between previous and next check if
                                   application needs to be. Defaults to 0.25s.
                                   [default: 0.25]
-
   --workers INTEGER               Number of worker processes. Defaults to the
                                   $WEB_CONCURRENCY environment variable if
                                   available, or 1. Not valid with --reload.
-
   --loop [auto|asyncio|uvloop]    Event loop implementation.  [default: auto]
   --http [auto|h11|httptools]     HTTP protocol implementation.  [default:
                                   auto]
-
   --ws [auto|none|websockets|wsproto]
                                   WebSocket protocol implementation.
                                   [default: auto]
-
   --ws-max-size INTEGER           WebSocket max size message in bytes
                                   [default: 16777216]
-
-  --ws-ping-interval FLOAT        WebSocket ping interval  [default: 20]
-  --ws-ping-timeout FLOAT         WebSocket ping timeout  [default: 20]
+  --ws-ping-interval FLOAT        WebSocket ping interval  [default: 20.0]
+  --ws-ping-timeout FLOAT         WebSocket ping timeout  [default: 20.0]
   --lifespan [auto|on|off]        Lifespan implementation.  [default: auto]
   --interface [auto|asgi3|asgi2|wsgi]
                                   Select ASGI3, ASGI2, or WSGI as the
                                   application interface.  [default: auto]
-
   --env-file PATH                 Environment configuration file.
   --log-config PATH               Logging configuration file. Supported
                                   formats: .ini, .json, .yaml.
-
   --log-level [critical|error|warning|info|debug|trace]
                                   Log level. [default: info]
   --access-log / --no-access-log  Enable/Disable access log.
@@ -83,7 +72,6 @@ Options:
                                   Enable/Disable X-Forwarded-Proto,
                                   X-Forwarded-For, X-Forwarded-Port to
                                   populate remote address info.
-
   --server-header / --no-server-header
                                   Enable/Disable default Server header.
   --date-header / --no-date-header
@@ -92,48 +80,36 @@ Options:
                                   proxy headers. Defaults to the
                                   $FORWARDED_ALLOW_IPS environment variable if
                                   available, or '127.0.0.1'.
-
   --root-path TEXT                Set the ASGI 'root_path' for applications
                                   submounted below a given URL path.
-
   --limit-concurrency INTEGER     Maximum number of concurrent connections or
                                   tasks to allow, before issuing HTTP 503
                                   responses.
-
   --backlog INTEGER               Maximum number of connections to hold in
                                   backlog
-
   --limit-max-requests INTEGER    Maximum number of requests to service before
                                   terminating the process.
-
   --timeout-keep-alive INTEGER    Close Keep-Alive connections if no new data
                                   is received within this timeout.  [default:
                                   5]
-
   --ssl-keyfile TEXT              SSL key file
   --ssl-certfile TEXT             SSL certificate file
   --ssl-keyfile-password TEXT     SSL keyfile password
   --ssl-version INTEGER           SSL version to use (see stdlib ssl module's)
                                   [default: 17]
-
   --ssl-cert-reqs INTEGER         Whether client certificate is required (see
                                   stdlib ssl module's)  [default: 0]
-
   --ssl-ca-certs TEXT             CA certificates file
   --ssl-ciphers TEXT              Ciphers to use (see stdlib ssl module's)
                                   [default: TLSv1]
-
   --header TEXT                   Specify custom default HTTP response headers
                                   as a Name:Value pair
-
   --version                       Display the uvicorn version and exit.
   --app-dir TEXT                  Look for APP in the specified directory, by
                                   adding this to the PYTHONPATH. Defaults to
                                   the current working directory.  [default: .]
-
   --factory                       Treat APP as an application factory, i.e. a
                                   () -> <ASGI app> callable.  [default: False]
-
   --help                          Show this message and exit.
 ```
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -104,50 +104,39 @@ Usage: uvicorn [OPTIONS] APP
 Options:
   --host TEXT                     Bind socket to this host.  [default:
                                   127.0.0.1]
-
   --port INTEGER                  Bind socket to this port.  [default: 8000]
   --uds TEXT                      Bind to a UNIX domain socket.
   --fd INTEGER                    Bind to socket from this file descriptor.
   --reload                        Enable auto-reload.
   --reload-dir PATH               Set reload directories explicitly, instead
                                   of using the current working directory.
-
   --reload-include TEXT           Set glob patterns to include while watching
                                   for files.
-
   --reload-exclude TEXT           Set glob patterns to exclude while watching
                                   for files.
-
   --reload-delay FLOAT            Delay between previous and next check if
                                   application needs to be. Defaults to 0.25s.
                                   [default: 0.25]
-
   --workers INTEGER               Number of worker processes. Defaults to the
                                   $WEB_CONCURRENCY environment variable if
                                   available, or 1. Not valid with --reload.
-
   --loop [auto|asyncio|uvloop]    Event loop implementation.  [default: auto]
   --http [auto|h11|httptools]     HTTP protocol implementation.  [default:
                                   auto]
-
   --ws [auto|none|websockets|wsproto]
                                   WebSocket protocol implementation.
                                   [default: auto]
-
   --ws-max-size INTEGER           WebSocket max size message in bytes
                                   [default: 16777216]
-
-  --ws-ping-interval FLOAT        WebSocket ping interval  [default: 20]
-  --ws-ping-timeout FLOAT         WebSocket ping timeout  [default: 20]
+  --ws-ping-interval FLOAT        WebSocket ping interval  [default: 20.0]
+  --ws-ping-timeout FLOAT         WebSocket ping timeout  [default: 20.0]
   --lifespan [auto|on|off]        Lifespan implementation.  [default: auto]
   --interface [auto|asgi3|asgi2|wsgi]
                                   Select ASGI3, ASGI2, or WSGI as the
                                   application interface.  [default: auto]
-
   --env-file PATH                 Environment configuration file.
   --log-config PATH               Logging configuration file. Supported
                                   formats: .ini, .json, .yaml.
-
   --log-level [critical|error|warning|info|debug|trace]
                                   Log level. [default: info]
   --access-log / --no-access-log  Enable/Disable access log.
@@ -156,7 +145,6 @@ Options:
                                   Enable/Disable X-Forwarded-Proto,
                                   X-Forwarded-For, X-Forwarded-Port to
                                   populate remote address info.
-
   --server-header / --no-server-header
                                   Enable/Disable default Server header.
   --date-header / --no-date-header
@@ -165,48 +153,36 @@ Options:
                                   proxy headers. Defaults to the
                                   $FORWARDED_ALLOW_IPS environment variable if
                                   available, or '127.0.0.1'.
-
   --root-path TEXT                Set the ASGI 'root_path' for applications
                                   submounted below a given URL path.
-
   --limit-concurrency INTEGER     Maximum number of concurrent connections or
                                   tasks to allow, before issuing HTTP 503
                                   responses.
-
   --backlog INTEGER               Maximum number of connections to hold in
                                   backlog
-
   --limit-max-requests INTEGER    Maximum number of requests to service before
                                   terminating the process.
-
   --timeout-keep-alive INTEGER    Close Keep-Alive connections if no new data
                                   is received within this timeout.  [default:
                                   5]
-
   --ssl-keyfile TEXT              SSL key file
   --ssl-certfile TEXT             SSL certificate file
   --ssl-keyfile-password TEXT     SSL keyfile password
   --ssl-version INTEGER           SSL version to use (see stdlib ssl module's)
                                   [default: 17]
-
   --ssl-cert-reqs INTEGER         Whether client certificate is required (see
                                   stdlib ssl module's)  [default: 0]
-
   --ssl-ca-certs TEXT             CA certificates file
   --ssl-ciphers TEXT              Ciphers to use (see stdlib ssl module's)
                                   [default: TLSv1]
-
   --header TEXT                   Specify custom default HTTP response headers
                                   as a Name:Value pair
-
   --version                       Display the uvicorn version and exit.
   --app-dir TEXT                  Look for APP in the specified directory, by
                                   adding this to the PYTHONPATH. Defaults to
                                   the current working directory.  [default: .]
-
   --factory                       Treat APP as an application factory, i.e. a
                                   () -> <ASGI app> callable.  [default: False]
-
   --help                          Show this message and exit.
 ```
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -111,9 +111,12 @@ Options:
   --reload-dir PATH               Set reload directories explicitly, instead
                                   of using the current working directory.
   --reload-include TEXT           Set glob patterns to include while watching
-                                  for files.
+                                  for files. Includes '*.py' by default, which
+                                  can be overridden in reload-excludes.
   --reload-exclude TEXT           Set glob patterns to exclude while watching
-                                  for files.
+                                  for files. Includes '.*, .py[cod], .sw.*,
+                                  ~*' by default, which can be overridden in
+                                  reload-excludes.
   --reload-delay FLOAT            Delay between previous and next check if
                                   application needs to be. Defaults to 0.25s.
                                   [default: 0.25]
--- a/docs/index.md
+++ b/docs/index.md
@@ -104,39 +104,50 @@ Usage: uvicorn [OPTIONS] APP
 Options:
   --host TEXT                     Bind socket to this host.  [default:
                                   127.0.0.1]
+
   --port INTEGER                  Bind socket to this port.  [default: 8000]
   --uds TEXT                      Bind to a UNIX domain socket.
   --fd INTEGER                    Bind to socket from this file descriptor.
   --reload                        Enable auto-reload.
   --reload-dir PATH               Set reload directories explicitly, instead
                                   of using the current working directory.
+
+  --reload-include TEXT           Set glob patterns to include while watching
+                                  for files.
+
+  --reload-exclude TEXT           Set glob patterns to exclude while watching
+                                  for files.
+
   --reload-delay FLOAT            Delay between previous and next check if
                                   application needs to be. Defaults to 0.25s.
                                   [default: 0.25]
-  --reload-include TEXT           Set glob patterns to include files in
-                                  reload watch.
-  --reload-exclude TEXT           Set glob patterns to exclude files from
-                                  reload watch.
+
   --workers INTEGER               Number of worker processes. Defaults to the
                                   $WEB_CONCURRENCY environment variable if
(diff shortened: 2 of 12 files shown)


uvicorn/supervisors/watchgodreload.py
```diff
@@ -1,54 +1,72 @@
...
 from uvicorn.supervisors.basereload import BaseReload
 
 logger = logging.getLogger("uvicorn.error")
 
 
 class CustomWatcher(DefaultWatcher):
-    ignore_dotted_file_regex = r"^\/?(?:\w+\/)*(\.\w+)"
-    ignored = []
+    def __init__(self, root_path: Path, config: Config):
+        default_includes = ["*.py"]
```
Roang-zero1, line 21: Currently only .py files are watched, this is different to the current default behavior.
Please provide feedback which behavior is desired.

uvicorn/supervisors/basereload.py
```diff
@@ -16,8 +17,8 @@
 
 
 class BaseReload:
-    def __init__(self, config, target, sockets):
-        self.config = config
+    def __init__(self, config: Config, target, sockets):
```
Roang-zero1: I've added the type to this class as it is easier to develop on the inheriting classes.
Can be removed if desired.

uvicorn/supervisors/statreload.py
```diff
@@ -11,31 +11,31 @@ class StatReload(BaseReload):
...
-                mtime = os.path.getmtime(filename)
+                mtime = file.stat().st_mtime
             except OSError:  # pragma: nocover
                 continue
 
-            old_time = self.mtimes.get(filename)
+            old_time = self.mtimes.get(file)
             if old_time is None:
-                self.mtimes[filename] = mtime
+                self.mtimes[file] = mtime
```
Roang-zero1, line 32: Here the dict is now bound to a Path instead of a string.
I don't have any preference either way, so please indicate which way to go.

docs/settings.md
```diff
@@ -19,9 +19,11 @@ equivalent keyword arguments, eg. `uvicorn.run("example:app", port=5000, reload=
...
+* `--reload-include <glob-pattern>` - Specify a glob pattern to match files or directories which will be watched. May be used multiple times. By default the following patterns are included: `.py`. These can be overwritten by explicitly excluding them.
+* `--reload-exclude <glob-pattern>` - Specify a glob pattern to match files or directories which will excluded from watching. May be used multiple times. By default the following patterns are excluded: `.*, .py[cod], .sw.*, ~*`. These can be overwritten by explicitly including them.
 
-By default Uvicorn uses simple changes detection strategy that compares python files modification times few times a second. If this approach doesn't work for your project (eg. because of its complexity), you can install Uvicorn with optional `watchgod` dependency to use filesystem events instead:
+By default Uvicorn uses simple changes detection strategy that compares python files modification times few times a second. If this approach doesn't work for your project (eg. because of its complexity), or you need watching of non python files, you can install Uvicorn with optional `watchgod` dependency to use filesystem events instead:
```
euri10: ok reading this I realize this is completely obsolete since we changed the packaging, so it needs to be removed.

```suggestion
By default Uvicorn uses simple changes detection strategy that compares python files modification times few times a second. If this approach doesn't work for your project (eg. because of its complexity), or you need watching of non python files
```

docs/settings.md
```diff
@@ -19,9 +19,11 @@ equivalent keyword arguments, eg. `uvicorn.run("example:app", port=5000, reload=
 ## Development
 
 * `--reload` - Enable auto-reload.
-* `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default all directories in current directory will be watched.
+* `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default the whole current directory will be watched.
+* `--reload-include <glob-pattern>` - Specify a glob pattern to match files or directories which will be watched. May be used multiple times. By default the following patterns are included: `.py`. These can be overwritten by explicitly excluding them.
+* `--reload-exclude <glob-pattern>` - Specify a glob pattern to match files or directories which will excluded from watching. May be used multiple times. By default the following patterns are excluded: `.*, .py[cod], .sw.*, ~*`. These can be overwritten by explicitly including them.
```
euri10, line 27: these are watchgod defaults as far as I know, in any case I'm not sure I'd enter that level of details in the docs unless there's a reason for it.

I recall we discussed at length this in another issue,

what happens in  the case no pattern is specified, which will be the most common use-case ?

tests/supervisors/test_statreload.py
```diff
@@ -11,37 +9,87 @@ def run(sockets):
...
-    try:
+    with tmpdir.as_cwd():
         config = Config(app=None, reload=True)
         reloader = StatReload(config, target=run, sockets=[])
         reloader.signal_handler(sig=signal.SIGINT, frame=None)
-        reloader.startup()
+        reloader.run()
+
+
+class TestWatchGodReload(TestBaseReload):
```
euri10:
```suggestion
class TestStatReload(TestBaseReload):
```

tests/supervisors/test_statreload.py
```diff
@@ -11,37 +9,87 @@ def run(sockets):
     pass
 
 
-def test_statreload():
+def test_statreload(tmpdir):
```
euri10: why this test is left out of the refactor ?

docs/settings.md
```diff
@@ -19,9 +19,11 @@ equivalent keyword arguments, eg. `uvicorn.run("example:app", port=5000, reload=
 ## Development
 
 * `--reload` - Enable auto-reload.
-* `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default all directories in current directory will be watched.
+* `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default the whole current directory will be watched.
+* `--reload-include <glob-pattern>` - Specify a glob pattern to match files or directories which will be watched. May be used multiple times. By default the following patterns are included: `.py`. These can be overwritten by explicitly excluding them.
```
euri10: shouldn't it be *.py ?

```suggestion
* `--reload-include <glob-pattern>` - Specify a glob pattern to match files or directories which will be watched. May be used multiple times. By default the following patterns are included: `*.py`. These can be overwritten by explicitly excluding them.
```

docs/settings.md
```diff
@@ -19,9 +19,11 @@ equivalent keyword arguments, eg. `uvicorn.run("example:app", port=5000, reload=
...
+* `--reload-exclude <glob-pattern>` - Specify a glob pattern to match files or directories which will excluded from watching. May be used multiple times. By default the following patterns are excluded: `.*, .py[cod], .sw.*, ~*`. These can be overwritten by explicitly including them.
 
-By default Uvicorn uses simple changes detection strategy that compares python files modification times few times a second. If this approach doesn't work for your project (eg. because of its complexity), you can install Uvicorn with optional `watchgod` dependency to use filesystem events instead:
+By default Uvicorn uses simple changes detection strategy that compares python files modification times few times a second. If this approach doesn't work for your project (eg. because of its complexity), or you need watching of non python files, you can install Uvicorn with optional `watchgod` dependency to use filesystem events instead:
 
 ```
 $ pip install uvicorn[watchgodreload]
```
euri10:
```suggestion

```

docs/deployment.md
```diff
@@ -36,6 +36,10 @@ Options:
   --reload                        Enable auto-reload.
   --reload-dir TEXT               Set reload directories explicitly, instead
                                   of using the current working directory.
+  --reload-include TEXT           Set glob patterns to exclude files in
```
euri10:
```suggestion
  --reload-include TEXT           Set glob patterns to include files in
```

docs/index.md
```diff
@@ -105,6 +105,10 @@ Options:
   --reload                        Enable auto-reload.
   --reload-dir TEXT               Set reload directories explicitly, instead
                                   of using the current working directory.
+  --reload-include TEXT           Set glob patterns to exclude files in
```
euri10:
```suggestion
  --reload-include TEXT           Set glob patterns to include files in
```

tests/supervisors/test_statreload.py
euri10:
```suggestion
class TestStatReload(TestBaseReload):
```

tests/supervisors/test_watchgodreload.py
```diff
@@ -12,96 +10,196 @@ def run(sockets):
 
 
 def test_watchgodreload(
-    tls_ca_certificate_pem_path, tls_ca_certificate_private_key_path
+    tls_ca_certificate_pem_path, tls_ca_certificate_private_key_path, tmpdir
```
euri10: could throw another PR to keep things separated but since it's not used here we should drop it
```suggestion
 tmpdir
```

tests/supervisors/test_watchgodreload.py
```diff
@@ -12,96 +10,196 @@ def run(sockets):
...
+            config = Config(app=None, reload=True)
+            reloader = self._setup_reloader(config)
 
-        assert not reloader.should_restart()
-        time.sleep(0.1)
+            assert self._reload_tester(reloader, update_file)
+
+            reloader.shutdown()
+
+    def test__not_should_reload_when_python_file_in_subdir_is_changed(self):
```
euri10:
```suggestion
    def test_should_not_reload_when_python_file_in_subdir_is_changed(self):
```

euri10 (commented):
thanks for this !

I begun reviewing this and realize it would be simpler to split it in 2:

1. the test refactor
2. the added functionality

in no particular order, this one seems simpler from my point of view.

Sorry but cant wrap my head around reviewing that much to be honest, and in any case it's a separate issue.

I left my comments anyway since I begun but the split would be much appreciated

Roang-zero1:
Sure, I'll start the work on separating out the test refactor.
I'll keep this PR for the reimplementation of the reload function and start a new one for the test refactor.

Roang-zero1 (commented):
Here are some points of consideration

docs/settings.md
Roang-zero1: This is in the docs in order to show which files are ignored. These ignores can be explicitly overwritten with an include.
I could also use super(), but then there would be no way to override the defaults.

Roang-zero1:
As requested I've separated out the test class refactor to https://github.com/encode/uvicorn/pull/833.

Ready for new round of review. Added some comments to point out where feedback is requesed.

euri10:
Will look at it over the weekend and get back to you, thanks !

hanneskuettner:
Any updates on this PR?

Kludex:
Closes #984 
Closes #966

Roang-zero1:
Just got the notification. I would be willing to revisit this PR, just let me know what is needed.

Kludex:
Hi @Roang-zero1 :wave: 

I'm going to review it over the weekend. If you'd like, merging master would be great! :smile: 

Sorry for the delay on it!

kennylajara:
> I'm going to review it over the weekend. 

Bump @Kludex

uvicorn/supervisors/statreload.py
Roang-zero1: I've updated the mtimes to be a path in the basereload, I don't think a string is appropriate anymore.

Roang-zero1:
I've rebased the changes onto the current master.
Unfortunately I don't know why the pipeline fails. I've run the lint and committed the changes, but no change.
Test Suite is successful on my local machine.

euri10:
hello @Roang-zero1 , you just need to run `scripts/lint` : that will automatically update the docs with the latest changes you made in the click options

Roang-zero1:
@euri10 That's what I did in fd185c4c2501ad3b2156172c28633d887f1cfc7c when I run it now there are no changes to commit.

The error is not present when I run scripts/check locally

euri10:
I'd reinstall the venv
Maybe you run an old click version ?

euri10:
I think that could be the reason the lint script behaves differently locally and in CI, iirc click 8 changed the cli help output format, so we need to be sure you have locally same version as in CI

if that's not the case I'm at a loss

Roang-zero1:
Typical issue of using the wrong venv... It's fixed now thanks!

tests/supervisors/test_reload.py
```diff
@@ -10,12 +10,11 @@
 from uvicorn.supervisors.watchgodreload import WatchGodReload
 
 
-@pytest.mark.parametrize("reloader_class", [StatReload, WatchGodReload])
 class TestBaseReload:
     tmp_path: Path
 
     @pytest.fixture(autouse=True)
-    def setup(self, tmpdir, reloader_class):
```
euri10, line 18: I'm gonna be annoying, but I'd prefer we drop `tmpdir` entirely, see the small change I made on my branch (sent you a PR on your feat branch if this helps), we therefore deal only with the `tmp_path` fixture which is a nice `Path` object and just have to write a small context manager to change the directory in tests, seems cleaner and more explicit imho.

tests/test_config.py
```diff
@@ -80,9 +80,36 @@ def test_config_should_reload_is_set(
...
-    assert config.reload_dirs == ["reload_me"]
+def test_reload_dir_is_set(tmp_path) -> None:
+    config = Config(app=asgi_app, reload=True, reload_dirs=[str(tmp_path)])
+    assert config.reload_dirs == [tmp_path]
+
+
+def test_non_existant_reload_dir_is_not_set(
+    tmpdir, caplog: pytest.LogCaptureFixture
+) -> None:
+    with tmpdir.as_cwd():
```
euri10: in light of the previous comment, maybe move the context manager as a test util and use it here too

uvicorn/config.py
```diff
@@ -224,13 +227,57 @@ def __init__(
...
+        self.reload_dirs_excludes: List[Path] = []
+        reload_dirs_list = list(set(reload_dirs)) if reload_dirs else []
+        reload_includes = list(set(reload_includes if reload_includes else []))
+        reload_excludes = list(set(reload_excludes if reload_excludes else []))
+
+        for pattern in reload_includes:
+            try:
+                if Path(pattern).is_dir():
+                    reload_dirs_list.append(pattern)
+                    reload_includes.remove(pattern)
```
euri10: do we cover this ? I failed to see where 

ok sorry it's done

uvicorn/config.py
```diff
@@ -224,13 +227,57 @@ def __init__(
...
+                )
+            reload_paths = [Path(os.getcwd())]
+
+        children = []
+        for j in range(len(reload_paths)):
+            for k in range(j + 1, len(reload_paths)):
+                if reload_paths[j] in reload_paths[k].parents:
+                    children.append(reload_paths[k])
+                elif reload_paths[k] in reload_paths[j].parents:
+                    children.append(reload_paths[j])
```
euri10: seems like 278 is not covered,
if I do a 
```
        children = []
        for j in range(len(reload_paths)):
            print(0)
            for k in range(j + 1, len(reload_paths)):
                if reload_paths[j] in reload_paths[k].parents:
                    print(1)
                    children.append(reload_paths[k])
                elif reload_paths[k] in reload_paths[j].parents:
                    print(2)
                    children.append(reload_paths[j])
```
and run the test suite it seems no 1 or 2 are printed, I'm not sure I get what this children is doing, would you show a test case that demonstrate its use ?

uvicorn/supervisors/watchgodreload.py
euri10: I'm ok with that, that would rquire a word or two in the changelog though

uvicorn/supervisors/watchgodreload.py
```diff
@@ -12,23 +10,59 @@
...
 
     def should_watch_file(self, entry: "DirEntry") -> bool:
-        return not any(r.search(entry.name) for r in self._ignored)
+        entry_path = Path(entry)
+        for include_pattern in self.includes:
+            if entry_path.match(include_pattern):
+                for exclude_pattern in self.excludes:
+                    if entry_path.match(exclude_pattern):
+                        return False
+                return True
```
euri10: that's beautiful

euri10:
I made a few comments @Roang-zero1 but this looks good !

tests/supervisors/test_reload.py
Roang-zero1: Nice, I could not figure out how to do that cleanly, so I reverted to tmpdir.
Will merge the branch.

uvicorn/config.py
Roang-zero1: I've added a specific test case for just these config lines.
Thanks for making me write this test case I discovered a rather nasty bug during testing it (Pro Tip: never modify a list you are currently iterating 😅 )
See: 4bd5fd10d700c0a34856e532bd1c8e638a34ae96

uvicorn/config.py
Roang-zero1: The use case is to find directories which are children of other watched directories.
- app
  - sub

and both app and sub are added to reloaddirs.
It would be a waste to create watchers for both app and sub, therefore I remove the children.
I have added a [new test case](https://github.com/Roang-zero1/uvicorn/commit/9a10bf80f7077ea9e78ca7cefe0d0217746a9ae6#diff-da24b355349b53e4c8047034767d008e4c2ce9ea6c8933a3acd50d0828f00528R132-R151), unfortunately I am [using a set](https://github.com/Roang-zero1/uvicorn/blob/9a10bf80f7077ea9e78ca7cefe0d0217746a9ae6/uvicorn/config.py#L255-L260) beforehand so I can't guarantee that both branches are met every time...
@euri10 Do you have input on how I could solve this in the test?

docs/settings.md
Roang-zero1: @euri10  Should we remove this from the docs or do we want to keep it in there to let users know about these patterns?

euri10: I like it to have it here as well as in the cli `--help` output

docs/settings.md
euri10: this one still needs to be removed from the docs, it's not relevant anymore

edit: I mean the line `$ pip install uvicorn[watchgodreload]`

tests/supervisors/test_reload.py
```diff
@@ -96,3 +170,88 @@ def test_should_reload_when_directories_have_same_prefix(self):
...
+            config = Config(app=None, reload=True, reload_includes=[str(app_dir)])
+            reloader = self._setup_reloader(config)
+
+            assert self._reload_tester(reloader, app_file)
+            assert not self._reload_tester(reloader, app_ext_file)
+
+            reloader.shutdown()
+
+    @pytest.mark.parametrize("reloader_class", [StatReload])
+    def test_should_not_parse_filetype_from_includes(self) -> None:
```
euri10: this  one test looks weird to me or I dont get the intent : you include all js files to be checked for a change, then assert that it's not reloaded, shouldn't it be the opposite ?

uvicorn/config.py
euri10: no this looks fine this way

euri10 (commented):
small comment and we soon gtg !

Roang-zero1:
There is a conceptional question I just realized must be discussed.
There are currently 3 settings:
- reload_dir a list setting where one can explicitly set directories which should be watched.
- reload_include a list of glob patterns to match files which trigger a reload.
- reload_exclude a list of glob patterns to match for files which to not trigger a reload.

As the pull request currently stands any directory in reload_dir and any pattern that exactly matches a directory in reload_includes is matched and added to reload_dir. This means it is currently not possible to glob for directories.
I think I should change this to be possible as it would be against my intuition when using these settings.

Example:
Directory structure
root
├── app_first
├── app_second
├── external
├── javascript
└── stylesheets

Let's say our user chose to include `javascript` and `stylesheets` in reload_dir  but decided to use a glob for the apps `app_*`. Then currently the app directories would *not* be watched. I think this is an problem.

I have four ways to resolve this:
- Separate the functionality of reload_dir and reload_include. Therefore the user has to explicitly set the directories they want to match.
- Resolve the directory glob patterns on startup and add them to the reload_dis if they are directories (Would increase startup costs, but is one time)
- Do not resolve the directory globs and let the `should_watch_dir` of `watchgod` handle the check if it is included
- A hybrid of 2 and 3 where the globs are resolved at the start, but watchgod still verifies the patterns against the triggered directories.

I would vote for the fourth solution. Although it adds complexity it is a good middle ground between speed and functionality.

tests/supervisors/test_reload.py
Roang-zero1: See my latest comment. This is more to the functionality as a whole that should be discussed.

docs/settings.md
Roang-zero1: Help text for include:
> Set glob patterns to include while watching
> for files. Includes '*.py' by default, which
> can be overridden in reload-excludes.

Help text for exclude:
>Set glob patterns to exclude while watching
>for files. Includes '.*, .py[cod], .sw.*,
>~*' by default, which can be overridden in
>reload-excludes.

Note I do not want to use the default parameter as it gives the impression that changing this will not include these patterns. Contrary to many other settings these patterns need to be explicitly negated.

euri10:
I'm also in favor of option `#4` @Roang-zero1 and I agree globs should include directories

Roang-zero1:
I've added the changes, unfortunately this meant reworking large parts of this PR 🔨. (Sorry @euri10 for the additional work)
There is now only one watcher for the current working directory. This enables watchgod to pick up new files and directories that match the globs that were not present before starting the server.

I think I've added test cases for all new functionality.
I am a bit on the fence as there is some feature duplication between `config.py` and the `watchgodreload.py`, but I believe this to be unavoidable.

I've added a cache to the watchers, this reduces the CPU load, but costs memory. Please indicate if that is ok.

There is now a fixture for testing all the reload related use cases, it is currently scoped for `package`, as I don't have a good enough understanding please verify it the scope is correct.

Roang-zero1:
Just saw the failed test. I'll look into what the problem with Windows is tomorrow.

euri10:
this is a subtle one:
this test works fine with this logger: `logger.info(f"Will watch for changes in these directories: {list(map(str, self.reload_dirs))}")`
```
        assert (
            caplog.records[-1].message
            == f"Will watch for changes in these directories: {[str(app_dir)]}"
        )
```
you cannot build manually the list `[]`  I think that was what caused the double escaping

tests/conftest.py
```diff
@@ -71,3 +71,43 @@ def tls_ca_ssl_context(tls_certificate_authority: trustme.CA) -> ssl.SSLContext:
     ssl_ctx = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
     tls_certificate_authority.configure_trust(ssl_ctx)
     return ssl_ctx
+
+
+@pytest.fixture(scope="package")
+def reload_directory_structure(tmp_path_factory: pytest.TempPathFactory):
+    """This fixture creates a directory structure to enable reload parameter tests"""
```
euri10:
```suggestion
    """
    This fixture creates a directory structure to enable reload parameter tests
    
    ├── app
    │ ├── css
    │ │ └── main.css
    │ ├── js
    │ │ └── main.js
    │ ├── src
    │ │ └── main.py
    │ └── sub
    │     └── sub.py
    ├── app_first
    │ ├── css
    │ │ └── main.css
    │ ├── js
    │ │ └── main.js
    │ ├── src
    │ │ └── main.py
    │ └── sub
    │     └── sub.py
    ├── app_second
    │ ├── css
    │ │ └── main.css
    │ ├── js
    │ │ └── main.js
    │ ├── src
    │ │ └── main.py
    │ └── sub
    │     └── sub.py
    ├── app_third
    │ ├── css
    │ │ └── main.css
    │ ├── js
    │ │ └── main.js
    │ ├── src
    │ │ └── main.py
    │ └── sub
    │     └── sub.py
    ├── ext
    │ └── ext.jpg
    └── main.py
    """
```

euri10: not sure about that suggestion, just something I may want to see because reading the test it was not obvious from me reading the fixture what the directory layout looked like, but this is very long so....

tests/supervisors/test_reload.py
```diff
@@ -36,63 +39,354 @@ def _reload_tester(self, reloader: BaseReload, file: Path) -> bool:
...
+        with as_cwd(self.reload_path):
+            config = Config(app="tests.test_config:asgi_app", reload=True)
+            reloader = self._setup_reloader(config)
+
+            assert self._reload_tester(reloader, file)
+
+            reloader.shutdown()
+
+    @pytest.mark.parametrize("reloader_class", [WatchGodReload])
+    def test_should_not_reload_when_python_file_in_subdir_is_changed(self) -> None:
```
euri10:
```suggestion
    def test_should_not_reload_when_python_file_in_excluded_subdir_is_changed(self) -> None:
```

tests/supervisors/test_reload.py
```diff
@@ -36,63 +39,354 @@ def _reload_tester(self, reloader: BaseReload, file: Path) -> bool:
...
+                reload_includes=[str(app_dir)],
+            )
+            reloader = self._setup_reloader(config)
+
+            assert self._reload_tester(reloader, app_file)
+            assert not self._reload_tester(reloader, app_first_file)
+
+            reloader.shutdown()
+
+    @pytest.mark.parametrize("reloader_class", [StatReload])
```
euri10: ok here I dont follow, I guess  that will be different here for the `WatchGodReload` case ? why not doing the same as in `test_reload_when_pattern_matched_file_is_changed` or put differently, what does this test differently than `test_reload_when_pattern_matched_file_is_changed`

tests/supervisors/test_reload.py
```diff
@@ -36,63 +39,354 @@ def _reload_tester(self, reloader: BaseReload, file: Path) -> bool:
...
+
+            app_first_dir.mkdir()
+            assert self._reload_tester(reloader, app_first_file)
+            assert caplog.records[-2].levelno == INFO
+            assert (
+                caplog.records[-2].message == "WatchGodReload detected a new reload "
+                f"dir '{app_first_dir.name}' in '{tmp_path}'; Adding to watch list."
+            )
+
+            reloader.shutdown()
```
euri10, line 350: this is neat

tests/test_config.py
```diff
@@ -80,9 +81,156 @@ def test_config_should_reload_is_set(
...
+) -> None:
+    app_dir = reload_directory_structure / "app"
+    with caplog.at_level(logging.INFO):
+        config = Config(
+            app="tests.test_config:asgi_app", reload=True, reload_dirs=[str(app_dir)]
+        )
+        assert len(caplog.records) == 1
+        assert (
+            caplog.records[-1].message
+            == f"Will watch for changes in these directories: ['{app_dir}']"
```
euri10: see my comment about not using `[]`

tests/test_config.py
```diff
@@ -80,9 +81,156 @@ def test_config_should_reload_is_set(
...
+        )
+        assert config.reload_dirs == [reload_directory_structure]
+        assert (
+            caplog.records[-1].message
+            == "Provided reload directories ['reload'] did not contain valid "
+            + "directories, watching current working directory."
+        )
+
+
+def test_reload_subdir_removal(reload_directory_structure: Path) -> None:
```
euri10, line 139:
```suggestion
def test_duplicate_direcories_are_removed(reload_directory_structure: Path) -> None:
```
not sure about this one I let you choose

uvicorn/config.py
```diff
@@ -132,6 +133,57 @@ def create_ssl_context(
...
+        return path.is_dir()
+    except OSError:
+        return False
+
+
+def resolve_reload_patterns(
+    patterns_list: List[str], directories_list: List[str]
+) -> Tuple[List[str], List[Path]]:
+
+    directories: List[Path] = list(set(map(Path, directories_list.copy())))
```
euri10, line 149: writing this without thinking too much or having read the rest, would a deepcopy be better or not ?

uvicorn/config.py
```diff
@@ -132,6 +133,57 @@ def create_ssl_context(
...
+    for pattern in patterns_list:
+        # Special case for the .* pattern, otherwise this would only match
+        # hidden directories which is probably undesired
+        if pattern == ".*":
+            continue
+        patterns.append(pattern)
+        if is_dir(Path(pattern)):
+            directories.append(Path(pattern))
+        else:
+            print("here")
```
euri10: let's remove that

uvicorn/supervisors/basereload.py
```diff
@@ -64,7 +64,6 @@ def startup(self) -> None:
         self.process.start()
 
     def restart(self) -> None:
-        self.mtimes: Dict[str, float] = {}
```
euri10, line 67: why this is dropped ?

uvicorn/supervisors/statreload.py
```diff
@@ -19,31 +18,35 @@ def __init__(
     ) -> None:
         super().__init__(config, target, sockets)
         self.reloader_name = "statreload"
-        self.mtimes = {}
+        self.mtimes: Dict[Path, float] = {}
```
euri10, line 21: because you put it here ?

uvicorn/supervisors/watchgodreload.py
```diff
@@ -12,23 +10,107 @@
...
+        self.dirs_includes = set(config.reload_dirs)
+        self.dirs_excludes = set(config.reload_dirs_excludes)
+        self.resolved_root = root_path
+        super().__init__(str(root_path))
 
     def should_watch_file(self, entry: "DirEntry") -> bool:
-        return not any(r.search(entry.name) for r in self._ignored)
+        cached_result = self.watched_files.get(entry.path)
+        if cached_result is not None:
+            return cached_result
```
euri10, line 49: is it made to be faster ?

uvicorn/config.py
Roang-zero1: No we don't need a deep copy here as we can keep the path elements. I just wanted to decouple the list inside from the outside as I still need the outside one to check what was passed to the config.

uvicorn/supervisors/basereload.py
Roang-zero1: It was moved to StartReload as it is not used anywhere else

uvicorn/supervisors/statreload.py
Roang-zero1: Correct

uvicorn/supervisors/watchgodreload.py
Roang-zero1: Yes, depending on the amount of patterns it might take a while to match all the files. I had to implement it for the directories anyway and reused it for the files.
For the files the cache is something I am on the fence about...

uvicorn/config.py
Roang-zero1: Removed

tests/conftest.py
Roang-zero1: I've taken the middle road and just shortened the directories down to one example with indication they are all the same

tests/supervisors/test_reload.py
Roang-zero1: Sorry Github is not showing the test name, which one is this comment for?

euri10: for     `def test_should_not_parse_filetype_from_includes(self) -> None:`

euri10:
for the failing test, let's deal with that sorting the files in the output for predictable behaviour:

```diff
diff --git a/tests/test_config.py b/tests/test_config.py
index a999b33..73f2e5f 100644
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -225,8 +225,10 @@ def test_reload_includes_exclude_dir_patterns_are_matched(
                 "Will watch for changes in these directories: "
                 in caplog.records[-1].message
             )
-            assert str(first_app_dir) in caplog.records[-1].message
-            assert str(second_app_dir) in caplog.records[-1].message
+            assert (
+                    caplog.records[-1].message
+                    == f"Will watch for changes in these directories: {sorted([str(first_app_dir), str(second_app_dir)])}"
+            )
             assert frozenset(config.reload_dirs) == frozenset(
                 [first_app_dir, second_app_dir]
             )

             )
             )
diff --git a/uvicorn/config.py b/uvicorn/config.py
index 4fae430..13f5898 100644
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -331,7 +331,7 @@ class Config:

             logger.info(
                 "Will watch for changes in these directories: %s",
-                list(map(str, self.reload_dirs)),
+                sorted(list(map(str, self.reload_dirs))),
             )

         if env_file is not None:
```

tests/supervisors/test_reload.py
Roang-zero1: Ok, it's an old duplicate test. I have removed it

uvicorn/supervisors/watchgodreload.py
```diff
@@ -41,26 +123,14 @@ def __init__(
...
-
-        self.watch_dir_set = watch_dirs_set
-        for w in watch_dirs_set:
-            self.watchers.append(CustomWatcher(str(w)))
+        reload_dirs = []
+        for directory in config.reload_dirs:
+            if Path.cwd() not in directory.parents:
+                reload_dirs.append(directory)
+        if Path.cwd() not in reload_dirs:
+            reload_dirs.append(Path.cwd())
```
euri10, line 140: I'm just wondering about those 2 lines because I was trying the branch on my ucivorn repo:
it has a dev folder I gitignored to contain all the dev app I'm loading to debug, and several dot files that are the result of mypy / pytest etc...
```
uvicorn
|--dev
|--.git
|--.idea
|--.pytest_cache
```
now if I do `dev.app:app --reload --log-level=trace --reload-dir=dev`
I get 
```
INFO:     Will watch for changes in these directories: ['/home/lotso/PycharmProjects/uvicorn/dev']
DEBUG:    WatchGodReload detected a new excluded dir '.github' in '/home/lotso/PycharmProjects/uvicorn'; Adding to exclude list.
DEBUG:    WatchGodReload detected a new excluded dir '.git' in '/home/lotso/PycharmProjects/uvicorn'; Adding to exclude list.
DEBUG:    WatchGodReload detected a new excluded dir '.eggs' in '/home/lotso/PycharmProjects/uvicorn'; Adding to exclude list.
DEBUG:    WatchGodReload detected a new excluded dir '.idea' in '/home/lotso/PycharmProjects/uvicorn'; Adding to exclude list.
DEBUG:    WatchGodReload detected a new excluded dir '.pytest_cache' in '/home/lotso/PycharmProjects/uvicorn'; Adding to exclude list.
INFO:     Started reloader process [1445703] using watchgod
```

so looking for changes in dev is ok to me, the unexpected here is the dot folders at the same level.
not sure if that's intended

Roang-zero1: Yeah that's intended, but I see that it is not very intuitive. I'll add a check for the exclude warning.
What is happening is the following:

- We add the reload dir to folders we watch
- We need to add the cwd for folders that could be later found via an include patterns
- The cwd is scanned and foders that are excluded via the default pattern .* are found in cwd, that leads to this messages

I'll check if found folders are in the reload dirs before firing of this warning, that should clear up the unnecessary information.

Roang-zero1: I've added the check, but it will still log for the default patterns (esp. hidden directories) in watched directories.
I think this is fine as it is DEBUG level output...

euri10 (approved):

euri10:
thanks a lot @Roang-zero1 for this, this is a very nice enhancement !

HansBrende:
@Roang-zero1 @euri10 @Kludex this PR introduced a regression and deleted the test for it (`test_reload_dir_is_set()`). Previously you could pass `reload_dirs` a string instead of a list. Now:
```python
Config('myapp:app', reload=True, reload_dirs='src')
```
prints:
```
WARNING:  Provided reload directories ['r', 'c', 's'] did not contain valid directories, watching current working directory.
```
https://github.com/encode/uvicorn/pull/820/files#diff-6982a5e270879281f9182303a9da9776148857b46325afb230e50ef8eb3264c9R294
(Of course, `set` expects an iterable... and strings are iterables of chars.)

PR #822, opened 2020-10-18
Support module:create_app() style factory pattern imports

### LINKED ISSUE #492: How to run app from app factory?
I try to run the server by command `uvicorn my_packege.app:create_app` and got error: `TypeError: __call__() missing 1 required positional argument: 'send'`. What could be the reason?

> [!IMPORTANT]

<a href="https://polar.sh/encode/uvicorn/issues/492">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/492/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

laco:
I think holding state in module level objects (created at import time) is a double-edged sword. It's a great python feature, but comes with a cost: makes things harder to test.

My PR allows the users to define a clear intent to use a callable.
I've implemented the same behavior gunicorn has with the explicit  `()` at the end of the import string.

I saw the previous discussion: #492 

**Edit (florimondmanca)**: Closes  #492

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -200,6 +200,54 @@ For a [PyPy][pypy] compatible configuration use `uvicorn.workers.UvicornH11Worke
 
 For more information, see the [deployment documentation](deployment.md).
 
+### Application factories
+
+You can use functions to instantiate and configure applications and the APP command-line argument can also be a function call. In this case the name will be imported from the module, then called to get the application object.
+
+```python
+import os
+
+
+class App:
+    def __init__(self, config):
+        self.config = config
+
+    async def __call__(self, scope, receive, send):
+        ...
+
+
+def default_config():
+    return {"FOO": os.environ.get("FOO", "BAR")}
+
+
+def create_app(config=None):
+    # Use env to configure and prepare the app before the first call
+    config = config or default_config()
+    app = App(config)
+    return app
+```
+
+Run the server by referencing the function call using the following pattern:
+
+```shell
+$ uvicorn 'example:create_app()'
+```
+
+Providing arguments in the APP string is not supported. This was a decision to keep the application import code simple and fast.
+You can always have multiple factory functions or use some logic inside the factory function to lookup the proper configuration.
+
+However, functions where all parameters have default values can be used as application factory.
+So, it's easier to test different configurations without monkeypatching the environment or any object:
+
+
+```python
+from example import create_app
+
+async def test_create_app_baz():
+    app = create_app(config={"FOO": "BAZ"})
+    ...
+```
+
 ## The ASGI interface
 
 Uvicorn uses the [ASGI specification][asgi] for interacting with an application.
--- a/tests/importer/test_importer.py
+++ b/tests/importer/test_importer.py
@@ -24,6 +24,13 @@ def test_invalid_attr():
     assert expected in str(exc_info.value)
 
 
+def test_invalid_factory_attr():
+    with pytest.raises(ImportFromStringError) as exc_info:
+        import_from_string("tempfile:tempdir()")
+    expected = 'Attribute "tempdir()" is not callable in module "tempfile".'
+    assert expected in str(exc_info.value)
+
+
 def test_internal_import_error():
     with pytest.raises(ImportError):
         import_from_string("tests.importer.raise_import_error:myattr")
@@ -34,6 +41,13 @@ def test_valid_import():
     from tempfile import TemporaryFile
 
     assert instance == TemporaryFile
+
+
+def test_valid_factory_import():
+    instance = import_from_string("tempfile:TemporaryFile()")
+
+    assert instance is not None
+    assert hasattr(instance, "read") and callable(instance.read)
 
 
 def test_no_import_needed():
--- a/uvicorn/importer.py
+++ b/uvicorn/importer.py
@@ -27,11 +27,19 @@ def import_from_string(import_str):
     instance = module
     try:
         for attr_str in attrs_str.split("."):
-            instance = getattr(instance, attr_str)
+            if attr_str.endswith("()"):
+                callable_ = getattr(instance, attr_str[:-2])
+                instance = callable_()
+            else:
+                instance = getattr(instance, attr_str)
     except AttributeError:
         message = 'Attribute "{attrs_str}" not found in module "{module_str}".'
         raise ImportFromStringError(
             message.format(attrs_str=attrs_str, module_str=module_str)
         )
-
+    except TypeError:
+        message = 'Attribute "{attrs_str}" is not callable in module "{module_str}".'
+        raise ImportFromStringError(
+            message.format(attrs_str=attrs_str, module_str=module_str)
+        )
     return instance


florimondmanca:
I'm personally happy to consider this PR. :) The change footprint is reasonably low, and the fact that this matches Gunicorn's support for application factories is quite nice.

Any relevant changes to documentation? We should probably advertise that this alternative style would now supported.

laco:
I'm happy to add some documentation. I'll write a dedicated paragraph under the "Usage" part later today. 

I would like to mention one key difference from gunicorn's implementation: I think gunicorn supports function parameters.  But they are also recommending environment variables instead. So, for simplicity we can say, that only functions with optional arguments or functions without any arguments supported as an application factory.

florimondmanca:
Yes, let's keep it plain and simple — the function will be called with no arguments. People can always wrap things up however they'd want to accommodate that API. :)

florimondmanca:
One thing I'm not super happy with is that this blurs the line for class-based applications. 🧐

Now people can write:

```python
class App:
    async def __call__(...):
        ...
```

And then run it with:

```python
uvicorn app:App()
```

But this is neither targeted by this "factory function use case", nor desirable, since in this class-based case users should still instantiate the app instance like they did before (and if their init function does weird dynamic loading stuff then it's a code smell on their end).

Should we put an additional restrictions in place, for example that the callable must be a function (not a class), or do we think highlighting the use case strongly enough in the docs ("if you're using a factory pattern, do this...") is enough? Or maybe we should do something else entirely to resolve the ambiguity more clearly?

Maybe a more explicit command line flag, eg:

```python
uvicorn app:create_app --factory
```

* Feels "heavier", so it's more obvious when to stick to the default usage.
* The name of the flag reflects the intended use case (factory functions).
* Doesn't even ask the question "but can I pass arguments into `()`?" because there is no function call in there at all.
* Keeps the feature scope as tight as possible, while not being utterly ugly on the eyes or hard to use for people that do need it.

Thoughts?

laco:
I don't see anything bad about simple class-based applications using this feature.  If the application constructor doesn't need any parameters then why not? 

For the documentation I'm thinking about the example below. What do you think? 

In the `app.py`:

```python
class App:
    def __init__(self, config):
        self.config = config

        ...


def default_config():
    return {"FOO": "BAR"}


def setup_app(config=None):
    config = config or default_config()
    app = App(config)
    return app

```
Then start the server with: `uvicorn 'app:setup_app()'`

But, as you pointed out, this can be done as well:

```python
class App2:
    def __init__(self, config=None):
        self.config = config or default_config()

        ...
```
and `uvicorn a:App2()`

I think it's not a big deal, because most of us already using some ASGI frameworks with Application classes already implemented. So in most cases we're going to stick to the factory function to configure these applications to a proper state before the first `__call__ `.

marianhlavac:
I'm really looking forward when this gets merged.

> `--factory`
> Thoughts?

Personally, I don't mind using an explicit flag/switch for the functionality. But if not required, it would reflect 1:1 to gunicorn's functionality and will be familiar for gunicorn users.

florimondmanca:
I guess I'm still way more open to considering this with a tightly-scoped `--factory` flag than I am for accepting Gunicorn's style.

(And actually, the flag would probably be better named as `--app-factory`, as originally suggested by Tom in https://github.com/encode/uvicorn/issues/492#issuecomment-561851598, so that it's clear what kind of "factory" we're talking about.)

I get that this would differ from Gunicorn, but also we don't want users to think we're aiming at mimicking Gunicorn's full set of capabilities, and then coming up in the future asking / proposing something like passing arguments to the factory when we _really_ don't want / need to support anything else then "call the factory with no arguments"

So, yup, tempted to push a bit towards an explicit flag so that once this gets merged, we don't need to expand anything and this feature can be considered "done". :) (Keeping the scope tight on Uvicorn is paramount to its sustainability as there's unfortunately little maintainer headspace.)

florimondmanca:
@laco Yes, your first example looks motivational enough. That being said I think we'll really want to stress the exact motivation for this (which I'm not entirely sure I've grokked myself, so curious to see how you'll write that up), as I'm personally seeing this in the realm of "you most likely don't need this unless you have good reasons to care about startup time".

laco:
I've added some docs to the PR.

In the meantime I've checked gunicorn's implementation for the argument parsing. They are using the `ast` module to parse and eval the arguments. It can be ported, but personally I've never needed that functionality.  It definitely slows down the import process, so I've mentioned in the documentation with a short reason why it's not implemented. It can be added later, but it's not a frequently used gunicorn feature, and not recommended for gunicorn users. Maybe they have this for legacy/backward compatibility reasons...

About the cli arguments:
I'm fully understand you concerns, and I'm not against the `--app-factory` cli flag, but....
The gunicorn worker is one of the recommended production deployment options, so keeping some level of consistency defining the importable APP maybe it isn't a bad idea. 
This was already pointed out in the earlier conversations: https://github.com/encode/uvicorn/issues/492#issuecomment-568757541

Anyhow.. 
The biggest win with the factory pattern is to avoid global mutable state and imports with side-effects. These are generally bad things. And If I can use functions instead of special purpose "import only in some environments" modules to achieve this, than I'm happy.  (Maybe because I'm coming from a functional programming background... :) )

ismaeIfm:
any news on this?

florimondmanca:
I haven't heard from other maintainers on this, but I'm okay _in principle_ with adding this new feature. But let's review this `--factory` vs `()` thing:

For now I see the main point for `()` is consistency with Uvicorn.

The idea is to have this:

```bash
uvicorn --port 5000 app:create_app()
```

So that it looks similar when using Gunicorn with the Uvicorn worker, like this:

```bash
gunicorn --bind :5000 -k uvicorn.workers.UvicornWorker app:create_app()
```

I'm ready to bet that adding this form will mean users would _also_ expect `uvicorn` to accept args and kwargs, and I don't think I want us to have to bother with that, when even the Gunicorn docs don't recommend it. (_"Positional and keyword arguments can also be passed, **but it is recommended to load configuration from environment variables** rather than the command line."_ [Source](https://docs.gunicorn.org/en/stable/run.html#gunicorn)) But then the same argument of being "consistent" would come back up, and if indeed we _don't_ provide args/kwargs support. That would only cause unnecessary frustration for users who (incorrectly) expect it to behave like Gunicorn, when we really just want to allow `() -> App` callables.

Which is why I'm suggesting this:

```bash
uvicorn --port 5000 --factory app:create_app

# …Compared to:
gunicorn --bind :5000 -k uvicorn.workers.UvicornWorker app:create_app()
```

_Yes_, that's a slight inconsistency. But Uvicorn and Gunicorn are two different projects after all. Uvicorn has `--host`, `--port` and `--uds` while Gunicorn has `--bind`: there's never been a contract whatsoever. So while we're at it, I'd prefer us choosing the API that not only is but also _looks_ feature complete, so we implement this once and for all and be done with it. Again, maintainer time has been critically sparse in Encode projects, and Uvicorn especially, so controlling feature creep is key…

If that sounds alright then I'd be happy to review an implementation for the `--factory` flag. (Obviously — thanks  to everyone involved here, but I hope this position makes sense.)

florimondmanca:
Cc'ing @euri10 or @rafalp in case they have any more thoughts, though I hold to a strong preference for a straight-forward `--factory` flag.

docs/index.md
```diff
@@ -200,6 +200,54 @@ For a [PyPy][pypy] compatible configuration use `uvicorn.workers.UvicornH11Worke
...
+
+```shell
+$ uvicorn 'example:create_app()'
+```
+
+Providing arguments in the APP string is not supported. This was a decision to keep the application import code simple and fast.
+You can always have multiple factory functions or use some logic inside the factory function to lookup the proper configuration.
+
+However, functions where all parameters have default values can be used as application factory.
+So, it's easier to test different configurations without monkeypatching the environment or any object:
```
florimondmanca, line 240: See — we shouldn't even have to bother justifying this. Let's just allow `--factory`, say this will call the app factory with no arguments, show an example, and we'll be good.

florimondmanca:
Closing in favor of #875. Thanks for putting this together and following up with discussions! Should be released soon as 0.13.0.

kristjanvalur:
FWIW, to help "grok" the need for application factories, here is where many people are coming from.  I have arrived at Fastapi/uvicorn out of flask and am having a hard time adjusting to the mentality of import-time application creation and global state.  It seems like a step backwards to me and will probably result in problems down the road.
https://hackersandslackers.com/flask-application-factory/

florimondmanca:
@kristjanvalur Thanks for pitching in. I'm aware of this convention in the Flask world, and that it's also what FastAPI uses. But here's my understanding and what I think was the original motivation for import-time apps.

Flask used an `app` instance declared ahead-of-time. After declaring this `app` users can proceed to declare the rest of the web app resources (such as route handlers, middleware, etc).

The need for a factory _arised_ from this ahead-of-time app declaration. One had to have an `app` (or blueprint) reference ready to define basically anything; and so an import-time style quickly led to absurd situations that typically involve circular dependencies.

Quite the contrary, Starlette has made a leap _forward_ by completely decoupling components from the final application instance. All components (routes, middleware, etc) implement the ASGI interface, and so they can be defined separately, in a modular fashion, and then assembled in the final `Starlette()` app instance.

Initially this wasn't the case; Starlette had an ahead-of-time app declaration style too. But that got switched in Starlette 0.11 or so I believe. All along Uvicorn had been encouraging import-style. Initially that was a bit of a stretch, because even Starlette would end up encouraging factories because of the ahead-of-time declaration. But when it got switched, then I believe import-time style became very sensible and acceptable.

Alas (imho :-)), FastAPI gave the "declare app first" a pretty loud comeback. So yeah, to support the diversity of users the best we can, introducing something like `--factory` was required. But to me the modular style of Starlette was the step _forward_ — removing the need for factories alltogether — while on the other hand the `@app.<...>` style from Flask, FastAPI and friends is basically the status quo. It causes a problem to which factories provide a solution — but we could very well do without it, as Starlette shows.

In the end, YMMV, and I'm happy if `--factory` makes Uvicorn easier to work with for users of FastAPI and other frameworks that result in requiring app factories.

I'm leaving a shamless link to one of my blog posts, [Structuring Starlette projects](https://florimond.dev/blog/articles/2020/05/structuring-starlette-projects/), which goes into further depth into how Starlette apps can typically be structured in a super-configurable, absolutely-testable way, without having to wrap everything around a factory function.

kristjanvalur:
Thank you for your thoughts.  This is useful information.  It's important, I think, to provide guidance and rationale for people like me who want to move over to a new and braver world. I've been struggling a bit with FastAPI, in trying to make my application more modular.

Now, being able to call an app factory fom Uvicorn is little more than syntactic sugar, because a wrapper module can always be created that instantiates a static app object.  The question is more about the style of the web app as a whole.

There is a separate issue that is often quoted about application factories, and that has to do with unittesting.  Basically to be able to create and tear down applications at will, with different configurations, from within the same testing script.  This ties again into modularity.  In regular python programs, import side effects are usually frowned upon.  The functionality of a module needs to be explicitly invoked.  The exception is usually the`if __name__ == "__main__` idiom, which also serves to indicate that one usually wants one's module to be able to function as a module in a larger application, _as well_ as a stand-alone application.

In the end, we obviously want to make things as simple and productive for people and not create mandatory steps that just get in the way,  Will be watching this with interest :)

PR #832, opened 2020-10-23
Cancel old keep-alive trigger before setting new one

### LINKED ISSUE #824: uvicorn closes client connection arbitrarily

cdeler:
This PR should fix #824. The second PR with tests https://github.com/encode/uvicorn/pull/831 should be merged after this PR (as you can see its build fails https://github.com/encode/uvicorn/pull/831/checks?check_run_id=1298976600)

I found that we do not cancel old keep-alive timeout, creating new one. So that sometimes we have the race-condition, where there are 2 keep-alive sweepers existing at the same time.

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -153,10 +153,13 @@ class H11Protocol(asyncio.Protocol):
     def eof_received(self):
         pass
 
-    def data_received(self, data):
+    def _unset_keepalive_if_required(self):
         if self.timeout_keep_alive_task is not None:
             self.timeout_keep_alive_task.cancel()
             self.timeout_keep_alive_task = None
+
+    def data_received(self, data):
+        self._unset_keepalive_if_required()
 
         self.conn.receive_data(data)
         self.handle_events()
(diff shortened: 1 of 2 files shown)


florimondmanca (commented):
Looks sensible, good catch!

florimondmanca (approved):

PR #835, opened 2020-10-23
Ensure non-zero exit code when startup fails

### LINKED ISSUE #780: uvicorn exit with success code on error in startup stage
### Checklist

### Describe the bug

In the stage of "startup", if some exception occurs, uvicorn just quit with exit code "0", instead of some non-zero code.

https://github.com/encode/uvicorn/blob/54d729ccc1638180a51e6ab600c4724e6424b048/uvicorn/main.py#L414-L418

With this exit code, my process manager (systemd, supervisord) cannot realize that my app exit abnormally, to restart my app.

### To reproduce

- Create a small FastAPI app with PostgreSQL database, like this example: https://fastapi.tiangolo.com/advanced/async-sql-databases/#connect-and-disconnect.

- Before running server, stop PostgreSQL, so that this code raises `ConnectionErrror`:

  ```py
  @app.on_event("startup")
  async def startup():
      await database.connect()
  ```
- Run the app with univorn.

### Expected behavior

uvicorn should quit with non-zero code.

### Actual behavior

uvicorn quit with zero code.

### Debugging material

<details>

```
$ ./runserver.py -vv
/usr/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: Linux supports fsync/fdsync with io_submit since 4.18 but current kernel 2.6.32
-754.el6.i686 doesn't support it. Related calls will have no effect.
  return f(*args, **kwds)
INFO:     Started server process [17004]
2020-09-16 17:13:25 |INFO api.main: To connect to postgresql:///vl-agent
  File "/home/quan/venv/lib/python3.6/site-packages/starlette/routing.py", line 526, in lifespan
  File "/home/quan/venv/lib/python3.6/site-packages/starlette/routing.py", line 467, in default_lifespan
    await self.startup()
  File "/home/quan/venv/lib/python3.6/site-packages/starlette/routing.py", line 502, in startup
    await handler()
  File "/home/quan/Vl-Agent/api/main.py", line 21, in startup
    await adb.connect()
  File "/home/quan/venv/lib/python3.6/site-packages/databases/core.py", line 84, in connect
    await self._backend.connect()                                                                                                       [0/175]
  File "/home/quan/venv/lib/python3.6/site-packages/databases/backends/postgres.py", line 65, in connect
    self._pool = await asyncpg.create_pool(str(self._database_url), **kwargs)
  File "/home/quan/venv/lib/python3.6/site-packages/asyncpg/pool.py", line 398, in _async__init__
    await self._initialize()
  File "/home/quan/venv/lib/python3.6/site-packages/asyncpg/pool.py", line 426, in _initialize
    await first_ch.connect()
  File "/home/quan/venv/lib/python3.6/site-packages/asyncpg/pool.py", line 125, in connect
    self._con = await self._pool._get_new_connection()
  File "/home/quan/venv/lib/python3.6/site-packages/asyncpg/pool.py", line 472, in _get_new_connection
    **self._connect_kwargs)
  File "/home/quan/venv/lib/python3.6/site-packages/asyncpg/connection.py", line 1727, in connect
    max_cacheable_statement_size=max_cacheable_statement_size)
  File "/home/quan/venv/lib/python3.6/site-packages/asyncpg/connect_utils.py", line 674, in _connect
    raise last_error
  File "/home/quan/venv/lib/python3.6/site-packages/asyncpg/connect_utils.py", line 666, in _connect
    connection_class=connection_class)
  File "/home/quan/venv/lib/python3.6/site-packages/asyncpg/connect_utils.py", line 633, in _connect_addr
    connector, timeout=timeout)
  File "/usr/lib/python3.6/asyncio/tasks.py", line 358, in wait_for
    return fut.result()
  File "uvloop/loop.pyx", line 1974, in create_connection
  File "uvloop/loop.pyx", line 1951, in uvloop.loop.Loop.create_connection
ConnectionRefusedError: [Errno 111] Connection refused

(venv) quan@virt-centos:~/Vl-Agent$ echo $?                                                                                                    
0
```

</details>

### Environment

mboutet:
Fixes #780 

I think this is a pretty straightforward and minimal fix. I tested it in my app and it works.

I'm using `uvicorn` like this:
```
uvicorn \
  --host "0.0.0.0" \
  --port "${SERVER_PORT}" \
  --access-log \
  --log-level debug \
  --limit-concurrency "${UVICORN_LIMIT_CONCURRENCY:-30}" \
  --backlog "${UVICORN_BACKLOG:-2048}" \
  myapp.main:app
```
In other words, I don't use a supervisor or something like that, so I'm not sure if those use cases are covered.

Let me know what you think.

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -388,6 +388,11 @@ def run(app, **kwargs):
         supervisor.run()
     else:
         server.run()
+        # If the server exited and it is not started,
+        # we can assume that something went wrong during startup.
+        # Thus we need to exit with non-zero exit code.
+        if not server.started:
+            sys.exit(1)
 
 
 class ServerState:


hongquan:
Could you return the exit code which follow some existing conventions? https://github.com/sixty-north/exit-codes/blob/master/exit_codes/exit_codes.py#L87

In this case, I think the code 78 (configuration error) is relevant.

mboutet:
Personally, I don't really mind using another non-zero code other than 1. However, it would be inconsistent with other parts of the codebase such as:
https://github.com/encode/uvicorn/blob/d78394e31310bb60defca35f0c6b1104820cf687/uvicorn/main.py#L373-L379
To be consistent, this exit code should also be 78 and I don't think that changing this would be a good idea since some users might rely on this behaviour. If you search `sys.exit(` in the codebase, we can see that exit code 1 is always used, mainly in cases where some kind of configuration error happened (i.e. before the server startup).

Let's wait for other opinions, and I'll change it if there's a consensus.

liketechnik:
I've personally had a similar, although different issue, where an exception in my startup handler did *not* cause uvicorn to return an exit code != 0. 

Therefore the current fix would not only catch configuration errors, but also application level errors. So either the handling of startup errors needs to be more fine grained or it is impossible to report the specific error cause via the exit code, imo.

hongquan (approved):

pquentin:
@florimondmanca @euri10 What are your thoughts on this? I got confused today again by this: I thought my app was shutdown externally when the issue was in fact an internal startup failure in my own code.

euri10:
yes this should be definitely be dealt with, I'm not sure the 1 exit is the most sensible way though,
and I dont have the time right now but it's on the todo !

florimondmanca:
Could someone interested in moving this forward share some before/after output for Uvicorn's behavior?

Edit: Actually I understand the issue is that the output shows that there was an issue on startup (and Uvicorn shuts down accordingly), but exit code is currently still zero in that case, right?

Is there anything we need to consider for the GunicornWorker? Does that handle startup/shutdown events too?

This might cause some breakage in case people falsely rely on "error on startup exits as OK", so let's just make sure to release this in a minor bump.

pquentin:
> Edit: Actually I understand the issue is that the output shows that there was an issue on startup (and Uvicorn shuts down accordingly), but exit code is currently still zero in that case, right?

Right!

I don't know about GunicornWorker, sorry.

mboutet:
@florimondmanca, I'm using the `uvicorn.workers.UvicornWorker` with gunicorn and the exit code is nonzero when there's a failure during startup if that's what you're asking.

euri10:
Long time I haven't thought about his but what would be the difference with a `raise click.Abort()` here ? I know we're using `sys.exit(1)` everywhere else instead of this pattern in the codebase, I just wondered what difference this could make, if that could be an improvement over just sending exit ?

Also this covers the only case where uvicorn is ran without --relaod or --workers , this would need to be considered as well in fixing the issue.

And finally the gunicorn case, but I think you alreaduy answered that

All that to say also a test for this would be awesome.

Kludex:
Gunicorn exits with code 3 if boot (startup) fails, and we accepted #1077.
I believe, for practicality, `uvicorn` should also exit with code 3.

Would you guys agree? @florimondmanca @euri10

Kludex:
I'd like to speed up this, and I don't agree with the exit code. Let's continue on #1278.

PR #838, opened 2020-10-29
Fix protocol import in test_config

euri10:
the test suite runs fine if you run all tests, but if you decide to just run test_config then you'll have this error

```
FAILED                          [ 27%]
tests/test_config.py:73 (test_concrete_http_class)
def test_concrete_http_class():
>       config = Config(app=asgi_app, http=protocols.http.h11_impl.H11Protocol)
E       AttributeError: module 'uvicorn.protocols.http' has no attribute 'h11_impl'

test_config.py:75: AttributeError
```

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -5,11 +5,11 @@ from copy import deepcopy
 import pytest
 import yaml
 
-from uvicorn import protocols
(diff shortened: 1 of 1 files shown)


cdeler (approved):
Fine! Approved

PR #839, opened 2020-11-01
Documentation: HTTPS settings add missing --ssl-keyfile-password

remster85:
Add-on to PR https://github.com/encode/uvicorn/pull/808.
The documentation is incomplete on settings.md, check https://www.uvicorn.org/settings.

image


docs/settings.md
```diff
@@ -63,6 +63,7 @@ connecting IPs in the `forwarded-allow-ips` configuration.
 ## HTTPS
 
 * `--ssl-keyfile <path>` - SSL key file
+* `--ssl-keyfile-password <str>` - SSL key file password
```
euri10: Good catch, could we just add that this is in case the key is encrypted ?
Something like this?
```suggestion
* `--ssl-keyfile-password <str>` - SSL key file password for encrypted keys
```

euri10 (approved):

docs/settings.md
remster85: hi @euri10, what do you think of  "Password to decrypt ssl key" ?

euri10: lgtm 😎

PR #842, opened 2020-11-08
Isolate server code into an asyncio-specific module

florimondmanca:
First step of the plan for expanding async library support, as laid out here… https://github.com/encode/uvicorn/issues/169#issuecomment-723556384

This PR takes the `Server` and `ServerState` classes defined in `uvicorn.main`, and moves them to `uvicorn._impl.asyncio`. Then we import those and define aliases in `uvicorn.main` to ensure 100% backwards compatibility.

(The plan mentioned `_impl/asyncio/server.py`, but this is in `_impl/asyncio.py`, since we don't need an `asyncio/` folder for now, given that there'd be a single script in there.)

### DIFF
--- a/uvicorn/_impl/asyncio.py
+++ b/uvicorn/_impl/asyncio.py
@@ -1,0 +1,256 @@
+import asyncio
+import functools
+import logging
+import os
+import platform
+import signal
+import socket
+import sys
+import time
+from email.utils import formatdate
+
+import click
+
+HANDLED_SIGNALS = (
+    signal.SIGINT,  # Unix signal 2. Sent by Ctrl+C.
+    signal.SIGTERM,  # Unix signal 15. Sent by `kill <pid>`.
+)
+
+logger = logging.getLogger("uvicorn.error")
+
+
+class AsyncioServerState:
+    """
+    Shared servers state that is available between all protocol instances.
+    """
+
+    def __init__(self):
+        self.total_requests = 0
+        self.connections = set()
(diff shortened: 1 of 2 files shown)


JayH5 (approved):
Seems reasonable (just moving code) and the plan sounds good. Keeping an eye on this work for Starlette maybe one day 👀

florimondmanca:
@JayH5 Thanks :-)

I've downgraded this to draft for now, and will close #843, #844 and #845 for now — it's good that we have those as reference now, but I'm thinking of a different approach.

The code in `protocols/` and `loops/` is only relevant to asyncio, and the rest is implemented in the `Server` class. So I'm less sure now that we need to be moving stuff _right now_, since adding `trio` support could pretty much mean add some `_impl/trio.py` module with whatever needs to be added, and then hook it up in `main.py` to switch between `Server` and `_impl.trio.TrioServer`.

In other words, _maybe_ the present PR is nice to have, but we don't _need_ it right now either. Eventually we'll have to put an if statement such as:

```python
if library == "asyncio":
    return Server(...)
elif library == "trio"
    return TrioServer(...)
```

But that time is not now _yet_.

(We _will_ still need the "make lifespan/WSGI asyncio-agnostic" part more early, though.)

florimondmanca:
Actually, sure, I think we can pull this in, so that eventually all the server implementations live in `_impl/<library>`. Thanks for the review @JayH5!

PR #843, opened 2020-11-08
Move protocols and loops to asyncio impl directory

florimondmanca:
Builds on #842 

Second step of https://github.com/encode/uvicorn/issues/169#issuecomment-723556384: move `uvicorn/{protocols,loops}/` to `uvicorn/_impl/asyncio/{protocols,loops}/`, keeping shims for backwards compatibility.

(I tried to `git mv` my way through things, but looks like the GitHub web UI doesn't recognize those. Trust me: I didn't change the content of the files; only edited the existing ones to turn them into shims.)

PR #844, opened 2020-11-08
Make lifespan implementation agnostic to async library

florimondmanca:
Builds on #843 

Third step of https://github.com/encode/uvicorn/issues/169#issuecomment-723556384: update `LifespanOn` implementation so that it's using an asyncio-agnostic API. This is done by introducing an `AsyncBackend` interface (much like we're doing in HTTPCore).

Note: Not using `anyio` here (yet) so that we have full control over the interface and we can move forward more easily. (But as in HTTPCore, we can consider switching later.)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,7 @@ env_marker_below_38 = "python_version < '3.8'"
 minimal_requirements = [
     "click==7.*",
     "h11>=0.8",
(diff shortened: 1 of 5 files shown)


florimondmanca:
Closing for now, see https://github.com/encode/uvicorn/pull/842#issuecomment-723660411

PR #845, opened 2020-11-08
Make WSGI implementation agnostic to async library

florimondmanca:
Builds on #844 

Last part of stage 1 of https://github.com/encode/uvicorn/issues/169#issuecomment-723556384: rework `WSGIMiddleware` so that it calls into `AsyncBackend` rather than `asyncio` directly.

This is the gnarliest of refactor PRs, since this component does some close concurrency. I know the `backend.call_soon()` method won't stay here when `trio` is added (trio doesn't have such a "schedule a callback from outside the event loop" API), and we'll need to refactor this, but for now the goal is to have some abstraction without changing too much the implementation logic. Tests pass locally, so I'm happy (but let's see what CI says).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,8 @@ minimal_requirements = [
     "h11>=0.8",
     "sniffio==1.*",
     "typing-extensions;" + env_marker_below_38,
+    "async_generator; python_version < 3.7",
+    "async_exit_stack; python_version < 3.7",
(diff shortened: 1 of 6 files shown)


florimondmanca:
Closing for now, see https://github.com/encode/uvicorn/pull/842#issuecomment-723660411

PR #851, opened 2020-11-17
make sure tests don't fail if a running uvicorn app is present

justindujardin:
Twice I've tried running the uvicorn tests only to have them freeze half-way through execution because I have a Uvicorn app running on my system that is already bound to port 8000 😅 

This PR updates the test suite to use a non-standard port so that the tests pass even if there's already an app running on port 8000.

### DIFF
--- a/tests/middleware/test_trace_logging.py
+++ b/tests/middleware/test_trace_logging.py
@@ -72,13 +72,14 @@ def test_trace_logging(capsys):
         limit_max_requests=1,
         log_config=test_logging_config,
         log_level="trace",
+        port=8917,
     )
     server = CustomServer(config=config)
     thread = threading.Thread(target=server.run)
     thread.start()
     while not server.started:
         time.sleep(0.01)
-    response = requests.get("http://127.0.0.1:8000")
+    response = requests.get("http://127.0.0.1:8917")
     assert response.status_code == 204
     thread.join()
     captured = capsys.readouterr()
@@ -111,13 +112,14 @@ def test_access_logging(capsys, http_protocol):
         http=http_protocol,
         limit_max_requests=1,
         log_config=test_logging_config,
+        port=8917,
     )
     server = CustomServer(config=config)
     thread = threading.Thread(target=server.run)
     thread.start()
     while not server.started:
         time.sleep(0.01)
-    response = requests.get("http://127.0.0.1:8000")
+    response = requests.get("http://127.0.0.1:8917")
     assert response.status_code == 204
     thread.join()
     captured = capsys.readouterr()
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -78,7 +78,7 @@ def test_concrete_http_class():
 
 
 def test_socket_bind():
-    config = Config(app=asgi_app)
+    config = Config(app=asgi_app, port=8917)
     config.load()
 
     assert isinstance(config.bind_socket(), socket.socket)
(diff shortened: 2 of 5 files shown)


tiangolo (approved):
LGTM! :rocket:

I'll wait a bit to make sure nobody opposes it before merging. :nerd_face:

euri10:
not opposed to it because I was in this situation once or twice, this said I'd be more in favor of leaving the default port as is and add the ability to run tests on a non-default one with an env variable.

justindujardin:
>not opposed to it because I was in this situation once or twice, this said I'd be more in favor of leaving the default port as is and add the ability to run tests on a non-default one with an env variable.

Thanks for the review! Would adding an env var help in any way? If we leave the default port, I'll still run into test failures and have to *change something* to get them to pass each time I have another app running. If that's the case, I don't think these changes do anything but add more complexity, and maybe we can just not merge them?

euri10:
In over a year on and off here, I never felt this was a problem, I think it's easy enough to just kill the 8000 port that is listening otherwise when developing for uvicorn, but maybe there's a part of the picture I dont get :drooling_face: 

I'm not opposed to it, I just don't see the value: I would have closed it for that reason, it solves no real issue, 1 test will quickly fail on `bind_socket` and you immediately know what's wrong, there are tons of way more interesting stuff to solve in here if you're interested :)

justindujardin:
Thanks for the quick response, I think I understand your point. A few comments:

> it solves no real issue

It solves the issue you say you've encountered as well: "I was in this situation once or twice."

> 1 test will quickly fail on bind_socket

One test does fail quickly, but if you run `sh scripts/test`, the tests keep going and hang for more than a minute unless you manually kill the test run. After that, you can view the errors to find the one that says "Address already in use." 

> you immediately know what's wrong

I think you mischaracterize how immediately or easily you know what's wrong because of your expertise with this project. Again you admit that something is wrong, so doesn't this change have value if it stops that thing from going wrong?

>I think it's easy enough to just kill the 8000 port that is listening otherwise when developing for uvicorn

I'm not a dedicated developer for this project, so if I'm working on a fix/test/change for it, I've likely got other projects going simultaneously. Telling me to stop my other projects to contribute my free time to yours because you don't like specifying a different port in tests isn't a great look. 😓 

> there are tons of way more interesting stuff to solve in here if you're interested :)

Yeah, but I'm unlikely to try larger changes if I can't get a small change accepted to make it easier for new users to contribute. 😅

euri10:
don't be offended when I say no real issue, I should have said no real issue for me if that suits you best :)

if you had taken the time to ask in gitter as it's advertised (maybe not enough) people would have immediately told you the reason for that hang, it's the same on all encode projects afaik, httpx, starlette, uvicorn, that's common knowledge shared, now telling you this I realize what should be done is probably to expand / create docs on contributing and pinpoint that issue.

>  Telling me to stop my other projects to contribute my free time to yours because you don't like specifying a different port in tests isn't a great look. sweat

nobody's asking you to stop anything, I'm also on my free time here and sometimes have the 8000 port taken, it's just simple to kill it, as a matter of fact over all the encode devs you're the first for which it seems to be what I badly call "a real issue", so be it

justindujardin:
You're here in your free time objecting to a tiny change that makes life better for developers trying to contribute tests for the first time because you want to bind to port 8000 in unit tests. 🤷

PR #854, opened 2020-11-18
Modernize test ASGI applications

florimondmanca:
Small refactoring PR that cleans up a bunch of test code, replacing the older ASGI 2.0 style:

```python
class App:
    def __init__(self, scope):
        ...

        ...
```

With an up-to-date ASGI 3.0 style:

```python
    ...
```

Also reduce code duplication in tests that were just using a sample "empty response" app.

### DIFF
--- a/tests/middleware/test_trace_logging.py
+++ b/tests/middleware/test_trace_logging.py
@@ -48,26 +48,23 @@ test_logging_config = {
 }
 
 
+async def app(scope, receive, send):
+    assert scope["type"] == "http"
+    await send({"type": "http.response.start", "status": 204, "headers": []})
+    await send({"type": "http.response.body", "body": b"", "more_body": False})
+
+
 @pytest.mark.skipif(
     sys.platform.startswith("win") or platform.python_implementation() == "PyPy",
     reason="Skipping test on Windows and PyPy",
 )
 def test_trace_logging(capsys):
-    class App:
-        def __init__(self, scope):
-            if scope["type"] != "http":
-                raise Exception()
-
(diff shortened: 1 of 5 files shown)


tests/middleware/test_trace_logging.py
```diff
@@ -48,26 +48,23 @@
...
-    class App:
-        def __init__(self, scope):
-            if scope["type"] != "http":
-                raise Exception()
-
-        async def __call__(self, receive, send):
-            await send({"type": "http.response.start", "status": 204, "headers": []})
-            await send({"type": "http.response.body", "body": b"", "more_body": False})
-
     class CustomServer(Server):
```
euri10, line 62: I found myself having to change the CustomServer in 4 or 5 files while working on a draft today, maybe a good refactor for another pr to keep things separated or here have no preference, what do you think?

euri10 (approved):
Much love 💗 for net deletion, I left a small suggestion but this can go without !

tests/middleware/test_trace_logging.py
florimondmanca: Not sure what you mean here exactly — the `CustomServer` is left untouched? Only the ASGI apps are modified and factored out. Did you mean that the individual ASGI apps were modified as well?

euri10:
Nah just wanted to say that in this PR or another the CustomServer could be
refactored as well, it's currently duplicated all over

florimondmanca:
> in this PR or another the CustomServer could be refactored as well, it's currently duplicated all over

Yup ;-)

PR #859, opened 2020-11-20
Remove scope from access log when using json logging formatter

### LINKED ISSUE #858: Scope sent in access logs by default
### Checklist

### Describe the bug

prompted by @alicederyn in [gitter](https://gitter.im/encode/community?at=5fb6492d2a60f731f764a554) you can leak headers from the scope in the access logs given that you use json formatting for your logging.

### To reproduce

logging.yaml

```yaml
version: 1
formatters:
  json_formatter:
    format: '%(levelname)s'
    "()": "pythonjsonlogger.jsonlogger.JsonFormatter"
handlers:
  console:
    class: logging.StreamHandler
    level: DEBUG
    formatter: json_formatter
    stream: ext://sys.stdout
loggers:
  uvicorn:
    level: DEBUG
    handlers: [console]
    propagate: no
root:
  level: DEBUG
  handlers: [console]
```

cookieapp.py

```python
from starlette.requests import Request
from starlette.responses import Response


import uvicorn


    request = Request(scope, receive)
    mycookie = request.cookies.get("mycookie")
    if mycookie:
        response = Response(mycookie, media_type="text/plain")
    else:
        response = Response("Hello, world!", media_type="text/plain")
        response.set_cookie("mycookie", "Hello, cookies!")

if __name__ == '__main__':
    uvicorn.run("cookieapp:app", log_level="debug", log_config="logging.yaml")
```

returns :

```json
{"levelname": "INFO", "color_message": "Uvicorn running on \u001b[1m%s://%s:%d\u001b[0m (Press CTRL+C to quit)"}
{"levelname": "INFO", "status_code": 200, "scope": {"type": "http", "asgi": {"version": "3.0", "spec_version": "2.1"}, "http_version": "1.1", "server": ["127.0.0.1", 8000], "client": ["127.0.0.1", 35808], "scheme": "http", "method": "GET", "root_path": "", "path": "/", "raw_path": "b'/'", "query_string": "b''", "headers": [["b'host'", "b'127.0.0.1:8000'"], ["b'connection'", "b'keep-alive'"], ["b'cache-control'", "b'max-age=0'"], ["b'dnt'", "b'1'"], ["b'upgrade-insecure-requests'", "b'1'"], ["b'user-agent'", "b'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36'"], ["b'accept'", "b'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9'"], ["b'sec-fetch-site'", "b'none'"], ["b'sec-fetch-mode'", "b'navigate'"], ["b'sec-fetch-user'", "b'?1'"], ["b'sec-fetch-dest'", "b'document'"], ["b'accept-encoding'", "b'gzip, deflate, br'"], ["b'accept-language'", "b'en-US,en;q=0.9,fr;q=0.8'"], ["b'cookie'", "b'mycookie=\"Hello\\\\054 cookies!\"'"]]}}
```

### Expected behavior

we should hide the headers like we do in by default _[[?? and provide a mechanism (flag ??) so that people who truly want the scope in logs can have it ]]_

https://github.com/encode/uvicorn/blob/79a72e3dc97db4803ea4e5c92c936fd45f885b38/uvicorn/middleware/message_logger.py#L7

### Environment

euri10:
Fixes https://github.com/encode/uvicorn/issues/858 as a 1st step that ensures no sensitive headers are leaked through scope in access logs in case people are using a json formatter

Will work on an implementation that let you access those headers / scope later when there is consensus / discussion about how it should be done.

Gunicorn let you print it with a [custom format](https://docs.gunicorn.org/en/stable/settings.html#access-log-format)

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -460,7 +460,6 @@ class RequestResponseCycle:
                     get_path_with_query_string(self.scope),
                     self.scope["http_version"],
                     status_code,
-                    extra={"status_code": status_code, "scope": self.scope},
                 )
 
             # Write response status line and headers
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -462,7 +462,6 @@ class RequestResponseCycle:
                     get_path_with_query_string(self.scope),
                     self.scope["http_version"],
                     status_code,
-                    extra={"status_code": status_code, "scope": self.scope},
                 )
 
             # Write response status line and headers


florimondmanca (approved):
Sounds good. Is there any chance this could be breaking to some users? Any case, this 1st step seems like a clear yes; we can reassess a proper implementation afterwards.

euri10:
> Sounds good. Is there any chance this could be breaking to some users? Any case, this 1st step seems like a clear yes; we can reassess a proper implementation afterwards.

I think it can. It requires quite a specific setup though so chances are slim, but you never know, people get fancy when they need something, at least we'll know :)

Screenshot_20201112-211749

alippai:
I used to put the username from request / response headers into the access log using a custom formatter. How should I do that in the new version?

euri10:
> I used to put the username from request / response headers into the access log using a custom formatter. How should I do that in the new version?

hi @alippai sorry to hear I broke your use case !

If you could open a feature request with what you previously we can discuss and see if we can provide a solution in uvicorn directly

in the meantime, I'd take a look at opentelemetry, I personally uses it to trace my apps and adding the username is rather simple. There are probably other solutions, but that's the 1st that comes to mind.

alippai:
@euri10 I'm always happy to move forward, opentelemetry sounds promising, there is no need to maintain legacy on a free, open-source project. That kind of support just burns out the developers and kills the project eventually.
Perhaps you have a tip for a few questions:
- Should I use this middleware for the integration? https://github.com/open-telemetry/opentelemetry-python-contrib/tree/master/instrumentation/opentelemetry-instrumentation-asgi
- Do you know good sinks or "frontends" for the opentelemetry collector which is appropriate for HTTP logs?

euri10:
> @euri10 I'm always happy to move forward, opentelemetry sounds promising, there is no need to maintain legacy on a free, open-source project. That kind of support just burns out the developers and kills the project eventually.
> Perhaps you have a tip for a few questions:
> 
> * Should I use this middleware for the integration? https://github.com/open-telemetry/opentelemetry-python-contrib/tree/master/instrumentation/opentelemetry-instrumentation-asgi
> * Do you know good sinks or "frontends" for the opentelemetry collector which is appropriate for HTTP logs?

Yes ! Use an instrumentation : it adds a middleware that produces "automatic" traces, the asgi one is probably what you want, I know there is one for Starlette and FastAPI too, I'm also using the asyncpg one for the db...

then should you need more granular / precise information you can always add things yourself to the traces.

For instance I'm doing on some endpoints: `trace.get_current_span().set_attribute("enduser.id", current_user.email)`

as for the collector I'm using Cloud Trace which works fine so far, but I must say I have a very superficial view of this part so I won't be your best advisor :)

It's a brand new world so pretty scary at first but the guys at opentelemetry-python on gitter are super reactive and helpful so I'd try and see if I were you :)

gregbeech:
This change has broken our custom loggers which were using the scope to send the relevant information to Datadog using its standard logging attributes. I can't see any way to access half the information now such as user agent, etc. in logs. We get some of the information in traces but those are sampled so it's not the full record of things happening. Is there going to be some way to access the request data like headers (for User-Agent) and so on brought back?

euri10:
> This change has broken our custom loggers which were using the scope to send the relevant information to Datadog using its standard logging attributes. I can't see any way to access half the information now such as user agent, etc. in logs. We get some of the information in traces but those are sampled so it's not the full record of things happening. Is there going to be some way to access the request data like headers (for User-Agent) and so on brought back?

sorry for that @gregbeech 
currently the only way I see to access the headers would be in a middleware at the app level, so that doesn't fit your reqs that it's found in the logs.
maybe @florimondmanca has some ideas !
there's currently no work on adding that back, but indeed that's something worth considering.

florimondmanca:
@gregbeech Hi,

Yeah, same as @euri10 I would say a custom middleware to dump any info you'd like to Datadog sounds more sensible. Relying on extras passed by the logger sounds... Well, relatively okay tbh, just too constrained when à middleware you have full control over should do just fine. Hope that helps.

Edit: hmm, looks like I misread your comment. You were sending logs from Uvicorn to DD and attaching attributes from extras, I see. Can you do things differently, perhaps using a parser if that info is present in the log lines? You might also be able to wrap the app to log things manually with whichever info from the ASGI `scope` you're interested in. I don't think we'll be adding this back in core given the security discussions above.

tonkolviktor:
Hi @florimondmanca, @euri10,

thanks for the discussion so far. I'd like to keep the discussion open.

I would reather easily agree that having the option to dump the whole scope object to the log is not ideal.
However I still see the value being able to customize log messages / the alternatives are also far from ideal.

My use-case is that, I get in a token via HTTP Header. I check/resolve it to a token name, now I want to add it to the access log (eg via logging a response header value). This way I would have a proper access log.

1. Hacking with logging
a)
One option is to disable the "official" access log and implement my from 0. (I cannot have two access logs)
In this case I have to copy/reimplement all the code/tought that went into how to do this properly (catching exception, formatting messages, etc)
b)
I could just add my extra information in a separate log line, however the correlation becomes difficult: matching process id, timestamp, etc.

2. Starlette (FastAPI)
I have a problem that a custom Starlette middleware cannot be the most outter one:
https://github.com/encode/starlette/blob/master/starlette/applications.py#L68
This means if I have my own custom access logging, but I change the response code in a 'Exception' handler I'll not see that change in my logging.

3. Using a telemetry/prometheus could be an option, but seems to be a bit of an overkill.
a) still fails on point 2.
b) if we now talk about bigger companies than a small startup, one has some requirements to preserve system acces information. This is typically done via logfiles.

It would make me happy if you'd consider the above points and think of a mutually acceptable way of customizing the logs when required.

Viktor

insurandoag:
@euri10 this change also broke our complete custom logging. We need to log Request-Header information in fastapi, so we wrote a custom logger to access the "scope" key to get to the header infos. Now we must revert back to 0.13.0 unfortunately and have no quick solution. Any ideas?

KlausGlueckert:
@euri10 : this change also broke our complete custom logging. We need to log Request-Header information in fastapi, so we wrote a custom logger to access the "scope" key to get to the header infos. Now we must revert back to 0.13.0 unfortunately and have no solution.

Here is how we did it: https://stackoverflow.com/questions/62894952/fastapi-gunicon-uvicorn-access-log-format-customization

Ideally, these infos should somehow be accessible programmatically when defining a custom logger to be used from FastAPI

euri10:
#947 is probably something worth looking at if I understand your issue correctly

KlausGlueckert:
@euri10  i checked quickly, not sure if it makes it possible to access header infos, it just talks about tracking request/response times. 

This is also an important needed feature. My hack for this was to write this in the header as an attribute from within the fast-api async function of the endpoint. In that way I can measure the "server-time", and not the response time which is biased by a lot of other things (location, bandwith etc). Then I can access als this info in the "scope" and print it to console. We use Splunk for parsing the logs, so I just need to print my logs to the console, done.

Therefore, I also agree with  @Kludex that this customization is probably best made from within uvicorn

Kludex:
In theory, that PR will allow writing header info in the access format as well.

euri10:
> In theory, that PR will allow writing header info in the access format as well.

this, a solid alternative being a middleware.

PR #863, opened 2020-11-23
[Living POC] Async-agnostic implementation

florimondmanca:
Refs #169 and my comments there.

I'm opening a PR so anyone can view/appreciate the diff more easily (including myself 😄), but I'm not expecting we consider this for merging, since it's basically unreviewable. :-) We _could_ pull in bits and pieces incrementally, eg start with a minimal HTTP/1.1 + `h11` + `trio` package… But I'm still figuring things out.


uvicorn/_async_agnostic/server.py
```diff
@@ -0,0 +1,268 @@
...
+            self._closed_event_obj = self._backend.create_event()
+        return self._closed_event_obj
+
+    def run(self, sockets: List[socket.SocketType] = None) -> None:
+        if self._config.async_library == "asyncio":
+            self._config.setup_event_loop()
+        self._backend.run(self._main, sockets)
+
+    @contextlib.contextmanager
+    def run_in_thread(self, sockets: List[socket.SocketType] = None) -> Iterator[None]:
```
florimondmanca, line 59: This general-purpose helper function is something people end up having to re-implement anytime they want to use Uvicorn in a tests setup. (We do it ourselves, we do it in HTTPX, in HTTPCore, I do it in my projects, etc.) We could consider adding this to the existing `Server` implementation, if we'd like.

uvicorn/_async_agnostic/server.py
```diff
@@ -0,0 +1,268 @@
...
+        if self._shutdown_event.is_set():
+            return True
+        if config.limit_max_requests is not None:
+            return state.total_requests >= config.limit_max_requests
+        return False
+
+    async def _listen_signals(self) -> None:
+        if threading.current_thread() is not threading.main_thread():
+            await self._closed_event.wait()
+            return
```
florimondmanca, line 188: The existing `Server` doesn't take this into account and blindly tries to listen on signals in all cases, which means any people that want to run Uvicorn in a thread need to deal with this problem. Hence overriding `.install_signal_handlers()` to a "do-nothing" implementation in a tests setup. We could consider an enhancement to the existing `Server` to do a check like this by default in `.install_signal_handlers()`.

uvicorn/_async_agnostic/http11/keepalive.py
```diff
@@ -0,0 +1,52 @@
...
+from typing import Optional
+
+from uvicorn.config import Config
+
+from ..backends.auto import AutoBackend
+from ..backends.base import TaskHandle, TaskStatus
+from .connection import HTTP11Connection
+
+
+class KeepAlive:
```
florimondmanca, line 11: I find it nice that this implementation uses a separate `KeepAlive` component that's somewhat testable in isolation. In `master` the keep-alive logic is entagled with the main server code, and there's low-level logic around managing asyncio tasks. This kind of isolated component makes it a bit easier to understand what's going on, I believe.

uvicorn/_async_agnostic/http11/handler.py
```diff
@@ -0,0 +1,155 @@
...
+            await keepalive.schedule()
+            conn.debug("Connection kept alive")
+
+    await keepalive.aclose()
+    await conn.shutdown_and_clean_up()
+    state.connections.discard(conn)
+    conn.debug("Connection closed")
+
+
+async def send_h11_response(
```
florimondmanca, line 97: This function _happens_ to call into an ASGI app, but interestingly at this point we could call into _anything else_ — which makes me think this work might end up being very useful for a server-side version of `httpcore`.

There's a strong similarity between:

```python
# (HTTPCore)
async def arequest(method, url, headers, stream, ext) -> ...:
    ...
```

And:

```python
# (This PR, roughly)
async def handle(conn, method, path, headers, stream) -> None:
    ...
```

uvicorn/_async_agnostic/backends/base.py
```diff
@@ -0,0 +1,160 @@
...
+        pass
+
+    async def get_value(self) -> Any:
+        return None
+
+
+TaskStatus.IGNORED = IgnoredTaskStatus()
+
+
+class AsyncBackend:
```
florimondmanca, line 103: This "backends" directory follows the same approach than we did in HTTPCore. It doesn't use `anyio` directly — although that would be a valid option — so that I have full control over the APIs I need for building this out. I'm still wary of requiring `anyio` to be added as an extra dependency, rather than us just writing down some glue code suited to our exact needs.

florimondmanca (commented):
Highlighting some things I think we could either consider improving _anyway_, or related thoughts and ideas I've had while working on this.

florimondmanca:
Aaaand the test suite fails on 3.6, as expected, since I'm using `AsyncExitStack` and `asynccontextmanager` in places but those are 3.7+ only. Heh, I'll add a backport so we can see how tests do.

florimondmanca:
@euri10 I'm getting a strange / unrelated error on Windows in this branch:

```
 E           RuntimeError: There is no current event loop in thread 'MainThread'.
```

in the original tests for `H11Protocol` and the like. I say it's strange because I haven't touched that test code or any of the other code… Does this ring a bell?

Edit: ah, that's probably related to installing `anyio` for tests… It's got its own pytest plugin shipped and activated by default, and that might clash somehow.

Edit: yup: by default the `anyio_backend` fixture is function-scoped, and on async test cases it invokes a coroutine that closes and unsets the event loop on each test: https://github.com/agronholm/anyio/blob/b5dffbe7ccf28afd97d43113ae9e98fca61c6627/src/anyio/_backends/_asyncio.py#L1316-L1322 For some reason `asyncio` on Windows doesn't seem to setup a new event loop when it's been set to `None`.

florimondmanca:
Heh, now tests hang on < 3.9, weird.

florimondmanca:
I'll close this since this is a great start but obviously this is unmergeable — the goal was just to get a better feeling or how we might be going about this. :-)

PR #866, opened 2020-11-24
Move server implementation to server.py

florimondmanca:
Follow-up to #842 

#842 moved the `Server` code to an `_impl/asyncio.py` module anticipating support for multiple async environments, but this was obviously premature. I've now learnt through #863 that this just won't be as easy as "one server implementation per async library" (well it could be, but that would be at a huge code duplication cost).

This PR reverts back to a more stable and simple state: a `server.py` module. The sole purpose is that we have this isolated in a module, and not cluggering up the `main.py` module — `main.py` should be focused on the CLI.

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -7,7 +7,6 @@ import typing
 import click
 
 import uvicorn
-from uvicorn.server import Server, ServerState  # noqa: F401  # Used to be defined here.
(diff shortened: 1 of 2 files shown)


JayH5 (approved):
Seems reasonable 😄 hope you get your Trio support eventually 🤞

Issue #526, opened 2019-12-13
[FEATURE] Make event-handling optional


sm-Fifteen:
I'm trying to get Uvicorn to run as a Windows service so event logging and automatic start/restart can be handled more cleanly than by just running `uvicorn` as a scheduled task. I've run into a number of roadblocks, though most of these I've been able to work my way around with some configuration. The main roadbloack I still have is event-loop compatibility. Due to some odd behavior of Python threads while running as a windows service (see mhammond/pywin32#1452), it's not currently possible for Python code to register signal handlers. That makes the default `ProactorEventLoop` unusable, but that can be solved by falling back to a `SelectorEventLoop`, which used to be the default event loop for Windows before Python 3.8. Not being able to handle the signal for Ctrl+C is not too much of a problem for my use case either, since I can just set the service to trip the `Server#should_exit` flag manually when the service is asked to stop.

So with all this said and done, if you ignore the service boilerplate wrapping my code, I'm essentially running Uvicorn like this:

```py
import asyncio
import logging
from logging.handlers import NTEventLogHandler

from uvicorn import Server, Config
from app import app

asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())

windows_log_handler = NTEventLogHandler('myASGIApp')
logging.getLogger().addHandler(windows_log_handler)

server_config = Config(
    app, host="0.0.0.0", port=8000,
    # Can't log to stdout as a service
    log_config={"version": 1},
    # Don't let uvicorn select what loop to use
    loop="asyncio",
)
self.app_server = Server(server_config)
self.app_server.run()
```

The main problem I still have with this is this line, which will try to setup event handlers even though doing so will immediately fail given my event loop configuration:

https://github.com/encode/uvicorn/blob/996aa2d459360f61507caa142b89a8f5b562cb6d/uvicorn/main.py#L360

https://github.com/encode/uvicorn/blob/996aa2d459360f61507caa142b89a8f5b562cb6d/uvicorn/main.py#L518-L527

I currently have to go and comment out that line in the uvicorn module code to get my application to start at all, which is less than ideal. Would it be possible to make signal handling optional if uvicorn is running from inside a Python script, maybe?

**EDIT:** I would have tried `import uvicorn; uvicorn.main.HANDLED_SIGNALS = ()`, but it turns out that `uvicorn.main:main` is imported by `uvicorn` in a way that shadows the real `uvicorn.main` package, so even this awful hack is not an option.

[...]

PR #871, opened 2020-11-30
Skip installation of signal handlers when not in main thread

florimondmanca:
This one has been bogging me for a while, so here's a tentative pull request for it:

Since signals can only be listened to from the main thread, this PR modifies `.install_signal_handlers()` to skip installation of signal handlers when not in the main thread:

https://docs.python.org/3/library/signal.html#signals-and-threads

> […] **Only the main thread of the main interpreter is allowed to set a new signal handler**.

This way, we can also get rid of a bunch of boilerplate overrides of `.install_signal_handlers()` in tests — and many other people will be able to as well (eg we do it in HTTPX, HTTPCore, many of my personal packages, and essentially anywhere one needs to run Uvicorn in a Python thread right now).

Surprisingly enough I haven't found an issue open for this yet. :-) But this is somewhat related to #526 (although I don't think it's _exactly_ fixing it).

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -240,7 +240,7 @@ class Server:
             await self.lifespan.shutdown()
 
     def install_signal_handlers(self) -> None:
-        if threading.current_thread() is threading.main_thread():
+        if threading.current_thread() is not threading.main_thread():
             # Signals can only be listened to from the main thread.
             return
 
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -6,6 +6,7 @@ import platform
 import signal
 import socket
 import sys
+import threading
 import time
 from email.utils import formatdate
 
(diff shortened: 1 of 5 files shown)


euri10:
weird this doesnt pass tests, looks it blocked on the config one ?

uvicorn/server.py
```diff
@@ -238,7 +239,11 @@ async def shutdown(self, sockets=None):
         if not self.force_exit:
             await self.lifespan.shutdown()
 
-    def install_signal_handlers(self):
+    def install_signal_handlers(self) -> None:
+        if threading.current_thread() is threading.main_thread():
```
euri10:
```suggestion
        if threading.current_thread() is not threading.main_thread():
```

euri10 (commented):
Probably you wanted that ?

uvicorn/server.py
florimondmanca: 🤦 Obviously! I guess I didn't actually take the chance to run this locally either. 😆  Thanks!

euri10 (approved):

florimondmanca:
@euri10 I'll merge this and then tentatively release as 0.13.0… I'm not sure whether this should be considered a new feature or a bug fix, but since it can potentially mean some people will have to undo some scaffolding, I'd treat this as an "Added" entry.

PR #875, opened 2020-12-07
Add --factory flag to support factory-style application imports

florimondmanca:
Closes #492

This PR adds support for projects that don't expose an `app` instance directly, but rather an application factory.

As per discussions in #822, this style is most typically preferred by some users to avoid side-effects at import time, which may be important for larger-size projects, and — but that's more of a taste preference — global state.

Alternative implementation to #822 that also handles edge cases, and updates the `Settings` documentation.

PR #876, opened 2020-12-07
Add tool to keep CLI usage in `index.md` in sync

florimondmanca:
This PR adds a simple Python-based tool to dump the output of `uvicorn --help` into the docs home page. Also supports a `--check` mode for CI purposes.

**Motivation**
While working on #875 I realized the CLI usage there was out of sync with the current options, and I thought adding an automated validation tool was the best way to keep it in sync moving forward. :-)

I could expand this later on to also validate that all options are listed in the "Settings" page, but that's a different thing so I'll keep it for another PR.


tools/gen_cli_usage.py
```diff
@@ -0,0 +1,45 @@
...
+        content = f.read()
+
+    lines = content.splitlines()
+    marker_lineno = lines.index("<!-- gen_cli_usage:marker -->")
+    assert lines[marker_lineno + 1] == "```"
+    start = marker_lineno + 2
+    end = next(
+        lineno for lineno, line in enumerate(lines[start:], start) if line == "```"
+    )
+    lines = lines[:start] + get_usage_lines() + lines[end:]
```
florimondmanca: Honestly, these kinds of < 10 LOC scripts that do useful tasks yet remain pretty readable is why I love Python so much. ❤️

euri10:
lgtm, 
but why you dont use mkdocs-click you wrote ???

euri10 (approved):

florimondmanca:
> mkdocs-click

😳 Absolutely! Definitley worth considering, I'll look at what that looks like. https://github.com/DataDog/mkdocs-click Though right now my thoughts would be "a 40-LOC script vs a new dependency?" tends to lean towards the former…

Also a note to myself that we also have a listing of the CLI options in the [Deployment docs](https://www.uvicorn.org/deployment/) (for some reason).

florimondmanca:
@euri10 After considering it, I won't be using `mkdocs-click`. It adds an `h1` with the name of the Click command, which we don't need, and it's an extra dependency which means if anything changes there then we might need to tweak things here, too. A simple script inside the repo is much more resilient. :-)

I updated the script to be a bit more robust, feel free to take a new look!

euri10:
nothing to add @florimondmanca  ! let's merge and see it deployed :)

Issue #877, opened 2020-12-08
uvicorn.run with reload or workers in ProcessPoolExecutor fails with Bad file descriptor error


likeanaxon:
Greetings!

I have run into an issue when trying to spawn several local `uvicorn`s (obviously bound to different ports) **from within Python**. Instead of launching each one manually in a separate terminal, I have set up a function that spawns the desired servers in a `ProcessPoolExecutor` (imported from `concurrent.futures`). However, if the servers are configured with the `workers` or `reload` options, I get the following error:

```
    Process SpawnProcess-1:1:
    Traceback (most recent call last):
      File "/usr/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
        self.run()
      File "/usr/lib/python3.8/multiprocessing/process.py", line 108, in run
        self._target(*self._args, **self._kwargs)
      File "/home/cantor/code/envs/tabula/lib/python3.8/site-packages/uvicorn/subprocess.py", line 55, in subprocess_started
        sys.stdin = os.fdopen(stdin_fileno)
      File "/usr/lib/python3.8/os.py", line 1023, in fdopen
        return io.open(fd, *args, **kwargs)
    OSError: [Errno 9] Bad file descriptor

```

Stating the obvious, there is an issue with reopening `stdin` in the child process.

Simple workarounds resulting in happy server spawning: 

- Commenting out lines `54` and `55` in `subprocess.py`

```
    # if stdin_fileno is not None:
        # sys.stdin = os.fdopen(stdin_fileno)
```

- Setting `'stdin_fileno': None` on line `35` in `subprocess.py`

### To reproduce

```
pip install fastapi uvicorn
```

(Assuming the code is located in `launch.py`)

With `workers`:

```
    import uvicorn
    from fastapi import FastAPI
    from concurrent.futures import ProcessPoolExecutor

    app = FastAPI()

    if __name__ == '__main__':

        with ProcessPoolExecutor() as executor:

            executor.submit(uvicorn.run, 'launch:app', workers = 3)
```

Error:

```
    Process SpawnProcess-1:3:
    Traceback (most recent call last):
    File "/usr/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
        self.run()
    File "/usr/lib/python3.8/multiprocessing/process.py", line 108, in run
        self._target(*self._args, **self._kwargs)
    File "/usr/lib/python3.8/site-packages/uvicorn/subprocess.py", line 55, in subprocess_started
[...]

PR #883, opened 2020-12-10
Fix Bad file descriptor error when using reload or workers flag and multiprocessing

likeanaxon:
Fix for #877. 

Note: Reopening `sys.stdin` in the child process might not even be necessary, but at any rate it works if we use `open('/dev/stdin')`. A warning is produced if for some reason `/dev/stdin` cannot be opened.

### DIFF
--- a/uvicorn/subprocess.py
+++ b/uvicorn/subprocess.py
@@ -44,7 +44,7 @@ def subprocess_started(config, target, sockets):
     # Re-open stdin.
     # This is required for some debugging environments.
     try:
-        sys.stdin = open('/dev/stdin')
+        sys.stdin=open("/dev/stdin")
     except:
         print("WARNING: Failed to open '/dev/stdin', debugging might not work.")
 
--- a/uvicorn/subprocess.py
+++ b/uvicorn/subprocess.py
@@ -3,7 +3,6 @@ Some light wrappers around Python's multiprocessing, to deal with cleanly
 starting child processes.
 """
 import multiprocessing
-import os
 import sys
 
 multiprocessing.allow_connection_pickling()
@@ -45,7 +44,7 @@ def subprocess_started(config, target, sockets):
     # This is required for some debugging environments.
     try:
         sys.stdin = open("/dev/stdin")
-    except:
+    except OSError:
         print("WARNING: Failed to open '/dev/stdin', debugging might not work.")
 
     # Logging needs to be setup again for each child.
--- a/uvicorn/subprocess.py
+++ b/uvicorn/subprocess.py
@@ -3,6 +3,7 @@ Some light wrappers around Python's multiprocessing, to deal with cleanly
 starting child processes.
 """
 import multiprocessing
+import os
 import sys
 
 multiprocessing.allow_connection_pickling()
@@ -43,9 +44,9 @@ def subprocess_started(config, target, sockets):
     # Re-open stdin.
     # This is required for some debugging environments.
     try:
-        sys.stdin = open("/dev/stdin")
+        sys.stdin = os.fdopen(0)
     except OSError:
-        print("WARNING: Failed to open '/dev/stdin', debugging might not work.")
+        print("WARNING: Failed to open stdin, debugging might not work.")
 
     # Logging needs to be setup again for each child.
     config.configure_logging()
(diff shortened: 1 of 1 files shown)


euri10:
I tested an app with a `pdb.set_trace()` in it and can use pdb with both workers and the reload flags, however I'm not very familiar with pdb and it's use so I'd like someone else to ponder here on what we should do. 

the above fix seems reasonable to me, another pair (or more :)) of eyes would be amazing.

uvicorn/subprocess.py
```diff
@@ -47,12 +39,13 @@ def subprocess_started(config, target, sockets, stdin_fileno):
...
                 by the parent process, and then passed to the child processes.
-    * stdin_fileno - The file number of sys.stdin, so that it can be reattached
-                     to the child process.
     """
     # Re-open stdin.
-    if stdin_fileno is not None:
-        sys.stdin = os.fdopen(stdin_fileno)
+    # This is required for some debugging environments.
+    try:
+        sys.stdin = open("/dev/stdin")
```
florimondmanca: Does /dev/stdin exist on all platforms?

likeanaxon: @florimondmanca Good question. I believe that it exists on OSX, but Windows is a different beast altogether, and I think it will require a special case (based on [this SO answer](https://stackoverflow.com/a/62913474); thank you @euri10 for pointing it out in the first place). Unfortunately, I don't have access to a Windows machine to test it. My goal was to get it to work without crashing first, and then work on a cross-platform solution.

Since the file descriptor for `stdin` is `0` on *nix and [apparently also on Windows](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fileno?view=msvc-160), I can replace `open("/dev/stdin")` with `os.fdopen(0)` - would that be OK?

likeanaxon: Follow-up: I went ahead and switched to `os.fdopen(0)`. Let me know if you need me to make any further changes.

likeanaxon:
@florimondmanca @euri10 Is there anything else I can do to progress this?

uvicorn/subprocess.py
```diff
@@ -47,12 +40,13 @@ def subprocess_started(config, target, sockets, stdin_fileno):
...
                 by the parent process, and then passed to the child processes.
-    * stdin_fileno - The file number of sys.stdin, so that it can be reattached
-                     to the child process.
     """
     # Re-open stdin.
-    if stdin_fileno is not None:
-        sys.stdin = os.fdopen(stdin_fileno)
+    # This is required for some debugging environments.
+    try:
+        sys.stdin = os.fdopen(0)
```
florimondmanca, line 47: I think I'm a bit confused. Does this version fixes the issue you were seeing? Asking because I would expect stdin.fileno() to return 0, as I believe is traditional on Unix systems, so this code would do the same as initially... ?

Were you able to see the value of the file descriptor that was throwing "Bad file descriptor"?

Do we need to add `os.dup(stdin_fileno)` to create a copy rather than reusing it directly?

florimondmanca (commented):
Leaving a question I'm basing off this SO thread... https://stackoverflow.com/questions/32199552/what-is-sys-stdin-fileno-in-python

uvicorn/subprocess.py
likeanaxon: >  Does this version fixes the issue you were seeing?

Yes, it does.

> I would expect stdin.fileno() to return 0, as I believe is traditional on Unix systems

I had exactly the same expectation.

> Were you able to see the value of the file descriptor that was throwing "Bad file descriptor"?

Yes, I was. In the original code, `stdin.fileno()` returns values greater than 2 (such as 9, 10 or 11; I mentioned that in the [original report](https://github.com/encode/uvicorn/issues/877#issuecomment-742133901)). The same thing happens if I use `os.dup`.

I am not sure if the issue runs deeper than that (for example, an issue with the way `stdin.fileno()` retrieves the file descriptor for `stdin`), but the fix definitely works - I've been using my branch for a couple of weeks now.

I would be curious to see if you can reproduce the issue - there is an MRE in the [bug report](https://github.com/encode/uvicorn/issues/877#issue-759008860).

Kludex:
Closing as stale.

PR #885, opened 2020-12-12
Test default logging

euri10:
While #859 bug went unoticed this test would have revealed it, nowhere we seem to test what the output of our default logging should be,
this is a small addition to do it

### DIFF
--- a/tests/middleware/test_trace_logging.py
+++ b/tests/middleware/test_trace_logging.py
@@ -84,7 +84,8 @@ def test_trace_logging(capsys, log_config, expected_204, expected_logger):
     assert response.status_code == 204
     thread.join()
     captured = capsys.readouterr()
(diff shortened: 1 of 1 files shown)


florimondmanca:
@euri10 This is a great idea. I had a suggestion to actually create a new and more thorough test case for this — figured it would be easier to send a new PR, so I filed this one: #887. WDYT?

euri10:
> @euri10 This is a great idea. I had a suggestion to actually create a new and more thorough test case for this — figured it would be easier to send a new PR, so I filed this one: #887. WDYT?

yep I prefer yours, there has been unnoticed regressions in the past where the startup/shutdown events were skipped (I think it was the shutdowns) and this one wont catch it, closing this in favor ofhttps://github.com/encode/uvicorn/pull/887

PR #887, opened 2020-12-12
Test output of default logging configuration

florimondmanca:
Alternative to #885

Adds a test that verifies the expected output from our default logging configuration on a basic startup-request-shutdown sequence.


florimondmanca:
Hmm. Looks like there's a weird interaction between the existing test cases and this new one. Only the access log line is captured when running all the tests in a row, rather than `-k test_default_logging`. 😕

euri10:
iirc I had to resort on using capsys instead of caplog when I sent the trace logging test but cant remember if I found the reason, and yes there might be an interaction from previious tests

tests/middleware/test_trace_logging.py
```diff
@@ -103,3 +106,49 @@ def test_access_logging(capsys, http_protocol):
...
+        assert response.status_code == 204
+        thread.join()
+
+        messages = [record.message for record in caplog.records]
+        assert "Started server process" in messages.pop(0)
+        assert "Waiting for application startup" in messages.pop(0)
+        assert "ASGI 'lifespan' protocol appears unsupported" in messages.pop(0)
+        assert "Application startup complete" in messages.pop(0)
+        assert "Uvicorn running on http://127.0.0.1:8000" in messages.pop(0)
+        assert '"GET / HTTP/1.1" 204' in messages.pop(0)
```
euri10: I'm just surprised here that it outputs 204 and not 204 No Content

florimondmanca: Yeah, same…

florimondmanca:
Even if using `capsys`, the captured output only contains the access log line. Weird…

tests/middleware/test_trace_logging.py
```diff
@@ -40,8 +42,8 @@
         },
     },
     "loggers": {
-        "": {"handlers": ["default"], "level": "INFO"},
-        "uvicorn.error": {"handlers": ["default"], "level": "INFO", "propagate": False},
+        "uvicorn": {"handlers": ["default"], "level": "INFO"},
+        "uvicorn.error": {"level": "INFO"},
```
florimondmanca: @euri10  There were two problems here:

* The test config overrides the `root` logger. Doesn't actually make the test fail if left as-is, but we had a PR to fix it in the default logging config: #674.
* It sets `propagate: False` on `uvicorn.error`, which is why when the new test runs after the others, `uvicorn.error` won't be captured. It _could_ be captured without changing the test config — I'd have to do the same trick than for `uvicorn.access` — but that sounds a bit absurd.

I think given that the test trace config aims at focusing on `uvicorn.asgi` in particular, keeping the rest in line with `LOGGING_CONFIG` sounds about right.

tests/middleware/test_trace_logging.py
euri10: I think it's because caplog uses its own formatter, and we use AccessFormatter for access logs,
I cant find a clean way to get it through caplog without messing up with pytest

euri10:
arg sorry @florimondmanca I didn't mean to push this here but to a new branch, hope this is not causing troubles, worst case we can revert my commit on this.
in any case I just renamed the filename to test_logging and leveraged your context manager a little bit more.

euri10:
@florimondmanca what do you think of this:

https://raw.githubusercontent.com/euri10/uvicorn/test_logging/tests/middleware/test_logging.py

i fu the last two commits on your PR, thought I was doing it on another branch, the above link is just the same as the present PR, just adapted for the new test suite,
it leverages you idea of the caplog context manager and test 3 things, trace, access and normal logger
feel free to take it, I dont want to "steal" PRs !!!

euri10:
closed in favor of https://github.com/encode/uvicorn/pull/964

PR #895, opened 2020-12-14
🐛 Fix resetting signals in Gunicorn UvicornWorker

### LINKED ISSUE #894: Subprocess returncode is not detected when running Gunicorn with Uvicorn (with fix PR companion)
### Checklist

### Describe the bug

When starting Gunicorn with Uvicorn worker(s), if the app uses `subprocess` to start other processes and captures the output, their `returncode` is in most cases `0`, even if the actual exit code was `1`.

### To reproduce

Take this minimal FastAPI app (or replace with Starlette), `main.py`:

```Python
import subprocess

from fastapi import FastAPI

app = FastAPI()

@app.post("/run")
def run_subprocess():
    result = subprocess.run(
        ["python", "-c", "import sys; sys.exit(1)"], capture_output=True
    )
    return {"returncode": result.returncode}
```

Then run it with:

```console
$ gunicorn -k uvicorn.workers.UvicornWorker main:app
```

Open the browser at http:127.0.0.1:8000/docs and send a request to `/run`.

### Expected behavior

The detected `returncode` should always be `1`, as the subprocess always exits with `1`.

### Actual behavior

In most of the cases it will return a `returncode` of `0`. Strangely enough, in some cases, it will return a `returncode` of `1`.

### Debugging material

This is because the `UvicornWorker`, which inherits from the base Gunicorn worker, declares a method `init_signals()` (overriding the parent method) but doesn't do anything. I suspect it's because the signal handlers are declared in the `Server.install_signal_handlers()` with compatibility with `asyncio`.

But the `UvicornWorker` process is started with `os.fork()` by Gunicorn (if I understand correctly) and by the point it is forked, the Gunicorn "Arbiter" class (that handles worker processes) already set its own signal handlers.

And the signal handlers in the Gunicorn base worker reset those handlers, but the `UvicornWorker` doesn't. So, when a process started with `subprocessing` is terminated, the `SIGCHLD` signal is handled by the Gunicorn `Arbiter` (as if the terminated process was a worker) instead of by the `UvicornWorker`.

Disclaimer: why the `SIGCHLD` signal handling in the Gunicorn `Arbiter` alters the `returncode` of a process run with `subprocess`, when capturing output, is still a mystery to me. But I realized the signal handler in the `Arbiter` is expected to handle dead worker processes. And worker subclasses all seem to reset the signal handlers to revert those signals set by the `Arbiter`.

I'm also submitting a PR to fix this: https://github.com/encode/uvicorn/pull/895. It's just 3 lines of code. But debugging it and finding it took me almost a week. :sweat_smile: 

### Environment

- OS / Python / Uvicorn version: just run `uvicorn --version`: `Running uvicorn 0.13.1 with CPython 3.8.5 on Linux` (it's actually installed from source, for debugging)
- Gunicorn version (also installed from source, for debugging): `gunicorn (version 20.0.4)`

```console
$ gunicorn -k uvicorn.workers.UvicornWorker main:app
```

### Additional context

I'm pretty sure this issue https://github.com/encode/uvicorn/issues/584 is related to the same problem.

tiangolo:
🐛 Fix resetting signals in Gunicorn `UvicornWorker` to fix subprocesses that capture output having an incorrect `returncode`.

Fixes #894 

Very short version:

When starting Gunicorn with Uvicorn worker(s), if the app uses `subprocess` to start other processes and captures the output, their `returncode` is in most cases `0`, even if the actual exit code was `1`.

Lengthy explanation and self-contained reproducible example in this issue: https://github.com/encode/uvicorn/issues/894

### DIFF
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -65,6 +65,7 @@ class UvicornWorker(Worker):
     def init_signals(self):
         # Reset signals so Gunicorn doesn't swallow subprocess return codes
         # other signals are set up by Server.install_signal_handlers()
+        # See: https://github.com/encode/uvicorn/issues/894
         for s in self.SIGNALS:
             signal.signal(s, signal.SIG_DFL)
 
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import signal
 
 from gunicorn.workers.base import Worker
 
@@ -62,7 +63,10 @@ class UvicornWorker(Worker):
         super(UvicornWorker, self).init_process()
 
     def init_signals(self):
-        pass
+        # Reset signals so Gunicorn doesn't swallow subprocess return codes
+        # other signals are set up by Server.install_signal_handlers()
+        for s in self.SIGNALS:
+            signal.signal(s, signal.SIG_DFL)
 
     def run(self):
         self.config.app = self.wsgi


uvicorn/workers.py
```diff
@@ -62,7 +63,10 @@ def init_process(self):
         super(UvicornWorker, self).init_process()
 
     def init_signals(self):
-        pass
+        # Reset signals so Gunicorn doesn't swallow subprocess return codes
+        # other signals are set up by Server.install_signal_handlers()
+        for s in self.SIGNALS:
+            signal.signal(s, signal.SIG_DFL)
```
euri10: what about removing entirely the `def init_signals(self):` override ?

it seems to me it was introduced at d0ff1cbd6a5979dca6ad21b859891b67b87704fc and looking at the version of the worker.py at that time it seemed necessary because signal handling was done differently.

it seems to me that the current override serves no purpose and is what prevents the base gunicorn Worker to reset its signal like it should, but I looked only briefly

tiangolo: Yeah, that was my first intuition as well.

But then I realized that the server installs the signal handlers it supports in a way that is compatible with asyncio here on `Server.install_signal_handlers()`: https://github.com/encode/uvicorn/blob/master/uvicorn/server.py#L242

I suspect that once those signals were added to git, this Gunicorn worker subclass method was changed to just override the parent class and have just the `pass`. To force it to *not* install the default handlers, as Uvicorn has its own. But then that previous change also removed the signal resets (which are still needed).

What I understand we need is 3 things:

* Reset the Gunicorn `Arbiter` signal handlers in the `UvicornWorker`, so that those signals are not used by the worker processes. That is done here in this PR.
* Override the default Gunicorn method `init_signals()`, to skip those signal handlers that are not compatible with asyncio. As those are not compatible with asyncio. That was done by the previous `pass` here, and is now done here by the fact that the method is still overridden and the previous signal handlers are not copied.
* Install the supported signal handlers in a way compatible with asyncio. That is done in [`Server.install_signal_handlers()`](https://github.com/encode/uvicorn/blob/master/uvicorn/server.py#L242) already.

uvicorn/workers.py
```diff
@@ -62,7 +63,10 @@ def init_process(self):
         super(UvicornWorker, self).init_process()
 
     def init_signals(self):
-        pass
+        # Reset signals so Gunicorn doesn't swallow subprocess return codes
+        # other signals are set up by Server.install_signal_handlers()
```
florimondmanca, line 67: I'm going ahead and adding a reference to the issue here too :)

```suggestion
        # other signals are set up by Server.install_signal_handlers()
        # See: https://github.com/encode/uvicorn/issues/894
```

florimondmanca (approved):
Great debugging, writeup and resolution! Thanks @tiangolo.

tiangolo:
Woohoo! Thanks @florimondmanca for checking, updating, and merging this! :rocket: :nerd_face: :cake: :coffee:

PR #898, opened 2020-12-17
Config coverage upgrade

euri10:
Another small step to #102 
brings config.py from 90 to 92%

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -65,10 +65,25 @@ def test_proxy_headers():
     assert isinstance(config.loaded_app, ProxyHeadersMiddleware)
 
 
-def test_app_unimportable():
+def test_app_unimportable_module():
     config = Config(app="no.such:app")
     with pytest.raises(ImportError):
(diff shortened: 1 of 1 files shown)


tests/test_config.py
```diff
@@ -65,12 +65,27 @@ def test_proxy_headers():
...
     config = Config(app="no.such:app")
     with pytest.raises(ImportError):
         config.load()
 
 
+def test_app_unimportable_other(caplog):
+    config = Config(app="test_config:app")
+    with pytest.raises(SystemExit):
+        config.load()
+        error_messages = [
```
florimondmanca: These lines need to be dedented, right now they most likely don't run because of the SystemExit exception thrown by config.load

florimondmanca (approved):
OK, just have a suggestion then G2G :)

tests/test_config.py
euri10: you're absolutely correct, my bad !

PR #899, opened 2020-12-18
Test config should_reload property

euri10:
another grind towards #102 

the should_reload property of the Config is set to True only if app is imported as a string and debug/reload is set

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -40,11 +40,16 @@ def wsgi_app(environ, start_response):
     pass  # pragma: nocover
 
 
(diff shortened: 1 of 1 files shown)


florimondmanca (commented):
Should we have a separate test, instead of hijacking the one for the debut flag?

euri10:
> Should we have a separate test, instead of hijacking the one for the debut flag?

yep it makes sense to have separated and more precise tests, I changed

florimondmanca (approved):

PR #901, opened 2020-12-18
Fix gunicorn worker not running if extras not installed

### LINKED ISSUE #816: UvicornWorker 'auto' loop and http implementations don't work on Linux without extras installed
### Checklist

### Describe the bug

After #666 , uvloop and and httptools are no longer installed unless you specify to install the 'standard' extra. Because of this when running with `loop='auto'` and `http='auto'` the application fails to start due to 'uvloop' and 'httptools' being selected. NOTE: This only happens when running the `uvicorn.workers.UvicornWorker` through something like Gunicorn.

### To reproduce

```bash
pip install uvicorn gunicorn
```

```python
#main.py

    await send({
        'status': 200,
        'headers': [
        ],
    })
    await send({
        'type': 'http.response.body',
        'body': b'Hello, world!',
    })
```

```bash
gunicorn -k uvicorn.workers.UvicornWorker main:app
```

### Expected behavior

'auto' should pick 'asyncio' and 'h11' for their respective settings on Linux if the 'standard' extra is not installed.

### Actual behavior

Application starts up but immediately throws `ModuleNotFoundError: No module named 'uvloop'`. If you install uvloop it then complains about not finding httptools. After installing both, no more issues occur.

### Environment

- Running uvicorn 0.12.1 with CPython 3.7.6 on Linux

euri10:
Fixes https://github.com/encode/uvicorn/issues/816

### DIFF
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -4,6 +4,7 @@ import logging
 from gunicorn.workers.base import Worker
 
 from uvicorn.config import Config
+from uvicorn.loops.auto import auto_loop_setup
 from uvicorn.main import Server
 
 
@@ -13,7 +14,7 @@ class UvicornWorker(Worker):
     rather than a WSGI callable.
     """
 
-    CONFIG_KWARGS = {"loop": "uvloop", "http": "httptools"}
+    CONFIG_KWARGS = {"loop": "auto", "http": "auto"}
 
     def __init__(self, *args, **kwargs):
         super(UvicornWorker, self).__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


florimondmanca:
I got hit by this too — getting errors when starting Gunicorn locally when `uvloop` and/or `httptools` were not installed. (httptools was particularly bad, since Gunicorn kept spinning up new processes to make up for the startup failure.)

I think switching to "auto" by default for both loop and HTTP parser sounds good. We probably want to keep this for `0.14` though?

euri10:
> I think switching to "auto" by default for both loop and HTTP parser sounds good. We probably want to keep this for `0.14` though?

yep I think you're the semver god here !

Kludex (approved):
> We probably want to keep this for 0.14 though?

Yes.

PR #902, opened 2020-12-19
Increase config coverage for ssl_ca_certs flag

euri10:
Another small grind towards #102 

this makes use of this flag


euri10:
closed for #982

PR #903, opened 2020-12-19
Fix factory if app uses a Middleware

### LINKED ISSUE #900: uvirorn.workers.UvicornWorker fails with certain ASGI middleware
### Description
Certain middleware is fails, either when launcing Uvicornworker, or when calling the ASGI handler.
A concrete case is the SentryAsgiMiddleware.
See https://github.com/getsentry/sentry-python/issues/947 for info.
Not sure if the problem is in Uvicorn or Sentry.

### To reproduce

Install the Sentry lib in a ASGI application, run with gunicorn/uviworker
```
from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
app = SentryAsgiMiddleware(app)
```
and 
```
gunicorn api:app -w 1 -k uvicorn.workers.UvicornWorker
```

### Expected behavior

Application launches and runs

### Actual behavior

As set up above, the following error is produced when launcing gunicorn:
```
File "/usr/lib/python3.7/inspect.py", line 2208, in _signature_from_callable
    raise TypeError('{!r} is not a callable object'.format(obj))
```
If I forcefully expose either a **asgi2** or **asgi3** _interface_ from the middleware, using either the `_run_asgi2()` or `_run_asgi3()` **methods** on the middleware object, the servers **starts up**, but then fails when handling a request, with e.g.:
```
TypeError: _run_asgi3() missing 2 required positional arguments: 'receive' and 'send'
```

Apparently, the introspection of the callable fails to recognize _method_ objects.

### Additional context

Using the **uvicorn** webserver it is possible to specify asgi2/3 interface using a commandline argument, and thus override the detection mechanism, which seems to be failing on this particular middleware.  Using the Uvicorn worker this appears to be not possible.

again, see issue here. See https://github.com/getsentry/sentry-python/issues/947 for info.

euri10:
Fixes #900 

I took deliberately quite a brutal way : removing the problematic piece of code and its test.

I dont understand the need for that test, so there might be a better or clever way of handling this !

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -94,11 +94,6 @@ def test_app_factory():
     config.load()
     assert config.loaded_app is asgi_app
 
-    # Flag missing.
-    config = Config(app=create_app)
-    with pytest.raises(SystemExit):
-        config.load()
-
     # App not a no-arguments callable.
     config = Config(app=asgi_app, factory=True)
     with pytest.raises(SystemExit):
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -316,12 +316,6 @@ class Config:
             except TypeError as exc:
                 logger.error("Error loading ASGI app factory: %s", exc)
                 sys.exit(1)
-        elif not inspect.signature(self.loaded_app).parameters:
-            logger.error(
-                "APP seems to be an application factory. "
-                "Run uvicorn with the --factory flag."
-            )
-            sys.exit(1)
 
         if self.interface == "auto":
             if inspect.isclass(self.loaded_app):


florimondmanca (commented):
What's the behavior now for this use case where users point at a factory directly but forget the `--factory` flag…?

```python
def create_app():
    return ...
```

```
uvicorn app:create_app
```

The intent of that error message was to be more proactive so users aren't left having to figure out what's happening.

We could also do "try to get the signature, and if that fails, assume an ASGI app instance because factory functions would _definitely_ have a signature"?

euri10:
> What's the behavior now for this use case where users point at a factory directly but forget the `--factory` flag…?
> 
> ```python
> def create_app():
>     return ...
> ```
> 
> ```
> uvicorn app:create_app
> ```
> 
> The intent of that error message was to be more proactive so users aren't left having to figure out what's happening.

yep in that case the app will load and return a 500 on 1st request with 
```
    instance = self.app(scope)
TypeError: create_app() takes 0 positional arguments but 1 was given
```

so it's quite deceptive ...

euri10:
and if we put the factory logic after?
nah bad idea

euri10:
I'm going to close this, given the excellent [explanation](https://github.com/encode/uvicorn/issues/900#issuecomment-748518295) given by @florimondmanca it's debatable the fix is needed here, let's see.

PR #908, opened 2020-12-22
Added docs example to tweak gunicorn worker config

### LINKED ISSUE #266: Configure workers
It would be great if the gunicorn workers could support uvicorn's config options.

I would like to run a Django Channels project using them, but cannot do so at this time because the `disable_lifespan` option from #262 cannot be set.

euri10:
Fixes #266 by writing an example on how to pass config params to gunicorn worker


florimondmanca (approved):
Nice addition, yes 👍

PR #909, opened 2020-12-23
remove redundant transform list -> list

SergBobrovsky:


euri10 (approved):
Thanks, seems logical

Issue #760, opened 2020-08-18
Server parts of websocket tests are not affecting test outcome


stefanw:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

While researching #757 I noticed that pytest assertions in server parts of websocket tests (`with run_server(...)`) are not working as intended in some cases.

Assertion failures are displayed (`pytest -s`) but do *not* affect the test run outcome. I suspect the reason lies in running the server in a loop in a thread and neither exceptions nor pytest failures are propagating to the main thread.

For example the `pytest.raises` in [`test_send_after_protocol_close`](https://github.com/encode/uvicorn/blob/093a1f7c1ef7e30ec0b8b63618a4d5ee59f323aa/tests/protocols/test_websocket.py#L296) is silently ignored.

### To reproduce

I added an `assert False` in the server part of a test in #759 and all tests are passing.

### Expected behavior

I expect the tests to fail when using `assert False` or `pytest.fail("...")` in the server parts of websocket tests.

### Actual behavior

Exception or failures are printed to standard out, but do not affect the overall test outcome.

### Debugging material

See test run for #759

### Environment

- OS / Python / Uvicorn version: Running uvicorn 0.11.8 with CPython 3.8.2 on Darwin

### Additional context

I suspect that the server running in a loop in a thread is not propagating exceptions or pytest failures.

Kludex:
We are on uvicorn 0.20.0, and by now we don't run the server in a thread anymore. We run it as an async task, but we still have the same behavior.

Kludex:
I think this is fine. When the request-response cycle fails in the server, we can have an exception, but that will only be valid for that cycle, and we can't actually retrieve that exception, unless we mock some object.

We need to be mindful that assertions need to happen outside the server on the test suite. :pray:

PR #910, opened 2020-12-23
Stricter pytest

euri10:
It seems to me we could benefit from a stricter pytest, especially the [pytest.PytestUnhandledThreadExceptionWarning](https://docs.pytest.org/en/stable/reference.html#pytest.PytestUnhandledThreadExceptionWarning) since we use threads a lot.

Moreover, test exceptions raised in a separate thread dont seem to be handled well with pytest, we have an example of this in the ws test suite, see https://github.com/encode/uvicorn/issues/760 for instance.
This is also one of the reason our test suite behave strangely as sometimes the whole suite passes fine but seperate tests triggered on their own can fail.

With this pytest configuration we can see 2 issues, fixed after:

1. the test_config left an unclosed socket when the bind_socket() method was called
2.  and the whole ws test suite failed with those stricter testing, as the context manager to run the ws server was not closed

this is the error we would get with the stricter config, without the 2 fixes.

```
>               warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
E               pytest.PytestUnraisableExceptionWarning: Exception ignored in: <socket.socket fd=-1, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0>
E               
E               Traceback (most recent call last):
E                 File "/home/lotso/PycharmProjects/uvicorn/venv/lib/python3.8/site-packages/_pytest/python.py", line 183, in pytest_pyfunc_call
E                   result = testfunction(**testargs)
E               ResourceWarning: unclosed <socket.socket fd=17, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('127.0.0.1', 8000)>

../venv/lib/python3.8/site-packages/_pytest/unraisableexception.py:78: PytestUnraisableExceptionWarning
```


euri10:
interestingly the ci fails on windows....on other tests !
have no idea as to why.... will investigate
if you get any idea I'm all ears, dont have a win machine at hand now

euri10:
will continue on this when I'll have a windows machine, those errors are hard to debug because it's failing on weird places and to be honest the pytest messages are not helping

euri10:
haha closing this, the test suite changed and the new approach with context manager leads to way less issues it seems, will resurrect when/if #918 lands

PR #912, opened 2020-12-23
Fix lifespan tests

### LINKED ISSUE #911: Lifespan tests are broken
### Checklist

### Describe the bug

Some lifespan tests are passing fine while they should fail

### To reproduce

this one for instance pass while the log suggests it should fail, if you change the assert scope to assert False it still passes !

```
@pytest.mark.parametrize("mode", ("auto", "on"))
def test_lifespan_scope_asgi3app(mode):
    async def asgi3app(scope, receive, send):
        assert scope == {"version": "3.0", "spec_version": "2.0"}

    async def test():
        config = Config(app=asgi3app, lifespan=mode)
        lifespan = LifespanOn(config)

        await lifespan.startup()
        await lifespan.shutdown()

    loop = asyncio.new_event_loop()
    loop.run_until_complete(test())
```

```
/home/lotso/PycharmProjects/uvicorn/venv/bin/python /home/lotso/.local/share/JetBrains/Toolbox/apps/PyCharm-P/ch-0/203.5981.165/plugins/python/helpers/pycharm/_jb_pytest_runner.py --target test_lifespan.py::test_lifespan_scope_asgi3app
Testing started at 2:47 PM ...
Launching pytest with arguments test_lifespan.py::test_lifespan_scope_asgi3app in /home/lotso/PycharmProjects/uvicorn/tests

============================= test session starts ==============================
platform linux -- Python 3.8.5, pytest-6.2.0, py-1.10.0, pluggy-0.13.1 -- /home/lotso/PycharmProjects/uvicorn/venv/bin/python
cachedir: .pytest_cache
rootdir: /home/lotso/PycharmProjects/uvicorn, configfile: setup.cfg
plugins: reraise-1.0.3, asyncio-0.14.0
collecting ... collected 2 items

test_lifespan.py::test_lifespan_scope_asgi3app[auto] 
test_lifespan.py::test_lifespan_scope_asgi3app[on] 

============================== 2 passed in 0.11s ===============================

PASSED              [ 50%]INFO:     Waiting for application startup.
PASSED                [100%]INFO:     Waiting for application startup.
  File "/home/lotso/PycharmProjects/uvicorn/uvicorn/lifespan/on.py", line 55, in main
  File "/home/lotso/PycharmProjects/uvicorn/uvicorn/middleware/proxy_headers.py", line 45, in __call__
  File "/home/lotso/PycharmProjects/uvicorn/tests/test_lifespan.py", line 161, in asgi3app
    assert scope == {"version": "3.0", "spec_version": "2.0"}
AssertionError: assert {'asgi': {'spec_version': '2.0', 'version': '3.0'}, 'type': 'lifespan'} == {'spec_version': '2.0', 'version': '3.0'}
```

### Expected behavior

test should fail

### Environment

euri10:
Fixes #911

it seems to me we have to raise the lifespan exception should it occur and call the lifespan main coroutine in the tests.

### DIFF
--- a/tests/test_lifespan.py
+++ b/tests/test_lifespan.py
@@ -115,6 +115,7 @@ def test_lifespan_on_with_error():
         await lifespan.startup()
         assert lifespan.error_occured
         assert lifespan.should_exit
+        await lifespan.main()
         await lifespan.shutdown()
 
     loop = asyncio.new_event_loop()
@@ -165,6 +166,7 @@ def test_lifespan_scope_asgi3app(mode):
         lifespan = LifespanOn(config)
 
         await lifespan.startup()
+        await lifespan.main()
         await lifespan.shutdown()
 
     loop = asyncio.new_event_loop()
@@ -186,6 +188,7 @@ def test_lifespan_scope_asgi2app(mode):
         lifespan = LifespanOn(config)
 
         await lifespan.startup()
+        await lifespan.main()
         await lifespan.shutdown()
 
     loop = asyncio.new_event_loop()
--- a/tests/test_lifespan.py
+++ b/tests/test_lifespan.py
@@ -115,7 +115,8 @@ def test_lifespan_on_with_error():
         await lifespan.startup()
         assert lifespan.error_occured
         assert lifespan.should_exit
-        await lifespan.main()
+        with pytest.raises(RuntimeError):
+            await lifespan.main()
         await lifespan.shutdown()
 
     loop = asyncio.new_event_loop()
@@ -159,7 +160,10 @@ def test_lifespan_with_failed_startup(mode, raise_exception, caplog):
 @pytest.mark.parametrize("mode", ("auto", "on"))
 def test_lifespan_scope_asgi3app(mode):
     async def asgi3app(scope, receive, send):
-        assert scope == {"version": "3.0", "spec_version": "2.0"}
+        assert scope == {
+            "asgi": {"spec_version": "2.0", "version": "3.0"},
+            "type": "lifespan",
+        }
 
     async def test():
         config = Config(app=asgi3app, lifespan=mode)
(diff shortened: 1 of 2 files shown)


tests/test_lifespan.py
```diff
@@ -158,13 +160,17 @@ async def test():
...
+            "asgi": {"spec_version": "2.0", "version": "3.0"},
+            "type": "lifespan",
+        }
 
     async def test():
         config = Config(app=asgi3app, lifespan=mode)
         lifespan = LifespanOn(config)
 
         await lifespan.startup()
+        await lifespan.main()
```
florimondmanca, line 173: Hmm, this feels weird. `lifespan.startup()` already schedules `lifespain.main()` on the event loop:

https://github.com/encode/uvicorn/blob/a9a46e87d8ba41f08b3a401b7b0cc572333a7193/uvicorn/lifespan/on.py#L29

I think we should instead add assertions:

```suggestion
        assert not lifespan.startup_failed
        assert not lifespan.error_occurred
        assert not lifespan.should_exit
```

There's also a small inconsistency here bc we test against `mode=auto` _and_ `mode=on`, but we always use `LifespanOn`. I think we don't necessarily need to bother with lifespan modes, just test the scope has the correct shape against `mode=on` (that would then properly show any exceptions in the error logs).

tests/test_lifespan.py
```diff
@@ -186,6 +195,7 @@ async def test():
...
```
florimondmanca, line 198:
```suggestion
        await lifespan.startup_event.wait()
```

tests/test_lifespan.py
florimondmanca: Filed #913 to illustrate. :)

euri10: yes this is way better I missed that error_occured !

PR #914, opened 2020-12-26
Tweak detection of app factories

### LINKED ISSUE #900: uvirorn.workers.UvicornWorker fails with certain ASGI middleware

florimondmanca:
Fixes #900, cleanup for #875, based on https://github.com/encode/uvicorn/issues/900#issuecomment-748959580

Instead of inspecting the `app`, attempt to call it without arguments each time.

* When it fails without `--factory`, consider it as normal — the app is meant to be an app instance.
* When it fails with `--factory`, exit with an error message — the factory failed to load.
* When it succeeds without `--factory`, keep Uvicorn running, but issue a warning — explicit flag is preferred:

```console
$ uvicorn debug.app:create_app
WARNING:  ASGI app factory detected. Using it, but please consider setting the --factory flag explicitly.
INFO:     Started server process [202]
```

This removes the `inspect.signature` code that was causing https://github.com/getsentry/sentry-python/issues/947.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import socket
 from copy import deepcopy
 
@@ -100,7 +101,7 @@ def test_app_unimportable_other(caplog):
     )
 
 
-def test_app_factory():
+def test_app_factory(caplog):
     def create_app():
         return asgi_app
 
@@ -108,10 +109,15 @@ def test_app_factory():
     config.load()
     assert config.loaded_app is asgi_app
 
-    # Flag missing.
-    config = Config(app=create_app)
-    with pytest.raises(SystemExit):
+    # Flag not passed. In this case, successfully load the app, but issue a warning
+    # to indicate that an explicit flag is preferred.
+    caplog.clear()
+    config = Config(app=create_app, proxy_headers=False)
+    with caplog.at_level(logging.WARNING):
         config.load()
+    assert config.loaded_app is asgi_app
+    assert len(caplog.records) == 1
+    assert "--factory" in caplog.records[0].message
 
     # App not a no-arguments callable.
     config = Config(app=asgi_app, factory=True)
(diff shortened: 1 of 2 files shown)


euri10 (approved):

euri10:
this rocks !

kristjanvalur:
Great, thanks!

PR #917, opened 2020-12-27
Test suite refactor using async server context manager instead of background thread

euri10:
WIP, but this looks promising, at least exceptions are bubbling up in ws tests !

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,6 +21,8 @@ cryptography
 coverage
 httpx
 pytest-asyncio
+asynccontextmanager; python_version < '3.7'
+
 
 # Documentation
 mkdocs
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,6 +19,8 @@ mypy
 trustme
 cryptography
 coverage
+httpx
+pytest-asyncio
 
 # Documentation
 mkdocs
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,7 +21,7 @@ cryptography
 coverage
 httpx
 pytest-asyncio
-asynccontextmanager; python_version < '3.7'
+async_generator; python_version < '3.7'
 
 
 # Documentation
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,8 +19,8 @@ mypy
 trustme
 cryptography
 coverage
-httpx
-pytest-asyncio
+httpx==0.16.*
+pytest-asyncio==0.14.*
 async_generator; python_version < '3.7'
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,10 @@
 import asyncio
-from contextlib import asynccontextmanager
+import sys
+
+if sys.version_info >= (3, 7):
+    from contextlib import asynccontextmanager
+else:
+    from async_generator import asynccontextmanager
 
 import pytest
 import trustme
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,24 @@
+import asyncio
+from contextlib import asynccontextmanager
+
 import pytest
 import trustme
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
+
+from uvicorn import Config, Server
+
+
+@asynccontextmanager
+async def run_server(config: Config):
+    server = Server(config=config)
+    cancel_handle = asyncio.ensure_future(server.serve())
+    await asyncio.sleep(0.1)
+    try:
+        yield
+    finally:
+        await server.shutdown()
+        cancel_handle.cancel()
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,29 +1,7 @@
-import asyncio
-import sys
-
-if sys.version_info >= (3, 7):
-    from contextlib import asynccontextmanager
-else:
-    from async_generator import asynccontextmanager
-
 import pytest
 import trustme
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
-
-from uvicorn import Config, Server
-
-
-@asynccontextmanager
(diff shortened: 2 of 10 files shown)


euri10:
ok this is almost ready but so far i like i a lot, minus 143 lines, exceptions are bubbling now. this feels way cleaner overall

few points to note before it's in a mergeable state:

1. FIXED ~~I removed 3.6 from the suite since asynccontextmanager appears in 3.7, there might be ways to add it though I didnt take time to look at that properly~~
2. 5 tests were skipped on windows (test_ssl and test_trace_logging), they are not anymore
3. I removed entirely one test [(test_run_with_shutdown)](https://github.com/encode/uvicorn/pull/917/files#diff-41d180f6f7233d175c2dfe36c45c4ea80eed4754fa7ed4eed46ccde1b2a778c7L62-L85) because it seems pointless now.
4. this [part is removed](https://github.com/encode/uvicorn/pull/917/files#diff-41d180f6f7233d175c2dfe36c45c4ea80eed4754fa7ed4eed46ccde1b2a778c7L62-L85) as it becomes dead code

euri10:
pretty happy with it now !

requirements.txt
```diff
@@ -19,6 +19,10 @@ mypy
 trustme
 cryptography
 coverage
+httpx
+pytest-asyncio
+async_generator; python_version < '3.7'
```
florimondmanca: Should we pin some of these, perhaps HTTPX and `pytest-asyncio` to the latest minor versions?

tests/conftest.py
```diff
@@ -1,8 +1,30 @@
+import asyncio
+import sys
+
+if sys.version_info >= (3, 7):
+    from contextlib import asynccontextmanager
+else:
+    from async_generator import asynccontextmanager
```
florimondmanca: From experience, a great approach for this kind of thing is to have a dedicated "compatibility" module.

Should we move this to a new `tests/compat.py`?

Also I think we could just try/except:

```suggestion
try:
    from contextlib import asynccontextmanager
except ImportError:  # pragma: no cover
    from async_generator import asynccontextmanager
```

tests/conftest.py
```diff
@@ -1,8 +1,30 @@
...
+@asynccontextmanager
+async def run_server(config: Config, sockets=None):
+    server = Server(config=config)
+    cancel_handle = asyncio.ensure_future(server.serve(sockets=sockets))
+    await asyncio.sleep(0.1)
+    try:
+        yield server
+    finally:
+        await server.shutdown()
+        cancel_handle.cancel()
```
florimondmanca: I _think_ we already have a `utils.py` module. What about move this there? If not, maybe we can consider it. Importing from conftest feels a bit strange given this is a special pytest module — things only we use are probably better in a separate module.

tests/test_main.py
```diff
@@ -23,69 +20,27 @@ async def app(scope, receive, send):
...
-        time.sleep(0.01)
-    response = requests.get("http://127.0.0.1:8000")
+    async with run_server(config):
+        async with httpx.AsyncClient() as client:
+            response = await client.get("http://127.0.0.1:8000")
     assert response.status_code == 204
-    thread.join()
-
-
-def test_run_with_shutdown():
```
florimondmanca, line 62: Could you clarify why we can do without this test now?

uvicorn/server.py
```diff
@@ -240,12 +239,7 @@ async def shutdown(self, sockets=None):
             await self.lifespan.shutdown()
 
     def install_signal_handlers(self) -> None:
-        if threading.current_thread() is not threading.main_thread():
-            # Signals can only be listened to from the main thread.
-            return
```
florimondmanca: We added this via #871, and I don't think we should be dropping it. Sure, maybe we don't really care about it in our own test suite anymore, but people who are running Uvicorn within threads certainly still do. Right?

tests/test_ssl.py
```diff
@@ -31,53 +11,37 @@ async def app(scope, receive, send):
...
     )
-    server = Server(config=config)
-    thread = threading.Thread(target=server.run)
-    thread.start()
-    while not server.started:
-        time.sleep(0.01)
-    with no_ssl_verification():
-        response = requests.get("https://127.0.0.1:8000")
+    async with run_server(config):
+        async with httpx.AsyncClient(verify=False) as client:
```
florimondmanca, line 24: Note for the future: maybe Requests didn't support it, but now with HTTPX I think we could pass `verify=<ca_cert>`, rather than disabling SSL verification.

florimondmanca (changes requested):
Looking _great_ ✨ Nice one!

A few comments, mostly "cosmetic" but not only. :-)

requirements.txt
euri10: yep I pinned both

tests/conftest.py
euri10: yes you're more than correct, I moved everything (import and run-server) in a new utils.py

uvicorn/server.py
euri10: Oo, I never think about that, I reverted

tests/test_ssl.py
euri10: I wanted to mimic the no verification, maybe another PR can use the certificates fixtures !

tests/test_main.py
euri10: ok maybe it's me misunderstanding what the test does but..

the way i get it is that it's equivalent to the new `async with run-server` with the slight exception that a socket `sock` was explicitly binded, and it just checks the fact no exception is raised in that case.

Now if the context manager was raising an exception all our tests would fail. hence why I didnt bother.

honestly at first I tried to run it  "the new way" and couldnt find a way that was really testing something.

florimondmanca: Yes, OK, sounds good :)

florimondmanca (approved):
Awesome. :)

PR #918, opened 2020-12-27
Remove requests dependency

euri10:
This build on #917 and removes requests dependency entirely, this feels also way cleaner given the neat httpx API, cant emphasize enough how easier it is with it !

Another 100ish lines slashed !

---------------------------------------------------------------------------------------------------
Not related to the PR still maybe interesting info for you @florimondmanca :
If I had only one slight ux critic it would be that if you forget to pass a `base_url` to the AsyncClient then you get a not very helpful urllib error.
This is only by experience I remembered that I had to pass it to test directly the asgi app, but I think it's not obvious if you dont know.

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../venv/lib/python3.8/site-packages/httpx/_client.py:1548: in get
    return await self.request(
../../venv/lib/python3.8/site-packages/httpx/_client.py:1371: in request
    response = await self.send(
../../venv/lib/python3.8/site-packages/httpx/_client.py:1406: in send
    response = await self._send_handling_auth(
../../venv/lib/python3.8/site-packages/httpx/_client.py:1444: in _send_handling_auth
    response = await self._send_handling_redirects(
../../venv/lib/python3.8/site-packages/httpx/_client.py:1476: in _send_handling_redirects
    response = await self._send_single_request(request, timeout)
../../venv/lib/python3.8/site-packages/httpx/_client.py:1524: in _send_single_request
    self.cookies.extract_cookies(response)
../../venv/lib/python3.8/site-packages/httpx/_models.py:1329: in extract_cookies
    urllib_request = self._CookieCompatRequest(response.request)
../../venv/lib/python3.8/site-packages/httpx/_models.py:1453: in __init__
    super().__init__(
/home/lotso/.asdf/installs/python/3.8.6/lib/python3.8/urllib/request.py:328: in __init__
    self.full_url = url
/home/lotso/.asdf/installs/python/3.8.6/lib/python3.8/urllib/request.py:354: in full_url
    self._parse()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <httpx._models.Cookies._CookieCompatRequest object at 0x7f4fa4d2b370>

    def _parse(self):
        self.type, rest = _splittype(self._full_url)
        if self.type is None:
>           raise ValueError("unknown url type: %r" % self.full_url)
E           ValueError: unknown url type: ''

/home/lotso/.asdf/installs/python/3.8.6/lib/python3.8/urllib/request.py:383: ValueError
```


euri10:
~~I'll undraft this if #917 lands, in the meantime there seems to be a slight glitch with headers encoding with httpx (or to be fair requests was ok with it but it fails with httpx), any idea @florimondmanca ?~~

tests/test_client.py
```diff
@@ -14,15 +15,21 @@ async def hello_world(scope, receive, send):
     )
 
 
-def test_explicit_base_url():
-    client = TestClient(hello_world, base_url="http://testserver:321")
-    response = client.get("/")
```
florimondmanca, line 19: I guess we can just remove this entire module, and the definition of the TestClient as well, right? We don't want to be re-testing HTTPX. :)

tests/middleware/test_proxy_headers.py
```diff
@@ -14,28 +17,34 @@ async def app(scope, receive, send):
...
-    client = TestClient(app)
-    headers = {
-        "X-Forwarded-Proto": "https",
-        "X-Forwarded-For": "\xf0\xfd\xfd\xfd, 1.2.3.4",
-    }
-    response = client.get("/", headers=headers)
+@pytest.mark.asyncio
+async def test_proxy_headers_invalid_x_forwarded_for():
+    async with httpx.AsyncClient(app=app, base_url="http://testserver") as client:
+        headers = Headers(
```
florimondmanca: We can do this to avoid importing Headers explicitly:

```suggestion
        headers = httpx.Headers(
```

florimondmanca (commented):
Looking good!

tests/test_client.py
euri10: done

tests/middleware/test_proxy_headers.py
euri10: done

euri10:
not sure what's happening here but the `test_send_before_handshake` seems flaky, everything passes fine locally, will have to do some super interesting CI live debugging,

florimondmanca:
FWIW there's an asyncio warning when I try to run `test_send_before_handshake` or `test_missing_handshake` in isolation. Doesn't show if I run either along with other tests, eg either the full test suite, with `-k websocket`, or the two tests together with `-k "missing_handshake or before_handshake"`.

```console
$ pytest -k test_missing_handshake
====================================================== test session starts =======================================================
platform darwin -- Python 3.9.0, pytest-6.1.1, py-1.9.0, pluggy-0.13.1
rootdir: /Users/florimond/Developer/python-projects/uvicorn, configfile: setup.cfg
plugins: mock-3.3.1, asyncio-0.14.0
collected 213 items / 211 deselected / 2 selected                                                                                

tests/protocols/test_websocket.py ..                                                                                       [100%]

=============================================== 2 passed, 211 deselected in 2.30s ================================================
task: <Task pending name='Task-11' coro=<WebSocketServerProtocol.handler() done, defined at /Users/florimond/Developer/python-projects/uvicorn/venv/lib/python3.9/site-packages/websockets/server.py:118> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x10a7e1760>()]>>
```

euri10:
I dont understand very well the sequence of events here.
It looks like, as you rightfully noticed, that the `WebSocketServerProtocol.handler` is  still pending is what causes the flakiness in those 2 tests.

I've tried to read https://websockets.readthedocs.io/en/stable/design.html?highlight=cancellation#cancellation but cant get a firm grasp on how to handle the situation. 

It seems we destroy this task in the server context manager when we do `cancel_handle.cancel()`

I think I hate ws

florimondmanca:
😆

> For these reasons, websockets never cancels connection handlers. Instead it expects them to detect when the connection is closed, execute finalization logic if needed, and exit.

I think it means that our `ws_handler` should be checking for disconnects and/or shutdowns on its own. Then then instead of cancelling, we do `server.should_exit = True` and then `await cancel_handle` (the cancel handle is actually an asyncio task) so the handler can handle eg a shutdown event and stop gracefully. Makes sense?

euri10:
ok I think I nailed it (well not sure but that's a mind-boggling stuff honestly...)

if you look closely @florimondmanca we effectively have that task pending on both `test_send_before_handshake` or `test_missing_handshake` ran in isolation

In both scenarios, when we hit the `connection_lost` method the Event `self.handshake_started_event` is **not** set.

With this last commit, the destroyed pending task disappears for `test_send_before_handshake` but it still persists for `test_missing_handshake`
So overall I've got the feeling we may forget to set that Event somewhere else in those "handshake failed" scenarios

euri10:
I just ran all CI jobs here after the flaky tests fix merge just to be sure, i think we're gtg on this one @florimondmanca !
thanks for your patience on all those tests refactor !

florimondmanca (approved):
Nice! Happy to have HTTPX here now. :-)

PR #925, opened 2020-12-29
Add httptools failing test for long url

### LINKED ISSUE #924: Httptools trims long urls
Prompted as part of the discussion in #344 , it seems httptools doenst handle well long url, see #925 for a failing test

I can say that for my case the issue is not URLs longer than 65,535 characters, but rather that `on_url` is being called multiple times for whatever reason: https://github.com/nodejs/http-parser#parsing-urls.

_Originally posted by @malthejorgensen in https://github.com/encode/uvicorn/issues/344#issuecomment-752074479_

euri10:
Does nothing for now, just shows the issue
May fix #924


Kludex:
@euri10 Can you rebase and see if we still have this issue, please? This should have been solved by `httptools==0.2.0`.

euri10:
it's still failing in `parsed_url = httptools.parse_url(url)` which returns None everywhere for that length

euri10:
~~kind of normal given the parser accepts url up to https://github.com/nodejs/http-parser/blob/ec8b5ee63f0e51191ea43bb0c6eac7bfbff3141d/http_parser.c#L31 so not sure we should handle that or ask it upstream~~ this is the old parser...

re-edit in fact httptools still does the url parsing with http_parser so using httptools 0.2.0 wont improve things for us, since it doesnt change anything on that front

euri10:
closing this 
see https://github.com/MagicStack/httptools/issues/68

PR #926, opened 2020-12-29
Improve user feedback if no ws library installed

### LINKED ISSUE #797: Error during WebSocket handshake: Unexpected response code: 400 (anonymous) @ (index):16
### Checklist

### Describe the bug

I am following the Fastapi websockets tutorial (https://fastapi.tiangolo.com/advanced/websockets/#await-for-messages-and-send-messages). With the exact code given (see link) I start uvicorn with `uvicorn main:app --reload` visit http://127.0.0.1:8000/. I get this error in my chrome console: 

```
(index):16 WebSocket connection to 'ws://localhost:8000/ws' failed: Error during WebSocket handshake: Unexpected response code: 400
```
### Expected behavior

When I click "send" I expect to see the message I typed appear on the screen
### Actual behavior

When I click "send" nothing happens
### Environment

macos 10.15.5
python 3.8.2
uvicorn 0.12.1 

`uvicorn main:app --reload`

euri10:
Fixes https://github.com/encode/uvicorn/issues/797

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -39,8 +39,7 @@ Install using `pip`:
 $ pip install uvicorn
 ```
 
-This will install uvicorn with minimal (pure Python) dependencies 
-(note: this does not enable websockets support, see below for "activating" it).
+This will install uvicorn with minimal (pure Python) dependencies.
 
 ```shell
 $ pip install uvicorn[standard]
--- a/docs/index.md
+++ b/docs/index.md
@@ -39,7 +39,8 @@ Install using `pip`:
 $ pip install uvicorn
 ```
 
-This will install uvicorn with minimal (pure Python) dependencies.
+This will install uvicorn with minimal (pure Python) dependencies 
+(note: this does not enable websockets support, see below for "activating" it).
 
 ```shell
 $ pip install uvicorn[standard]
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -266,11 +266,13 @@ class H11Protocol(asyncio.Protocol):
         if upgrade_value != b"websocket" or self.ws_protocol_class is None:
             msg = "Unsupported upgrade request."
             self.logger.warning(msg)
+
             from uvicorn.protocols.websockets.auto import AutoWebSocketsProtocol
 
             if AutoWebSocketsProtocol is None:
                 msg = "No supported WebSocket library detected. Please use 'pip install uvicorn[standard]', or install 'websockets' or 'wsproto' manually."  # noqa: E501
                 self.logger.warning(msg)
+
             reason = STATUS_PHRASES[400]
             headers = [
                 (b"content-type", b"text/plain; charset=utf-8"),
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -269,7 +269,7 @@ class H11Protocol(asyncio.Protocol):
             from uvicorn.protocols.websockets.auto import AutoWebSocketsProtocol
 
             if AutoWebSocketsProtocol is None:
-                msg = "You have no ws libary installed, either use pip install uvicorn[standard] or install websockets / wsproto depending on the ws library you want to use."  # noqa: E501
+                msg = "No supported WebSocket library detected. Please use 'pip install uvicorn[standard]', or install 'websockets' or 'wsproto' manually."  # noqa: E501
                 self.logger.warning(msg)
             reason = STATUS_PHRASES[400]
             headers = [
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -266,6 +266,11 @@ class H11Protocol(asyncio.Protocol):
         if upgrade_value != b"websocket" or self.ws_protocol_class is None:
             msg = "Unsupported upgrade request."
             self.logger.warning(msg)
+            from uvicorn.protocols.websockets.auto import AutoWebSocketsProtocol
+
+            if AutoWebSocketsProtocol is None:
+                msg = "You have no ws libary installed, either use pip install uvicorn[standard] or install websockets / wsproto depending on the ws library you want to use."  # noqa: E501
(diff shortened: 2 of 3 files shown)


docs/index.md
```diff
@@ -39,7 +39,8 @@ Install using `pip`:
 $ pip install uvicorn
 ```
 
-This will install uvicorn with minimal (pure Python) dependencies.
+This will install uvicorn with minimal (pure Python) dependencies 
+(note: this does not enable websockets support, see below for "activating" it).
```
florimondmanca: I find this note a bit confusing... FWIW, I think the existing prose here is sufficient, and the new warning makes it even easier for users to see that they should use the standard install for WebSocket support.

uvicorn/protocols/websockets/auto.py
```diff
@@ -1,10 +1,18 @@
...
+
 try:
     import websockets  # noqa
 except ImportError:  # pragma: no cover
     try:
         import wsproto  # noqa
     except ImportError:
         AutoWebSocketsProtocol = None
+        logger = logging.getLogger("uvicorn.error")
+        logger.warning(
```
florimondmanca: Doesn't this warning _always_ trigger when using a minimal install?

I believe the default WS mode is "auto", meaning that we import this "auto protocol", which will be None and trigger this warning by default...

I'm actually not sure if/how we can check for "your app is using WebSocket, but dependencies aren't available" before a WebSocket request comes in. 

So what we _can_ do instead is log this warning when the first WebSocket connection request is received. Maybe that means we _do_ need to define the "auto" protocol in this case as a "null protocol" class that logs this warning? I'm not sure how this `None` protocol is used by asyncio.

euri10: yes you nailed it, the ws flag is on auto by default, so `AutoWebSocketsProtocol` will be None, and we will get a warning set here if we try a handshake:
https://github.com/encode/uvicorn/blob/fa914bc8b3af38017f2438770616cab46f600141/uvicorn/protocols/http/h11_impl.py#L267

which sucks.

euri10: > So what we can do instead is log this warning when the first WebSocket connection request is received. Maybe that means we do need to define the "auto" protocol in this case as a "null protocol" class that logs this warning? I'm not sure how this None protocol is used by asyncio.

like in the `config.load()` ? if ws_protocol_class is None > warning ?

euri10: the current scenario is this:
- uvicorn minimal > auto ws > ws=None > auto http > h11 > handshake in h11 > `msg = "Unsupported upgrade request." `

The present PR is preemptively raising a warning:
- uvicorn minimal > auto ws > ws=None > `Warning blablabla ws cant be used`

We could also add this warning  to the existing h11 warning:
- uvicorn minimal > auto ws > ws=None > auto http > h11 > handshake in h11 > `msg = "Unsupported upgrade request. Warning blablabla ws cant be used" `

I just find it more deceptive when you want to use ws, but you'll have a useful info in both cases.

It's also true raising a warning for everyone when they dont want to use ws is probably annoying.

I have no preference, I let you choose ! :)

docs/index.md
euri10: ok I removed it

uvicorn/protocols/websockets/auto.py
euri10: ok after trying to put the warning in the httptools_impl.py I find it weird.
the scenario is very deceptive for a user, and on top of that for a regular user using ws and having a real upgrade issue he'll have the warning, so I prefer to put it here in the import logic

florimondmanca:
@euri10 I think you put it right in the earlier comments:

Between "somewhat deceptive message only when receiving a WS request w/o a WS library" and "raise a message for all minimal users", I think the latter is less acceptable.

Should we update the "Unsupported upgrade request" warning instead? Perhaps do a quick runtime check on libraries and add an extra warning hint?

euri10:
> @euri10 I think you put it right in the earlier comments:
> 
> Between "somewhat deceptive message only when receiving a WS request w/o a WS library" and "raise a message for all minimal users", I think the latter is less acceptable.
> 
> Should we update the "Unsupported upgrade request" warning instead? Perhaps do a quick runtime check on libraries and add an extra warning hint?

I had not thought about importing locally the `AutoWebSocketsProtocol` , it looks like a pretty alternative @florimondmanca

uvicorn/protocols/http/h11_impl.py
```diff
@@ -266,6 +266,11 @@ def handle_upgrade(self, event):
         if upgrade_value != b"websocket" or self.ws_protocol_class is None:
             msg = "Unsupported upgrade request."
             self.logger.warning(msg)
+            from uvicorn.protocols.websockets.auto import AutoWebSocketsProtocol
+
+            if AutoWebSocketsProtocol is None:
+                msg = "You have no ws libary installed, either use pip install uvicorn[standard] or install websockets / wsproto depending on the ws library you want to use."  # noqa: E501
```
florimondmanca: Small suggestion on the copywriting:

```suggestion
                msg = "No supported WebSocket library detected. Please use 'pip install uvicorn[standard]', or install 'websockets' or 'wsproto' manually."  # noqa: E501
```

uvicorn/protocols/http/h11_impl.py
```diff
@@ -266,6 +266,11 @@ def handle_upgrade(self, event):
         if upgrade_value != b"websocket" or self.ws_protocol_class is None:
             msg = "Unsupported upgrade request."
             self.logger.warning(msg)
+            from uvicorn.protocols.websockets.auto import AutoWebSocketsProtocol
```
florimondmanca, line 270: (Nit) Let's add empty lines around this block so it's clearer that it's related but separate from the rest of the error handling code?

florimondmanca (commented):
👍 I like this.

uvicorn/protocols/http/h11_impl.py
euri10: seems like we're gtg :)

florimondmanca (approved):

PR #928, opened 2020-12-29
Use ssl verification in tests

### LINKED ISSUE #927: SSL tests verification
Our ssl tests currently dont verify anything :)

Note for the future: maybe Requests didn't support it, but now with HTTPX I think we could pass `verify=<ca_cert>`, rather than disabling SSL verification.

_Originally posted by @florimondmanca in https://github.com/encode/uvicorn/pull/917#discussion_r549130563_

euri10:
Fixes #927 

begin to love httpx more and more, so easy to work with !

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -55,7 +55,7 @@ def tls_certificate_pem_path(tls_certificate):
 
 
 @pytest.fixture
-def ssl_ctx(tls_certificate):
-    ssl_ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+def tls_ca_ssl_context(tls_certificate):
+    ssl_ctx = ssl.SSLContext()
     tls_certificate.configure_cert(ssl_ctx)
     return ssl_ctx
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,5 @@
+import ssl
+
 import pytest
 import trustme
 from cryptography.hazmat.backends import default_backend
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -50,3 +52,10 @@ def tls_ca_certificate_private_key_encrypted_path(tls_certificate_authority):
 def tls_certificate_pem_path(tls_certificate):
     with tls_certificate.private_key_and_cert_chain_pem.tempfile() as cert_pem:
         yield cert_pem
+
+
+@pytest.fixture
+def ssl_ctx(tls_certificate):
```
florimondmanca: (Nit) What do you think about naming it `tls_ca_ssl_context`, for consistency with other TLS fixtures that start with `tls_ca_`?

tests/conftest.py
```diff
@@ -50,3 +52,10 @@ def tls_ca_certificate_private_key_encrypted_path(tls_certificate_authority):
...
+    ssl_ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
```
florimondmanca: I don't think there's actually a need for specifying anything, is there?

```suggestion
    ssl_ctx = ssl.SSLContext()
```

florimondmanca (commented):
Sweet!

tests/conftest.py
euri10: Indeed ! done

tests/conftest.py
euri10: it seems like it doenst

florimondmanca (approved):
💯

PR #930, opened 2020-12-31
Isolate server started message

florimondmanca:
Somewhat prompted by https://github.com/encode/uvicorn/pull/853#issuecomment-752934649

This is a small refactor that separates the "log a message indicating the server has started" logic from the "actually start listening on sockets" logic.

The motivation is that, according to #863, in the end we'll want to have one "start listening on sockets" implementation for each async library, but we don't want to have to duplicate the "started message" logic everywhere. So, decoupling. :-) The interface between the two is a list of effective sockets (`listeners`), which we use to create the started message.

I think this is somewhat easier to understand, as well.

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -9,6 +9,7 @@ import sys
 import threading
 import time
 from email.utils import formatdate
+from typing import List
 
 import click
 
@@ -110,6 +111,7 @@ class Server:
                     create_protocol, sock=sock, ssl=config.ssl, backlog=config.backlog
                 )
                 self.servers.append(server)
+            listeners = sockets
 
         elif config.fd is not None:
             # Use an existing socket, from a file descriptor.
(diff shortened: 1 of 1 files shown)


euri10:
Would it make sense to refactor those then too ?

https://github.com/encode/uvicorn/blob/34838e90311fea4f754d3fa893785444ebd2aac3/uvicorn/config.py#L376-L389

https://github.com/encode/uvicorn/blob/34838e90311fea4f754d3fa893785444ebd2aac3/uvicorn/server.py#L192-L205

florimondmanca:
@euri10 Not immediately obvious to me. I think it's true that the logging code for multiprocessing somewhat lacks in locality, but I can't think of a way to make it more local right now. I don't think keeping a bit of duplication here is harmful. Also we may have a clearer view once this initial "unblock" PR is merged. :)

euri10 (approved):

PR #935, opened 2021-01-12
fix: h11 only support Response status_code between [200, 600)

Ro0tk1t:
limited by [there](https://github.com/python-hyper/h11/blob/master/h11/_events.py#L221)

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -22,7 +22,7 @@ def _get_status_phrase(status_code):
 
 
 STATUS_PHRASES = {
-    status_code: _get_status_phrase(status_code) for status_code in range(200, 599)
+    status_code: _get_status_phrase(status_code) for status_code in range(200, 600)
 }
 
 HIGH_WATER_LIMIT = 65536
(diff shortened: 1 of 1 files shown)


uvicorn/protocols/http/h11_impl.py
```diff
@@ -22,7 +22,7 @@ def _get_status_phrase(status_code):
 
 
 STATUS_PHRASES = {
-    status_code: _get_status_phrase(status_code) for status_code in range(100, 600)
+    status_code: _get_status_phrase(status_code) for status_code in range(200, 599)
```
ycd: 100 to 200 is okay, but you should not change 600 to 599. 599 means `200-598`.

ycd (changes requested):

uvicorn/protocols/http/h11_impl.py
Ro0tk1t: ok

Kludex:
Thanks for the PR @Ro0tk1t, and @ycd for the review.

But this will give us `KeyError`, so I'll be closing this PR.

PR #942, opened 2021-01-22
Move `Config#load` out of `Server#serve` to enable Eager Asynchronous Construction of App

### LINKED ISSUE #941: Eager Asynchronous Construction of App Interferes with Running Event Loop
Because [`Server#run`](https://github.com/encode/uvicorn/blob/320fd6d975d47257aecde808d602c9a57fb74cba/uvicorn/server.py#L46) calls [`Config#load`](https://github.com/encode/uvicorn/blob/320fd6d975d47257aecde808d602c9a57fb74cba/uvicorn/server.py#L56) within [`loop.run_until_complete(self.serve(...))`](https://github.com/encode/uvicorn/blob/320fd6d975d47257aecde808d602c9a57fb74cba/uvicorn/server.py#L49), a config load that involves asynchronously constructing an app fails because the event loop is already running. Note that this config load will fail only if the app name is provided instead of the app itself. This is because in the latter case, the app had already been asynchronously constructed in its own `run_until_complete` prior to the `run_until_complete` launched by `Server`.

### To reproduce

1) `pip install -e git+https://github.com/encode/uvicorn#egg=uvicorn`
2) Create `main.py`:

```
import asyncio

async def new_app():
        await send({
            'status': 200,
            'headers': [
            ],
        })
        await send({
            'type': 'http.response.body',
            'body': b'Hello, world!',
        })

    return app

app = asyncio.get_event_loop().run_until_complete(new_app())
```
3) Run `uvicorn main:app`.

### Expected behavior
```
INFO: Started server process [xxx]
INFO: Waiting for application startup.
INFO: Application startup complete.
INFO: Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
```
And that a server request would return Hello, world!

### Actual behavior

```
  File "/home/mishac/dev/python/personal/async_init/venv/bin/uvicorn", line 33, in <module>
    sys.exit(load_entry_point('uvicorn', 'console_scripts', 'uvicorn')())
  File "/home/mishac/dev/python/personal/async_init/venv/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/mishac/dev/python/personal/async_init/venv/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/mishac/dev/python/personal/async_init/venv/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mishac/dev/python/personal/async_init/venv/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/mishac/dev/python/personal/async_init/venv/src/uvicorn/uvicorn/main.py", line 362, in main
    run(**kwargs)
  File "/home/mishac/dev/python/personal/async_init/venv/src/uvicorn/uvicorn/main.py", line 386, in run
    server.run()
  File "/home/mishac/dev/python/personal/async_init/venv/src/uvicorn/uvicorn/server.py", line 49, in run
  File "/home/mishac/dev/python/personal/async_init/venv/src/uvicorn/uvicorn/server.py", line 56, in serve
    config.load()
  File "/home/mishac/dev/python/personal/async_init/venv/src/uvicorn/uvicorn/config.py", line 308, in load
    self.loaded_app = import_from_string(self.app)
  File "/home/mishac/dev/python/personal/async_init/venv/src/uvicorn/uvicorn/importer.py", line 20, in import_from_string
    module = importlib.import_module(module_str)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "./main.py", line 19, in <module>
    app = asyncio.get_event_loop().run_until_complete(new_app())
  File "uvloop/loop.pyx", line 1450, in uvloop.loop.Loop.run_until_complete
  File "uvloop/loop.pyx", line 1443, in uvloop.loop.Loop.run_until_complete
  File "uvloop/loop.pyx", line 1351, in uvloop.loop.Loop.run_forever
  File "uvloop/loop.pyx", line 480, in uvloop.loop.Loop._run
RuntimeError: this event loop is already running.
```
Event loop started running before app could be asynchronously constructed.

### Environment
Ubuntu 20.04
CPython 3.8.5
Uvicorn 0.13.3

### Additional context
Replacing step 1 with `pip install -e git+https://github.com/MarioIshac/uvicorn@feat-async-construction#egg=uvicorn` will achieve the expected behavior.

The particular benefit I get from having the app asynchronously & eagerly constructed instead of lazily constructed within `async def app(...)` is that I can take advantage of gunicorn preloading. I can also have the server fail-fast instead of on-first-request in the case that the asynchronous construction fails (for example, database connection pool could not be created).

> [!IMPORTANT]

<a href="https://polar.sh/encode/uvicorn/issues/941">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/941/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

MarioIshac:
Fixes #941 by identifying the 3 call sites (test utility, gunicorn worker, uvicorn runner itself) `Server#serve` is used at and moving `Config#load` to those sites.

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -44,7 +44,12 @@ class Server:
         self.last_notified = 0
 
     def run(self, sockets=None):
-        self.config.setup_event_loop()
+        config = self.config
+        config.setup_event_loop()
+
+        if not config.loaded:
+            config.load()
+
         loop = asyncio.get_event_loop()
         loop.run_until_complete(self.serve(sockets=sockets))
 
@@ -52,8 +57,6 @@ class Server:
         process_id = os.getpid()
 
         config = self.config
-        if not config.loaded:
-            config.load()
 
         self.lifespan = config.lifespan_class(config)
 
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -11,6 +11,10 @@ from uvicorn import Config, Server
 @asynccontextmanager
 async def run_server(config: Config, sockets=None):
     server = Server(config=config)
+
+    if not config.loaded:
+        config.load()
+
     cancel_handle = asyncio.ensure_future(server.serve(sockets=sockets))
     await asyncio.sleep(0.1)
     try:
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -70,7 +70,12 @@ class UvicornWorker(Worker):
             signal.signal(s, signal.SIG_DFL)
 
     def run(self):
-        self.config.app = self.wsgi
+        config = self.config
+        config.app = self.wsgi
+
+        if not config.loaded:
+            config.load()
+
         server = Server(config=self.config)
         loop = asyncio.get_event_loop()
         loop.run_until_complete(server.serve(sockets=self.sockets))


florimondmanca:
@MarioIshac Thanks for this PR. I think this makes sense, but just need a bit of careful thinking.

How does this behave with Uvicorn multiprocessing? Eg `--workers 3`, and `--reload`? Do we need to change anything else there?

Edit: okay, answering my own question, it seems we're all good. Here's how we spawn for each case (normal, reload, multiple workers):

https://github.com/encode/uvicorn/blob/c6b80fce49e9ad53f514262f4c1eb7f9a6798620/uvicorn/main.py#L377-L386

In each case we target `server.run()`, so the scope of when we call `config.load()` just gets moved from `server.serve()` to `server.run()`.

MarioIshac:
@florimondmanca I'll recheck the interaction between this PR and uvicorn multiprocessing to make sure I didn't miss anything. I also will take a look at how this PR behaves with apps imported as the object itself (`uvicorn.run(main.app`) instead of by name (`uvicorn.run("main:app")`), since after our discussion I realized that this might affect the PR. I've got to go to work in a bit, but I'll have an update later today.

uvicorn/server.py
```diff
@@ -44,16 +44,19 @@ def __init__(self, config):
         self.last_notified = 0
 
     def run(self, sockets=None):
-        self.config.setup_event_loop()
+        config = self.config
+        config.setup_event_loop()
+
+        if not config.loaded:
+            config.load()
```
florimondmanca, line 51: Let's add a comment about why we're doing this here, perhaps?

```suggestion
        # Setup and load app before entering the async environment, so it's possible to
        # run any async setup on our side or in the app module.
        # See: https://github.com/encode/uvicorn/issues/941
        config = self.config
        config.setup_event_loop()

        if not config.loaded:
            config.load()
```

florimondmanca (commented):
I think this is looking good.

@euri10, any opinion by chance?

@MarioIshac Would it be possible to add a test for this, eg by passing an app factory (see associated existing tests) that runs something on the event loop?

euri10:
> I think this is looking good.
> 
> @euri10, any opinion by chance?
> 
> @MarioIshac Would it be possible to add a test for this, eg by passing an app factory (see associated existing tests) that runs something on the event loop?

yep, a test for that is a good idea

MarioIshac:
I am working on writing the test, but it looks like `@pytest.mark.asyncio` already runs the test function in its event loop, so constructing the app within the test function will fail there. But some test functions (such as within `test_lifespan.py`) are managing the event loop themselves, so I'll see if I can do something like that.

euri10:
I have kind of the feeling this could be resurrected for 3.10

graingert:
@euri10 I don't think this makes sense anymore, given `asyncio.get_event_loop().run_until_complete(...)` is no longer supported in python3.10+

PR #943, opened 2021-01-23
Document the default value of 1 for workers (#940)

### LINKED ISSUE #940: Docs: workers = 1 when WEB_CONCURRENCY is not set
Hello, thanks for this great ASGI server 🙂 

I have an application that must use only one worker (it's not thread-safe), and so I wanted to confirm that, by default, uvicorn only uses one worker. But I had to read the source code

https://github.com/encode/uvicorn/blob/afb2d565c8dae859bcef4c76b3c6dc3f3077314d/uvicorn/config.py#L177

...to confirm that when `WEB_CONCURRENCY` is not defined, and `workers` is not specified, 1 is used.

Would you accept a PR adding that information to the docs (there are 4-5 places where this can be added)?

pawamoy:
Fixes #940.

I tried to keep things simple to avoid language mistakes.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -45,7 +45,7 @@ Options:
 
   --workers INTEGER               Number of worker processes. Defaults to the
                                   $WEB_CONCURRENCY environment variable if
-                                  available. Not valid with --reload.
+                                  available, or 1. Not valid with --reload.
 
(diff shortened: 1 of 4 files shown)


euri10 (approved):
lgtm, thanks a lot !

PR #947, opened 2021-02-03
Implement gunicorn access log format

### LINKED ISSUE #527: Honor gunicorn access log format

### LINKED ISSUE #773: Ability to log the HTTP referer
Unless I've missed something, it looks like it isn't currently possible to configure Uvicorn to output the HTTP Referer header in the access logs:

https://github.com/encode/uvicorn/blob/e77e59612ecae4ac10f9be18f18c47432be7909a/uvicorn/logging.py#L109-L132

This would be really useful for me because I'm trying to figure out where traffic to my sites comes from without running Google Analytics!

immerrr:
This PR a very naïve and straightforward implementation of gunicorn access log format, which is currently ignored by uvicorn (see #527).

One notable omission is the WSGI variables: they should be rather straightforward to add, but I think it is important to agree on the approach first. And the approach goes as follows:

It puts the gunicorn log object all the way through to the `RequestResponseCycle`, and it unrolls `gunicorn_log.access` function

https://github.com/benoitc/gunicorn/blob/master/gunicorn/glogging.py#L331-L351

so that it is not necessary to recreate all the gunicorn primitives, like `gunicorn.http.wsgi.Response` and `gunicorn.http.message.Request`. 

The timing is implemented via `time.monotonic` so that it works well w.r.t. DST transitions and leap seconds. The timing values are added to the asgi scope. And obviously, to be able to measure the timing, the logging had to be moved to the end of response processing. 

~~Default uvicorn logging is kept intact to minimize the scope of this PR, and is only disabled when `gunicorn_log` is passed by gunicorn worker classes to avoid duplication.~~

`GunicornSafeAtoms` does several things:

- it provides the values for gunicorn log formatter in the most performant way I could imagine
- it aggregates response body size from ASGI messages
- and it does what [gunicorn.glogging.SafeAtoms](https://github.com/benoitc/gunicorn/blob/master/gunicorn/glogging.py#L97-L117]) does to provide fallbacks for missing values

Edit by @Kludex : closes #773, closes #527

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -46,7 +46,27 @@ By default Uvicorn uses simple changes detection strategy that compares python f
 * `--log-level <str>` - Set the log level. **Options:** *'critical', 'error', 'warning', 'info', 'debug', 'trace'.* **Default:** *'info'*.
 * `--no-access-log` - Disable access log only, without changing log level.
 * `--use-colors / --no-use-colors` - Enable / disable colorized formatting of the log records, in case this is not set it will be auto-detected. This option is ignored if the `--log-config` CLI option is used.
-
+* `--access-log-format` - Customized message for access log. Will override log formatter for `uvicorn.access` logger. Format string that supports the following placeholders:
+    * `%(h)s` - remote IP address
+    * `%(l)s` - prints `-`
+    * `%(u)s` - user name, currently not supported, prints `-`
+    * `%(t)s` - request timestamp (in Apache Common Log Format `[10/Oct/2000:13:55:36 -0700]`)
+    * `%(r)s` - status line (method, path & http version, e.g.)
+    * `%(m)s` - request method
+    * `%(U)s` - request path
+    * `%(q)s` - request query string
+    * `%(H)s` - request protocol
+    * `%(s)s` - response status code
+    * `%(B)s` - response body length (if body was empty, prints `0`)
+    * `%(b)s` - response body length (if body was empty, prints `-`)
+    * `%(f)s` - request referer
+    * `%(a)s` - request user agent
+    * `%(T)s` - request+response time (in seconds, integer number)
+    * `%(D)s` - request+response time (in milliseconds, integer number)
+    * `%(L)s` - request+response time (in seconds, decimal number)
+    * `%(p)s` - process id (pid) of request handler
+    * `%({...}i)s` - value of arbitrary request header, e.g. `%({Accept-Encoding}i)s`
+    * `%({...}o)s` - value of arbitrary response header, e.g. `%({Content-Encoding}o)s`
 
 ## Implementation
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,6 +23,8 @@ coverage==5.5
 coverage-conditional-plugin==0.4.0
 httpx==1.0.0b0
 pytest-asyncio==0.15.1
+async_generator==1.10 ; python_version < '3.7'
+gunicorn==20.1.0
 
 
 # Documentation
--- a/tests/middleware/test_logging.py
+++ b/tests/middleware/test_logging.py
@@ -154,3 +154,43 @@ async def test_unknown_status_code(caplog):
             if record.name == "uvicorn.access"
         ]
         assert '"GET / HTTP/1.1" 599' in messages.pop()
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "access_log_format,expected_output",
+    [
+        # TODO: add more tests
+        ("access: %(h)s", "access: 127.0.0.1"),
+        ('access: "%({test-request-header}i)s"', 'access: "request-header-val"'),
+        ('access: "%({test-response-header}o)s"', 'access: "response-header-val"'),
+    ],
+)
+async def test_access_log_format(access_log_format, expected_output, caplog):
+    async def app(scope, receive, send):  # pragma: no cover
+        assert scope["type"] == "http"
+        await send(
+            {
+                "type": "http.response.start",
+                "status": 204,
+                "headers": [(b"test-response-header", b"response-header-val")],
+            }
+        )
+        await send({"type": "http.response.body", "body": b"", "more_body": False})
+
+    config = Config(app=app, access_log_format=access_log_format)
+    with caplog_for_logger(caplog, "uvicorn.access"):
+        async with run_server(config):
+            async with httpx.AsyncClient() as client:
+                response = await client.get(
+                    "http://127.0.0.1:8000",
+                    headers={"test-request-header": "request-header-val"},
+                )
+        assert response.status_code == 204
+
+    access_log_messages = [
+        record.message for record in caplog.records if "uvicorn.access" in record.name
+    ]
+
+    assert len(access_log_messages) == 1
+    assert access_log_messages[0] == expected_output
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -388,6 +388,10 @@ class Config:
                     self.log_config["formatters"]["access"][
                         "use_colors"
                     ] = self.use_colors
+                if self.access_log_format:
+                    self.log_config["formatters"]["access"][
+                        "fmt"
+                    ] = "%(levelprefix)s %(message)s"
                 logging.config.dictConfig(self.log_config)
             elif self.log_config.endswith(".json"):
                 with open(self.log_config) as file:
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -240,6 +240,7 @@ class Config:
         ssl_ciphers: str = "TLSv1",
         headers: Optional[List[List[str]]] = None,
         factory: bool = False,
+        gunicorn_log=None,
     ):
         self.app = app
         self.host = host
@@ -283,6 +284,8 @@ class Config:
         self.headers: List[List[str]] = headers or []
         self.encoded_headers: List[Tuple[bytes, bytes]] = []
         self.factory = factory
+
+        self.gunicorn_log = gunicorn_log
 
         self.loaded = False
         self.configure_logging()
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -211,6 +211,7 @@ class Config:
         log_config: Optional[Union[dict, str]] = LOGGING_CONFIG,
         log_level: Optional[Union[str, int]] = None,
         access_log: bool = True,
+        access_log_format: Optional[str] = None,
         use_colors: Optional[bool] = None,
         interface: InterfaceType = "auto",
         debug: bool = False,
@@ -240,7 +241,6 @@ class Config:
         ssl_ciphers: str = "TLSv1",
         headers: Optional[List[List[str]]] = None,
         factory: bool = False,
-        gunicorn_log=None,
     ):
         self.app = app
         self.host = host
@@ -285,7 +285,7 @@ class Config:
         self.encoded_headers: List[Tuple[bytes, bytes]] = []
         self.factory = factory
 
-        self.gunicorn_log = gunicorn_log
+        self.access_log_format = access_log_format
 
         self.loaded = False
         self.configure_logging()
--- a/uvicorn/logging.py
+++ b/uvicorn/logging.py
@@ -2,12 +2,16 @@ import http
 import logging
 import sys
 import time
+import typing
 from collections import abc
 from copy import copy
 from os import getpid
-from typing import Optional
+from typing import Any, Callable, Dict, Iterator, Optional
 
 import click
+
+if typing.TYPE_CHECKING:
+    import uvicorn.protocols.utils
 
 TRACE_LOG_LEVEL = 5
 
@@ -130,17 +134,19 @@ class GunicornSafeAtoms(abc.Mapping):
     - escapes double quotes found in atom strings
     """
 
-    def __init__(self, scope, timing):
+    def __init__(
+        self, scope: dict, timing: "uvicorn.protocols.utils.RequestResponseTiming"
+    ):
         self.scope = scope
         self.timing = timing
         self.status_code = None
-        self.response_headers = {}
-        self.response_length = 0
+        self.response_headers: Dict[str, str] = {}
+        self._response_length = 0
 
-        self._request_headers = None
+        self._request_headers: Optional[Dict[str, str]] = None
 
     @property
-    def request_headers(self):
+    def request_headers(self) -> Dict[str, str]:
         if self._request_headers is None:
             self._request_headers = {
                 k.decode("ascii"): v.decode("ascii") for k, v in self.scope["headers"]
@@ -148,30 +154,39 @@ class GunicornSafeAtoms(abc.Mapping):
         return self._request_headers
 
     @property
-    def duration(self):
+    def duration(self) -> float:
         return self.timing.total_duration_seconds()
 
-    def on_asgi_message(self, message):
+    def on_asgi_message(self, message: Dict[str, Any]) -> None:
         if message["type"] == "http.response.start":
             self.status_code = message["status"]
             self.response_headers = {
                 k.decode("ascii"): v.decode("ascii") for k, v in message["headers"]
             }
         elif message["type"] == "http.response.body":
-            self.response_length += len(message.get("body", ""))
+            self._response_length += len(message.get("body", ""))
 
-    def _request_header(self, key):
+    def _request_header(self, key: str) -> Optional[str]:
         return self.request_headers.get(key.lower())
 
-    def _response_header(self, key):
+    def _response_header(self, key: str) -> Optional[str]:
         return self.response_headers.get(key.lower())
 
-    def _wsgi_environ_variable(self, key):
+    def _wsgi_environ_variable(self, key: str) -> None:
         # FIXME: provide fallbacks to access WSGI environ (at least the
         # required variables).
-        return None
+        raise NotImplementedError
 
-    def __getitem__(self, key):
+    @classmethod
+    def _log_format_atom(cls, val: Optional[str]) -> str:
+        if val is None:
+            return "-"
+        if isinstance(val, str):
+            return val.replace('"', '\\"')
+        return val
+
+    def __getitem__(self, key: str) -> str:
+        retval: Optional[str]
         if key in self.HANDLERS:
             retval = self.HANDLERS[key](self)
         elif key.startswith("{"):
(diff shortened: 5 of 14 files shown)


immerrr:
ref #773, ref #527, ref #389

immerrr:
@euri10 is there anything i can do to help get this in? or at least to help to get the discussion started, as i would assume there's a decision to be taken w.r.t. the approach.

euri10:
> @euri10 is there anything i can do to help get this in? or at least to help to get the discussion started, as i would assume there's a decision to be taken w.r.t. the approach.

back here from the issue discussion, I quickly looked on the PR, I can't pronounce on the gunicorn side of things, however what is pretty clear to me reading the PR is that the implementation is adding keys to the scope that are not in the asgi specification. (https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope).

I tried to dig a little in asgiref and didn't find a definitive answer as to whether or not it's "allowed", I would tend to say it is not, 
see https://github.com/django/asgiref/issues/178 and also https://github.com/django/asgiref/issues/112 for interesting points.

It would be better to have an implementation that does not violate the spec, if not possible this would need a discussion about that beforehand, I cant remember where but I think this was discussed before and frowned upon, maybe @tomchristie remembers.

chbndrhnns:
From what I see, these are the added keys:
- request_start_time
- request_end_time

Would the [extensions approach](https://asgi.readthedocs.io/en/latest/specs/main.html#extensions) from the spec work here?

> There are times when protocol servers may want to provide server-specific extensions outside of a core ASGI protocol specification, or when a change to a specification is being trialled before being rolled in.
For this use case, we define a common pattern for extensions - named additions to a protocol specification that are optional but that, if provided by the server and understood by the application, can be used to get more functionality.
This is achieved via an extensions entry in the scope dictionary, which is itself a dict. Extensions have a Unicode string name that is agreed upon between servers and applications.

euri10:
> Would the [extensions approach](https://asgi.readthedocs.io/en/latest/specs/main.html#extensions) from the spec work here?

I think yes that would be the most sensible thing to do to move things forward.

immerrr:
Heyy, now we're talking (literally), thanks!

I'll update the PR to use the extensions approach, or if you see another way to store the start/end timestamps within the scope of the request, i'm open to other ideas :+1:

immerrr:
@euri10 i have updated the code to use `scope['extensions']['uvicorn_request_duration']` to store the timestamps, but like i said previously, if there's a better place to put it, just let me know.

euri10:
you can use `./scripts/lint` if you have linter issues, not sure if you saw it

Kludex:
I really want this feature, but I'm not sure this is the way to go.

Some points on my mind:
- I don't think we should have any reference to `gunicorn` on our source code besides the worker.
- I think `uvicorn` should have its way to customize the logger format, which for convenience it's a good idea to match `gunicorn`'s way.
- I don't feel comfortable adding an extension on the ASGI scope just for the time, I think we can achieve that with a single variable, even if that variable needs to be sent to another object (cycle). :thinking:

Those are my concerns at first glance. What do you think? @euri10 @immerrr

immerrr:
I agree with the concerns listed above, as i mentioned earlier, it was done the way it was done to minimize the scope of change and to let the uvicorn team decide the fate of uvicorn's logging and formatting at your convenience. 

If I'm reading this correctly, the requested changes are:
- move tracking of request-response timing to a variable
- turn the gunicorn logging format handling into uvicorn's native logging formatting

If those were to happen, would it make this PR "the way to go"?

Kludex:
> If I'm reading this correctly, the requested changes are:
> 
> - move tracking of request-response timing to a variable
> - turn the gunicorn logging format handling into uvicorn's native logging formatting
>
> If those were to happen, would it make this PR "the way to go"?

Yeah, but also adding the `--access-log-format` on the CLI command (just being explicit here).

Besides if we can get the answer to "is it possible to read the `gunicorn`s `--access-log-format` in the `UvicornWorker`?" it would be cool.

EDIT: For the last question, maybe `self.cfg.access_log_format`? Not sure...

immerrr:
Not sure if i understand the last question correctly, but i think the format is available as

`self.log.cfg.access_log_format`

On the following line:
https://github.com/encode/uvicorn/pull/947/files#diff-7c8a9d71c98f0ed83c32a88a1b7e777ef85776c98869045d24c74e3f7714494aR41

And I am making use of it in this PR.

Kludex:
Ah, yeah! Perfect.

So, in summary:
- Store request-response timing in a variable (being an attribute of `RequestResponseCycle`?).
- As you said: "turn the `gunicorn` logging format handling into `uvicorn`'s native logging formatting". This is for convenience, my point is that we should have our formatting rules.
- Create the `--access-log-format` on `uvicorn`, and on the `UvicornWorker` send the `"access_log_format": self.log.cfg.access_log_format` instead of `"gunicorn_log": self.log`. In case we decide to not match the rules (exactly the same), we can create a converter on the `UvicornWorker` that will aim to match the `gunicorn` with `uvicorn` rules. But with this idea present, we can make use of `--access-log-format` either from `uvicorn` or `gunicorn` on `uvicorn`.

Does it make sense? :thinking:

immerrr:
It can be made to work. 

It's been a while since I wrote this, but i think the default uvicorn logging uses colours, whereas gunicorn doesn't support that, so we either need to disable colouring (break backward compat), or agree that colouring is not possible with custom log formats, or start bending gunicorn formatting features right away :)

Kludex:
> It's been a while since I wrote this, but i think the default uvicorn logging uses colours, whereas gunicorn doesn't support that, so we either need to disable colouring (break backward compat), or agree that colouring is not possible with custom log formats, or start bending gunicorn formatting features right away :)


@euri10 thoughts?

immerrr:
Just started implementing the requested changes, and ran into another point of contention/agreement. The existing uvicorn logging is done when response is started, whereas gunicorn logging is done when the response is ended (to be able to provide all response headers and total request-response duration time).

Do you want to move the default uvicorn logging to the end of the cycle? Or perhaps do the access log before or after response depending on whether or not `access-log-format` was passed?

This was one more question I was trying to avoid with this PR by keeping the gunicorn logging format gunicorn-specific :)

Kludex:
> Do you want to move the default uvicorn logging to the end of the cycle? Or perhaps do the access log before or after response depending on whether or not access-log-format was passed?

IMHO, we can move it. I also think that makes more sense to move it. But maybe we should take a look at how `hypercorn` does it or `gunicorn` itself with their main workers :thinking:

immerrr:
~~Oh, and another thing that should work right now and will not with the new suggested approach is the handling of gunicorn's `--access-logfile`. Current implementation takes the pre-configured logger object from gunicorn with all its handlers~~

https://github.com/encode/uvicorn/blob/4e7da5495d9713dcfb04d96ef82f828319e29e84/uvicorn/protocols/http/httptools_impl.py#L534-L536

~~And if the new version only uses the access log format, this behaviour will have to be reproduced later on.~~

EDIT: nvm, log file and levels are copied in [workers.py](https://github.com/encode/uvicorn/blob/f29f09e8b287c386fb8110e901d7fa9d68b55399/uvicorn/workers.py#L23-L31)

immerrr:
I have added a minimal test to exercise gunicorn worker logging, because it is true that previously there were no tests for this.

immerrr:
I have updated the branch according to the feedback, it's almost there:
- the old default logging format is kept as is
- the `--access-log-format` parameter still needs to be added to uvicorn itself

immerrr:
I have rebased to master and added `--access-log-format`, and looks like the test are passing:
https://github.com/immerrr/uvicorn/actions/runs/957952813

I'm not sure I like the result: now it looks like there are two ways to configure access log format, in the logging dict, that depends on the custom formatter class and a very specific set of parameters passed into the logging method, and then with the new command line parameter. Not sure, what's your preference on resolving this, I personally don't have a strong opinion on this.

@Kludex @euri10, I have enabled "allow edits from maintainers", so  if you agree that it is going to be more efficient than the review-comment-fix-review-comment-fix... loop, feel free to tweak this to your liking yourselves.

uvicorn/protocols/http/h11_impl.py
```diff
@@ -477,6 +482,27 @@ async def send(self, message):
...
+                            self.access_log_fields.status_code,
+                        )
+                    else:
+                        try:
+                            self.access_logger.info(
+                                self.access_log_format,
+                                self.access_log_fields,
+                            )
+                        except:  # noqa
+                            self.logger.error(traceback.format_exc())
```
Kludex, line 495: When can this happen? Only when the format specified is not valid?

I guess we need to check what/how `gunicorn` and `hypercorn` behaves in cases like this? :thinking:

uvicorn/main.py
```diff
@@ -177,6 +177,12 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     default=True,
     help="Enable/Disable access log.",
 )
+@click.option(
+    "--access-log-format",
+    type=str,
+    default=None,
+    help="Access log format (supports same fields as gunicorn).",
```
Kludex: I don't think we should reference `gunicorn` on the CLI. We can have: "Log format for the access logger, check documentation for details", or something shorter?

Kludex (commented):
> I'm not sure I like the result: now it looks like there are two ways to configure access log format, in the logging dict, that depends on the custom formatter class and a very specific set of parameters passed into the logging method, and then with the new command line parameter. Not sure, what's your preference on resolving this, I personally don't have a strong opinion on this.

I'm going to have a close look soon.

Just for the record, we also need to update the documentation. On the documentation we should mention that it can be used with `gunicorn`.

uvicorn/protocols/http/h11_impl.py
immerrr: This is exactly what gunicorn does:

https://github.com/benoitc/gunicorn/blob/cf55d2cec277f220ebd605989ce78ad1bb553c46/gunicorn/glogging.py#L348-L351

immerrr: Re: when can it happen: it means a failure in log formatter or log handler, e.g. when access log is written to a file that consumes all space in the storage, or when the custom formatter specified in logging config for `uvicorn.access` logger is buggy.

Kludex: It's fine then. Thanks!

Kludex:
I'm going to take a look at this later today, but I also want to have your opinion @euri10 .

Here you have hypercorn implementation: https://gitlab.com/pgjones/hypercorn/-/blob/master/src/hypercorn/logging.py
This one is from gunicorn: https://github.com/benoitc/gunicorn/blob/master/gunicorn/glogging.py

Maybe we want to use a simple dictionary (same as both implementations above) instead of the decorator idea around the _atoms_? Looks simpler and less line consuming? :thinking:

uvicorn/main.py
immerrr: Fixed!

immerrr:
@Kludex @euri10 please, let me know if I can help somehow to push this forward

drderuiter:
@immerrr I have a suggestion to make it possible to pass custom fields to the access logger. Specifically, after #859 it became impossible to access any fields that one might have been storing on the scope.

We could achieve this by making it possible to configure which class is assigned to `self.access_log_fields` in the `__init__` method of `RequestResponseCycle`:
```python
class RequestResponseCycle:
    def __init__(..., access_log_fields: Type[AccessLogFields], ...):
        self.access_log_fields = access_log_fields(self.scope, self.timing)
        ...
```
We would then read that class from the config, much like the `access_log_format`.

A user could then inherit from the `AccessLogFields` to pass whatever fields they like to the access logger, including information stored on the scope:
```python
class CustomAccessLogFields(AccessLogFields):
    @_register_handler("my_field")
    def my_field(self) -> Optional[str]:
        return self.scope["my_field"]
```

Curious to hear your thoughts!

immerrr:
> Curious to hear your thoughts!

To me it sounds like a nice and a straightforward solution. I would probably try to keep this PR focused on getting the gunicorn log formatter in first, and once that's done see if there is consensus on the way to extend the basic functionality with the core team (which I am not a part of).

immerrr:
@Kludex @euri10 please, let me know if I can help somehow to push this forward

immerrr:
> Just for the record, we also need to update the documentation. 

I have rebased the PR to latest master and taken a stab at the documentation. Please, let me know WDYT.

immerrr:
> I have rebased the PR to latest master and taken a stab at the documentation.

Just updated the PR again to resolve merge conflicts. @Kludex @euri10 would be great if you had a look when you have a chance.

requirements.txt
```diff
@@ -23,6 +23,8 @@ coverage==5.5
 coverage-conditional-plugin==0.4.0
 httpx==1.0.0b0
 pytest-asyncio==0.15.1
+async_generator==1.10 ; python_version < '3.7'
+gunicorn==20.1.0
```
br3ndonland, line 27: https://github.com/encode/uvicorn/pull/947#issuecomment-860077322

> I don't think we should have any reference to `gunicorn` on our source code besides the worker.

This comment applies to both the source code and tests. AFAIK, the Gunicorn worker in `uvicorn/workers.py` is not tested. There have been some attempts to add testing, but they were reverted.

- https://github.com/encode/uvicorn/issues/102#issuecomment-471904183
- https://github.com/encode/uvicorn/pull/631
- https://github.com/encode/uvicorn/pull/658

I would suggest removing Gunicorn from `requirements.txt`, removing `test_gunicorn_worker.py`, and not depending on Gunicorn directly in tests at this time. Log message format can be tested on `uvicorn.access` directly. This might simplify the PR and help things move forward.

uvicorn/protocols/http/h11_impl.py
```diff
@@ -357,6 +364,10 @@ def __init__(
         # Response state
         self.response_started = False
         self.response_complete = False
+        self.timing = RequestResponseTiming()
+
+        # For logging
+        self.access_log_fields = AccessLogFields(self.scope, self.timing)
```
br3ndonland, line 370: Comment on how the `AccessLogFields` class is used, and https://github.com/encode/uvicorn/pull/947#issuecomment-860077322

> I think `uvicorn` should have its way to customize the logger format, which for convenience it's a good idea to match `gunicorn`'s way.

I agree we should be able to customize the logger format, but I'm not sure we should do it in the same way as Gunicorn does. I think the way Uvicorn provides log formatters currently is actually preferable, because the `logging.Formatter` classes can be used to format other loggers.

<details><summary>For example, I can use <code>uvicorn.logging.DefaultFormatter</code> to format messages on the root logger like this <em>(expand)</em>.</summary>

```py
#!/usr/bin/env python3
# logging_example.py
import logging
import logging.config

LOG_LEVEL = "INFO"
LOGGING_CONFIG = {
    "version": 1,
    "disable_existing_loggers": True,
    "formatters": {
        "uvicorn": {
            "()": "uvicorn.logging.DefaultFormatter",
            "format": "%(levelprefix)s %(message)s",
            "use_colors": False,
        },
    },
    "handlers": {
        "default": {
            "class": "logging.StreamHandler",
            "formatter": "uvicorn",
            "level": LOG_LEVEL,
            "stream": "ext://sys.stdout",
        }
    },
    "root": {"handlers": ["default"], "level": LOG_LEVEL},
}


if __name__ == "__main__":
    logging.config.dictConfig(LOGGING_CONFIG)
    logger = logging.getLogger()
    logger.info("Hello, World")

```

```sh
❯ python -m venv venv
❯ . venv/bin/activate
venv ❯ python -m pip install uvicorn
venv ❯ python logging_example.py
INFO:     Hello, World
```

</details>

I would suggest that we add the Gunicorn formatting capability to the one of the Uvicorn formatter classes. Is there some way we could do that? Maybe we allow the Gunicorn-style format string to be passed in to the formatter `fmt` arg, then handle it when instantiating the formatter, instead of inside `logger.info` calls? This is what we do now with the non-[standard](https://docs.python.org/3/library/logging.html#logrecord-attributes) `levelprefix` field, for example.

uvicorn/config.py
```diff
@@ -385,6 +388,10 @@ def configure_logging(self) -> None:
                     self.log_config["formatters"]["access"][
                         "use_colors"
                     ] = self.use_colors
+                if self.access_log_format:
+                    self.log_config["formatters"]["access"][
+                        "fmt"
+                    ] = "%(levelprefix)s %(message)s"
```
br3ndonland, line 394: Shouldn't this be like:

```py
if self.access_log_format:
    self.log_config["formatters"]["access"][
        "fmt"
    ] = self.access_log_format

```

docs/settings.md
```diff
@@ -46,7 +46,27 @@ By default Uvicorn uses simple changes detection strategy that compares python f
 * `--log-level <str>` - Set the log level. **Options:** *'critical', 'error', 'warning', 'info', 'debug', 'trace'.* **Default:** *'info'*.
 * `--no-access-log` - Disable access log only, without changing log level.
 * `--use-colors / --no-use-colors` - Enable / disable colorized formatting of the log records, in case this is not set it will be auto-detected. This option is ignored if the `--log-config` CLI option is used.
-
+* `--access-log-format` - Customized message for access log. Will override log formatter for `uvicorn.access` logger. Format string that supports the following placeholders:
```
br3ndonland, line 49: Thank you for adding these docs.

uvicorn/logging.py
```diff
@@ -115,3 +124,194 @@ def formatMessage(self, record: logging.LogRecord) -> str:
             }
         )
         return super().formatMessage(recordcopy)
+
+
+class AccessLogFields(abc.Mapping):  # pragma: no cover
```
br3ndonland, line 129: Shouldn't we cover this class?

tests/middleware/test_logging.py
```diff
@@ -154,3 +154,43 @@ async def app(scope, receive, send):
             if record.name == "uvicorn.access"
         ]
         assert '"GET / HTTP/1.1" 599' in messages.pop()
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "access_log_format,expected_output",
+    [
+        # TODO: add more tests
```
br3ndonland, line 163: I agree with the `# TODO: add more tests` comment. If we want to support all the formatting options independently of Gunicorn, we should test every one, and probably many combinations. See [Gunicorn `tests/test_logger.py`](https://github.com/benoitc/gunicorn/blob/933b210f2eff943f4cd9502c676f3d93c82f171c/tests/test_logger.py) for some ideas.

br3ndonland (changes requested):
@immerrr thank you for taking on this challenging problem, and for your patience while the maintainers consider these changes.

I can add to the review here, to help this PR move forward.

jaronoff97:
@br3ndonland any updates on this PR? This would be really helpful for improved o11y in gunicorn with uvicorn workers

br3ndonland:
@jaronoff97 there have been a couple of attempts (this PR and #1191), but nothing implemented directly in Uvicorn yet.

In the short-term, you could use [`asgi-logger`](https://github.com/Kludex/asgi-logger) from @Kludex.

Longer-term, we're looking at implementing the Gunicorn access log format in the [Uvicorn 0.20 milestone](https://github.com/encode/uvicorn/milestone/8).

@Kludex is already doing a lot of work for our Encode projects, so I can pick up some of the work on this if needed.

Kludex:
Yeah, I'd appreciate help. I need to organize my ideas a bit about this. :pray:

staubina:
Has this PR and potential fix died off? Does anyone have any suggestions here using just Uvicorn?

immerrr:
I have been reluctant to update the PR, as it was repeatedly stated that this was not a priority, and the latest review requested sizeable changes, which I have been unable to expend for a work that was going to spend the next X months in the "low priority" bucket.

If my understanding is incorrect or things have changed and now there is tangible interest in this,  I can give the PR a push.

staubina:
I appreciate the effort you have put in here. I was looking to resolve an issue with passing on more detailed logs with access information in a JSON format and stumbled into this PR. Still looking to see if this is something I will need or if I just need to expand my Custom Format approach and add a middleware like [asgi-logger](https://github.com/Kludex/asgi-logger) to get more details about the state.

The benefit of this approach for users would be not need a separate package to format the logs. 

I think in my case, I might still need the Middleware to get that data and pass it on to the JSON log formatter.

tl;dr

I think this is not what I need to resolve my issue now that I have reviewed some of the other threads linked, like: https://github.com/encode/uvicorn/pull/859

br3ndonland:
> I have been reluctant to update the PR, as it was repeatedly stated that this was not a priority, and the latest review requested sizeable changes, which I have been unable to expend for a work that was going to spend the next X months in the "low priority" bucket.
> 
> If my understanding is incorrect or things have changed and now there is tangible interest in this, I can give the PR a push.

No problem @immerrr, that's understandable! I'm in the same boat myself - just haven't really been able to put the time in. It's one of those things that maintainers say isn't a priority, but then people keep asking about it, so... 🤷 

Honestly, as a maintainer myself, I'm not always sure what the priorities are. Most people just work on stuff they need, or on highly-requested features. The Gunicorn piece has always been particularly confusing to me, because the Uvicorn docs recommend deploying with Gunicorn, but we don't test the Gunicorn worker, and we don't integrate with Gunicorn's access log format, so we don't actually provide much support for Gunicorn. I built another project, [inboard](https://github.com/br3ndonland/inboard), that has a more thorough integration with Gunicorn, and lots of Gunicorn testing with and without Docker, so that's how I approached it. I might try implementing the Gunicorn access log format in inboard, then I could come back here with some more ideas.

uvicorn/config.py
immerrr: Not quite: in Python logging it happens in two stages. 

E.g. for `logger.info('message %(hello)s %(world)s', {'hello': 1, 'world': 2})`: 

- first `LogRecord.getMessage` will format the message with the args, i.e. expand `hello` and `world` and store it as `LogRecord.message`
```python
class LogRecord:
    def getMessage(self):
        msg = str(self.msg)
        if self.args:
            msg = msg % self.args
        return msg

    def format(self, record):
        record.message = record.getMessage()
        ...

``` 
- then LogRecord fields are formatted using the corresponding Formatter's `fmt`, that would expand `%(message)s` that contains `message 1 2` pre-formatted in the previous step
```python
class PercentStyle:
    def _format(self, record):
        return self._fmt % record.__dict__

# and

class Formatter:
    def formatMessage(self, record):
        return self._style.format(record)
```

So right now uvicorn's log formatting and gunicorn access log formatting happen in different stages of processing. They can be made to both work on `Formatter` level via a `LoggingAdapter` of sort, but we'll have to sort out some conflicts between the two log formats, `--access-log-format` coming from gunicorn and uvicorn's own `LOGGING_CONFIG`.

uvicorn/protocols/http/h11_impl.py
immerrr: > I agree we should be able to customize the logger format, but I'm not sure we should do it in the same way as Gunicorn does.

The idea of this PR is to fix the workflow where in the end (or in the beginning, if you prefer) there is some gunicorn code that provides the access log format parameter to the worker in a certain way that is beyond uvicorn's control. IMO it is easier to use it as-is, but there can be of course any amount of transformation and adaptation on top.

PR #948, opened 2021-02-04
Docs: Nginx + websockets

JoseKilo:
I was trying a new deployment using Nginx => Gunicorn => Uvicorn.
The app serves some websocket endpoints using Starlette.

I was getting 404 errors for every connection and didn't get any logging explaining what was happening.
After some debugging I found out that the Connection/Upgrade headers need to be set explicitly in the Nginx config, otherwise Nginx assumes that they shouldn't be forwarded.

More info: https://nginx.org/en/docs/http/websocket.html


euri10 (approved):
thanks @JoseKilo that will certainly save some headaches to someone else ! (make s me think I love traefik even more, ws and wss just worked out off the box)

PR #952, opened 2021-02-10
fix: pin uvloop to the lastest Python 3.6 supported version

Kludex:
Latest uvloop release (0.15.0) doesn't support Python 3.6 anymore. I don't know what decision `uvicorn` is going to take, but maybe before a deciison is made, we can pin uvloop 0.14.0. 

Alternatives:
- Uvicorn follows uvloop and stop supporting 3.6.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,10 +49,12 @@ minimal_requirements = [
     "typing-extensions;" + env_marker_below_38,
 ]
 
+
 extra_requirements = [
     "websockets==8.*",
     "httptools==0.1.* ;" + env_marker_cpython,
-    "uvloop==0.14.0;" + env_marker_cpython,
+    "uvloop==0.14.0; python_version == '3.6' and " + env_marker_cpython,
+    "uvloop>=0.14.0; python_version >= '3.7' and " + env_marker_cpython,
     "colorama>=0.4;" + env_marker_win,
     "watchgod>=0.6",
     "python-dotenv>=0.13",
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ minimal_requirements = [
 extra_requirements = [
     "websockets==8.*",
     "httptools==0.1.* ;" + env_marker_cpython,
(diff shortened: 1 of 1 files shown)


euri10:
could also use another pep508 environment-markers specifically for this line, something like, not sure this works
```
    "uvloop==0.14.0 ; python_version == '3.6'" + env_marker_cpython,
    "uvloop>=0.14.0 ;" + env_marker_cpython,
```

Kludex:
Sure, no problem!

Give me two minutes :)

EDIT: Going to take more time, cause it didn't work, and I need to check why (working time). :kissing:

euri10:
I may have forgot an and after the 3.6 !

euri10 (approved):
thanks @Kludex

Kludex:
Thank you! :)

Just for the record, I couldn't make it work without pinning the Python versions on both lines.

euri10:
hehe >_) seems like we have another parisian contributor ! that begins to look like a mafia :)_

abduakhatov:
@euri10 @Kludex still not working when pulling from docker hub. how to get latest patches?

euri10:
> @euri10 @Kludex still not working when pulling from docker hub. how to get latest patches?

I'm not sure where to start, this will be available through a release and you'll get it with pip install -U or something equivalent once the release is published, there's no link with docker

setup.py
```diff
@@ -49,10 +49,12 @@ def get_packages(package):
     "typing-extensions;" + env_marker_below_38,
 ]
 
+
 extra_requirements = [
     "websockets==8.*",
     "httptools==0.1.* ;" + env_marker_cpython,
-    "uvloop>=0.14.0 ;" + env_marker_cpython,
+    "uvloop==0.14.0; python_version == '3.6' and " + env_marker_cpython,
+    "uvloop>=0.14.0; python_version >= '3.7' and " + env_marker_cpython,
```
1st1, line 57: I'd set this to 0.15.1 as there are many important bug fixes in 0.15.

PR #959, opened 2021-02-19
unpin uvloop

graingert:
Now that uvloop sets python_requires the pinning isn't needed, but we do need to
ignore the packages with missing python_requires

https://github.com/MagicStack/uvloop/pull/401


euri10 (approved):
that is great, thanks @graingert

PR #964, opened 2021-02-24
Test logging

euri10:
Alternative to #887 which gone stale, which was an alternative to #885

While #859 bug went unnoticed this test would have revealed it, nowhere we seem to test what the output of our default logging should be, this is a small addition to do it, it leverages @florimondmanca idea implementation in #887


tests/middleware/test_logging.py
```diff
@@ -0,0 +1,89 @@
...
+    await send({"type": "http.response.start", "status": 204, "headers": []})
+    await send({"type": "http.response.body", "body": b"", "more_body": False})
+
+
+@pytest.mark.asyncio
+async def test_trace_logging(caplog):
+    config = Config(
+        app=app,
+        log_level="trace",
+    )
```
Kludex:
```suggestion
    config = Config(app=app, log_level="trace")
```

tests/middleware/test_logging.py
```diff
@@ -0,0 +1,89 @@
...
+        assert "ASGI [2] Send " in messages.pop(0)
+        assert "ASGI [2] Send " in messages.pop(0)
+        assert "ASGI [2] Completed" in messages.pop(0)
+
+
+@pytest.mark.asyncio
+async def test_access_logging(caplog):
+    config = Config(
+        app=app,
+    )
```
Kludex:
```suggestion
    config = Config(app=app)
```

tests/middleware/test_logging.py
```diff
@@ -0,0 +1,89 @@
...
+            if record.name == "uvicorn.access"
+        ]
+        assert '"GET / HTTP/1.1" 204' in messages.pop()
+
+
+@pytest.mark.asyncio
+async def test_default_logging(caplog):
+    config = Config(
+        app=app,
+    )
```
Kludex:
```suggestion
    config = Config(app=app)
```

Kludex (approved):
I've added some suggestions because `black` doesn't remove them by default. But LGTM.

PR #967, opened 2021-02-24
MessageLoggerMiddleware sometimes fails to return and hangs

euri10:
I discovered that accidentally playing rough with the test suite on trying to fix stuff for https://github.com/encode/uvicorn/pull/929

The 1st commit of this PR demonstrate the issue, if you just add `log_level="trace"` in the config in that test it hangs forever, for good reason since the middleware never returns.

### DIFF
--- a/uvicorn/middleware/message_logger.py
+++ b/uvicorn/middleware/message_logger.py
@@ -58,7 +58,7 @@ class MessageLoggerMiddleware:
         log_text = "%s [%d] Started scope=%s"
         self.logger.trace(log_text, prefix, task_counter, logged_scope)
         try:
-            await self.app(scope, inner_receive, inner_send)
+            r = await self.app(scope, inner_receive, inner_send)
         except BaseException as exc:
             log_text = "%s [%d] Raised exception"
             self.logger.trace(log_text, prefix, task_counter)
@@ -66,3 +66,5 @@ class MessageLoggerMiddleware:
         else:
             log_text = "%s [%d] Completed"
             self.logger.trace(log_text, prefix, task_counter)
+            return r
+
--- a/uvicorn/middleware/message_logger.py
+++ b/uvicorn/middleware/message_logger.py
@@ -67,4 +67,3 @@ class MessageLoggerMiddleware:
             log_text = "%s [%d] Completed"
             self.logger.trace(log_text, prefix, task_counter)
             return r
-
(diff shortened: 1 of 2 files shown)


uvicorn/middleware/message_logger.py
```diff
@@ -58,11 +58,12 @@ async def inner_send(message):
         log_text = "%s [%d] Started scope=%s"
         self.logger.trace(log_text, prefix, task_counter, logged_scope)
         try:
-            await self.app(scope, inner_receive, inner_send)
+            r = await self.app(scope, inner_receive, inner_send)
```
florimondmanca, line 61: I think ASGI apps aren't meant to return anything? This is equivalent to storing `None` and returning it in the `else` branch, which was already the case since Python functions return `None` by default.

Is the hang due to something else in the logger middleware? (Note that I didn't try to reproduce, reviewing from my phone 😄)

euri10: ok so asgi apps arent meant to return anything, I didn't know that.

I may have mis-diagnosed the issue, not sure, the thing is that if you run this specific test, it hangs for wsprotocol

this hangs forever:
`pytest 'tests/protocols/test_websocket.py::test_asgi_return_value[WSProtocol]'` 


and it fails for websockets protocol:
pytest 'tests/protocols/test_websocket.py::test_asgi_return_value[WebSocketProtocol]'
```
>           assert exc_info.value.code == 1006
E           AssertionError: assert 1000 == 1006
E            +  where 1000 = ConnectionClosedOK('code = 1000 (OK), no reason').code
E            +    where ConnectionClosedOK('code = 1000 (OK), no reason') = <ExceptionInfo ConnectionClosedOK('code = 1000 (OK), no reason') tblen=4>.value

tests/protocols/test_websocket.py:375: AssertionError
```


It hangs in `wsproto_impl.py` in `run_asgi`  after you end up with a None result from the `await self.app...`. It seems to block after, not sure why or where,

https://github.com/encode/uvicorn/blob/9dc5a43209fe081ba3e74135189252bfddf75587/uvicorn/protocols/websockets/wsproto_impl.py#L208-L226

euri10: well after playing with this I'm going to close it, I cant find a reproducible real-life example where it does that, so now I'm not sure it's not a "test-only" issue.

PR #973, opened 2021-03-03
Fix mkdocs syntax highlight

euri10:
Should be an easy one, same as https://github.com/encode/httpx/commit/06559f8266f52a8ce86842d5068e2ba3855a3f6a


Kludex:
Is this still necessary? I didn't see it when I implemented the other one :cold_sweat:

euri10:
> Is this still necessary? I didn't see it when I implemented the other one

no worries :) closing !

PR #978, opened 2021-03-06
Force reload_dirs to be a list

### LINKED ISSUE #976: improve user experience by complaining when reload_dirs is given a string instead of a list
Passing a string to reload_dirs as in the following, causes reload to silently not work :

```
    uvicorn.run(
        "dp.server:app",
        host="127.0.0.1", port=5000,
        log_level="info",
        reload=True,
        reload_dirs="/home/maxou/dev/proj",
        reload_delay=2.0
    )
```

Giving reload_dirs an array : ["/home/maxou/dev/proj"]  fixes the problem

It's not really a bug, but complaining with an error message would improve the user experience.

euri10:
Fixes #976 
If the user runs programmatically uvicorn he can do a:      ` uvicorn.run(.....,reload_dirs="/home/maxou/dev/proj")`
OP suggested we complain about that not being a list, I'd prefer we do it directly for him.
I added a test too
That kind of stuff makes me thing we should mypy the project, but this is another story

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -63,6 +63,11 @@ def test_config_should_reload_is_set(app, expected_should_reload):
     assert config_reload.should_reload is expected_should_reload
 
 
+def test_reload_dir_is_set():
+    config = Config(app=asgi_app, reload=True, reload_dirs="reload_me")
+    assert config.reload_dirs == ["reload_me"]
+
+
 def test_wsgi_app():
     config = Config(app=wsgi_app, interface="wsgi", proxy_headers=False)
     config.load()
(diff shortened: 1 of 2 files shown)


Kludex:
I can fix the mypy issues we currently have so we can add it. Maybe duplicate some of the code that is on asgiref (https://github.com/django/asgiref/pull/221) until its release? What do you think? @euri10 

For reference, running mypy without any options:
```bash
$ mypy uvicorn/
...
Found 186 errors in 27 files (checked 38 source files)
```

euri10:
It's a good idea @Kludex 
I already started it a while ago and we decided to do it gradually to avoid reviewer headaches.
The idea is to add files that will be checked one by one in mypy.ini and fix them.

one by one or more if that makes more sense

Kludex:
Ahhh! Ok, perfect. Makes sense. Thank you :)

Kludex (approved):

PR #979, opened 2021-03-07
refactor: change click implementation by typer

Kludex:
There's no need to accept this PR. Nothing really changes with it. It tries to be a positive refactor, but I'm not sure everybody sees it as such. I've implemented just to see how it would look like. That being said, feel free to close the PR if the first word that came to your mind was: "why". :kissing:  

It has some typo fixes and some periods being added. Nothing really important.

Also, there was a reduction of ~100 lines of code but I'm not sure if it's more readable now.


Kludex:
This is not necessary. I'm closing it.

PR #980, opened 2021-03-07
tests: ignore coverage on helper test functions

Kludex:
I'll open several issues for https://github.com/encode/uvicorn/issues/102

I don't think a single PR will be a good thing to do, but I can merge them in a single one if you think is the best.


tests/test_config.py
```diff
@@ -159,7 +159,7 @@ def test_ssl_config_combined(tls_certificate_pem_path):
     assert config.is_ssl is True
 
 
-def asgi2_app(scope):
+def asgi2_app(scope):  # pragma: no cover
```
euri10: this should not be skipped imho, we have code in config.py that makes the switch from asgi2 to asgi3 : https://github.com/encode/uvicorn/blob/d81b96ee0590df0b5111376ba38637645385a684/uvicorn/config.py#L331-L339

tests/supervisors/test_reload.py
```diff
@@ -20,7 +20,7 @@ def setup(self, tmpdir, reloader_class):
         self.tmp_path = Path(tmpdir)
         self.reloader_class = reloader_class
 
-    def run(self, sockets):
+    def run(self, sockets):  # pragma: no cover
         pass
```
euri10:
```suggestion
    def run(self, sockets):
        pass  # pragma: no cover
```

tests/supervisors/test_multiprocess.py
```diff
@@ -4,7 +4,7 @@
 from uvicorn.supervisors import Multiprocess
 
 
-def run(sockets):
+def run(sockets):  # pragma: no cover
     pass
```
euri10:
```suggestion
def run(sockets):
    pass  # pragma: no cover
```

euri10 (changes requested):

tests/test_config.py
Kludex: You're right. But the body of that function is not executed, because the `__call__` on the `ASGI2Middleware` is never reached. 

https://github.com/encode/uvicorn/blob/e6028a6b787f13692d04b4e5aca67a6319b8360e/uvicorn/middleware/asgi2.py#L5-L7

Kludex: I've updated and I'm now adding "pragma" on the exact lines that are needed. What do you think? @euri10

euri10 (approved):

tests/test_config.py
euri10: lgtm thanks !

PR #982, opened 2021-03-07
tests: add ssl coverage

Kludex:
I'm 0% sure about this. Comments are welcome.

The intention here is to cover this line: https://github.com/encode/uvicorn/blob/d5d62d49cb0ab1a46a4704d5f489ad1964e18747/uvicorn/config.py#L112

The type hints on the conftest are a plus, I can remove them if you wish.

### DIFF
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -55,6 +55,7 @@ async def test_run_password(
         ssl_keyfile=tls_ca_certificate_private_key_encrypted_path,
         ssl_certfile=tls_ca_certificate_pem_path,
         ssl_keyfile_password="uvicorn password for the win",
+        ssl_ca_certs=tls_ca_certificate_pem_path,
     )
     async with run_server(config):
         async with httpx.AsyncClient(verify=tls_ca_ssl_context) as client:
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -21,6 +21,7 @@ async def test_run(
         limit_max_requests=1,
         ssl_keyfile=tls_ca_certificate_private_key_path,
         ssl_certfile=tls_ca_certificate_pem_path,
+        ssl_ca_certs=tls_ca_certificate_pem_path,
     )
     async with run_server(config):
         async with httpx.AsyncClient(verify=tls_ca_ssl_context) as client:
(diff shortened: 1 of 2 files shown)


tests/test_ssl.py
```diff
@@ -55,6 +55,7 @@ async def test_run_password(
         ssl_keyfile=tls_ca_certificate_private_key_encrypted_path,
         ssl_certfile=tls_ca_certificate_pem_path,
         ssl_keyfile_password="uvicorn password for the win",
+        ssl_ca_certs=tls_ca_certificate_pem_path,
```
euri10, line 62: can we add this to all 3 tests ?

euri10 (changes requested):
Duplicates #902 but let's roll with this since it the other will rot since nobody's looking at it :)

tests/test_ssl.py
Kludex: Done.

Kludex:
> Duplicates #902 but let's roll with this since it the other will rot since nobody's looking at it :)

I should have checked :sweat: I was following the order. :kissing: 

A question kind of related: should I open a PR to change slightly the CI, so it can generate the coverage result from all the test environments instead for each of them? I see that `httpx` and `starlette` just change the flag `--fail-under` to 100 (we have 80 on `uvicorn`), but I'm not sure if we should follow that. On windows, for example, the coverage is 85%, but on ubuntu 95%.

euri10:
> > Duplicates #902 but let's roll with this since it the other will rot since nobody's looking at it :)
> 
> I should have checked  I was following the order. 

no worries :)

 
> A question kind of related: should I open a PR to change slightly the CI, so it can generate the coverage result from all the test environments instead for each of them? I see that `httpx` and `starlette` just change the flag `--fail-under` to 100 (we have 80 on `uvicorn`), but I'm not sure if we should follow that. On windows, for example, the coverage is 85%, but on ubuntu 95%.

for the --fail-under iirc I put it at that level originally to grind safely towards 100% and not be blocked by the CI, we can increase either step by step as progress is made, or put it at 100 once we are at that level, I'd prefer option 2 but it's not a strong preference

about having 1 report for all env, I'm no coverage pro so I dont know if/how it's possible. As we skip tests on windows sometimes, I'd better see the coverage level reference taken as the linux one and not bother with windows if that makes sense

euri10 (approved):
thanks @Kludex

PR #983, opened 2021-03-07
fix: mypy issues

Kludex:
This fixes some mypy issues, but there are still ~140~ 101 to fix (with the current configuration on `setup.cfg`).

I want to keep working on this and add the `mypy` to the CI, but I also want a confirmation about: https://github.com/encode/uvicorn/pull/978#issuecomment-792283188.

EDIT: I'm using the "Starlette types" slightly modified for the time being.

### DIFF
--- a/uvicorn/_types.py
+++ b/uvicorn/_types.py
@@ -1,10 +1,10 @@
 import sys
-from typing import Optional
+from typing import Any, Awaitable, Callable, Dict, Optional, Type, Union
 
 if sys.version_info < (3, 8):
-    from typing_extensions import Literal, TypedDict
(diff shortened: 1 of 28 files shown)


euri10:
ok this is quite a task, in order to simplify things for you and those who're going to review, I'd suggest doing what we begun in https://github.com/encode/uvicorn/pull/751/files ie. add one / two or more related files you want to mypy in `setup.cfg` then fix the errors.
also as you know types were added to the asgi ref, I've not followed if this was usable in current's state, but let's use that as the baseline,

Kludex:
Sure. It works for me.

I'm going to close this PR as it's already too messed up. I'll be opening small PRs over the week.

Thank you :)

PR #985, opened 2021-03-08
tests: add env file config coverage

Kludex:
Lines covered:
```python
        if env_file is not None:
            from dotenv import load_dotenv

            logger.info("Loading environment from '%s'", env_file)
            load_dotenv(dotenv_path=env_file)
```
`config.py`


Kludex:
ok... The very informative comment on https://github.com/encode/uvicorn/pull/470 makes sense now hahaha

Kludex:
I've found this: https://www.tutorialfor.com/questions-105322.htm

But I'm not sure how to act here... Any thoughts? @euri10

euri10:
I dont remember I had issue on this, you can take a look at https://github.com/encode/uvicorn/pull/626/files which did part of coverage increase last year and was reverted.

tests/test_config.py
```diff
@@ -247,3 +249,37 @@ def test_log_config_file(mocked_logging_config_module):
...
+    yield request.param
+    if os.getenv("FORWARDED_ALLOW_IPS"):
+        del os.environ["FORWARDED_ALLOW_IPS"]
+
+
+def test_env_file(web_concurrency: int, forwarded_allow_ips: str, caplog):
+    """
+    Test that one can load environment variables using an env file.
+    """
+    with tempfile.NamedTemporaryFile("w+") as fp:
```
euri10: why not use pytest fixtures for temporary files ?

tests/test_config.py
```diff
@@ -247,3 +249,37 @@ def test_log_config_file(mocked_logging_config_module):
...
+        content = (
+            f"WEB_CONCURRENCY={web_concurrency}\n"
+            f"FORWARDED_ALLOW_IPS={forwarded_allow_ips}\n"
+        )
+        fp.write(content)
+        fp.seek(0)
+        with caplog.at_level(logging.INFO):
```
euri10: why not following other tests pattern with config.load ?

Kludex: The test that checks if should reload (above on this file), is not calling it as well. So I thought it was fine?

But I'll add it, sure. 😗

tests/test_config.py
Kludex: Thanks, going to apply later.

tests/test_config.py

tests/test_config.py

Kludex:
> I dont remember I had issue on this, you can take a look at https://github.com/encode/uvicorn/pull/626/files which did part of coverage increase last year and was reverted.

Oops... I've read this after applying the changes. Is my implementation enough or should I match the previous PR?

euri10:
I'm not sure we need to test for a permutation of both `WEB_CONCURRENCY` and `FORWARDED_ALLOW_IPS` but would there be a good reason we need more explicit test names with `ids` or similar, the below is way too abstruse

20210314_1632_232x133_1615735975

so either keep it this way but name the tests more explicitely or just run 2 tests, one for the 1st env var, one for the 2nd one, hope this makes sense.

Kludex:
Do you mind sharing the way to see that cool interface (on the screenshot), please? :pleading_face:

euri10:
> Do you mind sharing the way to see that cool interface (on the screenshot), please?

it's just Pycharm ! on a side note using `pytest -rA` should show you the same

tests/test_config.py
```diff
@@ -249,6 +250,40 @@ def test_log_config_file(mocked_logging_config_module):
...
+
+
+@pytest.fixture(params=["127.0.0.1", "127.0.0.2"])
+def forwarded_allow_ips(request):
+    yield request.param
+    if os.getenv("FORWARDED_ALLOW_IPS"):
+        del os.environ["FORWARDED_ALLOW_IPS"]
+
+
+def test_env_file(web_concurrency: int, forwarded_allow_ips: str, caplog, tmpdir):
```
euri10: I would use a `tmp_path` fixture instead, I'm not against the tmpdir usage though it's just that I feel it's cleaner to use Path-like objects.
I'll approve and let you choose !

euri10 (approved):

tests/test_config.py
Kludex: thanks

PR #986, opened 2021-03-08
coverage: add log level config coverage

Kludex:
Covering:
```python
        if self.log_level is not None:
            if isinstance(self.log_level, str):
                log_level = LOG_LEVELS[self.log_level]
            else:
                log_level = self.log_level  # THIS
```
`config.py`

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -251,7 +251,10 @@ def test_log_config_file(mocked_logging_config_module):
 
 @pytest.mark.parametrize("log_level", [5, 10, 20, 30, 40, 50])
 def test_config_log_level(log_level):
-    Config(app=asgi_app, log_level=log_level)
+    config = Config(app=asgi_app, log_level=log_level)
+    config.load()
+
     assert logging.getLogger("uvicorn.error").level == log_level
     assert logging.getLogger("uvicorn.access").level == log_level
     assert logging.getLogger("uvicorn.asgi").level == log_level
+    assert config.log_level == log_level
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -247,3 +247,11 @@ def test_log_config_file(mocked_logging_config_module):
     mocked_logging_config_module.fileConfig.assert_called_once_with(
         "log_config", disable_existing_loggers=False
     )
+
+
+@pytest.mark.parametrize("log_level", [5, 10, 20, 30, 40, 50])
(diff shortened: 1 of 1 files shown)


tests/test_config.py
```diff
@@ -247,3 +247,11 @@ def test_log_config_file(mocked_logging_config_module):
     mocked_logging_config_module.fileConfig.assert_called_once_with(
         "log_config", disable_existing_loggers=False
     )
+
+
+@pytest.mark.parametrize("log_level", [5, 10, 20, 30, 40, 50])
```
euri10, line 267: Can we pass an int ? I have to say I always used this with a string level

tests/test_config.py
```diff
@@ -247,3 +247,11 @@ def test_log_config_file(mocked_logging_config_module):
...
+def test_config_log_level(log_level):
+    Config(app=asgi_app, log_level=log_level)
```
euri10: why config is not loaded ?

tests/test_config.py
```diff
@@ -247,3 +247,11 @@ def test_log_config_file(mocked_logging_config_module):
...
+    assert logging.getLogger("uvicorn.error").level == log_level
+    assert logging.getLogger("uvicorn.access").level == log_level
+    assert logging.getLogger("uvicorn.asgi").level == log_level
```
euri10, line 274: I'd say the purpose of testing the config is just to look at its attributes and check they're not changed after being loaded, this checks the logger but let's check the config itself

Kludex: Ah, perfect. Thanks.

tests/test_config.py
Kludex: yes, there's a cast in place... In case is different than `str` and `int`, you'll have a `KeyError` tho.

tests/test_config.py

tests/test_config.py
Kludex: Applied.

Kludex:
@euri10 Comments applied. Thank you. :)

euri10 (commented):
I think a rebase later and we gtg @Kludex , thanks !

Kludex:
Done, @euri10 . Thank you :)

Kludex:
@euri10 done. Thank you :)

euri10 (approved):

PR #987, opened 2021-03-08
coverage: add access log config coverage

Kludex:
Covers:
```python
        if self.access_log is False:
```
`config.py`

Info: `propagate` is False on each case (entering or not in the conditional).

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -249,7 +249,16 @@ def test_log_config_file(mocked_logging_config_module):
     )
 
 
-@pytest.mark.parametrize("access_log, handlers", [(True, 1), (False, 0)])
+@pytest.mark.parametrize(
+    "access_log, handlers",
+    [
+        pytest.param(True, 1, id="access log enabled should have single handler"),
+        pytest.param(False, 0, id="access log disabled shouldn't have handlers"),
+    ],
+)
 def test_config_access_log(access_log: bool, handlers: int):
-    Config(app=asgi_app, access_log=access_log)
+    config = Config(app=asgi_app, access_log=access_log)
+    config.load()
+
     assert len(logging.getLogger("uvicorn.access").handlers) == handlers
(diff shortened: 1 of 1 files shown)


tests/test_config.py
```diff
@@ -247,3 +247,9 @@ def test_log_config_file(mocked_logging_config_module):
     mocked_logging_config_module.fileConfig.assert_called_once_with(
         "log_config", disable_existing_loggers=False
     )
+
+
+@pytest.mark.parametrize("access_log, handlers", [(True, 1), (False, 0)])
+def test_config_access_log(access_log: bool, handlers: int):
+    Config(app=asgi_app, access_log=access_log)
```
euri10: same as previously, let's load the config and test it

tests/test_config.py
```diff
@@ -247,3 +247,9 @@ def test_log_config_file(mocked_logging_config_module):
     mocked_logging_config_module.fileConfig.assert_called_once_with(
         "log_config", disable_existing_loggers=False
     )
+
+
+@pytest.mark.parametrize("access_log, handlers", [(True, 1), (False, 0)])
```
euri10: not sure I understand the need for checking the handlers count, can you explain ? oh and if it's relevant, adding an `ids` kwarg to the parametrize would help remembering it

Kludex: Correct me if I'm wrong, but the access log variable purpose is to turn on/off the access logs, and the logic we use is to remove the handlers available if that value is False. 

I'm going to modify this test to make it easier to understand.

tests/test_config.py

tests/test_config.py
Kludex: Better now?

Kludex:
@euri10 Comments applied. Thank you!

euri10 (approved):

PR #988, opened 2021-03-08
tests: ignore test on supervisors init file

Kludex:
Both classes are already being tested on `test_reload.py`. I don't feel the need to reach those lines in the tests, do you?

I'm accepting suggestions on how to test it if so :)

### DIFF
--- a/uvicorn/supervisors/__init__.py
+++ b/uvicorn/supervisors/__init__.py
@@ -2,7 +2,7 @@ from uvicorn.supervisors.multiprocess import Multiprocess
 
 try:
     from uvicorn.supervisors.watchgodreload import WatchGodReload as ChangeReload
-except ImportError:
+except ImportError:  # pragma: no cover
     from uvicorn.supervisors.statreload import StatReload as ChangeReload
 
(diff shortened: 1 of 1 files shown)


euri10:
> Both classes are already being tested on `test_reload.py`. I don't feel the need to reach those lines in the tests, do you?
> 
> I'm accepting suggestions on how to test it if so :)

good question, this is not about testing the classes themselves here but more to check the import flow, so I agree I dont see a way to test it easily.
an idea @florimondmanca ?

Kludex:
We can work with pip and uninstall/install (back to the previous state) watchgod for testing.

Something like this (but organized and maybe being a fixture, ofc):
```python
import pip
pip.main(['uninstall', package]) 
# test for statreload
pip main(['install', package])
```

Or maybe mocking the globals object so it doesn't have the watchdog? I'm not even sure this is possible :kissing:

euri10:
to be fair I think the pragma here is well justified, let' see what other think but for me this is good enough, adding more is needlessly complex

florimondmanca (approved):
Yes, I think we can go with a simple no-cover here.

euri10 (approved):

PR #989, opened 2021-03-08
coverage: add test for duplicate watch directory on watchdogreload class

Kludex:
Add test for when you have your parent directory being watched on `WatchdogReload` class.

I've actually changed slightly the loop that contains the "parent check", if this is something _not cool_ I can remove it. Just let me know :)


Kludex:
Tests are passing locally on Python 3.7.6... :thinking:

euri10:
> Add test for when you have your parent directory being watched on `WatchdogReload` class.
> 
> I've actually changed slightly the loop that contains the "parent check", if this is something _not cool_ I can remove it. Just let me know :)

ok I'm not sure I follow here, this test was introduced in #817 as our watchgod logic would have removed one of the 2 directories if their names were similar, I dont get what this is trying to achieve

Kludex:
This line is not covered currently:
```python
                if compare_dir in watch_dir.parents:
                    watch_dirs_set.remove(watch_dir) # THIS!
```

The aim of this PR is to cover that line, **only that**.

What you see on `uvicorn/supervisors/watchgodreload.py` is what I judge for a **non-asked** readability improvement. This is what I meant by "if this is something not cool I can remove it.".

Kludex:
Wait... I think I get what you said... The test in place was already supposed to test what I'm trying to test. Is that it? I'm going to check later if I find the reason why is not (I'm sure that line above is not being covered).

euri10:
@Kludex what's the status on this please, trying to see where we should aim in priority to make that coverage project a thing of the past :)

PR #990, opened 2021-03-08
fix: mypy issues

Kludex:
Applied mypy to the following files:
- uvicorn/loops
- uvicorn/main.py
- uvicorn/workers.py
- uvicorn/protocols/http/auto.py
- uvicorn/protocols/websockets/auto.py
- uvicorn/supervisors/__init__.py

Let me know if you want me to divide the PR.


Kludex:
I'm not sure if the app parameter on `run()` should be str? :thinking: On the main I think it is, but on `run`, should it? On `Config`, I'm sure it's not.

Vibhu-Agarwal (commented):
The `app` parameter _might_ be of type other than `str`
- Over here, in the reverse condition (i.e. `not config.reload and config.workers == 1`), it seems non-`str` `app` **is** allowed.
https://github.com/encode/uvicorn/blob/bf1c64e2c141971c546671c7dc91b8ccf0afeb7d/uvicorn/main.py#L369-L374 
- Also, `app` is eventually passed to `import_from_string` (through `Config`), which may return the same module back if it's not `str`
https://github.com/encode/uvicorn/blob/bf1c64e2c141971c546671c7dc91b8ccf0afeb7d/uvicorn/config.py#L308
https://github.com/encode/uvicorn/blob/bf1c64e2c141971c546671c7dc91b8ccf0afeb7d/uvicorn/importer.py#L11-L12

uvicorn/main.py
```diff
@@ -359,10 +358,10 @@ def main(
         "use_colors": use_colors,
         "factory": factory,
     }
-    run(**kwargs)
+    run(app, **kwargs)
 
 
-def run(app, **kwargs):
+def run(app: str, **kwargs: typing.Any) -> None:
```
br3ndonland: https://github.com/encode/uvicorn/pull/990#issuecomment-794321115

> I'm not sure if the app parameter on `run()` should be str? 🤔 On the main I think it is, but on `run`, should it? On `Config`, I'm sure it's not.

As mentioned in the related PR encode/uvicorn#992, `app: Union[ASGIApplication, str]` could work, based on the definition added to `uvicorn._types`. If not, a less strict annotation like `Union[Callable, str]` could work.

It could be helpful to be slightly more specific with the annotation of `kwargs` also. The `kwargs` should be a `dict`, so `**kwargs: dict`.

uvicorn/workers.py
```diff
@@ -16,7 +17,7 @@ class UvicornWorker(Worker):
 
     CONFIG_KWARGS = {"loop": "uvloop", "http": "httptools"}
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
br3ndonland, line 20: Suggested: `*args: tuple, **kwargs: dict`

uvicorn/main.py
Kludex: > It could be helpful to be slightly more specific with the annotation of kwargs also. The kwargs should be a dict, so **kwargs: dict.

https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values

uvicorn/workers.py
Kludex: https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values

uvicorn/main.py
br3ndonland: Sorry for the confusion there, I was tired and thinking of the packed forms. I meant to put the `dict` annotation on the kwargs object under `def main`, because you would need an annotation there.

To use something more specific than `Any`, something like this could work, based on the annotations you added:

```py
def run(
    app: str, **kwargs: typing.Union[typing.Optional[list], bool, float, int, str]
) -> None:
    config = Config(app, **kwargs)
    ...

```

It's a bit clumsy to list off all those types, so a [type alias](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#type-aliases) could be helpful.

Kludex: I've applied the `app` one, thanks.

As for the `**kwargs`... I'm not sure the real benefit of doing that with so many types.

euri10 (approved):

PR #992, opened 2021-03-09
fix: 3rd mypy issues

Kludex:
I need help here :kissing: I've found those two issues not trivial to solve. It's kind of a waterfall, I'm able to fix those two, but what comes after... :boom:

I've tried to apply the ASGIApplication that we have on [asgiref](https://github.com/django/asgiref/blob/main/asgiref/typing.py). But I ended up with mypy conflicts with the `test_lifespan.py` file. 

```python
❯ mypy
uvicorn/config.py:115: error: Argument 1 to "load_cert_chain" of "SSLContext" has incompatible type "Optional[str]"; expected "Union[str, _PathLike[str]]"
uvicorn/config.py:125: error: Function is missing a type annotation for one or more arguments
Found 2 errors in 1 file (checked 16 source files)
```

Again... I can't target the right branch. On review consider only `7d6733c` and after, please.

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -7,7 +7,7 @@ import os
 import socket
 import ssl
 import sys
-from typing import List, Tuple
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import click
 
@@ -102,8 +102,14 @@ logger = logging.getLogger("uvicorn.error")
 
 
 def create_ssl_context(
-    certfile, keyfile, password, ssl_version, cert_reqs, ca_certs, ciphers
-):
+    certfile: Optional[str],
+    keyfile: Optional[str],
+    password: Optional[str],
+    ssl_version: int,
+    cert_reqs: int,
+    ca_certs: Optional[str],
+    ciphers: Optional[str],
+) -> ssl.SSLContext:
     ctx = ssl.SSLContext(ssl_version)
     get_password = (lambda: password) if password else None
     ctx.load_cert_chain(certfile, keyfile, get_password)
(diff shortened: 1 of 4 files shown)


Kludex:
Number of issues increased after trying to merge master back to it :sweat_smile: 

Turning it into a draft.

uvicorn/config.py
```diff
@@ -102,8 +102,14 @@
 
 
 def create_ssl_context(
-    certfile, keyfile, password, ssl_version, cert_reqs, ca_certs, ciphers
-):
+    certfile: Optional[str],
```
br3ndonland, line 105: `certfile` is a required positional argument when running [`SSLContext.load_cert_chain`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain), so annotating as `Optional` (which allows `None`) would not be ideal here. If path-like objects are acceptable, after `from pathlib import Path`, the annotation could be `certfile: Union[Path, str]`.

uvicorn/config.py
```diff
@@ -119,43 +125,43 @@ class Config:
     def __init__(
         self,
         app,
```
br3ndonland, line 127: `app` will need an annotation.

I agree that `ASGIApplication`, from `uvicorn._types`, would be suitable here. As mentioned in the related PR encode/uvicorn#990, it's a bit challenging, because `app` is sometimes also a string. Annotating as `Union[ASGIApplication, str]` should address these use cases. If not, relaxing the annotation with `Union[Callable, str]` could also work.

`app: Union[ASGIApplication, str]`

uvicorn/config.py
```diff
@@ -119,43 +125,43 @@ class Config:
...
+        forwarded_allow_ips: Optional[str] = None,
+        root_path: str = "",
+        limit_concurrency: Optional[int] = None,
+        limit_max_requests: Optional[int] = None,
+        backlog: int = 2048,
+        timeout_keep_alive: int = 5,
+        timeout_notify: int = 30,
+        callback_notify: Callable[..., None] = None,
+        ssl_keyfile: Optional[str] = None,
+        ssl_certfile: Optional[str] = None,
```
br3ndonland, line 157: As mentioned above, `ssl_certfile: Optional[Union[Path, str]] = None,` could work.

uvicorn/config.py
```diff
@@ -266,9 +272,10 @@ def configure_logging(self):
             logging.getLogger("uvicorn.access").handlers = []
             logging.getLogger("uvicorn.access").propagate = False
 
-    def load(self):
+    def load(self) -> None:
         assert not self.loaded
 
+        self.ssl: Optional[ssl.SSLContext]
         if self.is_ssl:
```
br3ndonland, line 279: Consider `if self.is_ssl and self.ssl_certfile:` instead, which will help ensure that the `self.ssl_certfile` required positional argument is present.

br3ndonland (changes requested):
I think you're actually almost there. The annotations just need some minor adjustments.

Kludex:
I'm closing this because it's a mess (too old, too many conflicts to fix) :sweat_smile: 

Thanks for the comments @br3ndonland. I'll take in consideration for my next PR.

br3ndonland:
Sounds good. Maybe the Django ASGI types could be added in their own PR? It would be great to have some consensus on the ASGI types, so that we can build with them in other PRs.

Kludex:
@br3ndonland What do you mean? `asgiref` was already added as dependency on `uvicorn` :thinking:

br3ndonland:
That's right, I see that now. I guess most of the new types added to uvicorn/_types.py in this PR are no longer necessary, because we can just import them from `asgiref`.

PR #993, opened 2021-03-09
fix: add mypy on watchgodreload.py

Kludex:
Target is `master` again... :sweat_smile:

I'll keep targeting `master`. I think it will be easier for you.


setup.cfg
```diff
@@ -12,6 +12,7 @@ files =
     uvicorn/supervisors/basereload.py,
     uvicorn/importer.py,
     uvicorn/protocols/utils.py,
+    uvicorn/supervisors/watchgodreload.py
```
Kludex:
```suggestion
    uvicorn/supervisors/watchgodreload.py,
```

uvicorn/supervisors/watchgodreload.py
```diff
@@ -1,31 +1,43 @@
...
-    def __init__(self, root_path):
+    def __init__(self, root_path: str) -> None:
         for t in self.ignored_file_regexes:
             self.ignored.append(t)
         self.ignored.append(self.ignore_dotted_file_regex)
         self._ignored = tuple(re.compile(r) for r in self.ignored)
         super().__init__(root_path)
 
-    def should_watch_file(self, entry):
+    def should_watch_file(self, entry: "DirEntry") -> bool:
```
euri10, line 30: seems weird to me, but it's 1st time I see DirEntry, I would have thought `should_watch_file` entry parameter would be a file, the type name suggests it  is a directory,  DirEntry can be a file ?

Kludex: Yeah, I'm just honoring the parent class here.

Reference: https://github.com/samuelcolvin/watchgod/blob/1f738352da0a38eed25f4730fde097f06eaaacd6/watchgod/watcher.py#L90-L98

Kludex: But yes, it can: https://docs.python.org/3/library/os.html#os.DirEntry

euri10: ok that's cool then

euri10 (approved):

PR #999, opened 2021-03-10
fix: add mypy on proxy_headers.py

Kludex:
Need help here. I have two issues.

## First
I have this:
```python
    async def __call__(
        self, scope: WWWScope, receive: ASGIReceiveCallable, send: ASGISendCallable
    ) -> None:
```
The right one should be `Scope`, but mypy is not able to identify that there was a conditional eliminating the `LifespanScope`. I don't know how to solve this.

## Second
Besides, there's a mypy issue on this line:
```python
    scope["scheme"] = x_forwarded_proto.strip()
```
Issue:
```
uvicorn/middleware/proxy_headers.py:44: error: Unsupported target for indexed assignment ("Union[HTTPScope, WebsocketScope]")
```

Which I was able to solve if I select `HTTPScope` instead of `WWWScope`. Meaning that the problem is when you use `Union` with those two `TypedDict` i.e. `WWWScope = Union[HTTPScope, WebsocketScope]`.


Kludex:
Those were the errors ignored:
```bash
+ mypy --show-error-codes
uvicorn/middleware/proxy_headers.py:46: error: TypedDict "LifespanScope" has no key 'client'  [misc]
uvicorn/middleware/proxy_headers.py:50: error: TypedDict "LifespanScope" has no key 'headers'  [misc]
uvicorn/middleware/proxy_headers.py:56: error: Unsupported target for indexed assignment ("Union[HTTPScope, WebsocketScope, LifespanScope]")  [index]
uvicorn/middleware/proxy_headers.py:68: error: Unsupported target for indexed assignment ("Union[HTTPScope, WebsocketScope, LifespanScope]")  [index]
Found 4 errors in 1 file (checked 27 source files)
```
I'd rather prefer to use `Scope` instead of `WWWScope` on the `scope` parameter because that's the right thing to do. Unfortunately the code is not able to infer that `LifespanScope` is not part of the conditional.

uvicorn/middleware/proxy_headers.py
```diff
@@ -8,11 +8,18 @@
...
+    ASGI3Application,
+    ASGIReceiveCallable,
+    ASGISendCallable,
+    Scope,
+)
 
 
 class ProxyHeadersMiddleware:
-    def __init__(self, app, trusted_hosts="127.0.0.1"):
+    def __init__(self, app: ASGI3Application, trusted_hosts: str = "127.0.0.1") -> None:
```
euri10: I'm surprised by this, `trusted_hosts` is a list of strings or a string I think
I dont get why mypy allows that given the tests on that  module is using either a list or a single string.
I now realize we added the whole typing without the corresponding tests, which is a mistake imho since we never check for usage in typing.
I'd be interested in seeing adding `test_proxy_headers.py` makes a difference

Kludex: > I now realize we added the whole typing without the corresponding tests, which is a mistake imho since we never check for usage in typing.

I can see it as a plus, but not as a mistake. :thinking: 

> I'd be interested in seeing adding test_proxy_headers.py makes a difference

I'm going to add it tonight then. Let's see. :+1:

euri10: yep mistake is probably too strong, this said, I'm no mypy pro but this:
```
        if isinstance(trusted_hosts, str):
            self.trusted_hosts = {item.strip() for item in trusted_hosts.split(",")}
        else:
            self.trusted_hosts = set(trusted_hosts)
```
is pretty clear to me: `trusted_hosts` it's a tr or something else, so I dont know why `trusted_hosts: str = "127.0.0.1"` is valid here !
it's probably my bad understanding of mypy though but I'd love to undertstand :)

Kludex: > is pretty clear to me: trusted_hosts it's a tr or something else

Yeah, to me as well. I didn't pay enough attention. 

> it's probably my bad understanding of mypy though but I'd love to undertstand :)

`mypy` can't infer based on "future conditionals". I've declared the `trusted_hosts` to be `str`, so that's it. What we understand seeing this tho, is that `trusted_hosts` will always be `str`, then on the conditional, we'll always be on the first logic branch. This is not the case... 

Anyway, you're are right. Covering the test would have helped me to see this.

euri10: I played a little bit with this, this is a hard one, I'll need more time

Kludex:
@euri10 As promised, here you have more information about the current issues (the previous comment was not up-to-date):
```bash
❯ ./scripts/check
+ black --check --diff --target-version=py36 uvicorn tests
All done! ✨ 🍰 ✨
65 files would be left unchanged.
+ flake8 uvicorn tests
+ mypy --show-error-codes
uvicorn/middleware/proxy_headers.py:48: error: Incompatible types in assignment (expression has type "object", variable has type "Optional[Tuple[str, int]]")  [assignment]
uvicorn/middleware/proxy_headers.py:58: error: Unsupported target for indexed assignment ("Union[HTTPScope, WebsocketScope, LifespanScope]")  [index]
uvicorn/middleware/proxy_headers.py:70: error: Unsupported target for indexed assignment ("Union[HTTPScope, WebsocketScope, LifespanScope]")  [index]
Found 3 errors in 1 file (checked 40 source files)
```

uvicorn/middleware/proxy_headers.py
```diff
@@ -21,28 +30,32 @@ def __init__(self, app, trusted_hosts="127.0.0.1"):
...
 
-    async def __call__(self, scope, receive, send):
+        return None
+
+    async def __call__(
+        self, scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable
+    ) -> None:
         if scope["type"] in ("http", "websocket"):
-            client_addr = scope.get("client")
+            client_addr: Optional[Tuple[str, int]] = scope.get("client")  # type: ignore
```
euri10:
```suggestion
            scope = cast(Union[HTTPScope, WebsocketScope], scope)
            client_addr: Optional[Tuple[str, int]] = scope.get("client")
```

uvicorn/middleware/proxy_headers.py
```diff
@@ -21,28 +30,32 @@ def __init__(self, app, trusted_hosts="127.0.0.1"):
...
             client_host = client_addr[0] if client_addr else None
 
             if self.always_trust or client_host in self.trusted_hosts:
-                headers = dict(scope["headers"])
+                headers = dict(scope["headers"])  # type: ignore
```
euri10:
```suggestion
                headers = dict(scope["headers"])
```

uvicorn/middleware/proxy_headers.py
```diff
@@ -21,28 +30,32 @@ def __init__(self, app, trusted_hosts="127.0.0.1"):
...
 
                 if b"x-forwarded-proto" in headers:
                     # Determine if the incoming request was http or https based on
                     # the X-Forwarded-Proto header.
                     x_forwarded_proto = headers[b"x-forwarded-proto"].decode("latin1")
-                    scope["scheme"] = x_forwarded_proto.strip()
+                    scope["scheme"] = x_forwarded_proto.strip()  # type: ignore[index]
```
euri10, line 61: these 2   `# type: ignore[index]` seem impossible to avoid

tests/middleware/test_proxy_headers.py
```diff
@@ -1,14 +1,21 @@
...
-    host, port = scope["client"]
-    addr = "%s://%s:%d" % (scheme, host, port)
+async def app(
+    scope: Scope,
+    receive: ASGIReceiveCallable,
+    send: ASGISendCallable,
+) -> None:
+    scheme = scope["scheme"]  # type: ignore
+    host, port = scope["client"]  # type: ignore
+    addr = "%s://%s:%d" % (scheme, host, port)  # type: ignore
```
euri10:
```suggestion
    addr = "%s://%s:%d" % (scheme, host, port)
```

euri10 (commented):
I think using a cast is the correct way to avoid some of the ignore here, but not all can be avoided

uvicorn/middleware/proxy_headers.py
Kludex: Thanks!

euri10 (approved):

euri10:
yeah another one ! thanks @Kludex !

PR #1002, opened 2021-03-11
fix: add mypy on debug.py

Kludex:
I need help here.

This is the issue:
```bash
uvicorn/middleware/debug.py:24: error: Type of TypedDict is ambiguous, could be any of ("HTTPResponseStartEvent", "WebsocketResponseStartEvent")
uvicorn/middleware/debug.py:24: error: Argument 1 has incompatible type "Dict[str, object]"; expected "Union[HTTPResponseStartEvent, HTTPResponseBodyEvent, HTTPServerPushEvent, HTTPDisconnectEvent, WebsocketAcceptEvent, WebsocketSendEvent, WebsocketResponseStartEvent, WebsocketResponseBodyEvent, WebsocketCloseEvent, LifespanStartupCompleteEvent, LifespanStartupFailedEvent, LifespanShutdownCompleteEvent, LifespanShutdownFailedEvent]"
uvicorn/middleware/debug.py:31: error: Type of TypedDict is ambiguous, could be any of ("HTTPResponseBodyEvent", "WebsocketResponseBodyEvent")
uvicorn/middleware/debug.py:31: error: Argument 1 has incompatible type "Dict[str, object]"; expected "Union[HTTPResponseStartEvent, HTTPResponseBodyEvent, HTTPServerPushEvent, HTTPDisconnectEvent, WebsocketAcceptEvent, WebsocketSendEvent, WebsocketResponseStartEvent, WebsocketResponseBodyEvent, WebsocketCloseEvent, LifespanStartupCompleteEvent, LifespanStartupFailedEvent, LifespanShutdownCompleteEvent, LifespanShutdownFailedEvent]"
uvicorn/middleware/debug.py:48: error: Type of TypedDict is ambiguous, could be any of ("HTTPResponseStartEvent", "WebsocketResponseStartEvent")
uvicorn/middleware/debug.py:48: error: Argument 1 has incompatible type "Dict[str, object]"; expected "Union[HTTPResponseStartEvent, HTTPResponseBodyEvent, HTTPServerPushEvent, HTTPDisconnectEvent, WebsocketAcceptEvent, WebsocketSendEvent, WebsocketResponseStartEvent, WebsocketResponseBodyEvent, WebsocketCloseEvent, LifespanStartupCompleteEvent, LifespanStartupFailedEvent, LifespanShutdownCompleteEvent, LifespanShutdownFailedEvent]"
uvicorn/middleware/debug.py:55: error: Type of TypedDict is ambiguous, could be any of ("HTTPResponseBodyEvent", "WebsocketResponseBodyEvent")
uvicorn/middleware/debug.py:55: error: Argument 1 has incompatible type "Dict[str, object]"; expected "Union[HTTPResponseStartEvent, HTTPResponseBodyEvent, HTTPServerPushEvent, HTTPDisconnectEvent, WebsocketAcceptEvent, WebsocketSendEvent, WebsocketResponseStartEvent, WebsocketResponseBodyEvent, WebsocketCloseEvent, LifespanStartupCompleteEvent, LifespanStartupFailedEvent, LifespanShutdownCompleteEvent, LifespanShutdownFailedEvent]"
```

It happens for this:
```python
        await send(
            {
                "status": self.status_code,
                "headers": [[b"content-type", b"text/html; charset=utf-8"]],
            }
        )
```

Ideas?


euri10:
> ```shell
> error: Type of TypedDict is ambiguous, could be any of 
> ```

have you seen last comment here : https://github.com/python/mypy/issues/8533#issuecomment-599118564

Kludex:
I've applied the fixes according to your comment (also, the same ignores are present on hypercorn).

Besides, I've changed the import of `urllib` to `urllib.parse`, as there were mypy issues. Details about it can be found here: https://stackoverflow.com/a/41502702

uvicorn/middleware/debug.py
```diff
@@ -81,6 +96,7 @@ async def inner_send(message):
                 raise exc from None
 
             accept = get_accept_header(scope)
+            response: Union[HTMLResponse, PlainTextResponse]
```
Kludex, line 99: It can be either this or `response: Response`. Any preference?

uvicorn/middleware/debug.py
```diff
@@ -1,22 +1,33 @@
...
+    async def __call__(
+        self, scope: WWWScope, receive: ASGIReceiveCallable, send: ASGISendCallable
+    ) -> None:
         await send(
-            {
+            {  # type: ignore
                 "type": "http.response.start",
                 "status": self.status_code,
                 "headers": [[b"content-type", b"text/html; charset=utf-8"]],
             }
```
euri10: here we could avoid ignoring the type and do what we already did elsewhere with TypedDicts, something like
```
response_start: HTTPResponseStart = {"type": "http.response.start", ......}
await send(response_start)
```
I'm doing this on top of my head and cant remember if `HTTPResponseStart` exists but you see the spirit !

uvicorn/middleware/debug.py
```diff
@@ -1,22 +1,33 @@
...
         )
         await send(
-            {
+            {  # type: ignore
```
euri10: same as above

uvicorn/middleware/debug.py
euri10: lgtm that way

uvicorn/protocols/utils.py
```diff
@@ -1,5 +1,5 @@
 import asyncio
-import urllib
+import urllib.parse
```
euri10, line 2: why do we need that ?

euri10 (changes requested):

uvicorn/middleware/debug.py
Kludex: It exists, but I had to change the list inside the headers, because it accepts `Iterable[Tuple[bytes, bytes]]`, but not the current `Iterable[List[bytes, bytes]]`, jfyk.

euri10: > It exists, but I had to change the list inside the headers, because it accepts `Iterable[Tuple[bytes, bytes]]`, but not the current `Iterable[List[bytes, bytes]]`, jfyk.

that's a good mypy catch then :)

uvicorn/protocols/utils.py
Kludex: `urllib` doesn't import the module `parse`, and the only call we make on this file is here:

```python
    path_with_query_string = urllib.parse.quote(
        scope.get("root_path", "") + scope["path"]
    )
```
With the current `import urllib`, you can't travel to `quote()` due to the reason above.

But if you want a better excuse: it restrict the scope of what is being imported.

euri10: ok quite unrelated to the current PR but that's minor ;)

euri10 (approved):

uvicorn/protocols/utils.py
Kludex: Oh, yeah, sorry. I don't know why I add it here :sweat_smile:

PR #1003, opened 2021-03-11
fix: add mypy on logging.py

Kludex:
Explaining ignore on this line:
```python
        status_code = self.get_status_code(status_code)  # type: ignore
```
I have this error message:
```bash
uvicorn/logging.py:106: error: Argument 1 to "get_status_code" of "AccessFormatter" has incompatible type "Union[Any, str]"; expected "int"
```
And the culprit is ` recordcopy.args`. :kissing: 

I don't know how to solve it. Any ideas?

EDIT: I've just added a cast around it:
```python
        status_code = self.get_status_code(int(status_code))
```


uvicorn/logging.py
```diff
@@ -96,7 +103,7 @@ def formatMessage(self, record):
             http_version,
             status_code,
         ) = recordcopy.args
-        status_code = self.get_status_code(status_code)
+        status_code = self.get_status_code(status_code)  # type: ignore
```
br3ndonland: If the status code is coming in from the log record as a string, like `"418"`, it could be cast to an `int`.

`status_code = self.get_status_code(int(status_code))`

This would work well with `http.HTTPStatus(status_code).phrase`.

```py
>>> import http
>>> status_code = "418"
>>> http.HTTPStatus(int(status_code)).phrase
"I'm a Teapot"
```

Kludex: Thanks @br3ndonland

euri10 (approved):
thanks @Kludex

PR #1005, opened 2021-03-11
Use `surrogateescape` to encode headers on `websockets` implementation

### LINKED ISSUE #968: WebSocket headers encoding problem
### Checklist

### Describe the bug

If there is a websocket connection with a non-ascii char in header, the server will explode. The problem seems to be from here:
https://github.com/encode/uvicorn/blob/9dc5a43209fe081ba3e74135189252bfddf75587/uvicorn/protocols/websockets/websockets_impl.py#L102
Also, I checked the [ASGI doc](https://buildmedia.readthedocs.org/media/pdf/asgi/latest/asgi.pdf) and in section 2.2.5 it states:
- These are byte strings of the exact byte sequences sent by the client/to be sent by the server. While
modern HTTP standards say that headers should be ASCII, older ones did not and allowed a wider range of
characters. Frameworks/applications should decode headers as they deem appropriate.

So, I think uvicorn should allow for non-ascii chars in header. It is definitely the only reason that we can't use uvicorn in production yet.

### Expected behavior

Doesn't explode.

### Actual behavior

Explodes with an exception.

### Environment

sephioh:
Currently, any non-ascii header value sent to uvicorn running with websockets implementation will raise a `UnicodeEncodeError`. 
As [websockets package represents non-ascii characters in headers using surrogate escapes](https://github.com/aaugustin/websockets/blob/3401751d4f1200d398c5e69f9c4a0ece862ff36b/src/websockets/legacy/http.py#L142), this PR makes websockets implementation to follow the same pattern when encoding them.

Edit by @Kludex:
- Closes #968

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -139,10 +139,13 @@ async def test_headers(protocol_cls):
             headers = self.scope.get("headers")
             headers = dict(headers)
             assert headers[b"host"].startswith(b"127.0.0.1")
+            assert headers[b"username"] == bytes("abraão", "utf-8")
             await self.send({"type": "websocket.accept"})
 
     async def open_connection(url):
-        async with websockets.connect(url) as websocket:
+        async with websockets.connect(
+            url, extra_headers=[("username", "abraão")]
+        ) as websocket:
             return websocket.open
 
     config = Config(app=App, ws=protocol_cls, lifespan="off")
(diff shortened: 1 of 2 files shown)


euri10:
closing this until we agree on a fix, let's discuss that in https://github.com/encode/uvicorn/issues/968

Kludex (approved):
This matches the behavior with `wsproto`.

You can confirm it with the following client code:

```python
#!/usr/bin/env python

import asyncio

import websockets


async def hello():
    async with websockets.connect(
        "ws://localhost:8000", extra_headers={"haha": "Abraão"}
    ) as websocket:
        await websocket.send("Hello world!")
        await websocket.recv()


asyncio.run(hello())
```
And on the `uvicorn` side, you can run the following using `wsproto`:

```bash
uvicorn main:app --port 8000 --reload --ws wsproto
```
And... Replace `wsproto` by `websockets` above to check the matching behavior.

@tomchristie can you please confirm that we are fine with this?

lovelydinosaur:
I don't have a problem with this.

PR #1006, opened 2021-03-12
coverage: pin higher coverage

Kludex:
The file `httptools_impl.py` is not covered by windows, the reason for it is the marker that was added on https://github.com/encode/uvicorn/pull/309. 

But after removing the marker, it worked... :thinking:

Previous Windows coverage: 86% (due mostly to the file mentioned above).
Coverage achieved: 95%.

Related issue: https://github.com/encode/uvicorn/issues/102

I still don't like this, I think the "heavyweight tooling approaches" makes more sense, but... Here you have @tomchristie . 

I've also learned about the `compat.py`, but it will not solve the issue with the `httptools_impl.py` and it will change slightly the structure of files (`loops/asyncio.py` might cease to exist). 

Well, I can just add a `no pragma` for the whole `httptools_impl.py` file if the removal of the marker is not accepted, but I don't think we want that...


euri10:
there is something going on with the windows CI, any idea ?

I have 
`uvicorn/protocols/http/httptools_impl.py            311     11    96%   115-116, 143-144, 148, 300, 306, 404, 407, 439, 441` on my machine and it's 2% on the windows CI

Kludex:
Yes, it was working before because I've removed the marker for `httptools` on Windows. I'm checking what I said previously (on the description above).

Kludex:
Ok... https://github.com/MagicStack/httptools/blob/40404907f2714c4463aca35f3d258efdbb417fb3/setup.py#L187

Looks like Windows is supported by `httptools`...

euri10 (approved):

euri10:
cool !

Issue #722, opened 2020-07-24
Reload option doesnt work with unix sockets


euri10:
Was about to report the same thing, but with unix sockets.

_Originally posted by @magnusjjj in https://github.com/encode/uvicorn/issues/553#issuecomment-663530183_

magnusjjj:
Yeah, uh, I have looked through the source.

main.py:346 
```
    if config.should_reload:
        sock = config.bind_socket()
        supervisor = ChangeReload(config, target=server.run, sockets=[sock])
        supervisor.run()
    elif config.workers > 1:
        sock = config.bind_socket()
        supervisor = Multiprocess(config, target=server.run, sockets=[sock])
        supervisor.run()
    else:
        server.run()
```

That. It seems like there are three kinds of 'processes', and they can't "stack".

The problem lies in that unix sockets, and passing a slew of premade sockets, require different calls to setup loops with asyncio, if I read the code correctly, as seen in main.py:445

So it's the same bug for both having workers and the reloader.
Creation of the sockets should *probably* be moved to the config.bind_socket (or rather, bind_sockets i guess), and then later code figuring out whether or not the sockets given to them are unix sockets or workers. This is my first glance at the code though.

euri10:
yes you're right I think this would need a refactor, a good example would be in hypercorn I think.
lacking time to tackle on this though

Kludex:
It looks like this was fixed by #1009.

Issue #586, opened 2020-03-09
Uvicorn ignores --uds for --reload


Roang-zero1:
As talked about in #368 the `--reload` setting ignores the `--uds` setting, thereby reload binds itself to the default localhost:8000.

In our dev deployment we try to stay as close to production as possible. Due to this issue we have to workaround our nginx config to fall back to a https socket in dev,

I would like to request this feature to be added.

lovelydinosaur:
Personally I'm going to treat this as out of scope at the moment. If someone wants to take a go at a PR for it, I'd happily review it, but otherwise there's no realistic chance of me being able to make time for it.

For now `--reload` is a development setting that shouldn't be used alongside `--uds` which is for use with in production.

Professional Tier Sponsors are able to request priority support... https://github.com/sponsors/encode

Issue #368, opened 2019-06-04
Uvicorn ignores --uds and --fd configs


sebastiaoLa:
I'm using uvicorn with supervisor and nginx

I'm trying to use the unix socket but it keep using the host and port

here is my supervisor conf file:

``` supervisor
[fcgi-program:tas]
enviroment=DJANGO_SETTINGS_MODULE="maq_tasks.settings",DJANGO_ASGI_MODULE="maq_tasks.asgi",LOG_LEVEL="debug",WORKERS="3"
directory=/home/tas/src/
socket=unix:///home/tas/run/uvicorn.sock
command=/home/tas/bin/uvicorn maq_tasks.asgi:application --workers 3 --log-level debug --fd 0
user=tas
autostart=true
autorestart=true
redirect_stderr=true
stdout_logfile=/home/tas/logs/uvicorn-error.log
```

and here is my error

```
Traceback (most recent call last):
  ....
  File "/home/tas/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/tas/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/tas/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/tas/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/tas/lib/python3.6/site-packages/uvicorn/main.py", line 257, in main
    run(**kwargs)
  File "/home/tas/lib/python3.6/site-packages/uvicorn/main.py", line 274, in run
    socket = config.bind_socket()
  File "/home/tas/lib/python3.6/site-packages/uvicorn/config.py", line 219, in bind_socket
    sock.bind((self.host, self.port))
OSError: [Errno 98] Address already in use
```

sebastiaoLa:
looking futher down on the code, when using supervisor the uvicorn executable file ignores --uds or --fd configs and uses the default host:port config instead of the unix socket
```python
def run(app, **kwargs):
    config = Config(app, **kwargs)
    server = Server(config=config)

    if config.reload and not isinstance(app, str):
        config.logger_instance.warn(
            "auto-reload only works when app is passed as an import string."
        )

    if isinstance(app, str) and (config.debug or config.reload):
        socket = config.bind_socket() # <- here
        supervisor = StatReload(config)
[...]

PR #1009, opened 2021-03-15
Adapt bind_socket to make it usable with multiple processes

### LINKED ISSUE #428: UNIX domain socket raises 503 when workers are used

euri10:
This makes it possible to use reload and workers flag with uds and fd.

fixes https://github.com/encode/uvicorn/issues/428 https://github.com/encode/uvicorn/issues/368 https://github.com/encode/uvicorn/issues/586 #722


euri10:
still WIP I noticed the uvicorn logger message needs to be adapted

euri10:
ready now @Kludex  !

uvicorn/config.py
```diff
@@ -366,37 +366,75 @@ def setup_event_loop(self):
...
+            )
+            protocol_name = "https" if self.is_ssl else "http"
+            logger.info(
+                message,
+                protocol_name,
+                self.host,
+                self.port,
+                extra={"color_message": color_message},
+            )
         sock.set_inheritable(True)
```
Kludex, line 454: Does this impact the socket from `socket.fromfd`? https://docs.python.org/3/library/socket.html#socket.fromfd

image

uvicorn/config.py
```diff
@@ -366,37 +366,75 @@ def setup_event_loop(self):
...
-            logger.error(exc)
-            sys.exit(1)
+            message = f"Uvicorn running on {addr_format} (Press CTRL+C to quit)"
+            color_message = (
+                "Uvicorn running on "
+                + click.style(addr_format, bold=True)
+                + " (Press CTRL+C to quit)"
+            )
+            protocol_name = "https" if self.is_ssl else "http"
+            logger.info(
```
Kludex: Those `logger.info` take a lot of lines now, can we have a single one at the end that accepts `logger_args` and the `color_message`?

This is just a small point not related to the logic, so you can close it if you think this is not necessary / overwork. :kissing:

tests/test_config.py
```diff
@@ -320,3 +321,25 @@ def test_ws_max_size():
...
+        (True, 1),
+        (False, 2),
+    ],
+    ids=["--reload=True --workers=1", "--reload=False --workers=2"],
+)
+@pytest.mark.skipif(sys.platform == "win32", reason="require unix-like system")
+def test_bind_unix_socket_works_with_reload_or_workers(tmp_path, reload, workers):
+    uds_file = tmp_path / "uvicorn.sock"
+    config = Config(
+        app=asgi_app, uds=uds_file.as_posix(), reload=reload, workers=workers
```
Kludex, line 372: Is there a reason for not adding any test for `fd`?

Kludex (commented):
I did my best here. It's a short PR, but I feel like I'm not qualified to approve it yet. I'm going to study some details before having the confidence to do so.

But I hope my current comments make sense.

uvicorn/config.py
euri10: this `sock.set_inheritable(True)` applies to all 3 cases so yes. it serves when we have multiple process and 1 socket to listen on iirc

uvicorn/config.py
euri10: yes that's a good idea, done

tests/test_config.py
euri10: no, I added one

euri10:
ready for another review @Kludex 
let me know if something not clear I can easily expain

uvicorn/config.py
Kludex: Ah, I understand what confused me now. Thanks.

uvicorn/main.py
```diff
@@ -391,6 +392,8 @@ def run(app: typing.Union[ASGIApplication, str], **kwargs: typing.Any) -> None:
         Multiprocess(config, target=server.run, sockets=[sock]).run()
     else:
         server.run()
+        if config.uds:
+            os.remove(config.uds)
```
Kludex: Maybe I'm mistaken, but can't we have a single one of those outside the conditional?

```suggestion
    if config.uds:
        os.remove(config.uds)
```

Kludex: I guess we can, but it's not right, because we bind inside them. :thinking: Right?

euri10: it looks simpler yes, I removed occurrences in the basereload and multiprocess shutdown method and put it outside

Kludex (approved):
I'm confident enough now.

But maybe we should wait for the `mypy` merge first? Anyway, your call.

euri10:
> But maybe we should wait for the `mypy` merge first? Anyway, your call.

the one on config.py ?

Kludex:
Yeah

euri10:
> Yeah

that's a good idea, he put a lot of efforts into it and it would be cruel to have to add that part on top of it ;)

uvicorn/config.py
```diff
@@ -395,37 +395,63 @@ def setup_event_loop(self) -> None:
             loop_setup()
 
     def bind_socket(self) -> socket.socket:
-        family = socket.AF_INET
-        addr_format = "%s://%s:%d"
+        logger_args: List[Union[str, int]]
```
euri10, line 398: just needed to add this post-merge with master @Kludex 
the rest stayed the same

euri10: So  I removed the hold button and feel free to merge if still ok with this !

Kludex: I find it so weird that the approval is not removed after a commit is pushed. I don't understand the reason behind it.

Anyway, LGTM! :)

Issue #742, opened 2020-08-03
Uvicorn cannot be shutdown programmatically


Andrew-lll:
### There is no documented way to shutdown uvicorn in python:

ex:

```
instance = uvicorn.run("example:app", host="127.0.0.1", port=5000, log_level="info")
instance.shutdown()
```

How do we shutdown uvicorn?

florimondmanca:
Hi,

Not documented indeed, but a multithreaded approach should do…

```python
import contextlib
import time
import threading
import uvicorn

class Server(uvicorn.Server):
    def install_signal_handlers(self):
        pass

    @contextlib.contextmanager
    def run_in_thread(self):
        thread = threading.Thread(target=self.run)
        thread.start()
        try:
            while not self.started:
                time.sleep(1e-3)
            yield
        finally:
            self.should_exit = True
            thread.join()

config = Config("example:app", host="127.0.0.1", port=5000, log_level="info")
server = Server(config=config)

with server.run_in_thread():
    # Server started.
    ...
# Server stopped.
```

Very handy to run a live test server locally using a pytest fixture…

```python
# conftest.py
import pytest

@pytest.fixture(scope="session")
def server():
    server = ...
    with server.run_in_thread():
        yield
```

Andrew-lll:
I do appreciate the reply/code, but:

Why is this not built in and documented?
Why do I have to implement this code?

It's a webserver.  Run_in_thread() and shutdown() are core functionality...am I wrong?

plazmakeks:
I'm following @florimondmanca s approach to run pact tests on the provider side. However when using this i get the following error from uvicorn's Server.run method:

`RuntimeError: There is no current event loop in thread 'Thread-7'.`

Any idea on that? Happens with python 3.6.9 as well as 3.7.7.

another edit:

I got it to work using

`config = Config("example:app", host="127.0.0.1", port=5000, log_level="info", loop="asyncio")`

Elijas:
> I do appreciate the reply/code, but:
> 
> Why is this not built in and documented?
> Why do I have to implement this code?
> 
> It's a webserver. Run_in_thread() and shutdown() are core functionality...am I wrong?

Fully agreed, this question of how to run and stop unicorn pops up [on stack overflow](https://stackoverflow.com/a/64521239/1544154) as well, it's quite obscure as it is now

florimondmanca:
I appreciate the feedback here, but then this brings the question — what would folks expect the usage API to be for something like this?

[...]

PR #1011, opened 2021-03-19
Add thread event primitives to track server start/exit

rmorshea:
Add `did_start`/`did_exit` thread events to `ServerState` and turn `Server.started` into alias of `Event.is_set` - downstream applications can then hook into `Server.server_state.did_*` events to `wait()` for certain state changes.

While this does not strictly solve #742 it makes it easier to implement a `run_in_thread` style method.

Instead of writing:

```python
import contextlib
import time
import threading
import uvicorn

class Server(uvicorn.Server):
        pass

    @contextlib.contextmanager
    def run_in_thread(self):
        thread = threading.Thread(target=self.run)
        thread.start()
        try:
            while not self.started:
                time.sleep(1e-3)
            yield
        finally:
            self.should_exit = True
            thread.join()
```

You would do:

```python
import contextlib
import uvicorn

@contextlib
def run_in_thread(server):
    ...

class Server(uvicorn.Server):
        pass

    @contextlib.contextmanager
    def run_in_thread(self):
        threading.Thread(target=self.run).start()
        try:
            self.server_state.did_start.wait()
            yield
        finally:
            self.should_exit = True
            self.server_state.did_exit.wait()
```

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -31,22 +31,33 @@ class ServerState:
         self.connections = set()
         self.tasks = set()
         self.default_headers = []
+        self.did_start = threading.Event()
+        self.did_exit = threading.Event()
 
 
 class Server:
     def __init__(self, config):
         self.config = config
         self.server_state = ServerState()
-
-        self.started = False
         self.should_exit = False
         self.force_exit = False
         self.last_notified = 0
 
+    @property
+    def started(self):
+        self.server_state.did_start.is_set()
+
+    @property
+    def exited(self):
+        return self.server_state.did_exit.is_set()
+
     def run(self, sockets=None):
-        self.config.setup_event_loop()
-        loop = asyncio.get_event_loop()
-        loop.run_until_complete(self.serve(sockets=sockets))
+        try:
+            self.config.setup_event_loop()
+            loop = asyncio.get_event_loop()
+            loop.run_until_complete(self.serve(sockets=sockets))
+        finally:
+            self.server_state.did_exit.set()
 
     async def serve(self, sockets=None):
         process_id = os.getpid()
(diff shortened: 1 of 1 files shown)


SommerEngineering (commented):
All changes are implemented correctly and without issues from my point of view.

rmorshea:
@SommerEngineering it looks like you haven't given explicit approval in the review, is there anything that should be fixed?

SommerEngineering:
@rmorshea In my opinion, nothing else needs to be done :slightly_smiling_face: I did not approve because I am unsure what the rules are in this repository. The following rules are linked on the PR page: [CONTRIBUTING.md](https://github.com/encode/.github/blob/e018bb630a92efd4609c2b60955e9c4678cd6c11/CONTRIBUTING.md)

Unfortunately, the rules do not say anything about this point :slightly_frowning_face: Can anyone give me some advice on this?

rmorshea:
Hmm, looks like I've gotta push this through some chat channels with maintainers.

euri10:
I have mixed feelings about this, the use of threading events in the server state is far from being an obvious choice to me, on top of that I dont know how this will play when we use multiple process or the reload machinery (nor are there tests to check that), and finally as you said it doesn't solve the issue in the first place, so maybe we should take a step back and as @florimondmanca said agree on the API people would likely see and mostly if we should support that use case.

rmorshea:
> mostly if we should support that use case.

@euri10 I see ample evidence to suggest that there's a need for an interface which would allow users to more easily run the server in a thread. #742 is a rather long lived issue that has seen relatively consistent activity since its creation. Additionally, as I pointed out in some out-of-band conversations, in many testing scenarios (i.e. when using sync tools Selenium) there isn't actually an alternative - you have to run the server in a thread.

While I can understand not wanting to add these primitives to make things easier in the short-term. In the long-run, I think there's good reasons to develop some interface that would meet these needs, if only to prevent people from having to subclass the server as is suggested in #742.

PR #1013, opened 2021-03-20
Added --reload-exclude-dir param

Aragroth:
---
New CLI feature
---
New `--reload-exclude-dir` parameter for uvicorn-CLI allows to exclude folder from being watched by reloader. As a `--reload-dir` param, it can be specified multiple times. 
## Purpose
The idea was already described in #984. I would also mention, that WatchGod (which is used by default) reloads project even if changes were made not only to source code (Unlike Statreload). So when you have a folder with static files and user uploads image, the project gets reloaded. And adding lots of folders with param `--reload-dir`, instead of just excluding one folder, can be really annoying.

## Working mechanism
I thought, that idea of pattern matching is not the best. Also it would be better for both of reloaders to have same excluding mechanism. So, I just compare absolute paths of files. You can not specify both included and excluded folders. Also added a few tests for this param.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -39,6 +39,9 @@ Options:
   --reload-dir TEXT               Set reload directories explicitly, instead
                                   of using the current working directory.
 
+  --reload-exclude-dir TEXT       Exclude directories from being watched by
+                                  auto-reloader.
+
   --reload-delay FLOAT            Delay between previous and next check if
                                   application needs to be. Defaults to 0.25s.
                                   [default: 0.25]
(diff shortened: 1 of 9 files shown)


robbymeals:
Strongly support this feature addition! 🙏 🥺

117503445:
I also want this feature 🙏

dusty-phillips:
I'm happy to rebase this and fix the conflicts if I can get some confirmation from the devs that it will be merged.

Kludex:
Hey @dusty-phillips !
Our thoughts are on [this comment](https://github.com/encode/uvicorn/issues/984#issuecomment-857913584) and on the next one. 

This PR has less priority than the #820, but we are not definitive yet. I really appreciate your intentions, but I'd recommend to not lose time rebasing right now.

Kludex:
#820 was merged. Closing this PR.

@Aragroth thanks for the efforts!

PR #1023, opened 2021-04-24
Improve user feedback if no ws library installed

### LINKED ISSUE #1022: Missing user feedback if no ws-library has been installed - with httptools installed
### Checklist

### Describe the bug

#797 still persists with `httptools` installed in the environment.

### To reproduce

1. ```pip install uvicorn httptools fastapi```
2. Create a simple app
```python
from fastapi import FastAPI, WebSocket

app = FastAPI()

@app.websocket("/ws")
async def websocket_endpoint(websocket: WebSocket):
    ...
``` 
3. Open a web-socket connection.
```js
new WebSocket(`ws://localhost:8000/ws`)
```

### Expected behavior

Proper user feedback if no ws-supported library has been installed.

### Actual behavior

```
INFO:     Started server process [19637]
WARNING:  Unsupported upgrade request.
^C
INFO:     Shutting down
INFO:     Finished server process [19637]
```

### Environment

- Running uvicorn (master) with CPython 3.9.0 on Linux
- `uvicorn main:app`

```
fastapi==0.63.0
h11==0.12.0
httptools==0.2.0
```

### Additional context
#926 Fixes #797 for `h11`

Vibhu-Agarwal:
Fixes #1022

I've just duplicated the work done in #926

Should we move this logic to a central place like `protocols/utils.py`?

### DIFF
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -181,6 +181,13 @@ class HttpToolsProtocol(asyncio.Protocol):
         if upgrade_value != b"websocket" or self.ws_protocol_class is None:
             msg = "Unsupported upgrade request."
             self.logger.warning(msg)
+
+            from uvicorn.protocols.websockets.auto import AutoWebSocketsProtocol
+
+            if AutoWebSocketsProtocol is None:
+                msg = "No supported WebSocket library detected. Please use 'pip install uvicorn[standard]', or install 'websockets' or 'wsproto' manually."  # noqa: E501
+                self.logger.warning(msg)
+
             content = [STATUS_LINE[400]]
             for name, value in self.default_headers:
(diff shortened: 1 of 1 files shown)


euri10 (commented):
LGTM

There's been a lot of discussion around this in #926 with @florimondmanca and the consensus was to be less deceptive (ie put a warning message when a user attempts a ws connection without a library) instead of put a warning message every time (which would be the behavior if we put the message above, would it be ?)

Any thoughts on this @florimondmanca , this is the case here a user doesn't install like we advertise, which can and will happen anyway so why not handle that case gently as well.

euri10 (approved):
looks good to me, we have this duplicated on both protocols but it feels nicer to have a warning on a ws connection attempt than on every single connection if a user doesnt intend to use websockets

euri10:
thanks @Vibhu-Agarwal

PR #1024, opened 2021-04-26
update httptools from 0.1.* to 0.2.*

hidaris:
Httptools 0.2 replaces the underlying HTTP parser with llhttp as http-parser is
no longer actively maintained. See https://github.com/MagicStack/httptools/pull/56

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ minimal_requirements = [
 
 extra_requirements = [
     "websockets>=8.*",
-    "httptools==0.1.* ;" + env_marker_cpython,
+    "httptools==0.2.* ;" + env_marker_cpython,
     "uvloop>=0.14.0,!=0.15.0,!=0.15.1; " + env_marker_cpython,
     "colorama>=0.4;" + env_marker_win,
     "watchgod>=0.6",


RobertoPrevato:
Note: as far as I know, `httptools` 0.2.0 is also required to add compatibility with `Python 3.10`. This is what I experienced with [`blacksheep`](https://github.com/Neoteroi/BlackSheep).

hidaris:
@RobertoPrevato Hi, both `uvicorn` and` httptools` currently do not support python 3.10, https://github.com/encode/uvicorn/blob/master/setup.py#L89
https://github.com/MagicStack/httptools/blob/master/.github/workflows/tests.yml#L18，
Maybe we still need to wait a while, or test if 3.10 needs some tweaking work.

Kludex (approved):
It was a long way for that upgrade on `httptools`, and they were concerned about us. :heart: :+1:  

LGTM.

euri10:
github has a problem, I get a 500 when clicking Approve and run on the workflow, any ideas appart from waiting ?

hidaris:
@euri10 I just rebase the code, now ci works properly.

Kludex (approved):

PR #1028, opened 2021-05-04
Update docs about reload_dirs

bechtold:
It is confusing that the reload_dirs parameter when running programmatically, differs from the reload-dir when using CLI. So I think it would be good to document this.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -20,7 +20,7 @@ equivalent keyword arguments, eg. `uvicorn.run("example:app", port=5000, reload=
 ## Development
 
 * `--reload` - Enable auto-reload.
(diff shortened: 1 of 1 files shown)


euri10:
I'm unsure what you mean @bechtold , would you clarify this please ?

bechtold:
@euri10 In the docs it says "If you're running using programmatically, using uvicorn.run(...), then use equivalent keyword arguments". But this is not 100% true. For the `--reload-dir` that can be used multiple times when using the CLI, the equivalent for programmatically running it is `reload-dirs=[]` with a list. So there are slight differences between CLI and programmatically running. I had to look through the code to realize this. It would have saved me time if this was stated in the docs, so I thought I'll add it. I basically just added one sentence to the docs explaining the parameter for programmatically running :-)

Is that clearer now?

euri10 (approved):
clearer thanks

PR #1034, opened 2021-05-10
Moved FlowControl class and globals to its own file in uvicorn/protocols/http/

ShahriyarR:
As I have spotted most of the parts of `h11_impl.py` and `httptools_impl.py` files are shared. The most notable duplication is FlowControl class and "constants", my suggestion is to move this part to separate file `flow_control.py` and then use it, this will reduce the duplicated code portion.

### DIFF
--- a/uvicorn/protocols/http/flow_control.py
+++ b/uvicorn/protocols/http/flow_control.py
@@ -1,0 +1,54 @@
+import asyncio
+
+
+CLOSE_HEADER = (b"connection", b"close")
+
+HIGH_WATER_LIMIT = 65536
+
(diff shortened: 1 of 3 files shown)


euri10 (changes requested):
makes sense, thanks, just need to fix your imports I think
and you can run `scipts/lint` so that CI wont break on that

ShahriyarR:
@euri10 I have run the scripts/lint as mentioned and pushed the changes.

euri10 (approved):

euri10:
thanks a lot @ShahriyarR !

PR #1037, opened 2021-05-10
Support clean shutdown on Windows in hot reloader

### LINKED ISSUE #1036: Reload skips shutdown on Windows
The reloader uses `process.terminate()` which is a kill on Windows, so shutdown steps are skipped.

Korijn:
Closes #1036


Korijn:
Nevermind, this is a stupid idea, I've overlooked more Windows subtleties

PR #1033, opened 2021-05-09
Relax click version upper bound to 8.0

mariobriggs:
Requested in https://github.com/encode/uvicorn/issues/1016

Click 8.0 has security patches

Issue #1016, opened 2021-04-10
Relax click version upper bound


szeleeteo:
Uvicorn can only use [click](https://click.palletsprojects.com/en/7.x/) package version 7.* as defined in `setup.py`

``` python
minimal_requirements = [
    "click==7.*",
    ...
]
```

However, there is currently a pre-release version [8.0.0a1](https://pypi.org/project/click/8.0.0a1/) which fixes many [issues](https://github.com/pallets/click/blob/master/CHANGES.rst#version-80)

Would like to request for uvicorn to relax the upper bound for click requirement, such that user can optionally use click version 8 if applicable
``` python
minimal_requirements = [
    "click>=7.*"
    ...
]
```

mariobriggs:
opened a PR

epogrebnyak:
> However, there is currently a pre-release version [8.0.0a1](https://pypi.org/project/click/8.0.0a1/) which fixes many [issues](https://github.com/pallets/click/blob/master/CHANGES.rst#version-80)

Just to mention for the purposes of #1033 - Click 8.0.0 is a major release since May 11, so locking at 7.* needs an update.

epogrebnyak:
My temp solution is `poetry add click=7.*`

mariobriggs:
click v8.0.0 released 2 days ago

simonw:
This is blocking me from upgrading [Datasette](https://github.com/simonw/datasette) to Click 8.0, since Datasette has `uvicorn` as a dependency.

euri10:
closed by #1042

PR #1042, opened 2021-05-23
Allow Click 8.0, refs #1016

simonw:
Similar to #1033 but updates the CLI documentation to reflect new whitespace.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ env_marker_win = "sys_platform == 'win32'"
 env_marker_below_38 = "python_version < '3.8'"
 
 minimal_requirements = [
-    "click==7.*",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -44,7 +44,7 @@ def get_packages(package):
 env_marker_below_38 = "python_version < '3.8'"
 
 minimal_requirements = [
-    "click==7.*",
+    "click>=7.*",
```
simonw, line 47: I wonder if this should also specify `<9` to avoid an accidental upgrade to Click 9.0 some time in the future.

euri10: We have time I hope

euri10 (approved):

PR #1044, opened 2021-05-24
♻️ Replace _types module by asgiref.typing

Kludex:
Following what was discussed on https://github.com/encode/uvicorn/pull/991 about adding `asgiref`.

This PR only does what the title suggests, and adapts the current code in the easiest way so it doesn't break and pass mypy.


uvicorn/lifespan/on.py
```diff
@@ -31,7 +49,7 @@ async def startup(self) -> None:
         # Keep a hard reference to prevent garbage collection
         # See https://github.com/encode/uvicorn/pull/972
 
-        await self.receive_queue.put({"type": "lifespan.startup"})
+        await self.receive_queue.put(LifespanStartupEvent(type="lifespan.startup"))
```
Kludex: Not doing this gives me: 
```bash
uvicorn/lifespan/on.py:52: error: Type of TypedDict is ambiguous, could be any of ("LifespanStartupEvent", "LifespanShutdownEvent")
uvicorn/lifespan/on.py:52: error: Argument 1 to "put" of "Queue" has incompatible type "Dict[str, str]"; expected "Union[LifespanStartupEvent, LifespanShutdownEvent]"
```
If you have suggestions, I'm open to change it :kissing:

euri10: I think if you do something like 
```
lste : LifespanStartupEvent = {"type": "lifespan.startup"}
await self.receive_queue.put(lste)
```
it should be ok, though I have not tested it

euri10 (changes requested):

uvicorn/lifespan/on.py
Kludex: Yeah, it should work. But I don't see any real improvement. Is there any?

euri10: > Yeah, it should work. But I don't see any real improvement. Is there any?

well not having a mypy error is one in my mind, I'm no mypy pro but found the TypedDict experience quite frustrating and solved most of my issues by doing that, I'm not married to it though so if you have a better idea I'm all ears.

Kludex: Well, the idea would be the current implementation on the PR, but it's the same to me. I've never done what I propose on this PR tho, I always do what you suggest, so I'm going to change the implementation.

Kludex: Done

uvicorn/lifespan/on.py
```diff
@@ -30,8 +48,8 @@ async def startup(self) -> None:
         main_lifespan_task = loop.create_task(self.main())  # noqa: F841
         # Keep a hard reference to prevent garbage collection
         # See https://github.com/encode/uvicorn/pull/972
-
-        await self.receive_queue.put({"type": "lifespan.startup"})
+        startup_event: LifespanStartupEvent {"type": "lifespan.startup"}
```
euri10:
```suggestion
        startup_event: LifespanStartupEvent = {"type": "lifespan.startup"}
```

euri10: should help ;)

Kludex: haha :kissing:

euri10 (approved):
thanks !

PR #1049, opened 2021-05-29
Improved wsgi middleware, mainly a copy of a2wsgi

### LINKED ISSUE #371: WSGI middleware should stream request body, rather than loading it all at once.
Hi! I'm deploying a Django app with uvicorn, running on k8s. Our containers were being killed, and I've found that when users upload large files, uvicorn increases memory usage, and slows to a crawl. Eventually causing an OOM.

I'm not sure where this is happening yet, and I suspect could be related to https://github.com/django/channels/pull/1251, but I did a bit more digging and I'm not totally sure. I've tried running uvicorn in wsgi mode, and completely removed channels from the django install, and I'm getting the same behavior. File uploads are being loaded into memory (rather than streaming to disk, as they should be), and upload speeds slow to a crawl. The same app running on gunicorn works fine.

The file I'm testing with is about 470mb.

euri10:
Should fix https://github.com/encode/uvicorn/issues/371
I added types as well

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,8 @@ files =
     uvicorn/middleware/debug.py,
     uvicorn/supervisors/watchgodreload.py,
     uvicorn/logging.py,
-    uvicorn/middleware/asgi2.py
+    uvicorn/middleware/asgi2.py,
+    uvicorn/middleware/wsgi.py
 
 
 [mypy-tests.*]
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ files =
     uvicorn/_handlers,
     uvicorn/__init__.py,
     uvicorn/__main__.py,
-    uvicorn/subprocess.py
+    uvicorn/subprocess.py,
     uvicorn/middleware/wsgi.py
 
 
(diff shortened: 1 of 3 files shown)


Kludex:
I guess we only need to remove the types on the `wsgi.py` file and rebase?

abersheeran:
I think it might be more convenient to directly use [a2wsgi](https://github.com/abersheeran/a2wsgi) as a dependency. As far as I know, Uvicorn and Starlette both have WSGIMiddleware, which is a duplicate. And it will cause the changes in one of the items to not be quickly reflected in the other item.

euri10:
that's indeed a good idea, my only concern is the types, not sure how that would work since we're not using the same base

Kludex:
We probably should:
1. Sync with Starlette. Should we only remove from `uvicorn` or removing on `starlette` also makes sense?    
2. If we want to remove it from `uvicorn` only, add a `DeprecationWarning`, recommending [a2wsgi](https://github.com/abersheeran/a2wsgi). Then later on we can remove it.
3. Update documentation.

Flauschbaellchen:
Hi,

I run into the same problem as described in #371 as my app needs to support uploads of large files which are often 1GB or much bigger.
I would like to support here as much as I can. Do you need someone for testing purposes?

PR #1051, opened 2021-05-30
🚨 Cover mypy on  _handlers subpackage

Kludex:
I ignore or add `TYPE_CHECKING` here. I'm all ears if there's alternatives.


euri10 (commented):
lgtm this way, it's just for 3.6 so likely will get dropped soon enough (I hope)

euri10 (approved):

PR #1053, opened 2021-05-30
🚨 Cover subprocess.py on mypy

Kludex:


scripts/check
```diff
@@ -10,6 +10,6 @@ set -x
 
 ${PREFIX}black --check --diff --target-version=py36 $SOURCE_FILES
 ${PREFIX}flake8 $SOURCE_FILES
-${PREFIX}mypy
+${PREFIX}mypy --show-error-codes
```
Kludex, line 13: I forgot to remove this. Should I? @euri10

euri10:
actually I like it, I like the --pretty flag also, and we could use
--warn-unused-ignores since this is an incremental work and if we ignore
stuff one day maybe in the future that would not be true anymore

scripts/check
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh -e
 
 export PREFIX=""
-if [ -d 'venv' ] ; then
```
euri10: seems unrelated

euri10 (approved):
small glitch in the check script but lgtm otherwise, approved

scripts/check
```diff
@@ -1,7 +1,7 @@
...
+if [ -d 'venv' ]; then
```
Kludex:
```suggestion
if [ -d 'venv' ] ; then
```

scripts/check
Kludex: fixed

PR #1054, opened 2021-05-30
🚨 Cover server.py on mypy

Kludex:


Kludex:
This failed on windows:
```bash
uvicorn\server.py:142: error: Module has no attribute "start_unix_server"; maybe "start_server"?  [attr-defined]
```
I've just ignored it.

uvicorn/server.py
```diff
@@ -130,7 +139,7 @@ def _share_socket(sock: socket.SocketType) -> socket.SocketType:
             uds_perms = 0o666
             if os.path.exists(config.uds):
                 uds_perms = os.stat(config.uds).st_mode
-            server = await asyncio.start_unix_server(
+            server = await asyncio.start_unix_server(  # type: ignore
```
euri10: probably we can `assert sys.platform != 'win32'` instead

Kludex: I've tried that, and I've also tried `assert hasattr(asyncio, "start_unix_server")`. Neither of those works.

I'm not confident enough to just ignore now. Any other suggestion?

euri10: > I'm not confident enough to just ignore now. Any other suggestion?

I'm out of ideas, did you try to ask on https://gitter.im/python/typing ?

Kludex: I've just asked: https://gitter.im/python/typing?at=60d71d56110daa37b1085b19

Let's wait. :)

uvicorn/server.py
```diff
@@ -126,6 +135,7 @@ def _share_socket(sock: socket.SocketType) -> socket.SocketType:
             uds_perms = 0o666
             if os.path.exists(config.uds):
                 uds_perms = os.stat(config.uds).st_mode
+            assert hasattr(asyncio, "start_unix_server")  # mypy
```
graingert:
```python
            assert sys.platform != "win32"  # mypy
```

graingert: ah you can't use `assert sys.platform` you need `if sys.platform != "win32"`

graingert: https://github.com/python/mypy/issues/9025

graingert:
```python
        elif config.uds is not None:
            # Create a socket using UNIX domain socket.
            uds_perms = 0o666
            if os.path.exists(config.uds):
                uds_perms = os.stat(config.uds).st_mode
            if sys.platform != "win32":
                server = await asyncio.start_unix_server(
                    handler, path=config.uds, ssl=config.ssl, backlog=config.backlog
                )
                os.chmod(config.uds, uds_perms)
                assert server.sockets is not None  # mypy
                listeners = server.sockets
                self.servers = [server]
 ```

works for me when I run with `mypy --platform=win32`

Comment (graingert) [reply uvicorn/server.py]: This might be a bit neater?

```patch
From 2b977657b86649222550daf1957493dd6ca40fc5 Mon Sep 17 00:00:00 2001
From: Thomas Grainger <tagrain@gmail.com>
Date: Sat, 26 Jun 2021 14:56:44 +0100
Subject: [PATCH] conditional typecheck start_unix_server

---
 scripts/check     |  2 +-
 uvicorn/server.py | 12 +++++++++---
 2 files changed, 10 insertions(+), 4 deletions(-)

diff --git a/scripts/check b/scripts/check
index e319add..eed005d 100755
--- a/scripts/check
+++ b/scripts/check
@@ -10,6 +10,6 @@ set -x
 
 ${PREFIX}black --check --diff --target-version=py36 $SOURCE_FILES
 ${PREFIX}flake8 $SOURCE_FILES
-${PREFIX}mypy --show-error-codes
+${PREFIX}mypy --platform=win32 --show-error-codes
 ${PREFIX}isort --check --diff --project=uvicorn $SOURCE_FILES
 ${PREFIX}python -m tools.cli_usage --check
diff --git a/uvicorn/server.py b/uvicorn/server.py
index 9ac565b..8a5993c 100644
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -9,7 +9,7 @@ import threading
 import time
 from email.utils import formatdate
 from types import FrameType
-from typing import List, Optional, Set, Tuple, Union
+from typing import Any, List, Optional, Set, Tuple, Union
 
 import click
 
@@ -30,6 +30,13 @@ logger = logging.getLogger("uvicorn.error")
 
 Protocols = Union[H11Protocol, HttpToolsProtocol, WSProtocol, WebSocketProtocol]
 
+if sys.platform != "win32":
+    from asyncio import start_unix_server as _start_unix_server
+else:
+
+    async def _start_unix_server(*args: object, **kwargs: object) -> Any:
+        raise NotImplementedError("Cannot start a unix server on win32")
+
 
 class ServerState:
     """
@@ -135,8 +142,7 @@ class Server:
             uds_perms = 0o666
             if os.path.exists(config.uds):
                 uds_perms = os.stat(config.uds).st_mode
-            assert hasattr(asyncio, "start_unix_server")  # mypy
-            server = await asyncio.start_unix_server(
+            server = await _start_unix_server(
                 handler, path=config.uds, ssl=config.ssl, backlog=config.backlog
             )
             os.chmod(config.uds, uds_perms)
-- 
2.32.0
```

Kludex: Thank you, @graingert ! Precise as always 😎

I do prefer your solution than using assert (it's not like I have an alternative 😅). But I wonder, isn't it too much to please `mypy`? Does my question even makes sense? 🤔

Anyway, I'm going to update this branch with the suggestion as soon as I get on my computer.

graingert: note AF_UNIX may come to win32 soon: https://bugs.python.org/issue33408

euri10 (approved):
thanks @Kludex and @graingert for the help !

PR #1057, opened 2021-05-31
Coverage logging

euri10:
Small grind towards #102


uvicorn/logging.py
```diff
@@ -44,13 +44,13 @@ def __init__(
 
     def color_level_name(self, level_name: str, level_no: int) -> str:
         def default(level_name: str) -> str:
-            return str(level_name)
+            return str(level_name)  # pragma: no cover
```
euri10, line 47: both this and the other def default most likely wont be hit since the protocol will likely crash on any status code > 600, we can remove it entirely and tests pass, but this wont hurt not covering them

Kludex (approved):

PR #1062, opened 2021-05-31
🔨 Add precision option on coverage command

Kludex:
To achieve 100% coverage, we should always increase that number. But as the number gets close to the goal, it may happen that we don't take some PRs into account and actually decrease its value.

This PR adds a `precision` parameter to be sure that we're not decreasing it by mistake.

### DIFF
--- a/scripts/coverage
+++ b/scripts/coverage
@@ -8,4 +8,4 @@ export SOURCE_FILES="uvicorn tests"
 
 set -x
 
-${PREFIX}coverage report --show-missing --skip-covered --fail-under=95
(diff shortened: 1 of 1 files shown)


Kludex:
I've found not that useful, as we have 95.73% now, but when I set `--fail-under=96` it passes :angry: 
On the [documentation](https://coverage.readthedocs.io/en/coverage-5.5/config.html#report) it states that `precision` affects `fail-under`... So I don't know why it doesn't fail with the numbers I've mentioned.

Anyway, I'm closing it.

PR #1065, opened 2021-06-01
Upgraded websockets requirements

euri10:
https://websockets.readthedocs.io/en/stable/changelog.html#id2


Kludex (approved):
Just for future reference, in case someone is lazy to open your link:
image

Issue #998, opened 2021-03-10
Type Annotation and mypy support


Kludex:
### Checklist

- [X] There are no similar issues or pull requests for this yet.
- [x] I discussed this idea on the [community chat](https://gitter.im/encode/community) and feedback is positive.

### Summary

I've seen somewhere here that we want to have mypy. To be more organized, I've created the table below, so we can divide and conquer :sunglasses: :+1: 

I'll be creating the PRs for those files in the following days.

### All files

    uvicorn
    ├── config.py ✅
    ├── _handlers ✅
    ├── importer.py ✅
    ├── __init__.py ✅
    ├── lifespan ✅
    ├── logging.py ✅
    ├── loops ✅
    ├── __main__.py ✅
    ├── main.py ✅
    ├── middleware ✅
    ├── protocols ➖
    │   ├── http ✅ 
    │   ├── __init__.py ✅
    │   ├── utils.py ✅
    │   └── websockets ➖
    │       ├── auto.py ✅
    │       ├── __init__.py ✅
    │       ├── websockets_impl.py ✅ 
    │       └── wsproto_impl.py ➖
    ├── server.py ✅
    ├── subprocess.py ✅
    ├── supervisors ✅
    └── workers.py ✅
    
✅: Merged.
➖: PR available / WIP.
❌: Available to pick.

EDIT: I've simplified the report.

jkklapp:
@Kludex are the files without a PR associated "up for grabs"?

Kludex:
Yes. But I think both h11 and httptools implementation files need others to be merged before (not sure). 

But yeah, you can go for it. Ping me on that PR once you've finished so I can update the list above, please. Also, be aware that we're trying to match this: https://github.com/django/asgiref/blob/main/asgiref/typing.py, which is on the main branch, but it was not released yet. 

EDIT: Looks like there will be a release on `asgiref` today. So we can start using it.

br3ndonland:
Thanks for initiating this, @Kludex! 👏 

It would be great to have type annotations in this project, and I'm happy to help. I'll start contributing PRs and code reviews as time allows.

br3ndonland:
This blog post by @tomchristie inspired me to be more diligent about using type annotations (and asyncio):

[Python async frameworks - Beyond developer tribalism](https://www.encode.io/articles/python-async-frameworks-beyond-developer-tribalism)

[...]

PR #1067, opened 2021-06-02
Type-annotate uvicorn/config.py

### LINKED ISSUE #1046: "ModuleType object is not callable" warnings from 'importer'-imported modules
### Checklist

### Importer => `ModuleType`
https://github.com/encode/uvicorn/blob/06fedab6c1da9ee2bfc3cee635fc772a9958a9c6/uvicorn/importer.py#L10
This has raised `'ModuleType' object is not callable` warnings all over, in my IDE (PyCharm).

For example:
- https://github.com/encode/uvicorn/blob/06fedab6c1da9ee2bfc3cee635fc772a9958a9c6/uvicorn/config.py#L319
- https://github.com/encode/uvicorn/blob/06fedab6c1da9ee2bfc3cee635fc772a9958a9c6/uvicorn/config.py#L359-L361
<hr>

- Is this a valid issue? 😕
- Should we switch to `Any` as return type annotation?

br3ndonland:
## Description

This PR will add type annotations and mypy type checking to _uvicorn/config.py_, the associated tests in _tests/test_config.py_, and the related code that is impacted by the new type annotations.

Supersedes the changes to _uvicorn/config.py_ in #992, and closes #1046.

### _uvicorn/config.py_

- Add changes from PR #992 (0db3bc2): This commit will bring in changes to _uvicorn/config.py_ added by @Kludex in PR #992, updating for the latest master branch.
- Correct SSL type annotations (eee7fba)
  - `certfile` is a required positional argument when running [`SSLContext.load_cert_chain`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain), so annotating as `Optional` (which allows `None`) would not be ideal. Path-like objects are acceptable, so after `from pathlib import Path`, the annotation is `certfile: Union[Path, str]`.
  - `if self.is_ssl and self.ssl_certfile` will help ensure that the `self.ssl_certfile` required positional argument is present.
- Simplify `log_config` type annotation (b2a1548)
- Add type annotation for app (12d323c):
  - app is sometimes a string, as described in _uvicorn/main.py_. In other cases, especially in the tests, app is an ASGI application callable.
  - This commit will add a type annotation for the app argument to address these use cases.
- Simplify if expression in `Config().headers` (b188fd1)
- Allow paths to be used for `Config(env_file)` (d31fddd)
- Add asyncio Protocol types to class `Config` kwargs (9c8e70d): Protocol classes are sometimes used for the app kwarg, such as in _tests/test_config.py_.
- Improve use of `Literal` type for ASGI interfaces (14cb546)
  - This commit will retain the use of [literals](https://mypy.readthedocs.io/en/stable/literal_types.html) to define ASGI protocol versions, but improve and correct the use of literal types.
  - As described in the mypy docs, `Literal["2.0", "3.0"]` is a simpler way to write `Union[Literal["2.0"], Literal["3.0"]]`.
- Add type annotation to `Config` kwargs in _uvicorn/workers.py_ (459544d)
- Add type annotations to _tests/test_config.py_ (c436bba)
- Type-annotate constants in _uvicorn/config.py_ (8d2cd6f): Helps prevent mypy `[has-type]` errors ("Cannot determine type of {object}")

### _uvicorn/importer.py_

Correct type annotations in _uvicorn/importer.py_ (e9e59e9)

The relationship between module imports and calls in _uvicorn/config.py_ was previously unclear. In encode/uvicorn#991, `import_from_string` was annotated as returning a `ModuleType`, but the function is used to return callables and other types. Mypy was raising "module not callable" errors, as reported in #1046.

https://github.com/encode/uvicorn/blob/87da6cf4082cd28306bdb78d7d42552d131cc179/uvicorn/config.py#L317-L325

Current use cases of `import_from_string` include:

- `uvicorn.Config.HTTP_PROTOCOLS`, `uvicorn.Config.WS_PROTOCOLS`: `Type[asyncio.Protocol]` (these are subclasses of asyncio protocols, so they are annotated with `Type[Class]` as explained in the [mypy docs](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#the-type-of-class-objects))
- `uvicorn.Config.LOOP_SETUPS`: `Callable` (each of the loops is a function)
- `uvicorn.Config().loaded_app`: `ASGIApplication` (should be an ASGI application instance, like `Starlette()` or `FastAPI()`)

Ideally, the return type of `import_from_string` would reflect these use cases, but the complex typing will be difficult to maintain. For easier maintenance, the return type will be `Any`, and objects returned by `import_from_string` will be annotated directly.

## Related

encode/uvicorn#990
encode/uvicorn#991
encode/uvicorn#992
encode/uvicorn#998
encode/uvicorn#1046

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,6 +15,7 @@ isort
 pytest
 pytest-mock
 mypy
+types-pyyaml
 trustme
 cryptography
 coverage
--- a/tests/importer/test_importer.py
+++ b/tests/importer/test_importer.py
@@ -3,40 +3,40 @@ import pytest
 from uvicorn.importer import ImportFromStringError, import_from_string
 
 
-def test_invalid_format():
+def test_invalid_format() -> None:
     with pytest.raises(ImportFromStringError) as exc_info:
         import_from_string("example:")
     expected = 'Import string "example:" must be in format "<module>:<attribute>".'
     assert expected in str(exc_info.value)
 
 
-def test_invalid_module():
+def test_invalid_module() -> None:
     with pytest.raises(ImportFromStringError) as exc_info:
         import_from_string("module_does_not_exist:myattr")
     expected = 'Could not import module "module_does_not_exist".'
     assert expected in str(exc_info.value)
 
 
-def test_invalid_attr():
+def test_invalid_attr() -> None:
     with pytest.raises(ImportFromStringError) as exc_info:
         import_from_string("tempfile:attr_does_not_exist")
     expected = 'Attribute "attr_does_not_exist" not found in module "tempfile".'
     assert expected in str(exc_info.value)
 
 
-def test_internal_import_error():
+def test_internal_import_error() -> None:
     with pytest.raises(ImportError):
         import_from_string("tests.importer.raise_import_error:myattr")
 
 
-def test_valid_import():
+def test_valid_import() -> None:
     instance = import_from_string("tempfile:TemporaryFile")
     from tempfile import TemporaryFile
 
     assert instance == TemporaryFile
 
 
-def test_no_import_needed():
+def test_no_import_needed() -> None:
     from tempfile import TemporaryFile
 
     instance = import_from_string(TemporaryFile)
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -3,10 +3,10 @@ import logging
 import os
 import socket
 import sys
+import typing
 from copy import deepcopy
 from pathlib import Path
 from types import TracebackType
-from typing import Callable, Iterable, Iterator, MutableMapping, Optional, Tuple, Type
 from unittest.mock import MagicMock
 
 if sys.version_info < (3, 8):
@@ -25,8 +25,13 @@ from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
 from uvicorn.middleware.wsgi import WSGIMiddleware
 from uvicorn.protocols.http.h11_impl import H11Protocol
 
-ExcInfo = Tuple[Type[BaseException], BaseException, Optional[TracebackType]]
-StartResponse = Callable[[str, Iterable[Tuple[str, str]], Optional[ExcInfo]], None]
+Environ = typing.MutableMapping[str, typing.Any]
+ExcInfo = typing.Tuple[
+    typing.Type[BaseException], BaseException, typing.Optional[TracebackType]
+]
+StartResponse = typing.Callable[
+    [str, typing.Iterable[typing.Tuple[str, str]], typing.Optional[ExcInfo]], None
+]
 
 
 @pytest.fixture
@@ -55,7 +60,7 @@ async def asgi_app(
     pass  # pragma: nocover
 
 
-def wsgi_app(environ: MutableMapping, start_response: StartResponse) -> None:
+def wsgi_app(environ: Environ, start_response: StartResponse) -> None:
     pass  # pragma: nocover
 
 
@@ -188,7 +193,7 @@ def test_ssl_config_combined(tls_certificate_pem_path: str) -> None:
     assert config.is_ssl is True
 
 
-def asgi2_app(scope: Scope) -> Callable:
+def asgi2_app(scope: Scope) -> typing.Callable:
     async def asgi(
         receive: ASGIReceiveCallable, send: ASGISendCallable
     ) -> None:  # pragma: nocover
@@ -219,8 +224,8 @@ def test_asgi_version(
 )
 def test_log_config_default(
     mocked_logging_config_module: MagicMock,
-    use_colors: Optional[bool],
-    expected: Optional[bool],
+    use_colors: typing.Optional[bool],
+    expected: typing.Optional[bool],
 ) -> None:
     """
     Test that one can specify the use_colors option when using the default logging
@@ -290,14 +295,14 @@ def test_log_config_file(mocked_logging_config_module: MagicMock) -> None:
 
 
 @pytest.fixture(params=[0, 1])
-def web_concurrency(request: pytest.FixtureRequest) -> Iterator[int]:
+def web_concurrency(request: pytest.FixtureRequest) -> typing.Iterator[int]:
     yield getattr(request, "param")
     if os.getenv("WEB_CONCURRENCY"):
         del os.environ["WEB_CONCURRENCY"]
 
 
 @pytest.fixture(params=["127.0.0.1", "127.0.0.2"])
-def forwarded_allow_ips(request: pytest.FixtureRequest) -> Iterator[str]:
+def forwarded_allow_ips(request: pytest.FixtureRequest) -> typing.Iterator[str]:
     yield getattr(request, "param")
     if os.getenv("FORWARDED_ALLOW_IPS"):
         del os.environ["FORWARDED_ALLOW_IPS"]
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -2,10 +2,21 @@ import json
 import logging
 import os
 import socket
+import sys
 from copy import deepcopy
+from pathlib import Path
+from typing import Any, Callable, Iterator, Optional
+from unittest.mock import MagicMock
+
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal
+else:
+    from typing import Literal
 
 import pytest
 import yaml
+from asgiref.typing import ASGIApplication, ASGIReceiveCallable, ASGISendCallable, Scope
+from pytest_mock import MockerFixture
 
 from uvicorn.config import LOGGING_CONFIG, Config
 from uvicorn.middleware.debug import DebugMiddleware
@@ -15,34 +26,36 @@ from uvicorn.protocols.http.h11_impl import H11Protocol
 
 
 @pytest.fixture
-def mocked_logging_config_module(mocker):
+def mocked_logging_config_module(mocker: MockerFixture) -> MagicMock:
     return mocker.patch("logging.config")
 
 
 @pytest.fixture(scope="function")
-def logging_config():
+def logging_config() -> dict:
     return deepcopy(LOGGING_CONFIG)
 
 
 @pytest.fixture
-def json_logging_config(logging_config):
+def json_logging_config(logging_config: dict) -> str:
     return json.dumps(logging_config)
 
 
 @pytest.fixture
-def yaml_logging_config(logging_config):
+def yaml_logging_config(logging_config: dict) -> str:
     return yaml.dump(logging_config)
 
 
-async def asgi_app(scope, receive, send):
+async def asgi_app(
+    scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable
+) -> None:
     pass  # pragma: nocover
 
 
-def wsgi_app(environ, start_response):
+def wsgi_app(environ: Any, start_response: Any) -> None:
     pass  # pragma: nocover
 
 
-def test_debug_app():
+def test_debug_app() -> None:
     config = Config(app=asgi_app, debug=True, proxy_headers=False)
     config.load()
 
@@ -54,7 +67,9 @@ def test_debug_app():
     "app, expected_should_reload",
     [(asgi_app, False), ("tests.test_config:asgi_app", True)],
 )
-def test_config_should_reload_is_set(app, expected_should_reload):
+def test_config_should_reload_is_set(
+    app: ASGIApplication, expected_should_reload: bool
+) -> None:
     config_debug = Config(app=app, debug=True)
     assert config_debug.debug is True
     assert config_debug.should_reload is expected_should_reload
@@ -64,12 +79,12 @@ def test_config_should_reload_is_set(app, expected_should_reload):
     assert config_reload.should_reload is expected_should_reload
 
 
-def test_reload_dir_is_set():
+def test_reload_dir_is_set() -> None:
     config = Config(app=asgi_app, reload=True, reload_dirs="reload_me")
     assert config.reload_dirs == ["reload_me"]
 
 
-def test_wsgi_app():
+def test_wsgi_app() -> None:
     config = Config(app=wsgi_app, interface="wsgi", proxy_headers=False)
     config.load()
 
@@ -78,7 +93,7 @@ def test_wsgi_app():
     assert config.asgi_version == "3.0"
 
 
-def test_proxy_headers():
+def test_proxy_headers() -> None:
     config = Config(app=asgi_app)
     config.load()
 
@@ -86,13 +101,13 @@ def test_proxy_headers():
     assert isinstance(config.loaded_app, ProxyHeadersMiddleware)
 
 
-def test_app_unimportable_module():
+def test_app_unimportable_module() -> None:
     config = Config(app="no.such:app")
     with pytest.raises(ImportError):
         config.load()
 
 
-def test_app_unimportable_other(caplog):
+def test_app_unimportable_other(caplog: pytest.LogCaptureFixture) -> None:
     config = Config(app="tests.test_config:app")
     with pytest.raises(SystemExit):
         config.load()
@@ -107,8 +122,8 @@ def test_app_unimportable_other(caplog):
     )
 
 
-def test_app_factory(caplog):
-    def create_app():
+def test_app_factory(caplog: pytest.LogCaptureFixture) -> None:
+    def create_app() -> ASGIApplication:
         return asgi_app
 
     config = Config(app=create_app, factory=True, proxy_headers=False)
@@ -125,19 +140,19 @@ def test_app_factory(caplog):
     assert len(caplog.records) == 1
     assert "--factory" in caplog.records[0].message
 
-    # App not a no-arguments callable.
+    # App not a no-arguments ASGIApplication.
     config = Config(app=asgi_app, factory=True)
     with pytest.raises(SystemExit):
         config.load()
 
 
-def test_concrete_http_class():
+def test_concrete_http_class() -> None:
     config = Config(app=asgi_app, http=H11Protocol)
     config.load()
     assert config.http_protocol_class is H11Protocol
 
 
-def test_socket_bind():
+def test_socket_bind() -> None:
     config = Config(app=asgi_app)
     config.load()
     sock = config.bind_socket()
@@ -145,7 +160,10 @@ def test_socket_bind():
     sock.close()
 
 
-def test_ssl_config(tls_ca_certificate_pem_path, tls_ca_certificate_private_key_path):
+def test_ssl_config(
+    tls_ca_certificate_pem_path: str,
+    tls_ca_certificate_private_key_path: str,
+) -> None:
     config = Config(
         app=asgi_app,
         ssl_certfile=tls_ca_certificate_pem_path,
@@ -156,7 +174,7 @@ def test_ssl_config(tls_ca_certificate_pem_path, tls_ca_certificate_private_key_
     assert config.is_ssl is True
 
 
-def test_ssl_config_combined(tls_certificate_pem_path):
+def test_ssl_config_combined(tls_certificate_pem_path: str) -> None:
     config = Config(
         app=asgi_app,
         ssl_certfile=tls_certificate_pem_path,
@@ -166,8 +184,10 @@ def test_ssl_config_combined(tls_certificate_pem_path):
     assert config.is_ssl is True
 
 
-def asgi2_app(scope):
-    async def asgi(receive, send):  # pragma: nocover
+def asgi2_app(scope: Scope) -> Callable:
+    async def asgi(
+        receive: ASGIReceiveCallable, send: ASGISendCallable
+    ) -> None:  # pragma: nocover
         pass
 
     return asgi  # pragma: nocover
@@ -176,7 +196,9 @@ def asgi2_app(scope):
 @pytest.mark.parametrize(
     "app, expected_interface", [(asgi_app, "3.0"), (asgi2_app, "2.0")]
 )
-def test_asgi_version(app, expected_interface):
+def test_asgi_version(
+    app: ASGIApplication, expected_interface: Literal["2.0", "3.0"]
+) -> None:
     config = Config(app=app)
     config.load()
     assert config.asgi_version == expected_interface
@@ -191,7 +213,11 @@ def test_asgi_version(app, expected_interface):
         pytest.param(False, False, id="use_colors_disabled"),
     ],
 )
-def test_log_config_default(mocked_logging_config_module, use_colors, expected):
+def test_log_config_default(
+    mocked_logging_config_module: MagicMock,
+    use_colors: Optional[bool],
+    expected: Optional[bool],
+) -> None:
     """
     Test that one can specify the use_colors option when using the default logging
     config.
@@ -206,8 +232,11 @@ def test_log_config_default(mocked_logging_config_module, use_colors, expected):
 
 
 def test_log_config_json(
-    mocked_logging_config_module, logging_config, json_logging_config, mocker
-):
+    mocked_logging_config_module: MagicMock,
+    logging_config: dict,
+    json_logging_config: str,
+    mocker: MockerFixture,
+) -> None:
     """
     Test that one can load a json config from disk.
     """
@@ -224,12 +253,12 @@ def test_log_config_json(
 
 @pytest.mark.parametrize("config_filename", ["log_config.yml", "log_config.yaml"])
 def test_log_config_yaml(
-    mocked_logging_config_module,
-    logging_config,
-    yaml_logging_config,
-    mocker,
-    config_filename,
-):
+    mocked_logging_config_module: MagicMock,
+    logging_config: dict,
+    yaml_logging_config: str,
+    mocker: MockerFixture,
+    config_filename: str,
+) -> None:
     """
     Test that one can load a yaml config from disk.
     """
@@ -244,7 +273,7 @@ def test_log_config_yaml(
     mocked_logging_config_module.dictConfig.assert_called_once_with(logging_config)
 
 
-def test_log_config_file(mocked_logging_config_module):
+def test_log_config_file(mocked_logging_config_module: MagicMock) -> None:
     """
     Test that one can load a configparser config from disk.
     """
@@ -257,20 +286,25 @@ def test_log_config_file(mocked_logging_config_module):
 
 
 @pytest.fixture(params=[0, 1])
-def web_concurrency(request):
+def web_concurrency(request: Any) -> Iterator[int]:
     yield request.param
     if os.getenv("WEB_CONCURRENCY"):
         del os.environ["WEB_CONCURRENCY"]
 
 
 @pytest.fixture(params=["127.0.0.1", "127.0.0.2"])
-def forwarded_allow_ips(request):
+def forwarded_allow_ips(request: Any) -> Iterator[str]:
     yield request.param
     if os.getenv("FORWARDED_ALLOW_IPS"):
         del os.environ["FORWARDED_ALLOW_IPS"]
 
 
-def test_env_file(web_concurrency: int, forwarded_allow_ips: str, caplog, tmp_path):
+def test_env_file(
+    web_concurrency: int,
+    forwarded_allow_ips: str,
+    caplog: pytest.LogCaptureFixture,
+    tmp_path: Path,
+) -> None:
     """
     Test that one can load environment variables using an env file.
     """
@@ -284,7 +318,7 @@ def test_env_file(web_concurrency: int, forwarded_allow_ips: str, caplog, tmp_pa
         config = Config(app=asgi_app, env_file=fp)
         config.load()
 
-    assert config.workers == int(os.getenv("WEB_CONCURRENCY"))
+    assert config.workers == int(str(os.getenv("WEB_CONCURRENCY")))
     assert config.forwarded_allow_ips == os.getenv("FORWARDED_ALLOW_IPS")
     assert len(caplog.records) == 1
     assert f"Loading environment from '{fp}'" in caplog.records[0].message
@@ -297,7 +331,7 @@ def test_env_file(web_concurrency: int, forwarded_allow_ips: str, caplog, tmp_pa
         pytest.param(False, 0, id="access log disabled shouldn't have handlers"),
     ],
 )
-def test_config_access_log(access_log: bool, handlers: int):
+def test_config_access_log(access_log: bool, handlers: int) -> None:
     config = Config(app=asgi_app, access_log=access_log)
     config.load()
 
@@ -306,7 +340,7 @@ def test_config_access_log(access_log: bool, handlers: int):
 
 
 @pytest.mark.parametrize("log_level", [5, 10, 20, 30, 40, 50])
-def test_config_log_level(log_level):
+def test_config_log_level(log_level: int) -> None:
     config = Config(app=asgi_app, log_level=log_level)
     config.load()
 
@@ -316,7 +350,7 @@ def test_config_log_level(log_level):
     assert config.log_level == log_level
 
 
-def test_ws_max_size():
+def test_ws_max_size() -> None:
     config = Config(app=asgi_app, ws_max_size=1000)
     config.load()
     assert config.ws_max_size == 1000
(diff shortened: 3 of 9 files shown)


Kludex:
Yeah, the `import_from_string` probably needs to be fixed.

Somehow related: https://github.com/encode/uvicorn/issues/1046

br3ndonland:
> Yeah, the `import_from_string` probably needs to be fixed.
> 
> Somehow related: #1046

_EDIT:_ I attempted to add a type alias for the return type of `import_from_string`, but there were many complex mypy errors and it was difficult to manage. Rather than trying to be specific with the return type of `import_from_string`, I made the return type `Any`, and shifted the specific type annotations to the objects returned by calls to `import_from_string`. Should be easier to maintain.

Mypy and tests are passing on my fork.

tests/importer/test_importer.py
```diff
@@ -3,41 +3,41 @@
...
 
-def test_no_import_needed():
-    from tempfile import TemporaryFile
+def test_no_import_needed() -> None:
+    from uvicorn.lifespan import on
 
-    instance = import_from_string(TemporaryFile)
-    assert instance == TemporaryFile
+    instance = import_from_string(on)
+    assert instance == on
```
euri10: not sure if it's the diff or me but I dont understand the change

tests/test_config.py
```diff
@@ -64,12 +67,12 @@ def test_config_should_reload_is_set(app, expected_should_reload):
     assert config_reload.should_reload is expected_should_reload
 
 
-def test_reload_dir_is_set():
-    config = Config(app=asgi_app, reload=True, reload_dirs="reload_me")
+def test_reload_dir_is_set() -> None:
+    config = Config(app=asgi_app, reload=True, reload_dirs=["reload_me"])
```
euri10: interesting, why we need to change it to a list ?

uvicorn/config.py
```diff
@@ -107,8 +109,14 @@
 
 
 def create_ssl_context(
-    certfile, keyfile, password, ssl_version, cert_reqs, ca_certs, ciphers
-):
+    certfile: Union[Path, str],
```
euri10: why using different types here and below than those provided in the typeshed ?
```suggestion
    certfile: StrPath,
```

uvicorn/config.py
```diff
@@ -123,45 +131,45 @@ def create_ssl_context(
...
+        port: int = 8000,
+        uds: Optional[str] = None,
+        fd: Optional[int] = None,
+        loop: str = "auto",
+        http: Union[Type[asyncio.Protocol], str] = "auto",
+        ws: Union[Type[asyncio.Protocol], str] = "auto",
+        ws_max_size: int = 16 * 1024 * 1024,
+        lifespan: str = "auto",
+        env_file: Optional[Union[Path, str]] = None,
+        log_config: Optional[Union[dict, str]] = LOGGING_CONFIG,
```
euri10, line 152: can be a file too

uvicorn/config.py
```diff
@@ -300,22 +313,27 @@ def load(self):
...
             self.ws_protocol_class = self.ws
 
         self.lifespan_class = import_from_string(LIFESPAN[self.lifespan])
 
         try:
-            self.loaded_app = import_from_string(self.app)
+            if isinstance(self.app, str):
+                self.loaded_app = import_from_string(self.app)
+            else:
+                self.loaded_app = self.app
```
euri10: why do we do that ? afair it's done already in the import_from_string def, isn't it ?

tests/importer/test_importer.py
br3ndonland: Yeah, this was to test the type annotation added in encode/uvicorn#991 (which said `import_from_str` would accept either a module or string). I relaxed the type annotation to just accept `Any`, and reverted the change here.

tests/test_config.py
br3ndonland: Thanks for pointing this out. After reading through encode/uvicorn#978, I understand this better, and I reverted the change here.

If the user passes in a string, they can only pass in a single directory though, right? `Config(reload_dirs="reload_me, reload_me_too")` would result in `["reload_me, reload_me_too"]`.

uvicorn/config.py
br3ndonland: `StrPath` is in `_typeshed`, and is not available at runtime by default, so attempting to annotate with `StrPath` may raise `error: Name 'StrPath' is not defined [name-defined]`.

I'm happy to consider it, though. How would I import and use `StrPath`?

uvicorn/config.py
br3ndonland: Yes, but the file path would be coming in as a string, right? The log config logic assumes string in some places, like where `.endswith()` is used:

https://github.com/encode/uvicorn/blob/87da6cf4082cd28306bdb78d7d42552d131cc179/uvicorn/config.py#L252-L259

uvicorn/config.py
br3ndonland: Yeah, good point, we can just let `import_from_str` handle this. Reverted.

euri10:
/rant on
this Approve and run button is really annoying
/rant off

tests/test_config.py
euri10: yes, with click we have reload-dir that can be set multiple times, in which case click will set it as a tuple:
`uvicorn app:app --reload --reload-dir=dir1 --reload-dir=dir2` => `reload_dirs = ('dir1', 'dir2')` which is then passed onto Config
if it's just set one time as a string then we'll force it to be a list in the Config init, should be a tuple now but cant bother

uvicorn/config.py
euri10: yes you're right, I had in minc click.Path could give you a Path object but it's not the case

uvicorn/config.py
euri10: something like this ?
```
def create_ssl_context(
    certfile: Union[str, os.PathLike[str]],
    keyfile: Optional[Union[str, os.PathLike[str]]],
    password: Optional[str],
    ssl_version: int,
    cert_reqs: int,
    ca_certs: Optional[Union[str, os.PathLike[str]]],
    ciphers: Optional[str],
```

br3ndonland: Sure, that works -> a2d9085. It's not subscriptable, so just has to be `os.PathLike` instead of `os.PathLike[str]`.

br3ndonland:
> /rant on
> this Approve and run button is really annoying
> /rant off

Hopefully almost done 🤞

tests/test_config.py
```diff
@@ -15,34 +26,36 @@
...
 
-async def asgi_app(scope, receive, send):
+async def asgi_app(
+    scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable
+) -> None:
     pass  # pragma: nocover
 
 
-def wsgi_app(environ, start_response):
+def wsgi_app(environ: Any, start_response: Any) -> None:
```
euri10: I've made a draft PR that touched a little bit on wsgi app and came up with
```
Environ = typing.MutableMapping[str, typing.Any]
ExcInfo = typing.Tuple[    typing.Type[BaseException], BaseException, typing.Optional[TracebackType]]
StartResponse = typing.Callable[[str, typing.Iterable[typing.Tuple[str, str]], typing.Optional[ExcInfo]], None]
```

```suggestion
def wsgi_app(environ: Environ, start_response: StartResponse) -> None:
```

tests/test_config.py
```diff
@@ -257,20 +286,25 @@ def test_log_config_file(mocked_logging_config_module):
...
-def web_concurrency(request):
+def web_concurrency(request: Any) -> Iterator[int]:
     yield request.param
     if os.getenv("WEB_CONCURRENCY"):
         del os.environ["WEB_CONCURRENCY"]
 
 
 @pytest.fixture(params=["127.0.0.1", "127.0.0.2"])
-def forwarded_allow_ips(request):
+def forwarded_allow_ips(request: Any) -> Iterator[str]:
```
euri10: not sure about this, worth checking if Any could be avoided
```suggestion
def forwarded_allow_ips(request: FixtureRequest) -> Iterator[str]:
```

euri10 (commented):
another round of comments, sorry this is a massive PR so I was not able to touch on everything at first sight :)
but we're getting there !

tests/test_config.py
br3ndonland: These are good suggestions. For the `Environ` type, I would modify it slightly, either simply `typing.MutableMapping` (which is what Starlette uses in [starlette/config.py](https://github.com/encode/starlette/blob/b6f3578bb2cf6c60e3efe110143409b47f368d36/starlette/config.py#L16)), or `typing.MutableMapping[str, str]`, based on [`os.environ`](https://github.com/python/cpython/blob/3e1c7167d86a2a928cdcb659094aa10bb5550c4c/Lib/os.py#L737) (which appears to expect all keys and values to be strings).

I went with `typing.MutableMapping`, hopefully that's cool.

tests/test_config.py
br3ndonland: I agree, it's good to be more specific than `Any`. `pytest.FixtureRequest` works. The `param` attribute is [optional](https://docs.pytest.org/en/latest/reference/reference.html#pytest.FixtureRequest), so mypy requires a check for the attribute before indexing into it, like `getattr(request, "param")`.

requirements.txt
```diff
@@ -15,6 +15,7 @@ isort
 pytest
 pytest-mock
 mypy
+types-pyyaml
```
br3ndonland, line 19: mypy is complaining about the yaml imports now in GitHub Actions:

```text
mypy --show-error-codes

uvicorn/config.py:21: error: Library stubs not installed for "yaml" (or incompatible with Python 3.8)  [import]
uvicorn/config.py:21: note: Hint: "python3 -m pip install types-PyYAML"
uvicorn/config.py:21: note: (or run "mypy --install-types" to install all missing stub packages)
uvicorn/config.py:21: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
tests/test_config.py:18: error: Library stubs not installed for "yaml" (or incompatible with Python 3.8)  [import]
```

Running `mypy --install-types` didn't seem to fix it (in GitHub Actions at least), so I installed the `types-pyyaml` package from the typeshed. Otherwise, we would have to add ignores, like `import yaml # type: ignore[import]`.

br3ndonland:
> another round of comments, sorry this is a massive PR so I was not able to touch on everything at first sight :)
> but we're getting there !

Cool, thanks for your reviews! Good feedback.

requirements.txt
Kludex: Mypy 0.900 was released yesterday, and we don't pin the mypy version, jfyk.
https://mypy-lang.blogspot.com/2021/06/mypy-0900-released.html

EDIT: You have the message on the GitHub Actions and not locally because you need to upgrade it.

Kludex: Can you try `mypy --install-types` instead of adding `types-pyyaml`?

euri10: kind of happy this doesnt blow up more than that !

tests/test_config.py
euri10: the starlette Environ you're referring to has nothing to do with the Environ here I think, appart from the name, the latter is for configuring starlette iirc, the former is the dict for the wsgi spec (https://www.python.org/dev/peps/pep-0333/#environ-variables)
I think `typing.MutableMapping[str, str]` is incorrect since one of the values it can take are bytes, for the wsgi.input key

tests/test_config.py
```diff
@@ -166,8 +188,10 @@ def test_ssl_config_combined(tls_certificate_pem_path):
     assert config.is_ssl is True
 
 
-def asgi2_app(scope):
-    async def asgi(receive, send):  # pragma: nocover
+def asgi2_app(scope: Scope) -> Callable:
```
Kludex:
```suggestion
def asgi2_app(scope: Scope) -> Callable[[ASGIReceiveCallable, ASGISendCallable], Awaitable[None]]:
```

tests/test_config.py
```diff
@@ -284,7 +322,7 @@ def test_env_file(web_concurrency: int, forwarded_allow_ips: str, caplog, tmp_pa
         config = Config(app=asgi_app, env_file=fp)
         config.load()
 
-    assert config.workers == int(os.getenv("WEB_CONCURRENCY"))
+    assert config.workers == int(str(os.getenv("WEB_CONCURRENCY")))
```
Kludex, line 322: Why this is needed? Isn't the return already a `str`?

uvicorn/config.py
```diff
@@ -123,45 +130,45 @@ def create_ssl_context(
...
-        ssl_keyfile=None,
-        ssl_certfile=None,
-        ssl_keyfile_password=None,
-        ssl_version=SSL_PROTOCOL_VERSION,
-        ssl_cert_reqs=ssl.CERT_NONE,
-        ssl_ca_certs=None,
-        ssl_ciphers="TLSv1",
-        headers=None,
-        factory=False,
+        app: Union[ASGIApplication, Awaitable, Callable, str],
```
Kludex: Why `Awaitable` and `Callable` should be accepted here?

uvicorn/config.py
```diff
@@ -123,45 +130,45 @@ def create_ssl_context(
...
+        http: Union[Type[asyncio.Protocol], str] = "auto",
+        ws: Union[Type[asyncio.Protocol], str] = "auto",
+        ws_max_size: int = 16 * 1024 * 1024,
+        lifespan: str = "auto",
+        env_file: Optional[Union[str, os.PathLike]] = None,
+        log_config: Optional[Union[dict, str]] = LOGGING_CONFIG,
+        log_level: Optional[Union[str, int]] = None,
+        access_log: bool = True,
+        use_colors: Optional[bool] = None,
+        interface: Literal["auto", "asgi3", "asgi2", "wsgi"] = "auto",
```
Kludex: I feel like this breaks your pattern (or you break it previously) comparing to:

```python
        http: Union[Type[asyncio.Protocol], str] = "auto",        
        ws: Union[Type[asyncio.Protocol], str] = "auto",
```
```python
        lifespan: str = "auto",
```
```python
        loop: str = "auto",
```
I mean, why don't make use of `Literal` on those as well?

uvicorn/config.py
```diff
@@ -360,12 +374,16 @@ def load(self):
...
-    def setup_event_loop(self):
-        loop_setup = import_from_string(LOOP_SETUPS[self.loop])
-        if loop_setup is not None:
-            loop_setup()
+    def setup_event_loop(self) -> None:
+        loop_setup_str: Optional[str] = LOOP_SETUPS[self.loop]
+        if loop_setup_str:
+            loop_setup: Callable = import_from_string(loop_setup_str)
+            if not inspect.isfunction(loop_setup):
+                raise ImportFromStringError("Asyncio event loop must be a callable.")
```
Kludex: @euri10 take special attention to this, plz
I'm not sure...

uvicorn/importer.py
```diff
@@ -1,13 +1,12 @@
...
-from typing import Union
+from typing import Any
 
 
 class ImportFromStringError(Exception):
     pass
 
 
-def import_from_string(import_str: Union[ModuleType, str]) -> ModuleType:
+def import_from_string(import_str: Any) -> Any:
```
Kludex, line 9: I understand the current behavior as being wrong but is `Any` the best approach here?

Kludex:
Thanks for the PR @br3ndonland :)
I've left some comments, I hope you can take a look at them.

tests/test_config.py
br3ndonland: `config.workers` is an `int`, so in order to compare, the environment variable should also be an `int`.

The return type of `os.getenv()` is `Optional[str]` (if the specified environment variable is not set, the function call will return `None`). Attempting to run `int(None)` will raise a `TypeError`:

```text
>>> int(None)
  File "<stdin>", line 1, in <module>
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

For type checking, either casting the object to a string first, or adding conditions like `if os.getenv("WEB_CONCURRENCY")` or `isinstance(os.getenv("WEB_CONCURRENCY"), str)` will work.

uvicorn/importer.py
br3ndonland: This is one of the few places where I think `Any` is actually appropriate.

If any object type other than string is passed to `import_from_string`, it will just return that object. Otherwise, it will parse the string, import the module (and maybe the attribute after `:`), and return that module or attribute.

I tried setting up a type alias (https://github.com/encode/uvicorn/pull/1067#issuecomment-853388871), but it was really difficult to get it right.

Also, even if we get the typing correct within Uvicorn's source code, we don't know how other applications are using `import_from_string` at this point. `Any` would help avoid breaking type checking in other applications.

uvicorn/config.py
br3ndonland: Yeah, I wasn't sure of the best approach here either. Open to suggestions.

tests/test_config.py
br3ndonland: Could we use the `asgiref` types instead?

```py
def asgi2_app(scope: Scope) -> ASGI2Application:
    async def asgi(
        receive: ASGIReceiveCallable, send: ASGISendCallable
    ) -> None:  # pragma: nocover
        pass

    return asgi  # pragma: nocover

```

Kludex: That doesn't work because the `ASGI2Application` is _all of it_. That part is only the `callable` that I've annotated.

uvicorn/config.py
br3ndonland: The tests pass callables in:

https://github.com/encode/uvicorn/blob/a8af0da26c921b27849701f2cedbf89b8f3cbd86/tests/test_config.py#L37-L47

It might help to set up some proper ASGI application fixtures, instead of these simple callables.

uvicorn/config.py
br3ndonland: Yes, good point. I set up some type aliases for these arguments (70694b5).

tests/test_config.py
br3ndonland: > That doesn't work because the `ASGI2Application` is _all of it_. That part is only the `callable` that I've annotated.

I see. This is related to https://github.com/encode/uvicorn/pull/1067#discussion_r648638750 - if we don't want `Config(app)` to accept a callable, we might have to refactor these test methods.

tests/test_config.py
br3ndonland: Ok that's fine, I accepted your suggestion (6e1c1dc).

uvicorn/config.py
Kludex: I guess it only passes because `asgi_app` is not annotated, but if we annotate, probably `Union[ASGIApplication, str]` is enough. :thinking:

uvicorn/config.py
Kludex: Cool! :tada:

br3ndonland:
> Thanks for the PR @br3ndonland :)
> I've left some comments, I hope you can take a look at them.

Thank you for your help!

uvicorn/importer.py
Kludex: I see.

@euri10 Do you have something to say here or can we stick with `Any`?

uvicorn/config.py
br3ndonland: Well, what I'm saying is that if I change _uvicorn/config.py_ like this:

```py
class Config:
    def __init__(
        self,
        app: Union[ASGIApplication, str],
        ...
```

mypy raises errors on _tests/test_config.py_, like this:

```text
tests/test_config.py:97: error: Argument "app" to "Config" has incompatible type "Callable[[MutableMapping[str, Any], Callable[[str, Iterable[Tuple[str, str]], Optional[Tuple[Type[BaseException], BaseException, Optional[TracebackType]]]], None]], None]"; expected "Union[Union[Type[ASGI2Protocol], Callable[[Union[HTTPScope, WebsocketScope, LifespanScope], Callable[[], Awaitable[Union[HTTPRequestEvent, HTTPDisconnectEvent, WebsocketConnectEvent, WebsocketReceiveEvent, WebsocketDisconnectEvent, LifespanStartupEvent, LifespanShutdownEvent]]], Callable[[Union[HTTPResponseStartEvent, HTTPResponseBodyEvent, HTTPServerPushEvent, HTTPDisconnectEvent, WebsocketAcceptEvent, WebsocketSendEvent, WebsocketResponseStartEvent, WebsocketResponseBodyEvent, WebsocketCloseEvent, LifespanStartupCompleteEvent, LifespanStartupFailedEvent, LifespanShutdownCompleteEvent, LifespanShutdownFailedEvent]], Awaitable[None]]], Awaitable[None]]], str]"
tests/test_config.py:138: error: Argument "app" to "Config" has incompatible type "Callable[[], Union[Type[ASGI2Protocol], Callable[[Union[HTTPScope, WebsocketScope, LifespanScope], Callable[[], Awaitable[Union[HTTPRequestEvent, HTTPDisconnectEvent, WebsocketConnectEvent, WebsocketReceiveEvent, WebsocketDisconnectEvent, LifespanStartupEvent, LifespanShutdownEvent]]], Callable[[Union[HTTPResponseStartEvent, HTTPResponseBodyEvent, HTTPServerPushEvent, HTTPDisconnectEvent, WebsocketAcceptEvent, WebsocketSendEvent, WebsocketResponseStartEvent, WebsocketResponseBodyEvent, WebsocketCloseEvent, LifespanStartupCompleteEvent, LifespanStartupFailedEvent, LifespanShutdownCompleteEvent, LifespanShutdownFailedEvent]], Awaitable[None]]], Awaitable[None]]]]"; expected "Union[Union[Type[ASGI2Protocol], Callable[[Union[HTTPScope, WebsocketScope, LifespanScope], Callable[[], Awaitable[Union[HTTPRequestEvent, HTTPDisconnectEvent, WebsocketConnectEvent, WebsocketReceiveEvent, WebsocketDisconnectEvent, LifespanStartupEvent, LifespanShutdownEvent]]], Callable[[Union[HTTPResponseStartEvent, HTTPResponseBodyEvent, HTTPServerPushEvent, HTTPDisconnectEvent, WebsocketAcceptEvent, WebsocketSendEvent, WebsocketResponseStartEvent, WebsocketResponseBodyEvent, WebsocketCloseEvent, LifespanStartupCompleteEvent, LifespanStartupFailedEvent, LifespanShutdownCompleteEvent, LifespanShutdownFailedEvent]], Awaitable[None]]], Awaitable[None]]], str]"
tests/test_config.py:145: error: Argument "app" to "Config" has incompatible type "Callable[[], Union[Type[ASGI2Protocol], Callable[[Union[HTTPScope, WebsocketScope, LifespanScope], Callable[[], Awaitable[Union[HTTPRequestEvent, HTTPDisconnectEvent, WebsocketConnectEvent, WebsocketReceiveEvent, WebsocketDisconnectEvent, LifespanStartupEvent, LifespanShutdownEvent]]], Callable[[Union[HTTPResponseStartEvent, HTTPResponseBodyEvent, HTTPServerPushEvent, HTTPDisconnectEvent, WebsocketAcceptEvent, WebsocketSendEvent, WebsocketResponseStartEvent, WebsocketResponseBodyEvent, WebsocketCloseEvent, LifespanStartupCompleteEvent, LifespanStartupFailedEvent, LifespanShutdownCompleteEvent, LifespanShutdownFailedEvent]], Awaitable[None]]], Awaitable[None]]]]"; expected "Union[Union[Type[ASGI2Protocol], Callable[[Union[HTTPScope, WebsocketScope, LifespanScope], Callable[[], Awaitable[Union[HTTPRequestEvent, HTTPDisconnectEvent, WebsocketConnectEvent, WebsocketReceiveEvent, WebsocketDisconnectEvent, LifespanStartupEvent, LifespanShutdownEvent]]], Callable[[Union[HTTPResponseStartEvent, HTTPResponseBodyEvent, HTTPServerPushEvent, HTTPDisconnectEvent, WebsocketAcceptEvent, WebsocketSendEvent, WebsocketResponseStartEvent, WebsocketResponseBodyEvent, WebsocketCloseEvent, LifespanStartupCompleteEvent, LifespanStartupFailedEvent, LifespanShutdownCompleteEvent, LifespanShutdownFailedEvent]], Awaitable[None]]], Awaitable[None]]], str]"
tests/test_lifespan.py:199: error: Argument "app" to "Config" has incompatible type "Callable[[Any], Any]"; expected "Union[Union[Type[ASGI2Protocol], Callable[[Union[HTTPScope, WebsocketScope, LifespanScope], Callable[[], Awaitable[Union[HTTPRequestEvent, HTTPDisconnectEvent, WebsocketConnectEvent, WebsocketReceiveEvent, WebsocketDisconnectEvent, LifespanStartupEvent, LifespanShutdownEvent]]], Callable[[Union[HTTPResponseStartEvent, HTTPResponseBodyEvent, HTTPServerPushEvent, HTTPDisconnectEvent, WebsocketAcceptEvent, WebsocketSendEvent, WebsocketResponseStartEvent, WebsocketResponseBodyEvent, WebsocketCloseEvent, LifespanStartupCompleteEvent, LifespanStartupFailedEvent, LifespanShutdownCompleteEvent, LifespanShutdownFailedEvent]], Awaitable[None]]], Awaitable[None]]], str]"
Found 4 errors in 2 files (checked 32 source files)
```

I think some better test fixtures that actually instantiate ASGI applications would help. I'll look into this, but not sure how to do it right now.

tests/test_config.py
```diff
@@ -125,27 +149,30 @@ def create_app():
     assert len(caplog.records) == 1
     assert "--factory" in caplog.records[0].message
 
-    # App not a no-arguments callable.
+    # App not a no-arguments ASGIApplication.
```
euri10: I dont get this change, I'd leave it as is

uvicorn/config.py
```diff
@@ -31,46 +32,52 @@
 from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
 from uvicorn.middleware.wsgi import WSGIMiddleware
 
-TRACE_LOG_LEVEL = 5
+HttpProtocolType = Literal["auto", "h11", "httptools"]
```
euri10:
```suggestion
HTTPProtocolType = Literal["auto", "h11", "httptools"]
```

uvicorn/config.py
```diff
@@ -31,46 +32,52 @@
...
+WsProtocolType = Literal["auto", "none", "websockets", "wsproto"]
```
euri10:
```suggestion
WSProtocolType = Literal["auto", "none", "websockets", "wsproto"]
```

uvicorn/config.py
```diff
@@ -280,11 +298,11 @@ def configure_logging(self):
             logging.getLogger("uvicorn.access").handlers = []
             logging.getLogger("uvicorn.access").propagate = False
 
-    def load(self):
+    def load(self) -> None:
         assert not self.loaded
 
-        if self.is_ssl:
-            self.ssl = create_ssl_context(
+        if self.is_ssl and self.ssl_certfile:
```
euri10: why do we need to add this ` and self.ssl_certfile` this seems odd, is_ssl check if we have a keyfile or certfile already ?

uvicorn/config.py
euri10: I would type ignore that part

uvicorn/importer.py
euri10: no sounds about the best option with my limited mypy understanding,

tests/test_config.py
Kludex: Yeah, me too.

br3ndonland: whoops, find and replace strikes again :zany_face:

uvicorn/config.py
br3ndonland: `certfile` is a required positional argument when running [`SSLContext.load_cert_chain`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain).

`is_ssl` could return `True` if `certfile` is not present:

https://github.com/encode/uvicorn/blob/f025efd4a853e87921b07acdb6efb1d25931a686/uvicorn/config.py#L239-L242

euri10: ok so the fix is wrong imho since if someone forgets to pass a certfile like `uvicorn dev.app:app --ssl-keyfile=dev/server.key` then the server will start but without ssl, which is deceiving

I'd propose instead :
```
        if self.is_ssl:
            assert self.ssl_certfile
            self.ssl = create_ssl_context(
```
so that the server does not start on that config mistake and it solves the mypy error
@Kludex ?

euri10:
I forgot to remove the type def in line 325 when merging master, can you update ?

br3ndonland:
> I forgot to remove the type def in line 325 when merging master, can you update ?

Sure, no problem, I updated and the CI checks should be passing now.

uvicorn/config.py
br3ndonland: I think the assertion would be reasonable here. I added it (8b38633), and I'm happy to consider alternatives also.

br3ndonland:
Thanks @euri10 and @Kludex for all your help on this PR so far.

I will be AFK for a few days, but will check in on this when I return.

tests/test_config.py
```diff
@@ -2,47 +2,69 @@
...
 from uvicorn.protocols.http.h11_impl import H11Protocol
 
+Environ = typing.MutableMapping[str, typing.Any]
+ExcInfo = typing.Tuple[
+    typing.Type[BaseException], BaseException, typing.Optional[TracebackType]
+]
+StartResponse = typing.Callable[
+    [str, typing.Iterable[typing.Tuple[str, str]], typing.Optional[ExcInfo]], None
+]
+
```
euri10: let's move this to _types.py so that we can use it for the wsgi middleware in the future, with a comment stating it's wsgi stuff

uvicorn/config.py
```diff
@@ -368,12 +389,11 @@ def load(self):
 
         self.loaded = True
 
-    def setup_event_loop(self):
-        loop_setup = import_from_string(LOOP_SETUPS[self.loop])
-        if loop_setup is not None:
-            loop_setup()
+    def setup_event_loop(self) -> None:
+        loop_setup: Callable = import_from_string(LOOP_SETUPS[self.loop])
+        loop_setup()
```
euri10: why the `if loop_setup is not None:` disappears ?

euri10 (commented):
ok 2 more comments on my side. I'm sorry this takes so long but it's a huge piece of review, hopefully we get to the end :)

euri10:
> Thanks @euri10 and @Kludex for all your help on this PR so far.
> 
> I will be AFK for a few days, but will check in on this when I return.

hopefully you're back  @br3ndonland  ;) can we try to finish this, I know it took already a while but I feel we're near the end !

tests/test_config.py
br3ndonland: Good idea, we can continue adding custom types there. (f9f509b)

uvicorn/config.py
br3ndonland: Restored (d93bf88)

br3ndonland:
> hopefully you're back @br3ndonland ;) can we try to finish this, I know it took already a while but I feel we're near the end !

I'm back! Yes I think we're almost done. Thanks for sticking with it!

euri10 (approved):
As far as I'm concerned I'm happy with it.
@Kludex can you please take another look and clear if necessary the conversation [here] (https://github.com/encode/uvicorn/pull/1067/files#r648495399) (will merge once I'm sure nothing's left unanswered!)
Thanks a lot for your time and patience with us on this @br3ndonland but this was not an easy one !!

uvicorn/config.py
Kludex: I see. But it should not accept the Awaitable here. Can we remove it?

Kludex:
Only the `Awaitable` in the `app` parameter on the `Config` and then we're ready! 🙌

uvicorn/config.py
br3ndonland: Done (2560310)

br3ndonland:
> Only the `Awaitable` in the `app` parameter on the `Config` and then we're ready! 🙌

Done (2560310). I was thinking about removing `Awaitable`. Thanks for pointing it out.

In a future PR, I will also look into removing `Callable` from the `app` annotation, as discussed in https://github.com/encode/uvicorn/pull/1067#discussion_r648653118.

br3ndonland:
Merged in latest `master`. Should be a clean PR merge with all CI checks passing.

euri10:
let's roll, thanks a lot @br3ndonland !

Kludex:
Thanks @br3ndonland ! :)

PR #1070, opened 2021-06-04
test under py3.10

graingert:
in 3.10 running asyncio.get_event_loop() when asyncio.get_running_loop() would fail is deprecated https://docs.python.org/3.10/library/asyncio-eventloop.html#asyncio.get_event_loop


uvicorn/server.py
```diff
@@ -46,8 +46,10 @@ def __init__(self, config):
 
     def run(self, sockets=None):
         self.config.setup_event_loop()
-        loop = asyncio.get_event_loop()
-        loop.run_until_complete(self.serve(sockets=sockets))
+        if sys.version_info > (3, 6):
```
graingert:
```suggestion
        if sys.version_info >= (3, 7):
```

uvicorn/loops/asyncio.py
```diff
@@ -1,19 +1,7 @@
...
-        and sys.version_info.minor >= 8
-        and platform.system() == "Windows"
-    ):
-        selector = selectors.SelectSelector()
-        loop = asyncio.SelectorEventLoop(selector)
-        asyncio.set_event_loop(loop)
-    else:
-        loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(loop)
+    if sys.version_info >= (3, 8) and sys.platform == "win32":
```
graingert, line 6: mypy only knows about sys.platform not platform.system() == "Windows"

euri10:
I've played a little with 3.10 and the test suite, it seems to me those changes makes sense, however there is a bunch of tests that will fail on 3.10 mainly on the way we setup the loop and get access to it (and on ssl protocol not set but it's unrelated to that and I think the fix is simpler), and unfortunately the PR does not solve it.
I'm not sure if the github workflow already has an image for 3.10 but what about adding it to the workflow and see how/if 3.10 can be added ?

Kludex:
I had to cancel the workflow for both Windows and Ubuntu 3.10, jfyk.

EDIT: Due to infinite loop... It was consuming our quota :kissing:

graingert:
PR for httpx here: https://github.com/encode/httpx/pull/1671

requirements.txt
```diff
@@ -18,12 +18,17 @@ mypy
...
 # Documentation
-mkdocs
+# from git due to https://github.com/mkdocs/mkdocs/pull/2209#issuecomment-855703485
+https://github.com/mkdocs/mkdocs/archive/b221df94fa7f29a2e63edfaeda56e3285b183f91.tar.gz#egg=mkdocs
 mkdocs-material
 
+# py3.10 workarounds
+https://github.com/aaugustin/websockets/archive/cb11516e0ed4fe2b67ab6c1511650bd42115d0b6.tar.gz#egg=websockets; python_version > '3.6'
+https://github.com/encode/httpx/archive/7eb610e606f4b95260c4835de4e9147477dcc6d7.tar.gz#egg=httpx
+https://github.com/python-trio/trustme/archive/6abfddd5f3139bde9ee4ba0f848fb7c84c9ac54c.tar.gz#egg=trustme
```
graingert:
```suggestion
```

trustme 0.8.0 got released

requirements.txt
```diff
@@ -18,12 +18,16 @@ mypy
 trustme
 cryptography
 coverage
-httpx==0.16.*
+httpx>=0.16.0
```
graingert:
```suggestion
httpx>=0.18.2
```

requirements.txt
```diff
@@ -18,12 +18,16 @@ mypy
...
 
 # Documentation
-mkdocs
+# from git due to https://github.com/mkdocs/mkdocs/pull/2209#issuecomment-855703485
+https://github.com/mkdocs/mkdocs/archive/b221df94fa7f29a2e63edfaeda56e3285b183f91.tar.gz#egg=mkdocs
 mkdocs-material
 
+# py3.10 workarounds
+https://github.com/aaugustin/websockets/archive/cb11516e0ed4fe2b67ab6c1511650bd42115d0b6.tar.gz#egg=websockets; python_version > '3.6'
+https://github.com/encode/httpx/archive/7eb610e606f4b95260c4835de4e9147477dcc6d7.tar.gz#egg=httpx
```
graingert:
```suggestion
```

requirements.txt
```diff
@@ -18,12 +18,15 @@ mypy
...
 
 
 # Documentation
-mkdocs
+# from git due to https://github.com/mkdocs/mkdocs/pull/2209#issuecomment-855703485
+https://github.com/mkdocs/mkdocs/archive/b221df94fa7f29a2e63edfaeda56e3285b183f91.tar.gz#egg=mkdocs
 mkdocs-material
 
+# py3.10 workarounds
+https://github.com/aaugustin/websockets/archive/cb11516e0ed4fe2b67ab6c1511650bd42115d0b6.tar.gz#egg=websockets; python_version > '3.6'
```
graingert:
```suggestion
https://github.com/aaugustin/websockets/archive/0bfa9f2ea86dbc12f24e4a2756cb339efafe76d8.tar.gz#egg=websockets; python_version > '3.6'
```

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -4,7 +4,8 @@
 from typing import Callable
 from urllib.parse import unquote
 
-import websockets
+# work around https://github.com/aaugustin/websockets/issues/989
+import websockets.exceptions
```
graingert:
```suggestion
import websockets
```

docs/deployment.md
```diff
@@ -86,7 +86,7 @@ Options:
   --ssl-certfile TEXT             SSL certificate file
   --ssl-keyfile-password TEXT     SSL keyfile password
   --ssl-version INTEGER           SSL version to use (see stdlib ssl module's)
-                                  [default: 2]
+                                  [default: 17]
```
euri10, line 95: this corresponds to ssl.PROTOCOL_TLS_SERVER right ?

tests/conftest.py
```diff
@@ -13,7 +13,7 @@ def tls_certificate_authority() -> trustme.CA:
 
 @pytest.fixture
 def tls_certificate(tls_certificate_authority: trustme.CA) -> trustme.LeafCert:
-    return tls_certificate_authority.issue_server_cert(
```
euri10, line 16: this is equivalent isn't it, why chage it ?

tests/conftest.py
```diff
@@ -49,13 +49,25 @@ def tls_ca_certificate_private_key_encrypted_path(tls_certificate_authority):
...
+def tls_certificate_server_cert_path(tls_certificate: trustme.LeafCert):
+    with tls_certificate.cert_chain_pems[0].tempfile() as cert_pem:
+        yield cert_pem
+
+
+@pytest.fixture
+def tls_ca_ssl_context(tls_certificate_authority: trustme.CA) -> ssl.SSLContext:
+    ssl_ctx = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+    tls_certificate_authority.configure_trust(ssl_ctx)
     return ssl_ctx
```
euri10, line 73: yeah that's way better, I did not know a vanilla context had by default CERT_NONE !

tests/conftest.py
```diff
@@ -49,13 +49,25 @@ def tls_ca_certificate_private_key_encrypted_path(tls_certificate_authority):
...
         yield cert_pem
 
 
 @pytest.fixture
-def tls_ca_ssl_context(tls_certificate: trustme.LeafCert) -> ssl.SSLContext:
-    ssl_ctx = ssl.SSLContext()
-    tls_certificate.configure_cert(ssl_ctx)
+def tls_certificate_server_cert_path(tls_certificate: trustme.LeafCert):
+    with tls_certificate.cert_chain_pems[0].tempfile() as cert_pem:
+        yield cert_pem
```
euri10, line 66: so here you're splitting what I called `tls_certificate_pem_path` in `tls_certificate_private_key_path` for just the private key and `tls_certificate_key_and_chain_path` for the bundle then add a test for just the bundle if I followed correctly ?

tests/protocols/test_http.py
```diff
@@ -161,13 +172,17 @@ def add_done_callback(self, callback):
...
     server_state = ServerState()
     protocol = protocol_cls(config=config, server_state=server_state, _loop=loop)
     protocol.connection_made(transport)
-    yield protocol
-    protocol.loop.close()
+    try:
+        yield protocol
+    finally:
+        protocol.loop.close()
+        asyncio._set_running_loop(None)
```
euri10, line 185: I'm certainly not familiar enough with all asyncio internals but why the last line here, isn't it enough to close the loop ?
edit just tested removing it and it seems like tests pass fine without, but I'm sure you have an explanation ;)

tests/test_auto_detection.py
```diff
@@ -39,15 +41,17 @@ def test_loop_auto():
     assert type(policy).__module__.startswith(expected_loop)
 
 
-def test_http_auto():
+@pytest.mark.asyncio
```
euri10, line 44: I dont get why those 2 tests suddenly need the async marker

docs/deployment.md
graingert: yeah I'm not sure what to do about this flag

tests/conftest.py
graingert: https://github.com/python-trio/trustme/blob/f9e13e03963260b32cab03f4e4c25cc515ddfad3/docs/source/index.rst#trustme-050-2019-01-21 it's a legacy name for compatibility

tests/protocols/test_http.py
graingert: You have to detach the running loop otherwise get_running_loop() will continue to return this MockLoop

tests/test_auto_detection.py
graingert: the WebsocketProtocol classes call get_event_loop() which is deprecated unless there's a running loop

graingert: Details here: 

https://github.com/aaugustin/websockets/issues/988#issuecomment-856034588

docs/deployment.md
graingert: I think this CLI flag should be changed to take either "1.2" or "1.3" and use a similar trick to https://github.com/encode/httpx/blob/master/httpx/_compat.py#L16

euri10: > I think this CLI flag should be changed to take either "1.2" or "1.3" and use a similar trick to https://github.com/encode/httpx/blob/master/httpx/_compat.py#L16

makes sense, but maybe let's keep that for a seperate PR, I wonder what gunicorn does about it, I remember I read about lenghty debate there but nothing was very actionnable iirc

tests/test_auto_detection.py
euri10: mmmm now I get it thanks

euri10 (approved):
Looks very good to me, approved and many thanks for that @graingert !

I just have one minor question / issue since I'm no semver pro, but would the ssl min default that is bumped from PROTOCOL_SSLv23 to PROTOCOL_TLS_SERVER qualify for a minor version bump or we good with just a patch ?

euri10:
cant wait to use the Rust match statements in 3.10 :turtle:  !!

.github/workflows/test-suite.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: "${{ matrix.os }}"
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10.0-beta.2"]
```
Kludex: Released 2 hours ago :)
```suggestion
        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10.0-beta.3"]
```

Kludex:
On the package metadata, we should add Python 3.10 information.

requirements.txt
```diff
@@ -18,12 +18,15 @@ mypy
...
 
 
 # Documentation
-mkdocs
+# from git due to https://github.com/mkdocs/mkdocs/pull/2209#issuecomment-855703485
+https://github.com/mkdocs/mkdocs/archive/b221df94fa7f29a2e63edfaeda56e3285b183f91.tar.gz#egg=mkdocs
 mkdocs-material
 
+# py3.10 workarounds
+https://github.com/aaugustin/websockets/archive/0bfa9f2ea86dbc12f24e4a2756cb339efafe76d8.tar.gz#egg=websockets; python_version > '3.6'
```
euri10: on hold until this is released, won't be able to upload to pypi anyway if merged

euri10 (changes requested):

tests/test_config.py
```diff
@@ -175,10 +175,10 @@ def test_ssl_config(
     assert config.is_ssl is True
 
 
-def test_ssl_config_combined(tls_certificate_pem_path: str) -> None:
+def test_ssl_config_combined(tls_certificate_key_and_chain_path: str) -> None
```
graingert:
```suggestion
def test_ssl_config_combined(tls_certificate_key_and_chain_path: str) -> None:
```

requirements.txt
graingert:
```suggestion
https://github.com/aaugustin/websockets/archive/d5cf1efb737a943583b1e5b4ceca5376bdc3995f.tar.gz#egg=websockets; python_version > '3.6'
```

requirements.txt
```diff
@@ -20,12 +20,15 @@ types-pyyaml
...
-httpx==0.16.*
+httpx>=0.18.2
 pytest-asyncio==0.14.*
 async_generator; python_version < '3.7'
 
 
 # Documentation
-mkdocs
+# from git due to https://github.com/mkdocs/mkdocs/pull/2209#issuecomment-855703485
+https://github.com/mkdocs/mkdocs/archive/b221df94fa7f29a2e63edfaeda56e3285b183f91.tar.gz#egg=mkdocs
```
graingert:
```suggestion
mkdocs>=1.2.2
```

tests/conftest.py
graingert: sort of tls_certificate_pem_path was the wrong file - it was the CA key

.github/workflows/test-suite.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: "${{ matrix.os }}"
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10.0-beta.3"]
```
graingert:
```suggestion
        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10.0-beta.4"]
```

requirements.txt
```diff
@@ -20,12 +20,14 @@ types-pyyaml
...
 async_generator; python_version < '3.7'
 
 
 # Documentation
-mkdocs
+mkdocs>=1.2.2
 mkdocs-material
 
+# py3.10 workarounds
+https://github.com/aaugustin/websockets/archive/d5cf1efb737a943583b1e5b4ceca5376bdc3995f.tar.gz#egg=websockets; python_version > '3.6'
```
graingert:
```suggestion
https://github.com/aaugustin/websockets/archive/d5cf1efb737a943583b1e5b4ceca5376bdc3995f.tar.gz#egg=websockets; python_version >= '3.10'
```

tests/test_auto_detection.py
```diff
@@ -41,15 +43,17 @@ def test_loop_auto():
     loop.close()
```
graingert: @euri10 what's this for `asyncio.get_event_loop()` can't be called here

requirements.txt
```diff
@@ -20,12 +20,14 @@ types-pyyaml
...
 async_generator; python_version < '3.7'
 
 
 # Documentation
-mkdocs
+mkdocs>=1.2.2
 mkdocs-material
 
+# py3.10 workarounds
+https://github.com/aaugustin/websockets/archive/d5cf1efb737a943583b1e5b4ceca5376bdc3995f.tar.gz#egg=websockets; python_version >= '3.10'
```
graingert:
```suggestion
```

graingert: @aaugustin the latest websockets is now hanging the uvicorn test suite, and I'm not sure why.

graingert:
The tests hang due to:
https://github.com/aaugustin/websockets/issues/1021#issuecomment-886191121

tests/test_auto_detection.py
```diff
@@ -37,19 +40,25 @@ def test_loop_auto():
     assert isinstance(policy, asyncio.events.BaseDefaultEventLoopPolicy)
     expected_loop = "asyncio" if uvloop is None else "uvloop"
     assert type(policy).__module__.startswith(expected_loop)
-    loop = asyncio.get_event_loop()
-    loop.close()
+
+    async def get_loop():
+        return asyncio.get_event_loop()
```
graingert:
```suggestion
        return anyio.lowlevel.current_token()
```

graingert:
@euri10 @Kludex anyio is now always available in the test deps, what do you think about replacing pytest-asyncio with anyio to save a dep?

requirements.txt
graingert:
```suggestion
https://github.com/aaugustin/websockets/archive/4e1dac362a3639b9cf0e5bcf382601e6e32cfede.tar.gz#egg=websockets; python_version >= '3.10'
```

requirements.txt
graingert: @euri10 I've changed this to use `; python_verion >= '3.10'` so we can upload this to pypi now

tests/test_auto_detection.py
```diff
@@ -37,19 +40,25 @@ def test_loop_auto():
...
     expected_loop = "asyncio" if uvloop is None else "uvloop"
     assert type(policy).__module__.startswith(expected_loop)
-    loop = asyncio.get_event_loop()
-    loop.close()
+
+    async def get_loop():
+        return anyio.lowlevel.current_token()
+
+    loop = anyio.run(get_loop)
+    assert loop.__module__.startswith(expected_loop)
```
euri10: I dont know anyio at all, what does this do and why introducing it, also if kept shouldn't it be added to the requirements.txt ?

uvicorn/loops/asyncio.py
```diff
@@ -1,19 +1,7 @@
...
+        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```
euri10, line 7: is it the same as the deleted part ?

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -76,7 +95,7 @@ def connection_made(self, transport):
         self.client = get_remote_addr(transport)
         self.scheme = "wss" if is_ssl(transport) else "ws"
 
-        if self.logger.level <= TRACE_LOG_LEVEL:
+        if self.logger.isEnabledFor(TRACE_LOG_LEVEL):
```
euri10, line 98: why change this ?

tests/test_auto_detection.py
graingert: it's a dep via httpx, it's a handy shim for `get_running_loop()` on py3.6-py3.10

uvicorn/protocols/websockets/websockets_impl.py
graingert: details here https://github.com/aaugustin/websockets/issues/1021

tests/test_auto_detection.py
graingert: it's to save writing:

```python
if sys.version_info >= (3, 7):
    async def get_loop():
        return asyncio.get_running_loop()

    loop = asyncio.run(get_loop())

else:

   with contextlib.closing(asyncio.get_event_loop()) as loop:
       pass
```

uvicorn/loops/asyncio.py
graingert: it behaves *almost* the same, it's more like https://github.com/graingert/uvicorn/blob/avoid-deprecated-get-event-loop/uvicorn/loops/uvloop.py#L7 and works on py3.10

euri10: ok, in essence the question was: we still force windows not to use the protractor event loop which is their default ?

graingert: Yes

tests/test_auto_detection.py
euri10: what about removing all that entirely, we just test the policy here, no loop is started if I'm correct and there should be no need to close it

uvicorn/loops/asyncio.py
euri10: ok so we're still NOT using the windows default with the chage, ie we are not on the protractor loop?

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -24,7 +25,26 @@ def is_serving(self):
...
+    class _LoggerMixin:
+        pass
+
+
+else:
+
+    class _LoggerMixin:
+        def __init__(self, *args, logger, **kwargs):
+            super().__init__(*args, **kwargs)
+            self.logger = logging.LoggerAdapter(logger, {"websocket": self})
```
euri10, line 44: TIL about `LoggerAdapter`, quite meat

uvicorn/protocols/websockets/websockets_impl.py
euri10: so if I follow correctly the logger inside this module is not the standard python logger anymore but a LoggerAdapter that we are kind of plugging from websockets ?

uvicorn/loops/asyncio.py
euri10: oups I'm making double comment  sorry @graingert

uvicorn/protocols/websockets/websockets_impl.py
graingert: it depends, but logger.level is private undocumented and you're supposed to use isEnabledFor

tests/test_auto_detection.py
graingert: Yeah we could just check the policy

graingert:
```suggestion
```

tests/test_auto_detection.py
```diff
@@ -1,5 +1,8 @@
 import asyncio
 
+import anyio
```
graingert:
```suggestion
```

euri10 (approved):

uvicorn/loops/asyncio.py
```diff
@@ -1,19 +1,7 @@
...
-        sys.version_info.major >= 3
-        and sys.version_info.minor >= 8
-        and platform.system() == "Windows"
-    ):
-        selector = selectors.SelectSelector()
-        loop = asyncio.SelectorEventLoop(selector)
-        asyncio.set_event_loop(loop)
-    else:
-        loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(loop)
```
Kludex, line 19: This is the cause of the httpx tests failing. @tomchristie

lovelydinosaur: Thanks @Kludex - No idea how you tracked that down, but... 👍

PR #1018, opened 2021-04-15
add types to wsgi module

jkklapp:
Part of #998

PR #1075, opened 2021-06-09
🚨 Cover middleware/wsgi.py on mypy

Vibhu-Agarwal:
Part of #998 
Extends #1018

I've tried to solve most of the issues, but some still remain - not sure how to solve them.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,6 +24,7 @@ files =
     uvicorn/supervisors/__init__.py,
     uvicorn/middleware/debug.py,
     uvicorn/middleware/wsgi.py,
+    tests/middleware/test_wsgi.py,
     uvicorn/supervisors/watchgodreload.py,
     uvicorn/logging.py,
     uvicorn/middleware/asgi2.py,
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,6 +20,7 @@ files =
     uvicorn/protocols/websockets/auto.py,
     uvicorn/supervisors/__init__.py,
     uvicorn/middleware/debug.py,
+    uvicorn/middleware/wsgi.py,
     uvicorn/supervisors/watchgodreload.py,
     uvicorn/logging.py,
     uvicorn/middleware/asgi2.py,
--- a/uvicorn/middleware/wsgi.py
+++ b/uvicorn/middleware/wsgi.py
@@ -5,7 +5,6 @@ import sys
 from typing import Iterable, List, Optional, Tuple
 
 from asgiref.typing import (
-    ASGI3Application,
     ASGIReceiveCallable,
     ASGIReceiveEvent,
     ASGISendCallable,
@@ -16,7 +15,7 @@ from asgiref.typing import (
     HTTPScope,
 )
 
-from uvicorn._types import Environ, ExcInfo, StartResponse
+from uvicorn._types import Environ, ExcInfo, StartResponse, WSGIApp
 
 
 def build_environ(scope: HTTPScope, message: ASGIReceiveEvent, body: bytes) -> Environ:
@@ -71,7 +70,7 @@ def build_environ(scope: HTTPScope, message: ASGIReceiveEvent, body: bytes) -> E
 
 
 class WSGIMiddleware:
-    def __init__(self, app: ASGI3Application, workers: int = 10):
+    def __init__(self, app: WSGIApp, workers: int = 10):
         self.app = app
         self.executor = concurrent.futures.ThreadPoolExecutor(max_workers=workers)
 
@@ -86,8 +85,8 @@ class WSGIMiddleware:
 class WSGIResponder:
     def __init__(
         self,
-        app: ASGI3Application,
-        executor: concurrent.futures.Executor,
+        app: WSGIApp,
+        executor: concurrent.futures.ThreadPoolExecutor,
         scope: HTTPScope,
     ):
         self.app = app
@@ -97,18 +96,18 @@ class WSGIResponder:
         self.response_headers = None
         self.send_event = asyncio.Event()
         self.send_queue: List[Optional[ASGISendEvent]] = []
-        self.loop: asyncio.AbstractEventLoop = None  # type: ignore
+        self.loop: asyncio.AbstractEventLoop = asyncio.get_event_loop()
         self.response_started = False
         self.exc_info: Optional[ExcInfo] = None
 
     async def __call__(
         self, receive: ASGIReceiveCallable, send: ASGISendCallable
     ) -> None:
-        message: HTTPRequestEvent = await receive()  # type: ignore
+        message: HTTPRequestEvent = await receive()  # type: ignore[assignment]
         body = message.get("body", b"")
         more_body = message.get("more_body", False)
         while more_body:
-            body_message: HTTPRequestEvent = await receive()  # type: ignore
+            body_message: HTTPRequestEvent = await receive()  # type: ignore[assignment]
             body += body_message.get("body", b"")
             more_body = body_message.get("more_body", False)
         environ = build_environ(self.scope, message, body)
@@ -147,7 +146,7 @@ class WSGIResponder:
         if not self.response_started:
             self.response_started = True
             status_code_str, _ = status.split(" ", 1)
-            status_code: int = int(status_code_str)
+            status_code = int(status_code_str)
             headers = [
                 (name.encode("ascii"), value.encode("ascii"))
                 for name, value in response_headers
--- a/uvicorn/middleware/wsgi.py
+++ b/uvicorn/middleware/wsgi.py
@@ -2,7 +2,7 @@ import asyncio
 import concurrent.futures
 import io
 import sys
-from typing import Awaitable, Iterable, List, Optional, Tuple
+from typing import Iterable, List, Optional, Tuple
 
 from asgiref.typing import (
     ASGI3Application,
@@ -16,7 +16,7 @@ from asgiref.typing import (
     HTTPScope,
 )
 
-from uvicorn._types import Environ, ExcInfo
+from uvicorn._types import Environ, ExcInfo, StartResponse
 
 
 def build_environ(scope: HTTPScope, message: ASGIReceiveEvent, body: bytes) -> Environ:
@@ -160,7 +160,7 @@ class WSGIResponder:
             self.send_queue.append(http_response_start_event)
             self.loop.call_soon_threadsafe(self.send_event.set)
 
-    def wsgi(self, environ: Environ, start_response: Awaitable[None]) -> None:
+    def wsgi(self, environ: Environ, start_response: StartResponse) -> None:
         for chunk in self.app(environ, start_response):  # type: ignore
             response_body: HTTPResponseBodyEvent = {
                 "type": "http.response.body",
--- a/uvicorn/middleware/wsgi.py
+++ b/uvicorn/middleware/wsgi.py
@@ -2,9 +2,18 @@ import asyncio
 import concurrent.futures
 import io
 import sys
+from typing import Awaitable, Dict, Iterable, Optional, Tuple
+
+from uvicorn._types import (
+    ASGI3Application,
+    ASGIReceiveCallable,
+    ASGISendCallable,
+    ASGISendEvent,
+    HTTPScope,
+)
 
 
-def build_environ(scope, message, body):
+def build_environ(scope: HTTPScope, message: ASGISendEvent, body: bytes) -> Dict:
     """
     Builds a scope and request message into a WSGI environ object.
     """
@@ -54,18 +63,25 @@ def build_environ(scope, message, body):
 
 
 class WSGIMiddleware:
-    def __init__(self, app, workers=10):
+    def __init__(self, app: ASGI3Application, workers: int = 10) -> None:
         self.app = app
         self.executor = concurrent.futures.ThreadPoolExecutor(max_workers=workers)
 
-    async def __call__(self, scope, receive, send):
+    async def __call__(
+        self, scope: HTTPScope, receive: ASGIReceiveCallable, send: ASGISendCallable
+    ) -> None:
         assert scope["type"] == "http"
         instance = WSGIResponder(self.app, self.executor, scope)
         await instance(receive, send)
 
 
 class WSGIResponder:
-    def __init__(self, app, executor, scope):
+    def __init__(
+        self,
+        app: ASGI3Application,
+        executor: concurrent.futures.ThreadPoolExecutor,
+        scope: HTTPScope,
+    ) -> Awaitable:
         self.app = app
         self.executor = executor
         self.scope = scope
@@ -75,9 +91,11 @@ class WSGIResponder:
         self.send_queue = []
         self.loop = None
         self.response_started = False
-        self.exc_info = None
+        self.exc_info: Optional[str] = None
 
-    async def __call__(self, receive, send):
+    async def __call__(
+        self, receive: ASGIReceiveCallable, send: ASGISendCallable
+    ) -> None:
         message = await receive()
         body = message.get("body", b"")
         more_body = message.get("more_body", False)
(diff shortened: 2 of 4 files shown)


Vibhu-Agarwal:
Not sure if this commit ^^^ is correct.
Will revert, if it's not.

uvicorn/middleware/wsgi.py
```diff
@@ -99,11 +100,11 @@ def __init__(
     async def __call__(
         self, receive: ASGIReceiveCallable, send: ASGISendCallable
     ) -> None:
-        message = await receive()
+        message: HTTPRequestEvent = await receive()
         body = message.get("body", b"")
```
Vibhu-Agarwal: Before the change, it was throwing this error:
```
uvicorn/middleware/wsgi.py:103: error: TypedDict "HTTPDisconnectEvent" has no key 'body'  [misc]
uvicorn/middleware/wsgi.py:103: error: TypedDict "WebsocketConnectEvent" has no key 'body'  [misc]
uvicorn/middleware/wsgi.py:103: error: TypedDict "WebsocketReceiveEvent" has no key 'body'  [misc]
uvicorn/middleware/wsgi.py:103: error: TypedDict "WebsocketDisconnectEvent" has no key 'body'  [misc]
uvicorn/middleware/wsgi.py:103: error: TypedDict "LifespanStartupEvent" has no key 'body'  [misc]
uvicorn/middleware/wsgi.py:103: error: TypedDict "LifespanShutdownEvent" has no key 'body'  [misc]
```

uvicorn/middleware/wsgi.py
```diff
@@ -46,43 +59,52 @@ def build_environ(scope, message, body):
...
 
 class WSGIMiddleware:
-    def __init__(self, app, workers=10):
+    def __init__(self, app: ASGI3Application, workers: int = 10):
         self.app = app
         self.executor = concurrent.futures.ThreadPoolExecutor(max_workers=workers)
 
-    async def __call__(self, scope, receive, send):
+    async def __call__(
+        self, scope: HTTPScope, receive: ASGIReceiveCallable, send: ASGISendCallable
```
Kludex, line 78: What's the error you have below? 
```suggestion
        self, scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable
```

uvicorn/middleware/wsgi.py
```diff
@@ -46,43 +59,52 @@ def build_environ(scope, message, body):
...
         instance = WSGIResponder(self.app, self.executor, scope)
         await instance(receive, send)
 
 
 class WSGIResponder:
-    def __init__(self, app, executor, scope):
+    def __init__(
+        self,
+        app: ASGI3Application,
+        executor: concurrent.futures.ThreadPoolExecutor,
```
Kludex, line 89:
```suggestion
        executor: concurrent.futures.Executor,
```

uvicorn/middleware/wsgi.py
Kludex: The `http.disconnect` never gets here?

uvicorn/middleware/wsgi.py
```diff
@@ -36,8 +49,8 @@ def build_environ(scope, message, body):
         environ["REMOTE_ADDR"] = client[0]
 
     # Go through headers and make them into environ entries
-    for name, value in scope.get("headers", []):
-        name = name.decode("latin1")
+    for name_, value_ in scope.get("headers", []):
```
Kludex: Can we use `key, value` instead? And on the `value` below, `value_str`? Or just maintain the same `name, value` here, but using the `name_str` and `value_str` below?

Kludex (commented):
Thanks for the PR @Vibhu-Agarwal ! :smiley: 

I've added some comments, can you check them, please?

uvicorn/middleware/wsgi.py
Vibhu-Agarwal: That's exactly what was bugging me 😅
Most probably `http.disconnect` could get here.
But if I remove the type annotation from `message`, the `body = message.get("body", b"")` line throws out the errors.
Not sure how to resolve that then.


PS. The same goes for `body_message` a couple of lines below that.

uvicorn/middleware/wsgi.py
Vibhu-Agarwal: There isn't any error in this line.
Also, I think `HTTPScope` should be fine as it mentions this one line down:
https://github.com/encode/uvicorn/blob/a8af0da26c921b27849701f2cedbf89b8f3cbd86/uvicorn/middleware/wsgi.py#L62

Kludex: Yeah, but the line below says that it should be `HTTPScope` **after** it, not before. Otherwise, we'd not need to assert it.

Vibhu-Agarwal: The `assert` has been put just after the function header - in the first line.
https://github.com/encode/uvicorn/blob/a8af0da26c921b27849701f2cedbf89b8f3cbd86/uvicorn/middleware/wsgi.py#L61-L62
My reasoning was that this should mean that the function is expecting the type of `scope` as HTTP, and therefore I put the typehint for that.

Also, I tried to replace that with `Scope`. That threw new mypy errors as the rest of the code is expecting attributes on `scope` which are only there in `HTTPScope` and not in `Scope`.

euri10:
I've not looked at this but you can maybe take inspiration from https://github.com/encode/uvicorn/pull/1049 where I typed everything in the same module I think

uvicorn/middleware/wsgi.py
```diff
@@ -111,31 +135,43 @@ async def sender(self, send):
...
+            }
+            self.send_queue.append(http_response_start_event)
             self.loop.call_soon_threadsafe(self.send_event.set)
 
-    def wsgi(self, environ, start_response):
-        for chunk in self.app(environ, start_response):
-            self.send_queue.append(
-                {"type": "http.response.body", "body": chunk, "more_body": True}
-            )
+    def wsgi(self, environ: Dict, start_response: Awaitable[None]) -> None:
```
euri10: Once https://github.com/encode/uvicorn/pull/1067 is merged (which is soon !) there will be a _types.py you can use for typing this

uvicorn/middleware/wsgi.py
```diff
@@ -37,52 +52,63 @@ def build_environ(scope, message, body):
...
-            corrected_name = "HTTP_%s" % name.upper().replace("-", "_")
+            corrected_name = "HTTP_%s" % name_str.upper().replace("-", "_")
         # HTTPbis say only ASCII chars are allowed in headers, but we latin1
         # just in case
-        value = value.decode("latin1")
+        value_str: str = value.decode("latin1")
         if corrected_name in environ:
-            value = environ[corrected_name] + "," + value
-        environ[corrected_name] = value
+            corrected_name_environ = environ[corrected_name]
```
Kludex, line 65: This looks a little too much just to satisfy `mypy` :thinking:

uvicorn/middleware/wsgi.py
```diff
@@ -111,31 +137,43 @@ async def sender(self, send):
...
+        response_headers: Iterable[Tuple[str, str]],
+        exc_info: Optional[ExcInfo] = None,
+    ) -> None:
         self.exc_info = exc_info
         if not self.response_started:
             self.response_started = True
-            status_code, _ = status.split(" ", 1)
-            status_code = int(status_code)
+            status_code_str, _ = status.split(" ", 1)
+            status_code: int = int(status_code_str)
```
Kludex: Why do we need to add the annotation here? :thinking:

Kludex: Ah, I guess we needed it before the renaming.

uvicorn/middleware/wsgi.py
Kludex: Should we just ignore it here?

Kludex (commented):
We're almost there! I've added some comments, do you mind taking a look @Vibhu-Agarwal ?

Thanks for your help so far!

setup.cfg
```diff
@@ -23,6 +23,7 @@ files =
     uvicorn/protocols/websockets/auto.py,
     uvicorn/supervisors/__init__.py,
     uvicorn/middleware/debug.py,
+    uvicorn/middleware/wsgi.py,
```
euri10, line 26:
```suggestion
    uvicorn/middleware/wsgi.py,
    tests/middleware/test_wsgi.py,
```
let's add the test also, this will show you some easy to spot errors in the types

uvicorn/middleware/wsgi.py
euri10: lgtm

uvicorn/middleware/wsgi.py
```diff
@@ -37,52 +52,63 @@ def build_environ(scope, message, body):
...
+            corrected_name_environ = environ[corrected_name]
+            assert isinstance(corrected_name_environ, str)
+            value_str = corrected_name_environ + "," + value_str
+        environ[corrected_name] = value_str
     return environ
 
 
 class WSGIMiddleware:
-    def __init__(self, app, workers=10):
+    def __init__(self, app: ASGI3Application, workers: int = 10):
```
euri10: this is wrong, app here is not an asgi application but a wsgi application:
`WSGIApp = typing.Callable[[Environ, StartResponse], typing.Iterable[bytes]]`

euri10: can be added to _types.py

uvicorn/middleware/wsgi.py
```diff
@@ -37,52 +52,63 @@ def build_environ(scope, message, body):
...
         assert scope["type"] == "http"
         instance = WSGIResponder(self.app, self.executor, scope)
         await instance(receive, send)
 
 
 class WSGIResponder:
-    def __init__(self, app, executor, scope):
+    def __init__(
+        self,
+        app: ASGI3Application,
```
euri10: same here, WSGIApp

uvicorn/middleware/wsgi.py
```diff
@@ -37,52 +52,63 @@ def build_environ(scope, message, body):
...
+        executor: concurrent.futures.Executor,
```
euri10: it is a `ThreadPoolExecutor`

uvicorn/middleware/wsgi.py
```diff
@@ -37,52 +52,63 @@ def build_environ(scope, message, body):
...
         self.app = app
         self.executor = executor
         self.scope = scope
         self.status = None
         self.response_headers = None
         self.send_event = asyncio.Event()
-        self.send_queue = []
-        self.loop = None
+        self.send_queue: List[Optional[ASGISendEvent]] = []
+        self.loop: asyncio.AbstractEventLoop = None  # type: ignore
```
euri10: no need to ignore that, we can instantiate it directly

uvicorn/middleware/wsgi.py
```diff
@@ -37,52 +52,63 @@ def build_environ(scope, message, body):
...
         self.response_started = False
-        self.exc_info = None
+        self.exc_info: Optional[ExcInfo] = None
 
-    async def __call__(self, receive, send):
-        message = await receive()
+    async def __call__(
+        self, receive: ASGIReceiveCallable, send: ASGISendCallable
+    ) -> None:
+        message: HTTPRequestEvent = await receive()  # type: ignore
```
euri10: worth maybe putting a note as to why we ignore this, iirc because mypy cannot infer the Literal well enough, or put the [assignment] so that we know

euri10 (changes requested):

euri10 (approved):
this lgtm @Vibhu-Agarwal thanks a lot
I let @Kludex finish it :)

edit: oups didnt see the CI was failing, will have to wait >_)

euri10:
you can use this patch @Vibhu-Agarwal I think this solves the rest

```patch
Index: tests/middleware/test_wsgi.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/tests/middleware/test_wsgi.py b/tests/middleware/test_wsgi.py
--- a/tests/middleware/test_wsgi.py	(revision 7610a04a8be1219b2b48eff20e73ad0049884c45)
+++ b/tests/middleware/test_wsgi.py	(revision 6b84705a15ace9d662a49f15921c54bdfd8287c1)
@@ -3,7 +3,7 @@
 
 import httpx
 import pytest
-from asgiref.typing import HTTPScope
+from asgiref.typing import HTTPRequestEvent, HTTPScope
 
 from uvicorn._types import Environ, StartResponse
 from uvicorn.middleware.wsgi import WSGIMiddleware, build_environ
@@ -16,7 +16,7 @@
         ("Content-Type", "text/plain; charset=utf-8"),
         ("Content-Length", str(len(output))),
     ]
-    start_response(status, headers)
+    start_response(status, headers, None)
     return [output]
 
 
@@ -27,7 +27,7 @@
         ("Content-Type", "text/plain; charset=utf-8"),
         ("Content-Length", str(len(output))),
     ]
-    start_response(status, headers)
+    start_response(status, headers, None)
     return [output]
 
 
@@ -45,7 +45,7 @@
             ("Content-Type", "text/plain; charset=utf-8"),
             ("Content-Length", str(len(output))),
         ]
-        start_response(status, headers, exc_info=sys.exc_info())
+        start_response(status, headers, sys.exc_info())  # type: ignore[arg-type]
         return [output]
 
 
@@ -108,7 +108,12 @@
         "root_path": "/文",
         "query_string": b"a=123&b=456",
         "headers": [(b"key", b"value1"), (b"key", b"value2")],
+    }  # type: ignore[typeddict-item]
+    message: HTTPRequestEvent = {
+        "type": "http.request",
+        "body": b"",
+        "more_body": False,
     }
-    environ = build_environ(scope, b"", b"")
+    environ = build_environ(scope, message, b"")
     assert environ["PATH_INFO"] == "/文".encode("utf8").decode("latin-1")
     assert environ["HTTP_KEY"] == "value1,value2"
Index: uvicorn/_types.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/uvicorn/_types.py b/uvicorn/_types.py
--- a/uvicorn/_types.py	(revision 7610a04a8be1219b2b48eff20e73ad0049884c45)
+++ b/uvicorn/_types.py	(revision 6b84705a15ace9d662a49f15921c54bdfd8287c1)
@@ -9,4 +9,6 @@
 StartResponse = typing.Callable[
     [str, typing.Iterable[typing.Tuple[str, str]], typing.Optional[ExcInfo]], None
 ]
-WSGIApp = typing.Callable[[Environ, StartResponse], typing.Iterable[bytes]]
+WSGIApp = typing.Callable[
+    [Environ, StartResponse], typing.Union[typing.Iterable[bytes], BaseException]
+]

```

Vibhu-Agarwal:
Failing on **3.6** and **3.7** 🤔

euri10:
typeddict is > 3.7 just need to fix the import

Vibhu-Agarwal:
Unable to figure out the fix 😕
Import doesn't seem like the issue here, as it's coming directly from `asgiref` where they're checking the version and aptly importing from `typing`/`typing_extensions`.
For some reason, `# type: ignore` isn't working 🤔

euri10:
you can fill the scope too
```
    scope: HTTPScope = {
        "type": "http",
        "asgi": {"version": "3.0", "spec_version": "2.0"},
        "scheme": "http",
        "http_version": "1.1",
        "method": "GET",
        "path": "/文",
        "raw_path": b"/\xe6\x96\x87",
        "root_path": "/文",
        "query_string": b"a=123&b=456",
        "client": None,
        "server": None,
        "headers": [(b"key", b"value1"), (b"key", b"value2")],
        "extensions": {},
    }
```

Vibhu-Agarwal:
Thanks for the reviews and support 👍
Cc @euri10 @Kludex

euri10 (approved):

euri10:
thanks a lot @Vibhu-Agarwal !

PR #1078, opened 2021-06-13
👷 Add Python 3.10 beta to the CI

Kludex:
Following the same idea as https://github.com/encode/httpx/issues/1678


Kludex:
I thought PRs could be separated. I was wrong. Just ignore this. :)

PR #1077, opened 2021-06-12
Exit with status 3 when worker starts failed

fanchunke:
This PR fixes #1066 .

When starting Gunicorn with Uvicorn worker, if exceptions raised at startup events, `UvicornWorker` exits with zero exit code, therefore Gunicorn will keep booting worker.

This PR sets exit code 3 when worker starts failed, Gunicorn will then handle `SIGCHLD` correctly and could shut it down to avoid infinite start/stop cycles.

PR #1079, opened 2021-06-13
e2e testing for gunicorn worker

euri10:


euri10:
interesting windows only error on an unrelated not-closed socket ;)

euri10:
> interesting windows only error on an unrelated not-closed socket ;)

solved in https://github.com/encode/uvicorn/pull/1080

euri10:
well the exit 3 solved in https://github.com/encode/uvicorn/pull/1077 is well demonstrated (and solved once that patch is applied)

euri10:
only big issue with that e2e approach is that it slows down the testing a lot.

Kludex:
Can we can run them in parallel? How slow it becomes?

euri10:
```
========================================================================================================== slowest durations ==========================================================================================================
77.30s setup    tests/e2e/test_gunicorn.py::test_gunicorn_default[default.py-0]
28.79s setup    tests/e2e/test_gunicorn.py::test_gunicorn_default[startup_failed.py-137]
2.60s call     tests/e2e/test_gunicorn.py::test_gunicorn_default[default.py-0]
2.10s call     tests/e2e/test_gunicorn.py::test_gunicorn_default[startup_failed.py-137]
```

settings up the docker images is what takes the most time

Kludex:
We should try to cache those images then, I guess. :thinking: 

EDIT: Did you mean the image build or the container startup? :thinking:

euri10:
> Can we can run them in parallel? How slow it becomes?

I tried pytest-xdist which blows the test suite, I think mostly because ports used for each run_in_server are the same so maybe something worh looking at, not sure

Kludex:
> I tried pytest-xdist which blows the test suite, I think mostly because ports used for each run_in_server are the same so maybe something worh looking at, not sure

I didn't mean on the pytest level, I mean on the CI, creating two jobs in parallel.

euri10:
> > I tried pytest-xdist which blows the test suite, I think mostly because ports used for each run_in_server are the same so maybe something worh looking at, not sure
> 
> I didn't mean on the pytest level, I mean on the CI, creating two jobs in parallel.

ok I seperated them into 2 jobs using markers in pytest, would like to see once [this](https://github.com/encode/uvicorn/pull/1077) is merged

PR #1082, opened 2021-06-18
Correct usage of param "message" in logger.info()

laggardkernel:
In one of the `logger.info()` usage, param `message` is defined but not used. Correct it.


Kludex (approved):
Nice catch! :eagle: 

Thanks! :tada:

PR #1083, opened 2021-06-18
Add missing trace log for websocket protocols

laggardkernel:
Formerly we only have trace log for http protocols on `connection_made` and 
`connection_lost` events. **Not very sure**, but I guess we forget
to add trace log for websocket protocols?

### DIFF
--- a/tests/middleware/test_logging.py
+++ b/tests/middleware/test_logging.py
@@ -3,6 +3,7 @@ import logging
 
 import httpx
 import pytest
+import websockets
 
 from tests.utils import run_server
 from uvicorn import Config
@@ -43,6 +44,55 @@ async def test_trace_logging(caplog):
         assert "ASGI [2] Send " in messages.pop(0)
         assert "ASGI [2] Send " in messages.pop(0)
         assert "ASGI [2] Completed" in messages.pop(0)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("http_protocol", [("h11"), ("httptools")])
+async def test_trace_logging_on_http_protocol(http_protocol, caplog):
+    config = Config(app=app, log_level="trace", http=http_protocol)
+    with caplog_for_logger(caplog, "uvicorn.error"):
+        async with run_server(config):
+            async with httpx.AsyncClient() as client:
+                response = await client.get("http://127.0.0.1:8000")
+        assert response.status_code == 204
+        messages = [
+            record.message
+            for record in caplog.records
+            if record.name == "uvicorn.error"
+        ]
+        assert any(" - Connection made" in message for message in messages)
+        assert any(" - Connection lost" in message for message in messages)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("ws_protocol", [("websockets"), ("wsproto")])
+async def test_trace_logging_on_ws_protocol(ws_protocol, caplog):
+    async def websocket_app(scope, receive, send):
+        assert scope["type"] == "websocket"
+        while True:
+            message = await receive()
+            if message["type"] == "websocket.connect":
+                await send({"type": "websocket.accept"})
+            elif message["type"] == "websocket.disconnect":
+                break
+
+    async def open_connection(url):
+        async with websockets.connect(url) as websocket:
+            return websocket.open
+
+    config = Config(app=websocket_app, log_level="trace", ws=ws_protocol)
+    with caplog_for_logger(caplog, "uvicorn.error"):
+        async with run_server(config):
+            is_open = await open_connection("ws://127.0.0.1:8000")
+        assert is_open
+        messages = [
+            record.message
+            for record in caplog.records
+            if record.name == "uvicorn.error"
+        ]
+        print(messages)
+        assert any(" - Connection made" in message for message in messages)
+        assert any(" - Connection lost" in message for message in messages)
 
 
 @pytest.mark.asyncio
--- a/tests/middleware/test_logging.py
+++ b/tests/middleware/test_logging.py
@@ -60,8 +60,8 @@ async def test_trace_logging_on_http_protocol(http_protocol, caplog):
             for record in caplog.records
             if record.name == "uvicorn.error"
         ]
-        assert any(" - Connection made" in message for message in messages)
-        assert any(" - Connection lost" in message for message in messages)
+        assert any(" - HTTP connection made" in message for message in messages)
+        assert any(" - HTTP connection lost" in message for message in messages)
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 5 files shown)


euri10 (changes requested):
looks good, this would look better with some tests added, I think we may have some in test_logging.py !

~~now one thing I just realize reading what seem to be a copy of what we do in the http protocols is the following (bear with me it's been a while I did not touch this) : self.logger here is a `uvicorn.error` logger, however iirc when we set `log_level` to trace in Config, we end up with a middleware that spits out those trace log levels on the `uvicorn.asgi` logger. so should we put those under the `uvicorn.asgi` logger ? not sure if that makes sense.~~

no, correcting myself, those are good under that uvicorn.error logger, those are not strictly speaking asgi message where we put the one sent by the middlware

tests/middleware/test_logging.py
```diff
@@ -45,6 +46,54 @@ async def test_trace_logging(caplog):
         assert "ASGI [2] Completed" in messages.pop(0)
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("protocol", [("h11"), ("httptools")])
+async def test_trace_logging_on_http_protocol(protocol, caplog):
+    config = Config(app=app, log_level="trace", http=protocol)
+    with caplog_for_logger(caplog, "uvicorn.protocol"):
```
euri10:
```suggestion
    with caplog_for_logger(caplog, "uvicorn.error"):
```
I merged your PR about that, should be good with the "correct" logger name ?

tests/middleware/test_logging.py
```diff
@@ -45,6 +46,54 @@ async def test_trace_logging(caplog):
         assert "ASGI [2] Completed" in messages.pop(0)
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("protocol", [("h11"), ("httptools")])
```
euri10:
```suggestion
@pytest.mark.parametrize("http_protocol", [("h11"), ("httptools")])
```
small naming variable fight :)

tests/middleware/test_logging.py
```diff
@@ -45,6 +46,54 @@ async def test_trace_logging(caplog):
...
+            record.message
+            for record in caplog.records
+            if record.name == "uvicorn.error"
+        ]
+        assert any(" - Connection made" in message for message in messages)
+        assert any(" - Connection lost" in message for message in messages)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("protocol", [("websockets"), ("wsproto")])
```
euri10:
```suggestion
@pytest.mark.parametrize("ws_protocol", [("websockets"), ("wsproto")])
```

tests/middleware/test_logging.py
```diff
@@ -45,6 +46,54 @@ async def test_trace_logging(caplog):
...
+                await send({"type": "websocket.accept"})
+            elif message["type"] == "websocket.disconnect":
+                break
+
+    async def open_connection(url):
+        async with websockets.connect(url) as websocket:
+            return websocket.open
+
+    config = Config(app=websocket_app, log_level="trace", ws=protocol)
+    with caplog_for_logger(caplog, "uvicorn.protocol"):
```
euri10: same as above for the logger's name

tests/middleware/test_logging.py
```diff
@@ -45,6 +46,54 @@ async def test_trace_logging(caplog):
...
+        async with run_server(config):
+            is_open = await open_connection("ws://127.0.0.1:8000")
+        assert is_open
+        messages = [
+            record.message
+            for record in caplog.records
+            if record.name == "uvicorn.error"
+        ]
+        assert any(" - Connection made" in message for message in messages)
+        assert any(" - Connection lost" in message for message in messages)
```
euri10: funny, in http test we have one connection made / one connection lost and in ws one we have 2 connection made message, probably the upgrade, not sure.
Not a blocker but I'd like to understand :)

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -7,6 +7,7 @@
 import websockets
 from websockets.extensions.permessage_deflate import ServerPerMessageDeflateFactory
 
+from uvicorn.config import TRACE_LOG_LEVEL
```
euri10:
```suggestion
from uvicorn.logging import TRACE_LOG_LEVEL
```

euri10: if someone ca review the relevant PR about that and we merge it first :)

euri10:
genuine question @laggardkernel : why did you need that ?

tests/middleware/test_logging.py
laggardkernel: Yes. Since GH-1086 has been merged, I could set the correct logger name.

tests/middleware/test_logging.py
laggardkernel: Yes. There's an upgrade. Http protocol but not ws protocol is set as the default protocol for transport in `handle_http()`.

Maybe we should specify the connection type in protocol trace log.

```python
self.logger.log(TRACE_LOG_LEVEL, "%sHttp connection made", prefix)
# or
self.logger.log(TRACE_LOG_LEVEL, "%sWebsocket connection made", prefix)
```

laggardkernel:
> genuine question @laggardkernel : why did you need that ?

What you mean by "that", this PR?

euri10:
> > genuine question @laggardkernel : why did you need that ?
> 
> What you mean by "that", this PR?

yes I mean the ws trace messages :)

tests/middleware/test_logging.py
laggardkernel: Another way to discriminate them is to trace log the "upgrade" request.

euri10: > Another way to discriminate them is to trace log the "upgrade" request.

would be cool yes, trace should be precise if needed, this is a good disambiguation

laggardkernel:
@euri10 Not sure. Because there's only trace log for http protocol. It made me wonder if we forgot to add the same level log for ws.

I did a searching about the history of trace log on http protocol.

Logging on connection made, connection lost was first introduced in uvicorn 0.2.0,
commit 07c6c18. At that time, it was categorized as access log. (In fact, there was
no separate "uvicorn.access" loggers yet.)

Later, with the coming of the `TRACE_LOG_LEVEL` in commit 1238455, http protocol connection logs was changed
to be logged with level 5.

Maybe @tomchristie could answer this, considering he's the guy designed the trace log and later integrated it into `MessageLoggerMiddleware`.

euri10:
Ok I just wondered if there was a use case behind that and if you needed it to debug some sort of issue !

euri10 (approved):
this looks beautiful, thanks a lot @laggardkernel  and the benefit of having that trace level declared everywhere is that thre's no need to wait for the refactor PR merge !

PR #1084, opened 2021-06-18
Moved TRACE_LOG_LEVEL in one spot

euri10:
we have it duplicated everywhere !

### DIFF
--- a/tests/middleware/test_message_logger.py
+++ b/tests/middleware/test_message_logger.py
@@ -1,9 +1,8 @@
 import httpx
 import pytest
 
+from uvicorn.config import TRACE_LOG_LEVEL
 from uvicorn.middleware.message_logger import MessageLoggerMiddleware
-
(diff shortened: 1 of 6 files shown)


laggardkernel:
Such a quick move!!! I found this problem during pushing GH-1083. When I was trying to push the same change, just found you've already done it.

I first chose to keep only one `TRACE_LOG_LEVEL` definition in `config.py`. But it caused a circular import when doing `from uvicorn.logging import TRACE_LOG_LEVEL` within `message_logger.py`. Seems you alleviated this by keeping `TRACE_LOG_LEVEL` definition both in `config.py` and `logging.py`.

euri10:
Did I let it in 2 spots ,will take a better look then

euri10:
ok I think I have it only in one spot now !

Kludex (approved):

PR #1087, opened 2021-06-21
Removed print from test

euri10:
Leftover from #1083 I reviewed too fast it seems


Kludex (approved):

Kludex:
I'm in favor of adding https://pypi.org/project/flake8-print/, jfyk.

PR #1088, opened 2021-06-21
Partial fix of ProxyMiddlware where the request comes from the proxy itself

### LINKED ISSUE #1068: Using ProxyHeadersMiddleware with client on trusted proxy results in invalid client information
### Checklist

### Describe the bug

Using `ProxyHeadersMiddleware` results in the client being `[None, 0]` when the original client runs on a trusted proxy.

In our setup, we have a client service which runs on one of our proxy servers. The proxy servers run on HAProxy with `option forwardfor` to add the `X-Forwarded-For` header. With today's upgrade to uvicorn `0.14.0`, the client fields are no longer present for requests coming from the mentioned server.

### To reproduce

This addition to the tests should illustrate the problem:

```patch
diff --git a/tests/middleware/test_proxy_headers.py b/tests/middleware/test_proxy_headers.py
index aacf789..b453892 100644
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -27,6 +27,8 @@ async def app(scope, receive, send):
         ("127.0.0.1, 10.0.0.1", "Remote: https://1.2.3.4:0"),
         # request from untrusted proxy
         ("192.168.0.1", "Remote: http://127.0.0.1:123"),
+        # request from client running on proxy server
+        (["127.0.0.1", "1.2.3.4"], "Remote: https://1.2.3.4:0"),
     ],
 )
 async def test_proxy_headers_trusted_hosts(trusted_hosts, response_text):
```

which results in:

```py
>       assert response.text == response_text
E       AssertionError: assert 'Remote: https://None:0' == 'Remote: https://1.2.3.4:0'
E         - Remote: https://1.2.3.4:0
E         ?                 ^^^^^^^
E         + Remote: https://None:0
E         ?                 ^^^^
```

### Expected behavior

The proper client address should be forwarded as before.

### Actual behavior

The client is set to `[None, 0]`.

### Debugging material

I assume this is a regression of https://github.com/encode/uvicorn/pull/591, more specifically, see the fallthrough discussion on https://github.com/encode/uvicorn/pull/591/files#r438008416. That said, I am not entirely sure what the best way to fix this would be - based on the discussion, the original code seems to be security-related. Would returning the last entry in `x_forwarded_for` be sufficient?

### Environment

```
Running uvicorn 0.14.0 with CPython 3.9.1 on Linux
```

I can provide proxy & detailed uvicorn configuration if needed, but I will need to sanitise it, that said, I am not sure if it's needed for this issue.

> [!IMPORTANT]

<a href="https://polar.sh/encode/uvicorn/issues/1068">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/1068/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

euri10:
Fixes https://github.com/encode/uvicorn/issues/1068

If you can take a look @b0g3r that would be appreciated !

It adresses only point 1 of the [comment](https://github.com/encode/uvicorn/issues/1068#issuecomment-855371576) you made though

### DIFF
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -15,28 +15,30 @@ async def app(scope, receive, send):
 
 @pytest.mark.asyncio
 @pytest.mark.parametrize(
-    ("trusted_hosts", "response_text"),
+    ("trusted_hosts", "x_forwarded_for", "response_text"),
     [
-        # always trust
-        ("*", "Remote: https://1.2.3.4:0"),
-        # trusted proxy
-        ("127.0.0.1", "Remote: https://1.2.3.4:0"),
-        (["127.0.0.1"], "Remote: https://1.2.3.4:0"),
-        # trusted proxy list
-        (["127.0.0.1", "10.0.0.1"], "Remote: https://1.2.3.4:0"),
-        ("127.0.0.1, 10.0.0.1", "Remote: https://1.2.3.4:0"),
-        # request from untrusted proxy
-        ("192.168.0.1", "Remote: http://127.0.0.1:123"),
+        # # always trust
+        # ("*", "1.2.3.4", "Remote: https://1.2.3.4:0"),
+        # # trusted proxy
+        # ("127.0.0.1", "1.2.3.4", "Remote: https://1.2.3.4:0"),
+        # (["127.0.0.1"], "1.2.3.4", "Remote: https://1.2.3.4:0"),
+        # # trusted proxy list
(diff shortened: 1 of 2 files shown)


pypae:
Why was this closed?

Background: I'm working on an implementation for [another issue](https://github.com/encode/uvicorn/issues/1068#issuecomment-1004813267) that came up in the same thread and wondered if I should also include fixes for this problem.

PR #1091, opened 2021-06-23
Allow use custom server class

abersheeran:
Related links:
- https://stackoverflow.com/questions/58133694/graceful-shutdown-of-uvicorn-starlette-app-with-websockets

This PR will allow passing `Server` to `uvicorn.run` via parameters. This will make it easier to handle the problem of long connections that cannot be closed normally.

```python
class Server(uvicorn.Server):
    def handle_exit(self, sig, frame):
        application = self.config.loaded_app
        while not isinstance(application, Starlette):
            application = application.app
        application.should_exit = True
        return super().handle_exit(sig, frame)


uvicorn.run(app, server_class=Server)
```

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -404,12 +404,17 @@ def main(
         "use_colors": use_colors,
         "factory": factory,
     }
-    run(app, **kwargs)
+    run(app, server_class=Server, **kwargs)
 
 
-def run(app: typing.Union[ASGIApplication, str], **kwargs: typing.Any) -> None:
+def run(
+    app: typing.Union[ASGIApplication, str],
+    *,
+    server_class: typing.Type[Server] = Server,
+    **kwargs: typing.Any,
+) -> None:
     config = Config(app, **kwargs)
-    server = Server(config=config)
+    server = server_class(config=config)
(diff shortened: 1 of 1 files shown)


euri10:
> This PR will allow passing `Server` to `uvicorn.run` via parameters. This will make it easier to handle the problem of long connections that cannot be closed normally.

do you have a reproducible application that demonstrate the issue @abersheeran please ?

euri10:
It brings memory, I think https://github.com/encode/uvicorn/issues/451 is the very same.

I'd prefer to solve the shutdown hanging on tasks not cancelled once ctrl+c has been sent, instead of this patch

but I need to look at it more before making hasty conclusions

abersheeran:
For the time being, I don’t have a minimal application that can be reproduced, only a large program. But through [related link](https://stackoverflow.com/questions/58133694/graceful-shutdown-of-uvicorn-starlette-app-with-websockets) on StackOverflow, this problem should be able to reproduce.

Kludex:
There's also #1012 to consider.

Kludex:
Closing this in favor of #1248.

PR #1092, opened 2021-06-23
Added some timeouts for long-running tasks or connections that won't close by themselves

### LINKED ISSUE #451: How do I cancel a streaming response on shutdown?

euri10:
Alternative to https://github.com/encode/uvicorn/pull/1091
Would that solve your issue @abersheeran ?

Fixes #451

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -43,6 +43,8 @@ class Server:
         self.should_exit = False
         self.force_exit = False
         self.last_notified = 0
+        self.shutdown_connections_timeout = 5
+        self.shutdown_background_tasks_timeout = 5
 
     def run(self, sockets=None):
         self.config.setup_event_loop()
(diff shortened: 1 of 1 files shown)


abersheeran:
Yes, this can solve the problem of infinite waiting.

But I think that giving the application a flag does not conflict with this, application can be handled by itself.

PR #1094, opened 2021-06-25
⬆️ Bump uvloop extra requirements to >=0.15.2

Kludex:
This is a recommendation made by @1st1 (@MagicStack member) in https://github.com/encode/uvicorn/pull/952#r578815980.


Kludex:
Ups! I forgot Python 3.6. Current pin should be fine until we lose support for 3.6.

PR #1097, opened 2021-06-26
🔧 Add funding option

Kludex:
Same as httpx: https://github.com/encode/httpx/blob/master/.github/FUNDING.yml


lovelydinosaur (approved):
Make sense really, yup.
Will need to think about how we're going to work sponsor ads on the docs at some point.

PR #1014, opened 2021-04-05
Typing/added typing to httptools_impl.py module

jkklapp:
Part of #998

PR #1102, opened 2021-06-29
Cover httptools_impl on mypy

Kludex:
Continuation of #1014 by @jkklapp .


uvicorn/_types.py
```diff
@@ -1,14 +1,44 @@
...
-WSGIApp = typing.Callable[
-    [Environ, StartResponse], typing.Union[typing.Iterable[bytes], BaseException]
-]
+Environ = MutableMapping[str, Any]
+ExcInfo = Tuple[Type[BaseException], BaseException, Optional[types.TracebackType]]
+StartResponse = Callable[[str, Iterable[Tuple[str, str]], Optional[ExcInfo]], None]
+WSGIApp = Callable[[Environ, StartResponse], Union[Iterable[bytes], BaseException]]
+
+
+class WebProtocol(Protocol):
```
Kludex, line 32: Here, I'm not sure if we should actually call it `WebProtocolProtocol`, because the `Protocol` is from typing. :thinking:

uvicorn/server_state.py
```diff
@@ -0,0 +1,16 @@
+import asyncio
```
Kludex, line 1: I had to move this due to circular import on the `WebProtocol`. :cry:

Kludex: 11 hours after doing this, I wonder if this is a good idea. It's a breaking change as we were exposing `ServerState` on `server.py`. I can't find a better solution, so suggestions are welcome.

PR #1106, opened 2021-07-01
🔨 Add script that fills CHANGELOG to release

Kludex:
It's incomplete :kissing: 

Still thinking about this.


Kludex:
Yeah... I'm going to close this, and think about other alternatives.

PR #1108, opened 2021-07-05
Annotate main run() function

Kludex:
## Context

It's a known fact that there are a lot of users who prefer the _code style_ `run()`, instead of using the CLI.

On this PR, we want to be able to be able to achieve an improvement of dev experience when running the `run()` function.

As a developer, I feel very comfortable when I (using autocomplete) am able to see the possible parameters available.

## Possible solutions

On this PR there are two solutions, only one is necessary to achieve the proposed goal.

1. Using stub files e.g. `main.pyi`.
2. Annotating the `run()` and passing the same arguments to `Config`.

@graingert do you have an alternative here? I didn't like either :sweat_smile: 

EDIT: After considering the proposed solution, I got a new one:
3. Use `**kwargs` on the `main()` function, and use the actual types on the `run()` one.


Kludex:
I'm happy to hear alternatives. I'm still not convinced with the results, but it does provide a better user experience.

graingert:
Hmm how about generating the annotations with https://github.com/nedbat/cog ?

uvicorn/config.py
```diff
@@ -145,8 +146,8 @@ def __init__(
         http: Union[Type[asyncio.Protocol], HTTPProtocolType] = "auto",
         ws: Union[Type[asyncio.Protocol], WSProtocolType] = "auto",
         ws_max_size: int = 16 * 1024 * 1024,
-        ws_ping_interval: int = 20,
-        ws_ping_timeout: int = 20,
+        ws_ping_interval: float = 20,
```
graingert, line 149: Eventually passed here:

https://github.com/aaugustin/websockets/blob/361f38066622636eff8d9e0d3d071e93e8e02b05/src/websockets/legacy/server.py#L994-L995

graingert (dismissed):
Should these be widened to Optional [float]?

Kludex:
> Hmm how about generating the annotations with https://github.com/nedbat/cog ?

You mean to not rewrite in every place? :thinking: 
Not sure if I get the idea for this PR. Would you mind explaining how?

PR #1109, opened 2021-07-06
✅ Increase coverage threshold from 95 to 96

Kludex:
It's important to keep updating the coverage threshold, so we don't lose coverage, and we can achieve the 100% that we always wanted.

EDIT: The first time it failed, it was because I've set to 97. Now it passes with 96. :smile:

### DIFF
--- a/scripts/coverage
+++ b/scripts/coverage
@@ -8,4 +8,4 @@ export SOURCE_FILES="uvicorn tests"
 
 set -x
 
-${PREFIX}coverage report --show-missing --skip-covered --fail-under=95
+${PREFIX}coverage report --show-missing --skip-covered --fail-under=97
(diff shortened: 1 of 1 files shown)


scripts/coverage
```diff
@@ -8,4 +8,4 @@ export SOURCE_FILES="uvicorn tests"
 
 set -x
 
-${PREFIX}coverage report --show-missing --skip-covered --fail-under=95
+${PREFIX}coverage report --show-missing --skip-covered --fail-under=97
```
Kludex:
```suggestion
${PREFIX}coverage report --show-missing --skip-covered --fail-under=96
```

Kludex:
The coverage on Windows is 95% yet. :sweat:

PR #1110, opened 2021-07-06
Import application before running the event loop

Kludex:
Replaces #347.

Text from @thmsmlr:
> Before this patch there was an inconsistency where if you ran uvicorn
> from the CLI versus programmatically the eventloop would either be or
> not be running when the code was imported. It is generally good practice
> to make sure that the event loop isn't running when code is imported
> since async code can only be run at import time using the
> loop.run_until_complete(...) command which requires the event loop to
> not be running.
> 
> This patch makes sure that the code is loaded after the event loop is
> configured, but before it starts running. This allows for consistent
> behavior across deployment options.
> 
> Solves: #346


Kludex:
For some reason, the test I've added here makes the `test_app_close` to fail on Windows. @euri10 do you know why? 

I had to add the `skipif` decorator restricting Windows.

euri10:
> For some reason, the test I've added here makes the `test_app_close` to fail on Windows. @euri10 do you know why?
> 
> I had to add the `skipif` decorator restricting Windows.

no idea I see no failed windows test nor stacktrace, you got one ?

Kludex:
It was on the previous commit: https://github.com/encode/uvicorn/runs/3002970222
I've added the skipif, that's why you don't see it now.

euri10:
> For some reason, the test I've added here makes the `test_app_close` to fail on Windows. @euri10 do you know why?
> 
> I had to add the `skipif` decorator restricting Windows.

 it seems to me we may still have unclosed ressources in tests, your addition just revealed it, either it's your new test that adds it, either something else, it is very hard to pinpoint

Kludex:
Maybe we can create a fixture that runs on every test function and check the resources created by pytest? :thinking:

graingert:
> Maybe we can create a fixture that runs on every test function and check the resources created by pytest? thinking

pytest is supposed to do this for you

tests/test_cli.py
```diff
@@ -26,5 +34,21 @@ def test_cli_headers():
...
-    pass
+@pytest.mark.skipif(sys.platform == "win32", reason="require unix-like system")
+def test_load_app_before_event_loop(tmp_path: Path):
+    runner = CliRunner()
+    fp = tmp_path / "main.py"
+    content = textwrap.dedent(
+        """
+        import asyncio
+
+        print("Event loop running:", asyncio.get_event_loop().is_running(), end="")
```
graingert, line 67: from 3.10 (with `-W error`) `asyncio.get_event_loop()` will only return a running loop

graingert (changes requested):
calling `get_event_loop()` without a running loop is no longer supported

tests/test_cli.py
```diff
@@ -26,5 +34,21 @@ def test_cli_headers():
     ]
 
 
-class App:
-    pass
+@pytest.mark.skipif(sys.platform == "win32", reason="require unix-like system")
```
graingert:
```suggestion
@pytest.mark.filterwarnings(r"ignore:unclosed .*:ResourceWarning")
```

Kludex: The suggestion was supposed to be an addition, not a replacement, right?

graingert: no this should be a replacement - if you ignore the ResourceWarning the test will pass on 3.9 or less on windows or linux

tests/test_cli.py
```diff
@@ -26,5 +34,21 @@ def test_cli_headers():
...
+        print("Event loop running:", asyncio.get_event_loop().is_running(), end="")
+
+        async def app(scope, receive, send):
+            pass
+        """
+    )
+    fp.write_text(content)
+    with mock.patch("uvicorn.server.Server.serve"):
+        result = runner.invoke(cli, ["main:app", "--app-dir", tmp_path])
+        assert "Event loop running: False" in result.stdout
```
graingert, line 76:
```suggestion
def test_load_app_before_event_loop(tmp_path: Path):
    runner = CliRunner()
    fp = tmp_path / "main.py"
    content = textwrap.dedent(
        """
        import anyio

        async def example():
            await anyio.sleep(0)
            print("All good!")

        anyio.run(example)

        async def app(scope, receive, send):
            pass
        """
    )
    fp.write_text(content)
    with mock.patch("uvicorn.server.Server.serve"):
        assert (
            runner.invoke(cli, ["main:app", "--app-dir", str(tmp_path)]).stdout
            == "All good!\n"
        )
```

graingert:
I think it might be worth considering how this would interact with https://github.com/encode/uvicorn/pull/1152

I think a separate `config.import_app()` and `async with config.app_context()` is needed

Kludex:
- Replaced by #1488

PR #1114, opened 2021-07-10
Read uvicorn settings from environment variables

### LINKED ISSUE #693: Include a environment variable to change arg commands

Kludex:
Closes #693

This is a further implementation proposed by #696. The solution proposed there only enabled `uvicorn` to read environment variables in case it was run using `python -m uvicorn` instead of the usual `uvicorn` entry point. This PR solves that issue.

Here we go further and add a test that loads one of the many settings options, and check that it's used successfully by `uvicorn`, and also shows a priority between CLI arguments and environment variables i.e. CLI arguments always have priority in case environment variables are present. 

### Do we have alternatives for this?

Yes. Gunicorn uses a single environment variable called `GUNICORN_CMD_ARGS`, instead of using a prefix. See more about it [here](https://docs.gunicorn.org/en/stable/settings.html#settings).

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -15,10 +15,12 @@ nav:
     - Server Behavior: 'server-behavior.md'
 
 markdown_extensions:
+  - admonition
   - codehilite:
       css_class: highlight
   - toc:
       permalink: true
+
 
 extra_javascript:
   - 'js/chat.js'
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -7,6 +7,13 @@ equivalent keyword arguments, eg. `uvicorn.run("example:app", port=5000, reload=
 Please note that in this case, if you use `reload=True` or `workers=NUM`,
 you should put `uvicorn.run` into `if __name__ == '__main__'` clause in the main module.
 
+You can also configure Uvicorn using environment variables with the prefix `UVICORN_`.
+For example, in case you want to run the app on port `5000`, just set the environment variable `UVICORN_PORT` to `5000`.
+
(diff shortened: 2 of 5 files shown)


mkdocs.yml
```diff
@@ -15,11 +15,13 @@ nav:
     - Server Behavior: 'server-behavior.md'
 
 markdown_extensions:
+  - admonition
```
Kludex, line 18: This enables `!!! note` on the docs. Reference: https://squidfunk.github.io/mkdocs-material/reference/admonitions/

kolypto:
+1 to this!

Kludex:
Close in favor of #1279

PR #1119, opened 2021-07-11
Implement asgiref tls extension

### LINKED ISSUE #1118: Implement TLS ASGI extension
### Checklist

No

Implementation of the TLS ASGI extension. Recently this extension has been formalized via https://github.com/django/asgiref/pull/192. It would be nice to get this implemented in Uvicorn so that applications can start using this info.

N/A

> [!IMPORTANT]

<a href="https://polar.sh/encode/uvicorn/issues/1118">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/1118/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

mdgilene:
This is a partial implementation of the new TLS extension added to the ASGI spec.

This PR partially implements the new TLS extension which defines a new set of attributes to be added to the request scope under `extensions.tls`.

#### `server_cert`
PEM encoded server certificate.

#### `client_cert_chain`
List of PEM encoded certificates in the client certificate chain.

It appears at the moment only the last certificate in the chain is available from the `SSLObject`. So this implementation will only ever provide 1 certificate in the cert chain.

#### `client_cert_name`
DN for the first certificate present in the `client_cert_chain`.

Without pulling in an additional dependency to handle X509 parsing, obtain this value is a bit more difficult. I have added in a very basic lookup table to be able to obtain the DN in string format. Please if anyone else has a better way to handle this I am all ears.

#### `client_cert_error`
Error if verification of client certificate chain failed.

I do not believe the `SSLObject` provides this information. If it doesn't I couldn't find it anywhere. So without it we just leave this blank at all times.

#### `tls_version`
TLS version number.

The spec stats this should be the integer for the version number defined in the TLS specification. Again, I have achieved this through a simple lookup table. If there is a more standard way of obtaining this I would appreciate any help here.

#### `cipher_suite`
TLS cipher suite.

Again, the spec states ` This is a 16-bit unsigned integer that encodes the pair of 8-bit integers specified in the relevant RFC, in network byte order`. Without direct access to openssl I don't know how to obtain this value. So for now I have simply set this to `SSLObject.cipher` which is off-spec but I don't know what else to do other than leave it blank.

## Related Issues
#1118

Edit by @Kludex : Closes #1118


mdgilene:
Note, I am new to contributing to this project so looking for input on this PR and if you see issues with code format/style/etc. please let me know and I will try to address it.

Also, I understand this addition to the spec is very new. If this is still blocked until a new release of asgiref that is understandable.

Thanks!

Kludex:
image

If the connection is not over TLS, we shouldn't have it in the `scope`, right?

Reference: https://github.com/django/asgiref/commit/6d19d453347ccb13b5cc8998d18b1a7f94ac523c

mdgilene:
I believe you're right. I missed this piece during my initial read through. I've updated it so that it's only added if the connection is made over TLS.

Kludex:
@mdgilene would you mind rebasing it? It's already released by asgiref, so I think we can start reviewing it. 

@tomchristie @euri10 I'd like this to be reviewed by you guys as well :kissing:

lovelydinosaur:
Interesting stuff, thanks.

Thoughts on stuff retrospectively, that might or might not be valuable to mention...

* It probably would have made sense for the ASGI spec addition there to also demo an implementation that meets the spec. It's a bit odd for us to try to wrangle out some information when there's not a clear indication that the info that's been added to the spec *is actually available to us*. That'd also help tease out if some of the design decisions in that bit of the spec made practical sense or not. (Eg.it looks a bit odd from *this standpoint* having to map TLS version strings into integers)
* It'd possibly be good to have a nice little simple *user-focused* example of "this is what we'd like to be able to do, and why. this new addition to the spec enables that". It's very easy to end up getting into cycles of adding tech for the sake of tech, without really making it clear what user-driven benefits we're attempting to provide for.

Anyways, those are just my high-level thoughts on this overall - perhaps not super useful at this point, but sometimes getting those sort of thought processes out in the open might help influence future changes?

That stuff aside, this seems to look decent. 👍

uvicorn/config.py
```diff
@@ -213,6 +213,7 @@ def __init__(
         self.callback_notify = callback_notify
         self.ssl_keyfile = ssl_keyfile
         self.ssl_certfile = ssl_certfile
+        self.ssl_cert_pem: Optional[str] = None
```
lovelydinosaur: Can we drop this? (And *exclusively populate it from `self.ssl_certfile`*?)

mdgilene:
> Interesting stuff, thanks.
> 
> Thoughts on stuff retrospectively, that might or might not be valuable to mention...
> 
> * It probably would have made sense for the ASGI spec addition there to also demo an implementation that meets the spec. It's a bit odd for us to try to wrangle out some information when there's not a clear indication that the info that's been added to the spec _is actually available to us_. That'd also help tease out if some of the design decisions in that bit of the spec made practical sense or not. (Eg.it looks a bit odd from _this standpoint_ having to map TLS version strings into integers)
> * It'd possibly be good to have a nice little simple _user-focused_ example of "this is what we'd like to be able to do, and why. this new addition to the spec enables that". It's very easy to end up getting into cycles of adding tech for the sake of tech, without really making it clear what user-driven benefits we're attempting to provide for.
> 
> Anyways, those are just my high-level thoughts on this overall - perhaps not super useful at this point, but sometimes getting those sort of thought processes out in the open might help influence future changes?
> 
> That stuff aside, this seems to look decent. 👍

On your points,

1. Yeah I wasn't involved in the writing of the spec but a few of the choices for which fields to provide and in what format don't make sense due to limitations of python's tls module. Proper parsing of the certificate could be done but it would require pulling a dependency. Ultimately like the spec states parsing of the cert should be left up to the end user so parsing the cert for the DN only seems a bit useless. As for the TLS version number idk...honestly as a user I would probably rather just have the string version.

2. That would have been nice yes. I agree things should have a use case to support adding tech so for the sake of argument here, the primary user focused use case that comes to mind that wanting to implement user authentication based on client certificates. This could be done through implementing an OAuth2/OIDC provider service that offers a login flow with X.509 authentication. (If you are familiar with something like Keycloak, it does exactly this). In order to accomplish this though you need access to the client certificate on each connection to be able to parse it for whatever fields you are using to match against existing users.

mdgilene:
> @mdgilene would you mind rebasing it? It's already released by asgiref, so I think we can start reviewing it.
> 
> @tomchristie @euri10 I'd like this to be reviewed by you guys as well 😗

Yep, won't get to it today, but I'll try to get to it tomorrow if I can

Edit: Nevermind, turns out github has an in browser tool for rebasing. How neat :) Was thinking I'd need to pull everything down again. Rebase should be done now

uvicorn/config.py
mdgilene: We need to provide the server cert contents in the scope so I did this to avoid needing to read the certfile in on each connection. If that isn't a concern I can move it to the connection code. Otherwise not sure where else to put it.

Unless I am misunderstanding. ssl_certfile just contains the path to the cert on disk no? I couldn't find any existing location where it is being read in. I assumed it's just being passed down to the underlying socket which handles reading the file.

ajoino:
> > Interesting stuff, thanks.
> > Thoughts on stuff retrospectively, that might or might not be valuable to mention...
> > 
> > * It probably would have made sense for the ASGI spec addition there to also demo an implementation that meets the spec. It's a bit odd for us to try to wrangle out some information when there's not a clear indication that the info that's been added to the spec _is actually available to us_. That'd also help tease out if some of the design decisions in that bit of the spec made practical sense or not. (Eg.it looks a bit odd from _this standpoint_ having to map TLS version strings into integers)
> > * It'd possibly be good to have a nice little simple _user-focused_ example of "this is what we'd like to be able to do, and why. this new addition to the spec enables that". It's very easy to end up getting into cycles of adding tech for the sake of tech, without really making it clear what user-driven benefits we're attempting to provide for.
> > 
> > Anyways, those are just my high-level thoughts on this overall - perhaps not super useful at this point, but sometimes getting those sort of thought processes out in the open might help influence future changes?
> > That stuff aside, this seems to look decent. +1
> 
> On your points,
> 
>     1. Yeah I wasn't involved in the writing of the spec but a few of the choices for which fields to provide and in what format don't make sense due to limitations of python's tls module. Proper parsing of the certificate could be done but it would require pulling a dependency. Ultimately like the spec states parsing of the cert should be left up to the end user so parsing the cert for the DN only seems a bit useless. As for the TLS version number idk...honestly as a user I would probably rather just have the string version.
> 
>     2. That would have been nice yes. I agree things should have a use case to support adding tech so for the sake of argument here, the primary user focused use case that comes to mind that wanting to implement user authentication based on client certificates. This could be done through implementing an OAuth2/OIDC provider service that offers a login flow with X.509 authentication. (If you are familiar with something like Keycloak, it does exactly this). In order to accomplish this though you need access to the client certificate on each connection to be able to parse it for whatever fields you are using to match against existing users.

I have a use case involving JWTs for authorization and authentication. I don't remember all details from the top of my head, but I think it was a JSON encrypted by the "server's" (it's a library for a SOA architecture) public key put into another JSON encrypted by the client's private key. To get the information in the JWT I need the client cert, which currently is unavailable unless running behind nginx or similar.

If you want help I could try to write an example based on my use-case.

EDIT: I remember the use case a bit more now. A service consumer, call it A, got a JWT from a central authorization system, which contained authorization information for a producer, called B. The JWT was signed by A's private key and the authorization system's public key. When A tries to consume a service of B, B must decrypt the JWT, which requires it to get A's public key. The way I tried to implement it was to write a middleware but since I couldn't get the cert from the transport I gave up and worked on other features.

amirkarimi:
Since this branch was behind and some tests were failing I rebased it and fixed the issues in my [fork](https://github.com/amirkarimi/uvicorn/tree/feature-asgiref-tls-extension). Here is the [compare](https://github.com/encode/uvicorn/compare/master...amirkarimi:uvicorn:feature-asgiref-tls-extension). @mdgilene feel free to merge my fork into this PR. If you don't get the time let me know and I can open a new PR.

Kludex:
Relevant reference for reviewer: https://github.com/pgjones/hypercorn/issues/62

mdgilene:
I'd be happy to look at this again as I've been out of the loop on this for a while. But from what it seems like is that we are potentially just waiting for additional updates/modifications to the spec as currently I do not believe the spec to be fully implementable.

amirkarimi:
> I'd be happy to look at this again as I've been out of the loop on this for a while. But from what it seems like is that we are potentially just waiting for additional updates/modifications to the spec as currently I do not believe the spec to be fully implementable.

I might have missed something but my impression from the last comment on https://github.com/pgjones/hypercorn/issues/62 is that the spec can be implemented.

jedrka:
Hello

I did not run code from this merge but after looking into diff I see some problems.
Firstly, string assigned to ssl_info["client_cert_name"] contains spaces after '=' and ',' characters, so the final string is like `OU = Sales + CN = J. Smith, DC = example, DC = net` instead of just `OU=Sales+CN=J.  Smith,DC=example,DC=net` as shown in example in rfc4514. Format with additional spaces can be seen in `openssl` command line tool output, but I think it's because of readability.

Second problem I see is that attribute names not present in RDNS_MAPPING are completely ignored and dropped from output. getpeercert()["subject"] returns attribute names in two formats: "long name" or "numerical. 

According to rfc4514 attributes without "short name" should be represented by 'stringified' numerical identifier, such as `2.5.4.3` (oid of CN attribute). I know that doing it will be complicated because openssl baked into python knows some (actually a lot of, see [openssl /include/openssl/obj_mac.h](https://github.com/openssl/openssl/blob/fdc5043d58900663b493147298e64f11353b35fe/include/openssl/obj_mac.h)) "long names" which get returned in getpeercert()["subject"]. And there is no way to access that table which would allow to convert "long name" given by SSL to the "short name" or "numerical" format.

mdgilene:
Finally got back around to this. I've addressed the merge conflicts and fixed up the tests. There is one linting error I still see but it's for a line I haven't touched so not sure what to do about that since it would see that issue should exist in master already. Any advice about what to do here would be appreciated.

As for the larger discussion as to whether or not this is a sufficient implementation of the specification, I would say it is. The fields that I am leaving set to None are allowed to be None by the spec so I would say it technically meets the requirements. Without pulling in a large library like `cryptography` as a dependency we simply cannot obtain that the required information. It would seem like the best route would be to leave that up to the application developers. If they desire detailed cert information the whole cert is provide in its raw form and they can parse it themselves.

jschlyter:
Any updates on this PR? I'm developing a FastAPI application using mTLS and really like to replace my workaround (monkey-patching `uvicorn.protocols.http.httptools_impl.HttpToolsProtocol`).

pugo:
I think this PR is exactly what I would need to differentiate our clients that use mTLS to connect. 

Is there anything I can do to help with this one?

commonism:
You may want to look at nonecorn or https://github.com/urllib3/hypercorn/tree/tls-extension.
I use nonecorn.

commonism:
> I'm developing a FastAPI application using mTLS

I'm interested.
I use nonecorn,  and (have to?) mangle the FastAPI description document to include mtls and extend the security for the Operation to allow or require mtls. It's part of my unit tests.
https://github.com/commonism/aiopenapi3/blob/master/tests/tls_test.py

Can you show me how you made FastAPI serve optional/required mtls?

mdgilene:
I re-synced this PR with the current master branch. Again, this is still just waiting on a re-review from a maintainer.

@Kludex @tomchristie

uvicorn/protocols/utils.py
```diff
@@ -54,3 +74,44 @@ def get_path_with_query_string(scope: WWWScope) -> str:
...
+
+    ssl_object = transport.get_extra_info("ssl_object", default=None)
+    peercert = ssl_object.getpeercert()
+
+    if peercert:
+        rdn_strings = []
+        for rdn in peercert["subject"]:
+            rdn_strings.append(
+                "+".join(
+                    [f"{RDNS_MAPPING[entry[0]]} = {entry[1]}" for entry in reversed(rdn) if entry[0] in RDNS_MAPPING]
```
jschlyter: I believe there should be no spaces around the `=` nor after the comma delimiter.

uvicorn/protocols/utils.py
```diff
@@ -54,3 +74,44 @@ def get_path_with_query_string(scope: WWWScope) -> str:
...
+                )
+            )
+
+        ssl_info["client_cert_chain"] = [ssl.DER_cert_to_PEM_cert(ssl_object.getpeercert(binary_form=True))]
+        ssl_info["client_cert_name"] = ", ".join(rdn_strings) if rdn_strings else ""
+        ssl_info["tls_version"] = (
+            TLS_VERSION_MAP[ssl_object.version()] if ssl_object.version() in TLS_VERSION_MAP else None
+        )
+        ssl_info["cipher_suite"] = list(ssl_object.cipher())
```
jschlyter: `cipher_suite` should be an integer.

uvicorn/protocols/http/h11_impl.py
```diff
@@ -96,6 +109,11 @@ def connection_made(  # type: ignore[override]
         self.client = get_remote_addr(transport)
         self.scheme = "https" if is_ssl(transport) else "http"
 
+        if self.config.is_ssl:
+            self.tls = get_tls_info(transport)
```
jschlyter, line 113: Given that `uvicorn` requires Python >= 3.8, I'd use the Walrus operator here to increase readability.

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -98,6 +116,11 @@ def connection_made(  # type: ignore[override]
         self.client = get_remote_addr(transport)
         self.scheme = "https" if is_ssl(transport) else "http"
 
+        if self.config.is_ssl:
+            self.tls = get_tls_info(transport)
```
jschlyter, line 120: Dito walrus

jschlyter (changes requested):
I've tested this PR using my own application and it works as expected with some minor nits (see comments). After these are fixed, I support this PR.

uvicorn/protocols/utils.py
mdgilene: After sitting down to look at this again, I'm wondering if we should just not set the `client_cert_name` property. It is marked as optional in the specification and I think properly constructing it is beyond the scope of this library.

Correcting the issue you mentioned is easy, but it did get me thinking about other edge cases involving reserved characters in the RDN values.  This led me to this page which describes how to handle escaping various characters that appear in the values. https://learn.microsoft.com/en-us/previous-versions/windows/desktop/ldap/distinguished-names

IMO this is beginning to get too complicated for a quick lookup table approach and without pulling in something like the `cryptography` library as a dependency which would have robust handling for this situation, I don't see a viable path forward.

I'm leaning towards just providing the bare minimum, i.e. the server cert and client cert in PEM encoded format and leave any parsing of the certificates as an exercise for the user.

uvicorn/protocols/utils.py
mdgilene: The issue here again lies with the specification. I would need a lookup table for every single cipher suite for TLSv1 through TLSv1.3.  The RFC referenced in the specification only has the ciphers for TLSv1.3.

This is another area where the specification doesn't easily align with what is readily available via the Python TLS module.

So rather than trying to collect and maintain a list of cipher mappings I opted for just passing through the result from the underlying SSLObject. This is obviously "off-spec" however. If we want to maintain strict compatibility with the spec then in a similiar regard to the `client_cert_name` we can just not set this value.

uvicorn/protocols/utils.py
jschlyter: I agree, leave it out.

uvicorn/protocols/utils.py
jschlyter: `cipher_suite` is mandatory so one must set it to be compliant. Generate cipher suite values via something like https://gist.github.com/jschlyter/555438387a502f08ed05724b56db60ad perhaps?

mdgilene: Pretty sure making a network request is a non-starter. This would immediately break for applications running in an offline/airgapped environment.

mdgilene: > cipher_suite (integer or None) -- The TLS cipher suite that is being used. This is a 16-bit unsigned integer that encodes the pair of 8-bit integers specified in the relevant RFC, in network byte order. For example [RFC8446 section B.4](https://tools.ietf.org/html/rfc8446#appendix-B.4) defines that the cipher suite TLS_AES_128_GCM_SHA256 is {0x13, 0x01}; that is encoded as a cipher_suite value of 0x1301 (equal to 4865 decimal). Some web server implementations may be unable to provide this (e.g. if TLS is terminated by a separate proxy or load balancer); in that case set to None. Mandatory.

From the spec, the way I read this is that a vale is required but we can just supply None.

jschlyter: > Pretty sure making a network request is a non-starter. This would immediately break for applications running in an offline/airgapped environment.

I'm not saying you should make a network request at runtime (that would be absurd…), but as part of a generated file when building uvicorn.

pugo: Fetching that XML document each build tingles my spider senses. Even though we should normally be able to trust `https://www.iana.org` it would mean that something not reviewed and tested would be included in the builds.

jschlyter: No need to fetch at each build, but a generated file with ciphers to code mapping is needed IMHO. How often it is regenerated and committed can be discussed. I would start adding the generated file to the repo, with a make target to rebuild on request. The list of ciphers is not updated that frequently.

pugo: That sounds like a fine plan!

jschlyter: @mdgilene do you want a PR for your code adding `cipher_suite` support?

jschlyter: @mdgilene https://github.com/jschlyter/uvicorn/tree/feature-asgiref-tls-extension-ciphers for cherry picking

mdgilene: Pulled in the commit with some minor changes.

uvicorn/protocols/utils.py
mdgilene: So, I did end up just pulling in an escape function I was able to find from the python-ldap library that seems to tackle this fairly concisely after I realized that the spec stats that this field is mandatory if the `client_cert_chain` is provided.

uvicorn/protocols/utils.py
```diff
@@ -103,15 +130,19 @@ def get_tls_info(transport: asyncio.Transport) -> dict[object, object]:
...
             )
 
         ssl_info["client_cert_chain"] = [ssl.DER_cert_to_PEM_cert(ssl_object.getpeercert(binary_form=True))]
-        ssl_info["client_cert_name"] = ", ".join(rdn_strings) if rdn_strings else ""
+        ssl_info["client_cert_name"] = ",".join(rdn_strings) if rdn_strings else ""
         ssl_info["tls_version"] = (
             TLS_VERSION_MAP[ssl_object.version()] if ssl_object.version() in TLS_VERSION_MAP else None
         )
-        ssl_info["cipher_suite"] = list(ssl_object.cipher())
+        ssl_info["cipher_suite"] = TLS_CIPHER_SUITES[ssl_object.cipher()[0]]
```
jschlyter: If the choosen cipher is not in the table, perhaps we should return None?

`TLS_CIPHER_SUITES.get(ssl_object.cipher()[0], None)`

mdgilene:
I think this PR is at the point where I have no other additional changes and I believe it adequately implements the TLS extension as it is currently written. Down the line I would probably like parts of the spec re-visited as they don't make much sense within what is readily available from the native python SSL module however that is outside the scope of this PR.

At this point just looking for a maintainer to sign off.

uvicorn/protocols/utils.py
```diff
@@ -54,3 +75,74 @@ def get_path_with_query_string(scope: WWWScope) -> str:
...
+                    ]
+                )
+            )
+
+        ssl_info["client_cert_chain"] = [ssl.DER_cert_to_PEM_cert(ssl_object.getpeercert(binary_form=True))]
+        ssl_info["client_cert_name"] = ",".join(rdn_strings) if rdn_strings else ""
+        ssl_info["tls_version"] = (
+            TLS_VERSION_MAP[ssl_object.version()] if ssl_object.version() in TLS_VERSION_MAP else None
+        )
+        ssl_info["cipher_suite"] = getattr(TLS_CIPHER_SUITES, ssl_object.cipher()[0], None)
```
jschlyter: This is not correct, please try:
```
ssl_info["cipher_suite"] = TLS_CIPHER_SUITES.get(ssl_object.cipher()[0], None)
```

jschlyter (changes requested):

pugo:
This is great work!

I have been using this branch for some days now during development. I haven't stressed any corner cases, but it does what I expect. With this I can read the client certificate to extract any extra data I have added to it. 

This allows us to adjust config and behavior depending on certificate info, without need to add user and session management.

uvicorn/protocols/utils.py
mdgilene: Thanks, I had just made a change in the tests that required `getattr` and had it on the mind and just went with it.

jschlyter (approved):

Kludex:
@synodriver We didn't talk much yet, but I think you are the only one who implemented this extension. Would you like to give a first review on this PR?

synodriver:
> @synodriver We didn't talk much yet, but I think you are the only one who implemented this extension. Would you like to give a first review on this PR?

Sorry for the late reply, I think this is definitely the first implementation and I just migrated it to hypercorn, hope it works as expected. (I am not very familiar with ssl). According to the feedback from users who are already use my fork, it should be fine... Or maybe we can release a beta version for this feature and wait for feedback from uvicron users?

lschneidpro:
What is the ETA ? Looking forward to use the tsl extension

Kludex:
People here, can you contribute to https://github.com/django/asgiref/issues/466, please?

PR #1121, opened 2021-07-12
Add socket annotation to utils functions

Kludex:
I was trying to solve one of the tests, but I couldn't. This is a consolation prize. :sunglasses:

### DIFF
--- a/uvicorn/protocols/utils.py
+++ b/uvicorn/protocols/utils.py
@@ -7,7 +7,7 @@ from asgiref.typing import WWWScope
 
 
 def get_remote_addr(transport: asyncio.Transport) -> Optional[Tuple[str, int]]:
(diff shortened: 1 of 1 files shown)


uvicorn/protocols/utils.py
```diff
@@ -1,12 +1,13 @@
...
 import urllib.parse
+from socket import socket
 from typing import Optional, Tuple
 
 from asgiref.typing import WWWScope
 
 
 def get_remote_addr(transport: asyncio.Transport) -> Optional[Tuple[str, int]]:
-    socket_info = transport.get_extra_info("socket")
+    socket_info: Optional[socket] = transport.get_extra_info("socket")
```
abersheeran, line 10: Maybe use https://docs.python.org/3/library/socket.html#socket.SocketType is better.

PR #1125, opened 2021-07-16
Add PyPy to the Test Suite

Kludex:
The pipeline fails due to the `codecs.h`, more info can be found [here](https://foss.heptapod.net/pypy/pypy/-/issues/3350).

Next steps on this branch:
- [x] Remove `black` from PyPy installation - and check if another package uses `typed-ast` .


tests/test_auto_detection.py
```diff
@@ -27,10 +27,6 @@ async def app(scope, receive, send):
     pass  # pragma: no cover
 
 
-# TODO: Add pypy to our testing matrix, and assert we get the correct classes
-#       dependent on the platform we're running the tests under.
```
Kludex, line 31: @tomchristie Do you remember what do you mean by "assert we get the correct classes dependent..." ?

Kludex:
`httptools` doesn't support `PyPy`, so for this to work we'd have to add conditionals around to avoid its usage. Is it worth it?

Edit: The error pointed out what I just said, but if that is true, then I guess we should add a marker on the setup.py for `httptools`?

Kludex:
The only remaining issue here is the coverage on `httptools_impl.py` that causes it to be under 95%. 

What should we do about it? Just ignoring the coverage of `httptools_impl.py` doesn't look right.

#1135 should be merged first.

euri10:
I thought uvloop wasn't working with pypy, did that change ?

graingert:
> and check if another package uses typed-ast .

black doesn't use typed-ast anymore

Kludex:
uvloop doesn't support pypy, but uvloop is optional on uvicorn itself.

PR #1127, opened 2021-07-18
✏️ Fix ping parameters annotation in Config class

Kludex:
Hard to find :sweat_smile:


graingert:
https://github.com/aaugustin/websockets/blob/361f38066622636eff8d9e0d3d071e93e8e02b05/src/websockets/legacy/protocol.py#L1097 actually pretty sure they should be optional to allow for disabling ping?

uvicorn/config.py
```diff
@@ -145,8 +145,8 @@ def __init__(
         http: Union[Type[asyncio.Protocol], HTTPProtocolType] = "auto",
         ws: Union[Type[asyncio.Protocol], WSProtocolType] = "auto",
         ws_max_size: int = 16 * 1024 * 1024,
-        ws_ping_interval: int = 20,
-        ws_ping_timeout: int = 20,
+        ws_ping_interval: float = 20,
```
graingert: Eventually passed here:

https://github.com/aaugustin/websockets/blob/361f38066622636eff8d9e0d3d071e93e8e02b05/src/websockets/legacy/server.py#L994-L995

graingert (changes requested):
I think these should be Optional [float]

graingert (approved):

Kludex:
@graingert True! Thanks for the review! I'm going to push the fix later on.

PR #1128, opened 2021-07-18
Allow app-dir parameter on the run() function

### LINKED ISSUE #1126: allow uvicorn.run to also accept app-dir kwarg
### Checklist

The issue is: I have a script using uvicorn.run, but I can only run it from a specific cwd, else it will fail with ImportError.

I would like the --app-dir functionality to be available in uvicorn.run() as well.
I.e.: to be able to run a script containing a call to uvicorn.run regardless of the cwd.

Doing it myself, i.e. calling sys.path.insert(0, abs_path_to_my_project_root_folder), but it feels to me that if this functionality is implemented in main() but not in run() (which from reading the code are the same thing) there must be something wrong with the uvicorn code.

## Additional context

https://github.com/encode/uvicorn/issues/549
https://github.com/encode/uvicorn/pull/619
https://github.com/encode/uvicorn/commit/77468df7697ec2f2962e77d7a02e0bc22b6f9c79

My guess is that this commit implemented the behaviour in the wrong place. Instead of 'consuming' the app_dir argument inside main(), it should have been passed down as-is to run() where it will be processed.

(I volunteer to do the job, if we agree that there is a problem)

Kludex:
Closes #1126

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -303,7 +303,6 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
 )
 @click.option(
     "--app-dir",
-    "app_dir",
     default=".",
     show_default=True,
     help="Look for APP in the specified directory, by adding this to the PYTHONPATH."
@@ -360,8 +359,6 @@ def main(
     app_dir: str,
     factory: bool,
 ) -> None:
-    sys.path.insert(0, app_dir)
-
     kwargs = {
         "host": host,
         "port": port,
@@ -403,11 +400,13 @@ def main(
         "headers": [header.split(":", 1) for header in headers],
         "use_colors": use_colors,
         "factory": factory,
+        "app_dir": app_dir,
     }
     run(app, **kwargs)
 
 
 def run(app: typing.Union[ASGIApplication, str], **kwargs: typing.Any) -> None:
+    sys.path.insert(0, kwargs.pop("app_dir"))
     config = Config(app, **kwargs)
     server = Server(config=config)
 
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -406,7 +406,7 @@ def main(
 
 
 def run(app: typing.Union[ASGIApplication, str], **kwargs: typing.Any) -> None:
-    sys.path.insert(0, kwargs.pop("app_dir"))
+    sys.path.insert(0, kwargs.pop("app_dir", "."))
     config = Config(app, **kwargs)
     server = Server(config=config)
 


uvicorn/main.py
```diff
@@ -303,7 +303,6 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
 )
 @click.option(
     "--app-dir",
-    "app_dir",
```
Kludex, line 306: This is not needed - the removal makes us follow the same pattern as above.

uvicorn/main.py
```diff
@@ -403,11 +400,13 @@ def main(
         "headers": [header.split(":", 1) for header in headers],
         "use_colors": use_colors,
         "factory": factory,
+        "app_dir": app_dir,
     }
     run(app, **kwargs)
 
 
 def run(app: typing.Union[ASGIApplication, str], **kwargs: typing.Any) -> None:
+    sys.path.insert(0, kwargs.pop("app_dir", "."))
```
HansBrende, line 409: Might want to insert at index 1 rather than 0 so as not to muck with the default sys.path[0]!
See: https://stackoverflow.com/a/10097543/2599133 for rationale.

From the [docs](https://docs.python.org/3/library/sys.html#sys.path):
> As initialized upon program startup, the first item of this list, path[0], is the directory containing the script that was used to invoke the Python interpreter. If the script directory is not available (e.g. if the interpreter is invoked interactively or if the script is read from standard input), path[0] is the empty string, which directs Python to search modules in the current directory first.

Kludex: Nop, we want to change the priority between "current directory" and `app_dir`.

HansBrende: @Kludex ok, but if app_dir is not specified at all (i.e. you default to "."), should you be changing the priority in that case? I don't think so...

Kludex: I'm not sure, but then we'd want something like this:

```python
    app_dir = kwargs.pop("app_dir")
    if app_dir is not None:
        sys.path.insert(0, app_dir)
```

This PR doesn't aim to change that logic tho, the aim here is to match the behavior from the CLI to the `run()` function, so I guess that would be a separate discussion.

HansBrende: @Kludex your suggested change in the above comment is (at the least) what is needed (and it is _not_ changing the logic). The `main` function still gives a default value of `.`, so cli logic is no different than it was before. No need to add _another_ default of `.` in the `run` function as well (because if you do, then the run function would have backwards-incompatible behavior, as well as giving devs no way to opt-out of modifying sys.path... and _worse_: no way to opt-out of modifying sys.path[0] [which is not a good idea in the first place](https://stackoverflow.com/a/10097543/2599133)).

Kludex: Yeah, you're right. Thanks. I'll apply the changes.

Kludex:
Replaced by #1271 :/

PR #1130, opened 2021-07-22
Format changelog file

Kludex:
This PR addresses the following:
- Fix typos - grammar mistakes.
- Set `-` as default for bullet point instead of `*`.


euri10:
> * Set `-` as default for bullet point instead of `*`.

what is the rationale for this ?

Kludex:
The only reason here is conformity.

But thinking twice... It's just polluting the history.

PR #1131, opened 2021-07-22
Remove non reachable return statement

Kludex:
I'm not 100% sure about this.

I was the one who added the `return None` a month ago, but I've noticed it's not covered, so I went back to see what would be a good test for it. But I don't think it should return `None`. I'm also uncertain if it's possible to not get a `host` on the `get_trusted_client_host`.

I'm going to check more closely about this later on, but thoughts are already welcome.


Kludex:
Replaced by #1349

Issue #494, opened 2019-11-22
Failing `socket_info.getpeername()` with uvloop against a UDS.


lovelydinosaur:
When running against a unix domain socket, we seem to occassionaly get an OSError when looking up the remote address. This appears to only occur with `uvloop`.

```
ERROR:    Exception in callback HttpToolsProtocol.connection_made(<_SelectorSoc...e, bufsize=0>>)
handle: <Handle HttpToolsProtocol.connection_made(<_SelectorSoc...e, bufsize=0>>)>
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/asyncio/events.py", line 127, in _run
    self._callback(*self._args)
  File "/Users/tomchristie/GitHub/encode/uvicorn/uvicorn/protocols/http/httptools_impl.py", line 129, in connection_made
    self.client = get_remote_addr(transport)
  File "/Users/tomchristie/GitHub/encode/uvicorn/uvicorn/protocols/utils.py", line 8, in get_remote_addr
    info = socket_info.getpeername()
OSError: [Errno 22] Invalid argument
```

PR #1132, opened 2021-07-22
Ignore coverage on get_remote_addr() condition

Kludex:
This exception was added on #495, and it's really needed, as we can see that [getpeername](https://docs.python.org/3/library/socket.html#socket.socket.getpeername) is not supported on some systems.

Related: 
- #494


euri10:
My take on this: if you find a way to reproduce on a the mentionned system (I never was able to where trying to cover this) why not write a test for it ?
reading the PR it seems the issue is encountered on a mac, maybe we could add it to the CI and see if this fails : uds + uvloop instead of just not covering this,

PR #1133, opened 2021-07-22
Add project links on setup.py for PyPI

Kludex:
Idea "copied" from [django-rest-framework](https://github.com/encode/django-rest-framework/blob/98e56e0327596db352b35fa3b3dc8355dc9bd030/setup.py#L109-L112) thanks to @graingert points. :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ extra_requirements = [
 setup(
     name="uvicorn",
     version=get_version("uvicorn"),
-    url="https://github.com/encode/uvicorn",
+    url="https://www.uvicorn.org/",
     license="BSD",
     description="The lightning-fast ASGI server.",
     long_description=get_long_description(),
@@ -95,4 +95,8 @@ setup(
     [console_scripts]
     uvicorn=uvicorn.main:main
     """,
+    project_urls={
+        "Funding": "https://github.com/sponsors/encode",
+        "Source": "https://github.com/encode/uvicorn",
+    },
 )


euri10 (approved):

setup.py
```diff
@@ -95,4 +95,8 @@ def get_packages(package):
     [console_scripts]
     uvicorn=uvicorn.main:main
     """,
+    project_urls={
+        "Funding": "https://github.com/sponsors/encode",
+        "Source": "https://github.com/encode/uvicorn",
```
graingert, line 100: changelog?

graingert (changes requested):
Changelog?

euri10:
> Changelog?

agreed, that goes with my comment here https://github.com/encode/uvicorn/pull/1134

euri10:
also my brain farts seeing that many emojis in PR titles, but this is unrelated

graingert (approved):

Kludex:
@euri10 I can stop using gitmoji if it bothers you. I don't mind. Just let me know 😗

graingert:
I'm not sure what ♿ means in terms of setup.py change. This is because it changes wheel metadata?

Kludex:
Well, maybe it was not the best choice... That one means "improve accessibility". But it could be the 📦.

Ref.: https://gitmoji.dev/

euri10:
> @euri10 I can stop using gitmoji if it bothers you. I don't mind. Just let me know

you do as you want @kludex I just find all those emoji eveywhere for everything quite unreadable, unnecessary.
good read: https://www.tandfonline.com/doi/abs/10.1080/014492999119129

Kludex:
I'm going to read it later. Thanks!

But as I said, I don't mind. So if it's improve your experience, I'll stop using it.

PR #1134, opened 2021-07-22
Add release notes to docs

Kludex:
Add release notes, analogous to what `starlette` [does](https://www.starlette.io/release-notes/).

Missing:
- [x] Add hyperlink to PRs reference.
- [x] Remove date and commit hash from sentences.
- [x] Add missing PR numbers.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,4 +95,5 @@ setup(
     [console_scripts]
     uvicorn=uvicorn.main:main
     """,
+    project_urls={"Changelog": "https://www.uvicorn.org/release-notes/"},
 )
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -2,168 +2,168 @@
 
 ### Added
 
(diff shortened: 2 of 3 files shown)


Kludex:
I've found this script: https://github.com/MagicStack/uvloop/blob/master/.github/release_log.py
Should we use it or do something similar?

setup.py
```diff
@@ -95,4 +95,5 @@ def get_packages(package):
     [console_scripts]
     uvicorn=uvicorn.main:main
     """,
+    project_urls={"Changelog": "https://www.uvicorn.org/release-notes/"},
```
euri10, line 98: what about we just add the github link of the Changelog.md to this ?

Kludex: I think we can start from there. But I think we should have it on the docs as well at some point, with the reference to the PRs, it's easy to travel around.

Kludex:
Closing this in favor of #1139.

PR #1147, opened 2021-08-12
Support Python 3.10-rc.1

Kludex:


setup.py
```diff
@@ -88,6 +88,7 @@ def get_packages(package):
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
```
graingert, line 91:
```suggestion
```
we can't add the trove until https://github.com/aaugustin/websockets/issues/1000 is on pypi

Kludex: But they have added as well https://github.com/aaugustin/websockets/blob/6edb363af07365867b4c372405b6ab3177a57830/setup.py#L53

Kludex: Ah, it's on master.

Kludex: But `websockets` is not a mandatory dependency, so it shouldn't be a blocker, right?

graingert (approved):

Kludex:
Thanks @graingert !

I guess you'll bump the websockets on a new PR?

PR #1150, opened 2021-08-13
use asyncio.run for gunicorn worker

graingert:
for some reason this didn't trigger the deprecation warnings in py3.10

### DIFF
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -72,7 +72,7 @@ class UvicornWorker(Worker):
         for s in self.SIGNALS:
             signal.signal(s, signal.SIG_DFL)
 
-    async def _serve(self):
+    async def _serve(self) -> None:
         self.config.app = self.wsgi
         server = Server(config=self.config)
         await server.serve(sockets=self.sockets)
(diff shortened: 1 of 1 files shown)


graingert:
I used github `.` aka codespaces but they didn't sign my commit

uvicorn/workers.py
```diff
@@ -72,17 +72,18 @@ def init_signals(self) -> None:
         for s in self.SIGNALS:
             signal.signal(s, signal.SIG_DFL)
 
-    def run(self) -> None:
+    async def _serve():
```
graingert:
```suggestion
    async def _serve(self):
```

uvicorn/workers.py
```diff
@@ -72,17 +72,18 @@ def init_signals(self) -> None:
         for s in self.SIGNALS:
             signal.signal(s, signal.SIG_DFL)
 
-    def run(self) -> None:
+    async def _serve(self):
```
graingert:
```suggestion
    async def _serve(self) -> None:
```

Kludex:
> for some reason this didn't trigger the deprecation warnings in py3.10

You mean in the pipeline? We don't have tests for the gunicorn worker, so it's normal.

Kludex (approved):

PR #1152, opened 2021-08-14
support an asynccontextmanager create_app factory

### LINKED ISSUE #1151: support @asynccontextmanager app_factory functions
### Checklist

Currently the lifespan task runs in a sibling task to any request tasks, but a number of use cases require wrapping a context manager around all the request tasks, eg:

support app factories like this:

```python
@contextlib.asynccontextmanager
async def app_factory() -> AsyncGenerator[App, None]:
    async with anyio.create_task_group() as tg:
        app = FastAPI(__name__)
        app.include_router(items.router)
        yield app
```

or :

```python
var: ContextVar[int] = ContextVar('var', default=42)

@contextlib.asynccontextmanager
async def app_factory():
    token = var.set(999)
    try:
        app = FastAPI(__name__)
        app.include_router(items.router)
        yield app
    finally:
        var.reset(token)
```

or including the `__aenter__/__aexit__` directly on the app instance:

```python
class ACMGRFastAPI(FastAPI):
    async def __aenter__(self) -> "ACMGRFastAPI":
        async with AsyncExitStack() as stack:
            self.database = stack.enter_async_context(database())
            self.cache = stack.enter_async_context(cache())
            self.bot = stack.enter_async_context(bot())
            self._stack = stack.pop_all()

        return self

    async def __aexit__(self, *exc_info) -> bool | None:
        return await self._stack.__aexit__(*exc_info)
```

run lifespan as a parent task to all the request tasks

## Additional context

see https://gitter.im/encode/community?at=610989bc8fc359158c4f959e

> [!IMPORTANT]

<a href="https://polar.sh/encode/uvicorn/issues/1151">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/1151/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

graingert:
fixes https://github.com/encode/uvicorn/issues/1151


tests/protocols/test_http.py
```diff
@@ -169,20 +170,57 @@ def add_done_callback(self, callback):
...
 
 
+def run_non_trapping_async_fn(*async_fn_args, **kwargs):
+    async_fn, *args = async_fn_args
+    with contextlib.closing(async_fn(*args, **kwargs).__await__()) as gen:
+        try:
+            gen.send(None)
+            raise AssertionError(f"{async_fn} did not stop")
+        except StopIteration as e:
+            return e.value
```
graingert:
```suggestion
    with contextlib.closing(async_fn(*args, **kwargs).__await__()) as gen:
        try:
            gen.send(None)
        except StopIteration as e:
            return e.value
        else:
            raise AssertionError(f"{async_fn} did not stop")
```

setup.py
```diff
@@ -42,12 +42,15 @@ def get_packages(package):
...
 env_marker_below_38 = "python_version < '3.8'"
+env_marker_below_310 = "python_version < '3.10'"
 
 minimal_requirements = [
     "asgiref>=3.4.0",
     "click>=7.0",
     "h11>=0.8",
     "typing-extensions;" + env_marker_below_38,
+    # contextlib2.nullcontext().__aenter__/__aexit__
+    "contextlib2 >= 21.6.0;" + env_marker_below_310,
```
adriangb, line 55: Is this really required for 3.9? Is the comment highlighting an issue with the implementation in 3.7, 3.8 & 3.9?

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -43,9 +43,6 @@ class HttpToolsProtocol(asyncio.Protocol):
     def __init__(
         self, config, server_state, on_connection_lost: Callable = None, _loop=None
     ):
-        if not config.loaded:
-            config.load()
-
```
adriangb, line 48: Could you explain why config loading changes are needed? I'd think all we need is to add a flag, or possibly not even that (since it is easy to inspect a function and check if it is an async generator context manager)

adriangb (approved):
I tested this locally (with a simple example), and it seems to work great!

There is a lot going on in the PR, and this is my first time looking at the internals of the codebase, so I don't fully understand all of the changes. Perhaps one of the Uvicorn maintainers can understand it better, but it might help overall to split this up (if possible).

setup.py
graingert: on python3.9 the  contextlib.nullcontext class is missing `__aenter__` and `__aexit__`

uvicorn/protocols/http/httptools_impl.py
graingert: This is because config's app can't be loaded without a context manager and so now it's always loaded by the time it reaches one of these classes

florimondmanca:
Hi,

Looking at the broadness of the changes here compared to the "additive" nature of #1151, I think we need to be _much more wary_ about the approach taken here, so as to minimize breaking change potential (in theory, that should be _zero_ for a feature addition).

Some example pointers:

* `config.load()` disappearing w/o deprecation is probably a no-no: many people might be depending on it, causing breaking changes when upgrading.
* This PR changes a bunch of tests that don't seem to _require_ changing. These changes are a strong signal that merging this PR would require users to go and change a bunch of _their own code_ to upgrade Uvicorn, whereas they might not care about the new feature. I'd suggest no existing tests should be changed here (they can be later, if we deem it relevant), and only _new tests_ be added for the new feature.
* Is the new `config.app_context()` supposed to be public API? If so, it should probably be documented. If documenting it doesn't make much sense, then it probably means that's an implementation detail that should be hidden from public API in some way.
* Should there be any change not-strictly-relevant to immediately resolving #1151 (eg refactoring, shuffling things around), ideally that should go in its own PR.

I've got more general comments and questions about the problem described in #1151, I'll post those there instead. :-)

PR #1154, opened 2021-08-15
Add 100% cover to uvicorn/main.py

humrochagf:
- Mostly test main as a cli client
- Test main functions directly when it's impossible to do is as a cli client (i.e. app non string case)
- Add no cover to the module entrypoint main call

Related to: #102


Kludex (approved):
@humrochagf Thank you very much for this! :)

PR #1155, opened 2021-08-15
Replace `pytest-asyncio` marker by `anyio`

graingert:
uvicorn currently depends on both pytest-asyncio and anyio in tests, so we can save a dep by swapping pytest-asyncio for anyio.


scripts/install
```diff
@@ -15,4 +15,5 @@ else
     PIP="pip"
 fi
 
+"$PIP" install -r "requirements-pypa.txt"
```
florimondmanca: Perhaps let's just do...?

```
"$PIP" install -U pip wheel setuptools
```

For the sake of simplicity, if that's what we need as a prerequisite.

florimondmanca:
It doesn't seem like we "_already_" depend on anyio, do we? We're adding it as a new dependency here.

If so, would the rationale better be to prepare the ground for future trio support?

graingert:
> It doesn't seem like we "already" depend on anyio, do we? We're adding it as a new dependency here.

no the anyio dep is already present

scripts/install
graingert: this is just for debugging locally - `python3.6 -m venv` creates an environment that can't install `cryptography`

Kludex:
@graingert where is it present? :thinking:

graingert:
> @graingert where is it present? 🤔

The test suite depends on httpx depends on anyio

requirements.txt
```diff
@@ -31,3 +30,5 @@ mkdocs-material
 
 # py3.10 workarounds
 https://github.com/aaugustin/websockets/archive/4e1dac362a3639b9cf0e5bcf382601e6e32cfede.tar.gz#egg=websockets; python_version >= '3.10'
+
+anyio @ git+https://github.com/graingert/anyio.git@b907c72ffae4e89c09fc9d7d81ff8c45fee43ccd
```
Kludex: Why adding this, then?

graingert: This is investigating https://github.com/agronholm/anyio/pull/358 it won't be needed if python 3.6 is dropped

Kludex: Python 3.6 was dropped. Removing this.

tests/protocols/test_http.py
```diff
@@ -189,6 +189,12 @@ def get_connected_protocol(app, protocol_cls, event_loop, **kwargs):
         asyncio._set_running_loop(None)
 
 
+@pytest.fixture
+def event_loop():
+    with contextlib.closing(asyncio.new_event_loop()) as loop:
+        yield loop
```
Kludex: @euri10 this is introduced on this PR, jfyk

tests/conftest.py
```diff
@@ -138,6 +138,11 @@ def reload_directory_structure(tmp_path_factory: pytest.TempPathFactory):
     yield root
 
 
+@pytest.fixture
+def anyio_backend() -> str:
+    return "asyncio"
```
Kludex, line 143: @euri10 this as well.

Kludex (dismissed):
@euri10 Please check if you agree with this PR as well.

I've removed all the unnecessary changes from this PR, and focused on what it proposes, jfyk @graingert

tests/protocols/test_http.py
euri10: the corresponding pytest-asyncio fixture `event_loop` is here

https://github.com/pytest-dev/pytest-asyncio/blob/4c7da65d6fcf9d725eccba28ad1ed2083524ee16/tests/async_fixtures/test_async_fixtures_with_finalizer.py#L19-L25

With this change, we don't close the loop after the yield, I'm surprised then pytest does not raise Resources not closed warnings, would be probably better to do it anyways so we dont get caught further down the road with such warning that are insanely hard to debug

and the scope is different, I'm not sure we need a loop for each test in that module, do we ?

Kludex: > With this change, we don't close the loop after the yield ...

We do, the `contextlib.closing` calls `loop.close()`. Is that what you mean? 🤔

agronholm: What is this fixture even used for here? It feels like an anti-pattern.

euri10: > We do, the `contextlib.closing` calls `loop.close()`. Is that what you mean? thinking

yes, well if it passes tests then the ressources might be closed properly !

> What is this fixture even used for here? It feels like an anti-pattern.

iirc we use this pytest_asyncio fixture only in those "protocol" tests to create a bunch of mocks, mocked loop, mocked transport and then we create a protocol we interact with and look at its buffer,
is it an anti-pattern, idk :)

tests/conftest.py
euri10, line 143: is it necessary to have this, I dont see it used anywhere

agronholm: It's used by anyio's pytest plugin to determine which backends are used. If uvicorn doesn't support trio, then this is necessary.

tests/protocols/test_http.py
graingert: Would be better to use a real event loop if that's what's used here

agronholm: I think this fixture should be removed. If it breaks protocol tests, we can surely update those tests.

euri10: I'd be interested in seeing how, cant wrap my head around it, we need a loop to built the protocols, how can we get it if not mocked ?

graingert: Use a real running event loop?

euri10: I might be misunderstanding something, we use pytest-asyncio event_loop fixture, which is to me a real event loop
```
@pytest.fixture
def event_loop(request):
    """Create an instance of the default event loop for each test case."""
    loop = asyncio.get_event_loop_policy().new_event_loop()
    yield loop
    loop.close()
```

we could avoid using the `MockLoop` doing something like this, is that what you mean ?

```diff
@@ -175,18 +175,15 @@ class MockTask:
 
 @contextlib.contextmanager
 def get_connected_protocol(app, protocol_cls, event_loop, **kwargs):
-    loop = MockLoop(event_loop)
-    asyncio._set_running_loop(loop)
     transport = MockTransport()
     config = Config(app=app, **kwargs)
     server_state = ServerState()
-    protocol = protocol_cls(config=config, server_state=server_state, _loop=loop)
+    protocol = protocol_cls(config=config, server_state=server_state, _loop=event_loop)
     protocol.connection_made(transport)
     try:
         yield protocol
     finally:
         protocol.loop.close()
-        asyncio._set_running_loop(None)
 
 
 @pytest.mark.parametrize("protocol_cls", HTTP_PROTOCOLS)
@@ -195,7 +192,8 @@ def test_get_request(protocol_cls, event_loop):
 
     with get_connected_protocol(app, protocol_cls, event_loop) as protocol:
         protocol.data_received(SIMPLE_GET_REQUEST)
-        protocol.loop.run_one()
+        tasks = asyncio.all_tasks(loop=event_loop)
+        protocol.loop.run_until_complete(*tasks)
         assert b"HTTP/1.1 200 OK" in protocol.transport.buffer
         assert b"Hello, world" in protocol.transport.buffer
```

agronholm: The loop is closed twice here (once in the fixture, once in `get_connected_protocol()`. Looks like the latter could be made into a fixture instead.

Does the protocol create tasks?

graingert: yes https://github.com/encode/uvicorn/blob/4be108592669ce8b289c2512689544dfea6ea7d9/uvicorn/protocols/http/h11_impl.py#L234-L237

graingert: @euri10 I just pushed a fix to make MockLoop wrap the currently running loop `get_connected_protocol` can be made into a plain function but it makes the PR more difficult to review

euri10 (approved):
I like those changes

it's just `test_http` whose diff is larger, but in any case I find it way cleaner than before.

I think we can give it a shot this way, all good on my side, if you prefer to separate the rework on the MockLoop from the anyio switch I'd be ok too but this way is also ok to me :)

requirements.txt
```diff
@@ -22,8 +22,6 @@ cryptography==3.4.8
 coverage==6.4
 coverage-conditional-plugin==0.5.0
 httpx==1.0.0b0
-pytest-asyncio==0.15.1
```
florimondmanca, line 25: Should we add an explicit test dependency on `anyio`, since we now explicitly import it?

```
anyio==3.6.1
```

florimondmanca (approved):
LGTM :+1:

Looks like I had already viewed this back when it was initially opened. Changes outside `test_http.py` are straightforward. I reviewed that file as well and the changes are in fact pretty systematic too, so giving my LGTM, though I haven't checked if there's nitty-gritty discussion still ongoing. Nice one!

Kludex (approved):

tests/protocols/test_http.py
graingert: as promised, here's the followup: https://github.com/encode/uvicorn/pull/1520

PR #1156, opened 2021-08-16
Upgrade test coverage from 95 to 96 percent

Kludex:


Kludex:
`master` is already with 97%. This PR is useless.

PR #1158, opened 2021-08-16
[Alternative] Apply coverage only on Python 3.8 using Linux

Kludex:
Alternative to #1157

IMO This solution is worse than the original, as developers on their environment would always have their test set failing.


Kludex:
I don't think it makes sense to have this. The reason is in the description above.

PR #1161, opened 2021-08-18
Create event loop on non-main threads 

Kludex:
When we create a new thread, we need to create an event loop.

This behavior was removed on https://github.com/encode/uvicorn/pull/1070/files#r691270112

Edit: I've also added the same behavior for `uvloop`. 
There's a huge discussion about all of this here: https://github.com/encode/uvicorn/issues/742#issuecomment-798766443

### DIFF
--- a/uvicorn/loops/asyncio.py
+++ b/uvicorn/loops/asyncio.py
@@ -5,3 +5,6 @@ import sys
 def asyncio_setup() -> None:  # pragma: no cover
     if sys.version_info >= (3, 8) and sys.platform == "win32":
         asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+    else:
+        loop = asyncio.new_event_loop()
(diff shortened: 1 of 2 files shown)


Kludex:
I guess now both this and #1162 solve the issue with `uvloop` as well.

Kludex:
Closed in favor of #1162 

I think going to #1162 as we'll have the same behavior on all the versions we support. At the EOL of 3.6 we'll remove the backward compatibility that is created from that PR and only use the `asyncio.run`.

Kludex:
I'm reopening this PR as it's the quickest fix to the issue introduced on 0.15.0, and #1162 introduced a new issue on this line https://github.com/encode/uvicorn/blob/f407d5147cb727e394a5bf59a88641bb3420504e/uvicorn/_compat.py#L57 (and the fix for it is on #1166 introduces further complexity on #1166). 

Also, I've opened #1171 to revert #1162.

PR #1162, opened 2021-08-18
backport asyncio.run for py3.6

graingert:


uvicorn/_compat.py
```diff
@@ -0,0 +1,70 @@
+import sys
```
Kludex, line 6:
```suggestion
"""
This module backports `asyncio.run` to Python 3.6. 
You can check the original implementation on:
https://github.com/python/cpython/blob/3.7/Lib/asyncio/runners.py
"""
import sys
```

uvicorn/_compat.py
```diff
@@ -0,0 +1,70 @@
...
+
+        loop = events.new_event_loop()
+        try:
+            events.set_event_loop(loop)
+            loop.set_debug(debug)
+            return loop.run_until_complete(main)
+        finally:
+            try:
+                _cancel_all_tasks(loop)
+                loop.run_until_complete(loop.shutdown_asyncgens())
```
Kludex, line 51: Why is this line missing? https://github.com/python/cpython/blob/ebe7e6d86cf8f54d449d49866698d7f4c700cc7c/Lib/asyncio/runners.py#L49

Kludex (commented):
For reference:

This PR enables both event loops (`asyncio` and `uvloop`) to run on new threads.
It was already possible to do that for `asyncio` since some releases ago, but on the last release we removed this behavior. For `uvloop` it was never possible (out of the box I mean).

In contrast to this PR, we have #1161 that only reverts the modifications on 0.15.0 that caused the removed behavior for `asyncio`..

uvicorn/_compat.py
graingert: I backported the asyncio.run code from py3.7 rather than main:

https://github.com/python/cpython/blob/3.7/Lib/asyncio/runners.py#L47
https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.shutdown_default_executor (new in 3.9)

uvicorn/_compat.py
```diff
@@ -0,0 +1,75 @@
+"""
+This module backports `asyncio.run` to Python 3.6. 
+You can check the original implementation on:
+https://github.com/python/cpython/blob/3.7/Lib/asyncio/runners.py
```
Kludex:
```suggestion
This module backports `asyncio.run` to Python 3.6.
You can check the original implementation on:
https://github.com/python/cpython/blob/3.7/Lib/asyncio/runners.py
```

Kludex:
Ok, last thought: can we create a test that prevent us to remove this behavior again?

graingert:
> Ok, last thought: can we create a test that prevent us to remove this behavior again?

the test would involve calling uvicorn run from a thread, which is currently problematic, so I think adding a test here would be best: 
https://github.com/encode/uvicorn/pull/1163

uvicorn/server.py
```diff
@@ -64,9 +65,7 @@ def __init__(self, config: Config) -> None:
 
     def run(self, sockets: Optional[List[socket.socket]] = None) -> None:
```
Kludex, line 66: The return is not `None` anymore, right?

Kludex:
@graingert Would you mind adding `_compat.py` to the mypy files on `setup.cfg`? After that, I'll happily merge it.

uvicorn/server.py
graingert: yes it's `-> None`, `asyncio.run(serve(...))` returns synchronously the `StopIteration.value` of the coro from `serve(...)`

uvicorn/_compat.py
```diff
@@ -0,0 +1,75 @@
...
+import sys
+
+__all__ = ("run",)
+
+if sys.version_info >= (3, 7):
+    from asyncio import run
+else:
+    from asyncio import coroutines, events, tasks
+
+    def run(main, *, debug=False):
```
graingert:
```suggestion
    from typing import TypeVar, Coroutine, Any

    _T = TypeVar("_T")

    def run(main: Coroutine[Any, Any, _T], *, debug: bool=False) -> _T:
```

Kludex (approved):
Thanks!

PR #1164, opened 2021-08-19
Add FastAPI sponsorship

lovelydinosaur:
Screenshot 2021-08-19 at 15 12 00


Kludex:
Should we consider something fancier in the future? Like a page for sponsors?

lovelydinosaur:
Yes absolutely, something like that would be fantastic.

PR #1165, opened 2021-08-19
Revert "When receiving a SIGTERM supervisors should terminate their processes before joining them"

### LINKED ISSUE #1160: Shutdown process is broken in 0.15
### Checklist

### Describe the bug

My FastAPI ASGI server cannot shutdown properly with uvicorn==0.15 while it can with 0.14

### To reproduce

Setup minimal FastAPI app and add some functions with logs(prints) to shutdown event

### Expected behavior

You see all logs(prints) from functions on shutdown

### Actual behavior

Get `ASGI 'lifespan' protocol appears unsupported.` without --lifespan on
Get error trace with --lifespan on

### Debugging material

uvicorn scheduler.main:app --host=0.0.0.0 --port ${WEB_PORT:-8000} --reload --lifespan on
INFO:     Will watch for changes in these directories: ['/home/dmytro/storage/chimplie/projects/raok-main/raok-scheduler']
INFO:     Uvicorn running on http://0.0.0.0:8004 (Press CTRL+C to quit)
INFO:     Started reloader process [177653] using statreload
INFO:     Started server process [177655]
INFO:     Tortoise-ORM started, {'default': <tortoise.backends.asyncpg.client.AsyncpgDBClient object at 0x7f63d4a10e50>}, {'models': {'Task': <class 'scheduler.models.task.Task'>, 'Aerich': <class 'aerich.models.Aerich'>}}
^CINFO:     Shutting down
INFO:     Finished server process [177655]
  File "/home/dmytro/.local/share/virtualenvs/raok-scheduler-hpGGYNLi/lib/python3.8/site-packages/uvicorn/lifespan/on.py", line 84, in main
  File "/home/dmytro/.local/share/virtualenvs/raok-scheduler-hpGGYNLi/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
  File "/home/dmytro/.local/share/virtualenvs/raok-scheduler-hpGGYNLi/lib/python3.8/site-packages/fastapi/applications.py", line 199, in __call__
    await super().__call__(scope, receive, send)
  File "/home/dmytro/.local/share/virtualenvs/raok-scheduler-hpGGYNLi/lib/python3.8/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/dmytro/.local/share/virtualenvs/raok-scheduler-hpGGYNLi/lib/python3.8/site-packages/starlette/middleware/errors.py", line 146, in __call__
  File "/home/dmytro/.local/share/virtualenvs/raok-scheduler-hpGGYNLi/lib/python3.8/site-packages/starlette/middleware/cors.py", line 70, in __call__
  File "/home/dmytro/.local/share/virtualenvs/raok-scheduler-hpGGYNLi/lib/python3.8/site-packages/starlette/exceptions.py", line 58, in __call__
  File "/home/dmytro/.local/share/virtualenvs/raok-scheduler-hpGGYNLi/lib/python3.8/site-packages/starlette/routing.py", line 569, in __call__
  File "/home/dmytro/.local/share/virtualenvs/raok-scheduler-hpGGYNLi/lib/python3.8/site-packages/starlette/routing.py", line 544, in lifespan
    await receive()
  File "/home/dmytro/.local/share/virtualenvs/raok-scheduler-hpGGYNLi/lib/python3.8/site-packages/uvicorn/lifespan/on.py", line 135, in receive
  File "/usr/lib64/python3.8/asyncio/queues.py", line 163, in get
    await getter
INFO:     Stopping reloader process [177653]

image

### Environment

- Fedora 34 / Python 3.8 / Uvicorn version: 0.15 - bug, 0.14 - ok
- command to run: `uvicorn main:app --host=0.0.0.0 --port 8000 --reload`

Kludex:
Reverts encode/uvicorn#1069

Closes #1160

I've taken my time to understand the situation. I'm going to explain it and then suggest a solution to the previous issue.

Right now, the issue on #1160 (`CancelledError`) is caused because the `CTRL + C` sends a `SIGINT (2)` to both parent and child processes, and given that we merged #1069, we have that the parent is also sending a `SIGTERM (15)` to the child. In other words, the child is receiving two signals, and the way `uvicorn` deals with multiple signals (two) is to forcefully exit the process.

To be more precise, when we press `CTRL + C` we send a `kill` signal to the process group, not to a single process.

Ok. Now, let's go back to the original issue: if we send `SIGINT` to the parent process, it doesn't terminate the children. And that's expected, because `kill -2 <uvicorn_pid>` will only send a signal to the `<uvicorn_pid>`.

The solution here would be to use `kill -2 -<uvicorn_pid>`, which sends a signal to the process group instead of only the parent. That being said, this solves the issue that #1069 (`process.terminate()`) solved, but we also avoid #1160 (`CancelledError`). 

Reference: https://stackoverflow.com/a/392155

### DIFF
--- a/uvicorn/supervisors/basereload.py
+++ b/uvicorn/supervisors/basereload.py
@@ -74,7 +74,6 @@ class BaseReload:
         self.process.start()
 
     def shutdown(self) -> None:
-        self.process.terminate()
         self.process.join()
         message = "Stopping reloader process [{}]".format(str(self.pid))
         color_message = "Stopping reloader process [{}]".format(
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -64,7 +64,6 @@ class Multiprocess:
 
     def shutdown(self) -> None:
         for process in self.processes:
-            process.terminate()
             process.join()
 
         message = "Stopping parent process [{}]".format(str(self.pid))


Kludex:
I don't think we can avoid reverting that PR, but for reference, we can also do something like this:

https://github.com/benoitc/gunicorn/blob/1299ea9e967a61ae2edebe191082fd169b864c64/gunicorn/workers/gthread.py#L176-L181

In other words, check if the parent is alive, and then terminate the worker if it's not.

Kludex:
hypercorn just ignores the SIGINT before creating the processes: https://gitlab.com/pgjones/hypercorn/-/blob/master/src/hypercorn/run.py#L49-52. Then it adds a handler after: https://gitlab.com/pgjones/hypercorn/-/blob/master/src/hypercorn/run.py#L70-72.

Kludex:
@euri10 Can you check this, please?

Asday:
> The solution here would be to use `kill -2 -<uvicorn_pid>`, which sends a signal to the process group instead of only the parent. 

I was originally interested in this issue (i.e., the previous one) because it was annoying to wait for my docker container to get killed on a timeout.  Keeping in mind that I am a stupid stupid idiot who can barely read my own name, does this mean that my docker environment is going to become rubbish again, or is there some way to add a hyphen to me running `docker-compose stop` or `^C`ing `docker-compose up`?

HansBrende:
@Asday if you want the container to be killed immediately without waiting for a graceful shutdown I think you could do: 
```
docker-compose kill
```

Asday:
That's not at all what I want, I want the SIGTERM to gracefully shut down uvicorn, including its workers.

HansBrende:
Hm... if there is really no way to signal to docker to terminate the process group instead of only the parent process (which I don't know if that's the case or not because I'm not an expert on the ins and outs of docker (but if it is, that would seem a bit ridiculous in light of [standard CTRL+C behavior](https://unix.stackexchange.com/a/149756/502695))), one other option would be, instead of reverting the old PR, to never force shutdown on SIGTERMs... only on the second SIGINT, specifically. Which is basically equivalent to a monkeypatch I did to get around this:

```python
def _handle_exit(self, sig, frame):
    if self.should_exit and sig == signal.Signals.SIGINT:
        self.force_exit = True
    else:
        self.should_exit = True
 
uvicorn.server.Server.handle_exit = _handle_exit
```

Oooh, another option I just thought of: rather than starting the server process as a foreground process, the reloader process could start the server as a _background_ process. In that case, CTRL+C would _not_ send a SIGINT to the server process since it specifically only sends SIGINTs to "the **foreground** process group" (See [spec](https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap11.html#tag_11_01_09)), which would align behavior between CTRL + C and docker containers (I think). But I really have only spent about a half-hour researching this so I could be totally off-base. Any thoughts @Kludex @euri10 ?

PR #1168, opened 2021-08-21
docs: httptools has wheels for Win now

imba-tjd:
See https://github.com/MagicStack/httptools/issues/42


Kludex:
You're right! Thanks! We even remove the Python marker on `httptools` some time ago on #1006.

Kludex (approved):

PR #1170, opened 2021-08-22
Add macOS to the pipeline

Kludex:
Following @euri10 's recommendation on https://github.com/encode/uvicorn/pull/1132#issuecomment-889670278, I also think we should add macOS to the pipeline. 

The path we use to bind to the socket is 127 bytes length. [AF_UNIX](https://man7.org/linux/man-pages/man7/unix.7.html) has only 108 bytes allocated for it, so we keep receiving `OSError: AF_UNIX path too long`. 

~The way I solved it is using the `--basetemp` parameter on `pytest`. You can check more about it [on the docs](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-default-base-temporary-directory) and on [this issue](https://github.com/pytest-dev/pytest/issues/5802).~ Doesn't work on Windows.

I've used [tmp_path_factory](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-factory-fixture) instead.


Kludex:
Maybe https://github.com/encode/uvicorn/issues/494 is because of the path name size? :thinking:

tests/test_config.py
```diff
@@ -513,10 +514,12 @@ def test_ws_max_size() -> None:
...
 @pytest.mark.skipif(sys.platform == "win32", reason="require unix-like system")
-def test_bind_unix_socket_works_with_reload_or_workers(tmp_path, reload, workers):
-    uds_file = tmp_path / "uvicorn.sock"
+def test_bind_unix_socket_works_with_reload_or_workers(
+    tmp_path_factory, reload, workers, loop
+):
+    uds_file = tmp_path_factory.mktemp("tmp") / "uvicorn.sock"
     config = Config(
-        app=asgi_app, uds=uds_file.as_posix(), reload=reload, workers=workers
+        app=asgi_app, uds=uds_file.as_posix(), reload=reload, workers=workers, loop=loop
```
Kludex, line 522: I'm not able to reproduce #494.

euri10:
would be cool to add this to the pipeline but I dont get where you're at with this @Kludex ?

PR #1173, opened 2021-08-23
Fix h11 event conditional

Kludex:
Hard to find.

We should merge this _as is_, but I'm going to add some tests in the following days.


Kludex:
Closed in favor of #1273

PR #1177, opened 2021-09-06
Add check if childs are still alive and shutdown instead of hanging

### LINKED ISSUE #1115: [BUG] Exit at startup failure on multiple workers
### Checklist

### Describe the bug

When the startup fails, either on reload mode or multiple workers, the main process is not terminated.

### To reproduce

Application:
```python
# test.py
from fastapi import FastAPI

app = FastAPI()

@app.on_event("startup")
def startup():
    raise Exception("Hi")
```

```bash
uvicorn test:app --workers 2
# or
uvicorn test:app --reload
```

### Expected behavior

All the processes (parent and children) should be terminated.

### Actual behavior

```bash
❯ uvicorn test:app --reload
INFO:     Started reloader process [40026] using watchgod
INFO:     Started server process [40028]
  File "/home/marcelo/anaconda3/envs/uvicorn/lib/python3.8/site-packages/starlette/routing.py", line 540, in lifespan
  File "/home/marcelo/anaconda3/envs/uvicorn/lib/python3.8/site-packages/starlette/routing.py", line 481, in default_lifespan
    await self.startup()
  File "/home/marcelo/anaconda3/envs/uvicorn/lib/python3.8/site-packages/starlette/routing.py", line 518, in startup
    handler()
  File "/home/marcelo/Development/./test.py", line 7, in startup
    raise Exception("Hi")
Exception: Hi

```
After this log, it hangs forever. On the `reload`, it doesn't prevent the failure from happening multiple times if you modify a watched file. See:

```bash
❯ uvicorn test:app --reload            
INFO:     Started reloader process [40674] using watchgod
INFO:     Started server process [40676]
  File "/home/marcelo/anaconda3/envs/uvicorn/lib/python3.8/site-packages/starlette/routing.py", line 540, in lifespan
  File "/home/marcelo/anaconda3/envs/uvicorn/lib/python3.8/site-packages/starlette/routing.py", line 481, in default_lifespan
    await self.startup()
  File "/home/marcelo/anaconda3/envs/uvicorn/lib/python3.8/site-packages/starlette/routing.py", line 518, in startup
    handler()
  File "/home/marcelo/Development/./test.py", line 7, in startup
    raise Exception("Hi")
Exception: Hi

WARNING:  WatchGodReload detected file change in '['/home/marcelo/Development/test.py']'. Reloading...
INFO:     Started server process [40734]
  File "/home/marcelo/anaconda3/envs/uvicorn/lib/python3.8/site-packages/starlette/routing.py", line 540, in lifespan
  File "/home/marcelo/anaconda3/envs/uvicorn/lib/python3.8/site-packages/starlette/routing.py", line 481, in default_lifespan
    await self.startup()
  File "/home/marcelo/anaconda3/envs/uvicorn/lib/python3.8/site-packages/starlette/routing.py", line 518, in startup
    handler()
  File "/home/marcelo/Development/./test.py", line 7, in startup
    raise Exception("Hi")
Exception: Hi

```

In the case of multiple workers, it just hangs forever.

### Environment

- OS / Python / Uvicorn version: Running uvicorn 0.14.0 with CPython 3.8.10 on Linux

humrochagf:
Fix: #1115 

Since server closes with `sys.exit(1)` on child process without communicating with the parent supervisor we need to check if the process is still alive.

I used the same interval as the one already setup for the reloader, but it could be another one as well.

Also added the tests for the direct call from the command line, and also ensured sockets are closed at the end of the shutdown. Since child dies before being able to properly close their sockets.

### DIFF
--- a/uvicorn/supervisors/basereload.py
+++ b/uvicorn/supervisors/basereload.py
@@ -45,6 +45,9 @@ class BaseReload:
             if self.should_restart():
                 self.restart()
 
+            if self.process.exitcode is not None:
+                break
+
         self.shutdown()
 
     def startup(self) -> None:
@@ -76,6 +79,10 @@ class BaseReload:
     def shutdown(self) -> None:
         self.process.terminate()
         self.process.join()
+
+        for _socket in self.sockets:
+            _socket.close()
+
         message = "Stopping reloader process [{}]".format(str(self.pid))
         color_message = "Stopping reloader process [{}]".format(
             click.style(str(self.pid), fg="cyan", bold=True)
--- a/uvicorn/supervisors/basereload.py
+++ b/uvicorn/supervisors/basereload.py
@@ -80,8 +80,8 @@ class BaseReload:
         self.process.terminate()
         self.process.join()
 
-        for _socket in self.sockets:
-            _socket.close()
+        for sock in self.sockets:
+            sock.close()
 
         message = "Stopping reloader process [{}]".format(str(self.pid))
         color_message = "Stopping reloader process [{}]".format(
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -42,7 +42,11 @@ class Multiprocess:
 
     def run(self) -> None:
         self.startup()
-        self.should_exit.wait()
+
+        while not self.should_exit.wait(self.config.reload_delay):
+            if not any(p.exitcode is None for p in self.processes):
+                break
+
         self.shutdown()
 
     def startup(self) -> None:
@@ -66,6 +70,9 @@ class Multiprocess:
         for process in self.processes:
             process.terminate()
             process.join()
+
+        for _socket in self.sockets:
+            _socket.close()
 
         message = "Stopping parent process [{}]".format(str(self.pid))
         color_message = "Stopping parent process [{}]".format(
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -42,11 +42,7 @@ class Multiprocess:
 
     def run(self) -> None:
         self.startup()
-
-        while not self.should_exit.wait(self.config.reload_delay):
-            if not any(p.exitcode is None for p in self.processes):
-                break
-
+        self.should_exit.wait()
         self.shutdown()
 
     def startup(self) -> None:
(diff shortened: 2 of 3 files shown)


uvicorn/supervisors/basereload.py
```diff
@@ -76,6 +79,10 @@ def restart(self) -> None:
     def shutdown(self) -> None:
         self.process.terminate()
         self.process.join()
+
+        for _socket in self.sockets:
+            _socket.close()
```
Kludex: Let's use the same notation as the `server.py`.

```suggestion
        for sock in self.sockets:
            sock.close()
```

uvicorn/supervisors/multiprocess.py
```diff
@@ -42,7 +42,11 @@ def signal_handler(self, sig: signal.Signals, frame: FrameType) -> None:
 
     def run(self) -> None:
         self.startup()
-        self.should_exit.wait()
+
+        while not self.should_exit.wait(self.config.reload_delay):
+            if not any(p.exitcode is None for p in self.processes):
+                break
```
Kludex: > I used the same interval as the one already setup for the reloader, but it could be another one as well.

Yep, I don't think we should use `self.config.reload_delay` here.

Kludex (changes requested):
@humrochagf Thanks for the PR! :tada: 

First, sorry for taking so long to review this.

Now, I do think the "reload" part makes sense, and we should get that merged. I've added a single comment on it (that I can change it myself).

As for the "multiprocess" part, looks fine as well. Just a remark about the `reload_delay`, as I don't think it should be used here. But... Although it looks fine, I don't want to merge this part, because I'm working on a process manager for uvicorn (which will be located on the Multiprocess class).

For the reason stated in the previous paragraph, would you mind removing the logic around the Multiprocess, so we can merge the "reload" part?

humrochagf:
> @humrochagf Thanks for the PR! tada
> 
> First, sorry for taking so long to review this.
> 
> Now, I do think the "reload" part makes sense, and we should get that merged. I've added a single comment on it (that I can change it myself).
> 
> As for the "multiprocess" part, looks fine as well. Just a remark about the `reload_delay`, as I don't think it should be used here. But... Although it looks fine, I don't want to merge this part, because I'm working on a process manager for uvicorn (which will be located on the Multiprocess class).
> 
> For the reason stated in the previous paragraph, would you mind removing the logic around the Multiprocess, so we can merge the "reload" part?

Hey, @Kludex no worries for the delay, and I don't mind removing the multiprocessing part, I agree with you, having a process manager would be much better :smiley: 

I'll do the changes.

Kludex:
Thanks! ✨

humrochagf:
Done :rocket:

Kludex (approved):

PR #1180, opened 2021-09-09
use officially released websockets v10 (and py3.10rc2)

graingert:


Kludex (approved):

Kludex:
Next release will be minor.

PR #1183, opened 2021-09-14
Ignore `asyncio.sleep` warnings on tests

Kludex:
On Python 3.9.7, `asyncio.sleep` and `asyncio.gather` were raising warnings. Next patch will solve the issue.

Check https://bugs.python.org/issue45097 for more information.

## Alternative solution

Pin Python 3.9.6 on the pipeline.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,5 +1,6 @@
 import asyncio
 import os
+import warnings
 from contextlib import contextmanager
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


graingert (approved):
Looks good I think I'd prefer it in the pytest config instead though

lovelydinosaur:
Ugh. But yes, let's just roll with this - this way around at least it's tightly scoped to *only* be around the case where it's needed.

PR #1185, opened 2021-09-16
 ignore loop deprecation warnings originating inside asyncio due to bpo-45097 

graingert:
This reverts commit 1f87ad4701f831127adbf8ea25132e4b442e10a0.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -62,7 +62,7 @@ xfail_strict=True
 filterwarnings=
     # Turn warnings that aren't filtered into exceptions
     error
-    # https://bugs.python.org/issue45097
+    # Workaround for Python 3.9.7 (see https://bugs.python.org/issue45097)
     ignore:The loop argument is deprecated since Python 3\.8, and scheduled for removal in Python 3\.10\.:DeprecationWarning:asyncio
 
 [coverage:run]
--- a/setup.cfg
+++ b/setup.cfg
@@ -62,6 +62,8 @@ xfail_strict=True
 filterwarnings=
     # Turn warnings that aren't filtered into exceptions
     error
+    # https://bugs.python.org/issue45097
+    ignore:The loop argument is deprecated since Python 3\.8, and scheduled for removal in Python 3\.10\.:DeprecationWarning:asyncio
 
 [coverage:run]
 omit = venv/*
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,6 +1,5 @@
 import asyncio
 import os
-import warnings
 from contextlib import contextmanager
 from pathlib import Path
 
@@ -16,11 +15,7 @@ from uvicorn import Config, Server
 async def run_server(config: Config, sockets=None):
     server = Server(config=config)
     cancel_handle = asyncio.ensure_future(server.serve(sockets=sockets))
-
-    # Workaround for Python 3.9.7 (see https://bugs.python.org/issue45097)
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", category=DeprecationWarning)
-        await asyncio.sleep(0.1)
+    await asyncio.sleep(0.1)
     try:
         yield server
     finally:


tests/utils.py
```diff
@@ -16,11 +15,7 @@
 async def run_server(config: Config, sockets=None):
     server = Server(config=config)
     cancel_handle = asyncio.ensure_future(server.serve(sockets=sockets))
-
-    # Workaround for Python 3.9.7 (see https://bugs.python.org/issue45097)
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", category=DeprecationWarning)
-        await asyncio.sleep(0.1)
```
graingert, line 23: this context manager applies the warning filter to all other tasks (and threads) running concurrently

graingert: see https://bugs.python.org/issue37604

setup.cfg
```diff
@@ -62,6 +62,8 @@ xfail_strict=True
 filterwarnings=
     # Turn warnings that aren't filtered into exceptions
     error
+    # https://bugs.python.org/issue45097
```
Kludex: So we can remember to remove in the future.

```suggestion
    # Workaround for Python 3.9.7 (see https://bugs.python.org/issue45097)
```

setup.cfg
```diff
@@ -62,6 +62,8 @@ xfail_strict=True
 filterwarnings=
     # Turn warnings that aren't filtered into exceptions
     error
+    # Workaround for Python 3.9.7 (see https://bugs.python.org/issue45097)
+    ignore:The loop argument is deprecated since Python 3\.8, and scheduled for removal in Python 3\.10\.:DeprecationWarning:asyncio
```
graingert, line 66: this extra `:asyncio` means that only warnings that are caused by asyncio itself are ignored:

I tested this by adding my own test case like this:

```python
@pytest.mark.asyncio
async def test_foo():
    async def bar():
        pass

    loop = asyncio.get_running_loop()
    await asyncio.gather(bar(), bar(), loop=loop)
```

```
=========================================================================================== test session starts ===========================================================================================
platform linux -- Python 3.9.7, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
rootdir: /home/graingert/projects/uvicorn, configfile: setup.cfg
plugins: mock-3.6.1, anyio-3.3.1, asyncio-0.14.0
collected 1 item                                                                                                                                                                                          

tests/test_auto_detection.py F                                                                                                                                                                      [100%]

================================================================================================ FAILURES =================================================================================================
________________________________________________________________________________________________ test_foo _________________________________________________________________________________________________

    @pytest.mark.asyncio
    async def test_foo():
        async def bar():
            pass
    
        loop = asyncio.get_running_loop()
>       await asyncio.gather(bar(), bar(), loop=loop)

tests/test_auto_detection.py:68: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

loop = <_UnixSelectorEventLoop running=False closed=False debug=False>, return_exceptions = False
coros_or_futures = (<coroutine object test_foo.<locals>.bar at 0x7f4f28e82a40>, <coroutine object test_foo.<locals>.bar at 0x7f4f28e82ac0>)

    def gather(*coros_or_futures, loop=None, return_exceptions=False):
        """Return a future aggregating results from the given coroutines/futures.
    
        Coroutines will be wrapped in a future and scheduled in the event
        loop. They will not necessarily be scheduled in the same order as
        passed in.
    
        All futures must share the same event loop.  If all the tasks are
        done successfully, the returned future's result is the list of
        results (in the order of the original sequence, not necessarily
        the order of results arrival).  If *return_exceptions* is True,
        exceptions in the tasks are treated the same as successful
        results, and gathered in the result list; otherwise, the first
        raised exception will be immediately propagated to the returned
        future.
    
        Cancellation: if the outer Future is cancelled, all children (that
        have not completed yet) are also cancelled.  If any child is
        cancelled, this is treated as if it raised CancelledError --
        the outer Future is *not* cancelled in this case.  (This is to
        prevent the cancellation of one child to cause other children to
        be cancelled.)
    
        If *return_exceptions* is False, cancelling gather() after it
        has been marked done won't cancel any submitted awaitables.
        For instance, gather can be marked done after propagating an
        exception to the caller, therefore, calling ``gather.cancel()``
        after catching an exception (raised by one of the awaitables) from
        gather won't cancel any other awaitables.
        """
        if loop is not None:
>           warnings.warn("The loop argument is deprecated since Python 3.8, "
                          "and scheduled for removal in Python 3.10.",
                          DeprecationWarning, stacklevel=2)
E           DeprecationWarning: The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.

/usr/lib/python3.9/asyncio/tasks.py:755: DeprecationWarning
============================================================================================ 1 failed in 0.17s ============================================================================================
Exception ignored in: <coroutine object test_foo.<locals>.bar at 0x7f4f28e82ac0>
RuntimeWarning: coroutine 'test_foo.<locals>.bar' was never awaited
Exception ignored in: <coroutine object test_foo.<locals>.bar at 0x7f4f28e82a40>
RuntimeWarning: coroutine 'test_foo.<locals>.bar' was never awaited
```

Kludex (approved):
Thanks @graingert ! :tada:

PR #1187, opened 2021-09-18
Pin development requirements

Kludex:
Same idea as https://github.com/encode/httpx/pull/1721.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,30 +1,30 @@
 -e .[standard]
 
 # Explicit optionals
-wsproto==1.0.*
+wsproto==1.0.0
 
 # Packaging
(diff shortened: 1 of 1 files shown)


abersheeran:
The last time I operated uvicorn locally was last year. Yesterday I installed new dependencies after updating the code, but found that the latest version of dependencies could not be installed automatically. I had to manually delete the virtual environment and reinstall it. If you can clarify the dependent version can reduce the trouble in this regard, great work!

lovelydinosaur (commented):
Seems pretty reasonable to me, yup.

abersheeran (approved):
Can not wait anymore!

PR #1188, opened 2021-09-18
Add dependabot

Kludex:
Following https://github.com/encode/uvicorn/pull/1187.


Kludex:
Looks like this is not necessary, as the dependabot is installed on this repository (or at least looks like).

PR #1190, opened 2021-09-18
Remove debug flag

### LINKED ISSUE #480: Phase out `debug` in favour of `reload`
Really we ought to phase out the usage of `--debug`/`debug=True`, in favour of...

* `--reload`/`reload=True` (A server-level concern. Reload the server on code changes.)
* `app = Starlette(debug=True)` (An application level concern. Issue nice HTML tracebacks for the "500 Server Error" case.)

Note that the Starlette package has a standalone `ServerErrorMiddleware` that you can use in isolation if you want just use debug tracebacks together with some custom ASGI app.

Kludex:
Closes https://github.com/encode/uvicorn/issues/480.


euri10:
is the functionality in Starlette now ?

Kludex:
Same parameter is there, if that's what you mean: https://github.com/encode/starlette/blob/10529a3b39e2db21b03600991117c564332f31d2/starlette/applications.py#L41

I've never used this parameter either here or on Starlette tbh. I'm not attached to any of those, but I've created this PR because there was an issue for it. If we decide to not remove it, I'm fine with it, but then we should close the issue.

euri10:
a quick look at starlette and I dont see the same middleware deleted from here `class DebugMiddleware:`
is there the same functionnality under another name ?

PR #1191, opened 2021-09-18
Implement access log format

Kludex:
~Check comments on: https://github.com/Kludex/uvicorn/pull/5~ Unfortunately, I've deleted my fork, and lost all the comments I made. There were a lot of them! 😢 

---
There's still an issue: my idea was to **not** remove the default formatter, but I'm unable to make it work with them.

I need help here. I've been on this problem for some time now, and the values in the `recordcopy` seems correct, but I lost the default colors...


RohanMishra97:
It'd be really nice to get this merged. Would save a lot of overengineering to add custom headers to access log. Is there any workaround as of now for this use-case?
Thanks!😊

Kludex:
Yep, erasing the current "uvicorn.access" logger, creating a new one, using a middleware to collect and print the new data. 

You can check a draft here: https://github.com/Kludex/uvicorn-logger

Kludex:
Conceptually, this PR is wrong. The `AccessLogAtoms` can't be used properly with the logger that we have, specifically the headers.

Issue #160, opened 2018-07-25
Keep alive timeout should trigger for first request too.


lovelydinosaur:

Kludex:
Can you add an explanation, please? I'm willing to give it a try. :) @tomchristie @euri10

euri10:
I've no idea sorry !

Kludex:
@tomchristie When you have time, can you please clarify what do you mean here?

lovelydinosaur:
Complete clarity.

PR #1192, opened 2021-09-19
Fix keepalive timeout task will not execute to close transport

pacoyang:
When transport can not kick off the HTTP protocol, the `on_response_complete` method will not be executed to start a keep-alive timeout task, in this situation if the client-side stream sends EOF, the server-side will not close this transport by timeout task and this TCP connection will keep `CLOSE_WAIT` status.

Start a TCP connection with telnet can replicate this issue. Expect the connection should be close after the keep-alive is timeout.
```bash
telnet 127.0.0.1 8000
```
Relates to #160. Some applications like Chrome or wrk will start a TCP connection to send empty-bytes at first before start the HTTP protocol TCP connection.

### DIFF
--- a/uvicorn/_handlers/http.py
+++ b/uvicorn/_handlers/http.py
@@ -79,8 +79,7 @@ async def handle_http(
     # yet: all data that the client might have already sent since the connection has
     # been established is in the `_buffer`.
     data = reader._buffer  # type: ignore
-    if data:
-        protocol.data_received(data)
+    protocol.data_received(data)
 
     # Let the transport run in the background. When closed, this future will complete
     # and we'll exit here.
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -126,6 +126,11 @@ class H11Protocol(asyncio.Protocol):
 
     def data_received(self, data):
         self._unset_keepalive_if_required()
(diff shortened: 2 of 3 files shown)


Kludex:
You said this on #1332:

> The keep-alive timeout task trigger still has issue, when a TCP connection is established and the client does not send any bytes, the server does not close this connection when keep-alive timeout. It can be reproduced by telnet.

You were referring to this PR/the issue the PR solves, right?

pacoyang:
> You were referring to this PR/the issue the PR solves, right?

@Kludex As #1332 has merged, Not sure this PR is there any conflicts. I will check and submit another PR later.

bevinhex:
this still exists on 0.23.2

bevinhex:
image
image

bevinhex:
log_level="trace" shows "HTTP connection made" when telnet connected, but it keeps that way forever, never timeout

bevinhex:
tested on both h11, httptools

bevinhex:
traced the error to 
image

somehow keep alive handler only set when response complete, not when request just received, thus people can connect and keep tcp connection open , do slow loris attack

bevinhex:
image
starting timer once connection_made seems to be fixing the issue for me, not sure if I am doing it correctly though, will see

PR #1193, opened 2021-09-23
Forward gunicorn reload parameter to uvicorn worker

Kludex:
Basically a copy/paste from https://github.com/benoitc/gunicorn/issues/2339#issuecomment-867481389 made by @jvasi

This solution pops up the following on reload:
```
[INFO] Error while closing socket [Errno 9] Bad file descriptor
```

Notes:
- I haven't investigated the issue above.
- I'm not sure if I want this to be merged, as it can be extended the way it was in the mentioned comment.

### DIFF
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -1,7 +1,10 @@
 import asyncio
 import logging
+import os
 import signal
 import sys
+import threading
+import time
 from typing import Any
 
 from gunicorn.arbiter import Arbiter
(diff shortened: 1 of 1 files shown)


motcke (approved):

Kludex:
This is not a satisfatible solution considering the _reload system_ gunicorn uses. They make use of SIGHUP on the reload, and we should leverage that here.

Kludex:
SIGHUP is not available on Windows, but gunicorn doesn't run on Windows anyway.

odravison:
Has anyone solved this?

PR #1200, opened 2021-09-29
Add type hint for lifespan_class attribute

Kludex:
We'll be able to see the definition of `lifespan` attribute on the `server.py`. 

Changes:
- Add type hint on `lifespan_class`.
- Sort imports on `config.py` (bonus - I can open a new PR if needed).

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -8,17 +8,37 @@ import socket
 import ssl
 import sys
 from pathlib import Path
-from typing import Awaitable, Callable, Dict, List, Optional, Tuple, Type, Union
+from typing import (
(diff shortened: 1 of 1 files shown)


euri10:
> Sort imports on config.py (bonus - I can open a new PR if needed).

would be better yes, cant follow things here
dont we have isort that should do that already ?

Kludex:
The problem is that the following was in the middle:

```python
if sys.version_info < (3, 8):
    from typing_extensions import Literal
else:
    from typing import Literal
```
So isort couldn't act properly.

I'm going to remove it from this PR.

PR #1202, opened 2021-09-30
Deprecate `ServerState` within the main module

Kludex:
Make use of [PEP 562](https://www.python.org/dev/peps/pep-0562/) to deprecate `ServerState` on the `main` module.

The PEP is only available starting on Python 3.7.


Kludex:
My previous fork was deleted. Need to create a new one. This PR is still valid.

PR #1205, opened 2021-10-02
Create ProcessManager

### LINKED ISSUE #517: Master process should restart expired workers.
Uvicorn stays alive when all workers are dead. Is this intended behaviour? Could someone explain the benefit of this?

Uvicorn does not seem to restart individual workers if they are dead. Therefore, shouldn't Uvicorn itself exit when there are zero workers left?

A container orchestrator could automatically reload Uvicorn on exit, which would be awesome.

> [!IMPORTANT]

<a href="https://polar.sh/encode/uvicorn/issues/517">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/517/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

### LINKED ISSUE #1115: [BUG] Exit at startup failure on multiple workers

### LINKED ISSUE #1116: [BUG] Received duplicated signal from Gunicorn
### Checklist

### Describe the bug

Currently, when we send `SIGINT` to the `gunicorn` process that is running with `uvicorn` workers, the `SIGINT` is **not** propagated to the workers. On its turn, a `SIGQUIT` is sent to them. By the time of this issue, the reason for that decision is unclear, but you can follow this [gunicorn issue](https://github.com/benoitc/gunicorn/issues/2604) to understand more.

The mentioned behavior has an effect on `uvicorn`, that I'm unsure how we should act. The effect is that when running `gunicorn` with `uvicorn` workers, the `shutdown` event doesn't happen. Which, in theory, is fine. As `SIGINT` and `SIGQUIT`, should be terminating the worker immediately, but that is not what is in place for standalone `uvicorn` i.e. `SIGNIT` triggers the shutdown event.

Well, all of what I mentioned above should actually be a separated issue, but I was not able to solve it without workarounds, that I'm unsure if we want them. The reason for that is the `uvicorn` worker receiving two signals when we send `SIGINT` to the `gunicorn` process, one is `SIGINT`, which is sent directly to the `uvicorn` worker and the other is the `SIGQUIT` which is sent by the `gunicorn` process. 

As you see, even if we fix `SIGQUIT` by sending `SIGINT`, the issue is not solved i.e. `uvicorn` will receive double `SIGINT`, and we will force exit (via `force_exit` attribute). 

### To reproduce

Just create any ASGI app, with a shutdown event:
```python
# test.py
from fastapi import FastAPI

app = FastAPI()

@app.on_event("shutdown")
async def shutdown():
    print("This will not be triggered")
```

Then run `gunicorn` with `uvicorn` workers:
```bash
gunicorn -k uvicorn.workers.UvicornWorker test:app
```

Feel free to press `CTRL + C` and you'll see this log:
```bash
❯ gunicorn -k uvicorn.workers.UvicornWorker test:app
[2021-07-10 20:08:01 +0200] [43323] [INFO] Starting gunicorn 20.1.0
[2021-07-10 20:08:01 +0200] [43323] [INFO] Listening at: http://127.0.0.1:8000 (43323)
[2021-07-10 20:08:01 +0200] [43323] [INFO] Using worker: uvicorn.workers.UvicornWorker
[2021-07-10 20:08:01 +0200] [43325] [INFO] Booting worker with pid: 43325
[2021-07-10 20:08:01 +0200] [43325] [INFO] Started server process [43325]
[2021-07-10 20:08:01 +0200] [43325] [INFO] Waiting for application startup.
[2021-07-10 20:08:01 +0200] [43325] [INFO] Application startup complete.
^C[2021-07-10 20:08:01 +0200] [43323] [INFO] Handling signal: int
[2021-07-10 20:08:01 +0200] [43325] [INFO] Shutting down
[2021-07-10 20:08:01 +0200] [43325] [INFO] Error while closing socket [Errno 9] Bad file descriptor
[2021-07-10 20:08:01 +0200] [43325] [INFO] Finished server process [43325]
[2021-07-10 20:08:01 +0200] [43325] [INFO] Worker exiting (pid: 43325)
[2021-07-10 20:08:01 +0200] [43323] [INFO] Shutting down: Master
```

As you see, the message on the shutdown event doesn't appear.

### Expected behavior

If we want to have consistency with the standalone `uvicorn`, which I'm really unsure (as it doesn't feel right to wait the process to finish gracefully with `SIGINT`), then we should match this behavior and `gunicorn` with `uvicorn` workers should trigger the shutdown event as well.

In case we decide that the shutdown event should actually not be triggered by standalone `uvicorn`, then the issue here is no more an issue, and we should stop handling the `SIGINT` on the `server.py`. I believe this is unlikely to happen, because of practicality when developing.

### Actual behavior

The shutdown event is not triggered running `gunicorn` with `uvicorn` workers when `SIGINT` is sent. 

### Environment

- OS / Python / Uvicorn version: Running uvicorn 0.14.0 with CPython 3.8.10 on Linux
- gunicorn (version 20.1.0)

### Additional context

More context can be found on Gitter. Here are some messages, but you can follow the discussion there:
https://gitter.im/encode/community?at=60e3592b457e19611a37a5d6
https://gitter.im/encode/community?at=60e35dd8f862a72a30eeb8dc
https://gitter.im/encode/community?at=60e5966e24f0ae2a244a5e0d

\cc @tomchristie @florimondmanca

Kludex:
~It's a draft, but comments about the `NOTES.md` are welcome. :)~
Reviews are welcome.

## Summary

The idea of this PR is to get a process manager that will kill and spawn the child processes as needed.

The issue #517 is the first that introduced the idea.

## Changes
- Deprecate `workers` module, and move worker to another package, as [suggested by Tom](https://github.com/encode/uvicorn/issues/517#issuecomment-564090865).
  Package created and PyPI name registered already: https://github.com/Kludex/uvicorn-worker - Ownership transference will take place if this PR gets merged.
- Replace `Multiprocess` class by `ProcessManager` on Unix systems.

## Decisions
- `should_exit` and `force_exit` aren't the only ones controlling the shutdown anymore. I've also added `terminate_called` inspired by [this article](https://www.cloudcity.io/blog/2019/02/27/things-i-wish-they-told-me-about-multiprocessing-in-python/), and after further thinking, it does make sense. As the signal will be propagated to the workers (and the master wants to let the worker know it should terminate). - This is actually a bug that happens pre this PR as you can see on #1116, but I've just come with this solution when creating this PR, which means I can also create another PR for it.

## Issues
- Closes #517
- Closes #1116
- Closes #1115

## Posterior work
- Worker health check.
- Accept `--pidfile`.
- Accept `--graceful-timeout`.
- Add further signal handling.

#### What happens if the startup event fails on a single worker?
The manager is notified and then terminate gracefully the workers i.e. it will give time to the other workers to perform the shutdown event.

#### Should we have an "age" attribute on the manager?
Age is used to choose which worker to kill.

#### Why is there a random delay on gunicorn `spawn_workers` function?
IDK

#### Should we implement the `maybe_promote_worker` logic? With USR2 signal?
Yes! References:
    - https://docs.gunicorn.org/en/stable/signals.html#upgrading-to-a-new-binary-on-the-fly
    - https://github.com/benoitc/gunicorn/issues/1267

#### On the `signal()` method on the `Arbiter`, why is there a limit of 5 signals on the queue?
IDK

#### Do I need to have a health check? I see that on gunicorn, a temporary file is written from time to time by the workers, so the manager is able to detect if the workers are alive or not.
Yes, we need to have a health check. Not sure if we can use a PIPE instead of a temporary file here.

#### Do we need `KeyboardInterruption` exception catch on the main loop?
IDK

## References
- [Things I wish they told me about multiprocessing in Python](https://www.cloudcity.io/blog/2019/02/27/things-i-wish-they-told-me-about-multiprocessing-in-python/)
- [Stackoverflow Python dynamic multiprocessing and signaling issues](https://stackoverflow.com/questions/40672264/python-dynamic-multiprocessing-and-signalling-issues)
- [Python multiprocessing](https://docs.python.org/3/library/multiprocessing.html)
- [Python signal](https://docs.python.org/3/library/signal.html)
- [Gunicron Arbiter source code](https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py)
- [Gunicorn settings](https://docs.gunicorn.org/en/stable/settings.html)
- [Gunicorn signals](https://docs.gunicorn.org/en/stable/signals.html#upgrading-to-a-new-binary-on-the-fly)
- [Encountering some Python trickery](https://medium.com/pipedrive-engineering/encountering-some-python-trickery-683bd5f66750)
- [Reap zombie processes using a SIGCHLD handler](http://www.microhowto.info/howto/reap_zombie_processes_using_a_sigchld_handler.html)
- [Windows signals](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=msvc-160)

**Design highly inspired by Gunicorn! Reviewers can use [this](https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py) to help.**


uvicorn/supervisors/manager.py
```diff
@@ -0,0 +1,157 @@
...
+    def handle_chld(self, sig: signal.Signals, frame: FrameType) -> None:
+        self.reap_processes()
+
+    def signal_handler(self, sig: signal.Signals) -> Optional[Callable[..., None]]:
+        sig_name = self.SIGNALS.get(sig)
+        return getattr(self, f"handle_{sig_name}", None)
+
+    def reap_processes(self) -> None:
+        for process in self.processes:
+            if not process.is_alive():
```
Kludex, line 132: I need to further check if the `wait()` here is enough to reap! It worked on the cases I've tested, but I'm not sure if there are edge cases!

Kludex:
I still need to add tests, but I'd like reviews in the meantime. I think it has the minimal workable setup that I wanted, and the `Posterior work` section above list what I want to do next.

uvicorn/supervisors/__init__.py
```diff
@@ -11,4 +11,9 @@
     except ImportError:  # pragma: no cover
         from uvicorn.supervisors.statreload import StatReload as ChangeReload
 
+if sys.platform == "win32":
+    from uvicorn.supervisors.multiprocess import Multiprocess
```
abersheeran, line 15: If there are plans to support Windows, I would be happy to help. (I read the code carefully, it is not difficult to support Windows)

Kludex: My problem is with the signals not supported by Windows, what do you expect to do on that part? I just added the minimal signal handlers, but further work should not work.

abersheeran: We can use [process.terminate()](https://docs.python.org/zh-cn/3/library/multiprocessing.html#multiprocessing.Process.terminate) replace SIGINT. Other signals are transmitted as they are.

abersheeran: In fact, a large number of other signals cannot be triggered on Windows. We can give Windows users a near-identical development experience as long as we can handle the signals raised by Ctrl-C in a unified manner.

Kludex: I was not referring to `SIGINT`, `SIGINT` and `SIGTERM` are not an issue, and I don't see any issue of calling `process.terminate()` on `SIGINT`. I mean, I don't think there's an issue on that part for windows users.

~The issue right now is the `init_signals()` method, that adds a handler for signals that Windows doesn't have. If we just add a conditional to create `SIGNALS` attribute according to the OS, then it will work.~ The issue is with the `SIGCHLD`, that Windows doesn't support.

Kludex: When you say "development experience" I imagine the user wanting to use the reload feature, which should not be recommended here? :thinking:

I mean, if the user wants to use the class implemented here in production, I imagine that it also wants to use the reload feature in development.

abersheeran: > ~The issue right now is the `init_signals()` method, that adds a handler for signals that Windows doesn't have. If we just add a conditional to create `SIGNALS` attribute according to the OS, then it will work.~ The issue is with the `SIGCHLD`, that Windows doesn't support.

https://github.com/Kludex/uvicorn/blob/a4377a14e3360fa487c506b49e8c5537fb0057a1/uvicorn/supervisors/manager.py#L71-L74

Maybe don't need SIGCHLD in Windows? The delay of 0.25 is not high.

abersheeran: > When you say "development experience" I imagine the user wanting to use the reload feature, which should not be recommended here? 🤔
> 
> I mean, if the user wants to use the class implemented here in production, I imagine that it also wants to use the reload feature in development.

Yes, I agree. It is a good idea to incorporate the reload here.

abersheeran: > I was not referring to `SIGINT`, `SIGINT` and `SIGTERM` are not an issue, and I don't see any issue of calling `process.terminate()` on `SIGINT`. I mean, I don't think there's an issue on that part for windows users.

https://github.com/encode/uvicorn/issues/684 Sometime, `os.kill()` maybe got an error.

Kludex: >https://github.com/Kludex/uvicorn/blob/a4377a14e3360fa487c506b49e8c5537fb0057a1/uvicorn/supervisors/manager.py#L71-L74
>
> Maybe don't need SIGCHLD in Windows? The delay of 0.25 is not high.

I see! You mean that it's redundant to have the handler as we already perform the same logic in case we don't handle the `SIGCHLD`.

I need to recheck if there's further logic to be implemented on the `SIGCHLD` handler that I didn't add because it's not needed on this initial step. If I can't find anything, then yes, we can remove it.

Kludex: > Yes, I agree. It is a good idea to incorporate the reload here.

I was not defending the idea of adding the reload here, as I think it's a posterior discussion. But I do agree with that, if we do that, we're going to have only this `ProcessManager` class, and both the Reload classes and the `Multiprocess` class should be removed.

PR #1189, opened 2021-09-18
Remove Python 3.6

Kludex:
Python 3.6 EOL is 23 Dec 2021, and this PR ~starts the process of removing it~ removes it. 

Reference: https://endoflife.date/python

PR #1210, opened 2021-10-05
Support zerocopy send

### LINKED ISSUE #35: Support for sendfile
Support `sendfile` usage.

Would require an ASGI extension.

abersheeran:
- [x] Implement with httptools.
- [x] Implement with h11: https://h11.readthedocs.io/en/latest/api.html#sendfile

Close #35


uvicorn/protocols/http/httptools_impl.py
```diff
@@ -480,31 +490,64 @@ async def send(self, message):
...
+                    if not more_body:
+                        content.append(b"0\r\n\r\n")
+                    self.transport.write(b"".join(content))
                 else:
-                    content = []
-                if not more_body:
-                    content.append(b"0\r\n\r\n")
-                self.transport.write(b"".join(content))
+                    self.transport.write(b"%x\r\n" % sendfile_count)
+                    os.sendfile(socket_fd, file_fd, sendfile_offset, sendfile_count)
```
abersheeran: I noted https://github.com/encode/uvicorn/pull/1189. If uvicorn would drop support for 3.6, we can use loop.sendfile in here. What do you think? @Kludex

Kludex: Feel free to use `loop.sendfile` :+1:

requirements.txt
```diff
@@ -23,6 +23,7 @@ coverage
 httpx>=0.18.2
 pytest-asyncio==0.14.*
 async_generator; python_version < '3.7'
+baize>=0.11
```
lovelydinosaur: It'd be good if we could avoid this additional dependancy.

lovelydinosaur:
Hi @abersheeran I've invited you as a member of the Encode org. 👍

abersheeran:
> Hi @abersheeran I've invited you as a member of the Encode org. 👍

😀Yeah, I joined.

requirements.txt
abersheeran: Well, it is not difficult. I will modify it next time I submit it.

Kludex:
> Hi @abersheeran I've invited you as a member of the Encode org. +1

I was about to ask that. :+1: Thanks!

uvicorn/protocols/http/httptools_impl.py
lovelydinosaur: We could make the sendfile support be conditional on Python 3.7, without having to outright drop 3.6 support right?
If that keeps the implementation simple, then I think that'd be worth doing.

lovelydinosaur:
Do we know what the deal is with the hanging CI tests?

Kludex:
It should be related with this branch. The CI is fine on uvicorn now.

abersheeran:
The sendfile cannot be implemented on Windows, so the test coverage cannot reach 95%. I am not familiar with whether it can be selectively skipped in this case?

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -44,6 +44,40 @@ def _get_status_line(status_code):
     HAS_SENDFILE = True
 except AttributeError:
     HAS_SENDFILE = False
+else:
+    # Note: because uvloop don't implements loop.sendfile, so use os.sendfile as here
+    # Related link: https://github.com/MagicStack/uvloop/issues/228
```
lovelydinosaur: I'd suggest that we keep things simple instead. If the underlying implementation doesn't support `sendfile`, then let's just not support it in `uvicorn` either.

So...

* You can have sendfile support on Python 3.7+, without uvloop.
* You can have sendfile support with uvloop once they support it.

Kludex:
> The sendfile cannot be implemented on Windows, so the test coverage cannot reach 95%. I am not familiar with whether it can be selectively skipped in this case?

Yeah, we were adding `pragma: no cover` on non Python 3.8, as discussed [here](https://github.com/encode/uvicorn/issues/102#issuecomment-797365601) and the comment below.
But I wish we could do this: https://github.com/encode/uvicorn/pull/1159 - More context on the mentioned PRs there.

abersheeran:
> > The sendfile cannot be implemented on Windows, so the test coverage cannot reach 95%. I am not familiar with whether it can be selectively skipped in this case?
> 
> Yeah, we were adding `pragma: no cover` on non Python 3.8, as discussed [here](https://github.com/encode/uvicorn/issues/102#issuecomment-797365601) and the comment below. But I wish we could do this: #1159 - More context on the mentioned PRs there.

Thanks for your reply, it is very helpful!

uvicorn/protocols/http/httptools_impl.py
abersheeran: This reminds me of one thing, maybe `sendfile` is also available on Windows.

synodriver:
Thank you for your pr! Now [hypercorn](https://gitlab.com/synodriver/hypercorn/-/tree/zerocopy) supports zerocopysend too.

docs/server-behavior.md
```diff
@@ -56,6 +56,10 @@ Applications should generally treat `HEAD` requests in the same manner as `GET`
 
 One exception to this might be if your application serves large file downloads, in which case you might wish to only generate the response headers.
 
+### Zero copy send
+
+When you install httptools, Uvicorn supports [ASGI Zero Copy Send Extension](https://asgi.readthedocs.io/en/latest/extensions.html#zero-copy-send). Using the method specified by the extension, you can call the zero-copy interface of the operating system to send files.
```
Kludex: You've also implemented for h11.

abersheeran: Haha, I almost forgot. I'm going to modify the document now.

docs/server-behavior.md
```diff
@@ -56,6 +56,12 @@ Applications should generally treat `HEAD` requests in the same manner as `GET`
 
 One exception to this might be if your application serves large file downloads, in which case you might wish to only generate the response headers.
 
+### Zero copy send
+
+Uvicorn supports [ASGI Zero Copy Send Extension](https://asgi.readthedocs.io/en/latest/extensions.html#zero-copy-send). Using the method specified by the extension, you can call the zero-copy interface of the operating system to send files.
+
+Note: This is currently only available on Python3.7+ and is temporarily unavailable with uvloop.
```
euri10: could we link to a uvloop issue that tracks sendfile implementation ?

euri10 (commented):
alos I see a lot of `# pragma: no cover` is there a reason ?

uvicorn/protocols/http/h11_impl.py
```diff
@@ -367,6 +379,11 @@ def __init__(
         self.response_started = False
         self.response_complete = False
 
+        # Sendfile
+        if self.allow_sendfile:  # pragma: no cover
+            # Set the buffer to 0 to avoid the problem of sending file before headers.
+            transport.set_write_buffer_limits(0)
```
lovelydinosaur, line 376: I'm not really comfortable with the fact that this'll change the behaviour throughout, regardless of if `http.response.zerocopysend` is actually being used or not for any given request/response.

uvicorn/protocols/http/h11_impl.py
```diff
@@ -454,20 +471,53 @@ async def send(self, message):
...
+                                self.transport, data.file, data.offset, data.count
+                            )
+                        else:
+                            self.transport.write(data)
             else:
                 event = h11.Data(data=body)
-            output = self.conn.send(event)
-            self.transport.write(output)
+                output = self.conn.send(event)
+                self.transport.write(output)
```
lovelydinosaur, line 513: This looks like it unintentionally changes the behaviour for the `"HEAD"` case.

lovelydinosaur:
Okay, so this is clearly a neatly done little bit of work, nonetheless I feel like I ought to be the voice of pushback here.

Personally I think that in the trade-off between functionality vs. increased complexity, increased ongoing maintenance cost, scope for bug introduction, this is a possibly poor trade-off for us.

I'd perhaps be more comfortable with this if it was obvious that it *didn't* affect any other behaviour other that just for that one message case, but currently the write buffer change makes me uncomfortable, as does the largish footprint of the change within the `# Sending response body` block.

uvicorn/protocols/http/h11_impl.py
abersheeran: If this option is not configured, the headers may be sent after the body is sent.

docs/server-behavior.md
abersheeran: https://github.com/encode/uvicorn/pull/1210/commits/9971c6bc48254018db55f17a25eefabf313831cd done.

uvicorn/protocols/http/h11_impl.py
abersheeran: https://github.com/encode/uvicorn/pull/1210/commits/ddb3e58672531cd3034513bd704921c047d2dd7a done.

abersheeran:
> I'd perhaps be more comfortable with this if it was obvious that it _didn't_ affect any other behaviour other that just for that one message case, but currently the write buffer change makes me uncomfortable, as does the largish footprint of the change within the `# Sending response body` block.

Sendfile has to face the buffer, there seems to be no good way to avoid interference with other types of responses-we have no means to predict this behavior before zerocopysend is sent.

Kludex:
Closing this as stale.

Issue #1136, opened 2021-07-28
Queue operations on lists


Vibhu-Agarwal:
### Checklist

- [X] The bug is reproducible against the latest release and/or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Issue
I noticed that `list` is being used for queue-based operations in a couple of places:
- https://github.com/encode/uvicorn/blob/3a5c3b871379236f2867d7799d3480b94fcc5e60/uvicorn/middleware/wsgi.py#L98
- https://github.com/encode/uvicorn/blob/3a5c3b871379236f2867d7799d3480b94fcc5e60/uvicorn/protocols/http/httptools_impl.py#L77

These operations, therefore, incur O(n) memory movement costs:
- https://github.com/encode/uvicorn/blob/3a5c3b871379236f2867d7799d3480b94fcc5e60/uvicorn/middleware/wsgi.py#L131
- https://github.com/encode/uvicorn/blob/3a5c3b871379236f2867d7799d3480b94fcc5e60/uvicorn/protocols/http/httptools_impl.py#L262

### Question
Is there a reason why `list` is being used and not a native queue like [asyncio.Queue](https://docs.python.org/3/library/asyncio-queue.html)?  
If there is a need of synchronous operation, we can use [collections.deque](https://docs.python.org/3/library/collections.html#collections.deque).

Kludex:
I don't think there is. PR welcome.

PR #1214, opened 2021-10-08
[middleware/wsgi.py] Replace list with deque for `send_queue`

Vibhu-Agarwal:
Fix for the issue mentioned in #1136


Kludex (approved):
Thanks @Vibhu-Agarwal ! 🎉

Issue #1223, opened 2021-10-20
Cant logging real client ip


monkeyWie:
I'm using nginx to reverse uvicorn server, the nginx conf like this:
```
location /api {
        proxy_pass http://127.0.0.1:28888;

        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
}
```
and the uvicorn start command is `["uvicorn","app.main:app","--host","0.0.0.0","--proxy-headers","--forwarded-allow-ips='*'"]`,but I see that the log ip is still the local ip

![image](https://user-images.githubusercontent.com/13160176/138018170-940dce78-e42e-4c09-bde4-38b3d726e37f.png)

So what should I do?

Thanks~

manlix:
Hi @monkeyWie

What IP address did you expect to see?

I think you use **Docker** or **Docker Compose** to run `nginx`. 

If it's so: `172.17.0.1` is correct IP address because your default network as `bridge` and your IP address is `172.17.0.1` (**gateway**) by default to connect to containers:

```
$ sudo docker network inspect bridge | jq ".[].IPAM.Config"
[
  {
    "Subnet": "172.17.0.0/16",
    "Gateway": "172.17.0.1"
  }
]
```

Try to use `host` network as workaround:

For **Docker**: `--network=host`
```
$ sudo docker run --network=host --rm -it IMAGE
```

For **Docker Compose**: `network_mode=host`
```
services:
  nginx:
  ...
  network_mode: host
```

But keep in mind in this case you use host network not isolated docker network.

monkeyWie:
Actually my nginx is running at host machine and uvicorn server runing in docker, I use tcpdump capture http data that nginx reverse  request uvicorn server,like this:
![image](https://user-images.githubusercontent.com/13160176/138070644-e0dd7339-0493-4d44-80e0-0e5243f1eb43.png)

You can see that `X-Forwarded-For` is fine, but the uvicorn log ip is still `172.17.0.1`.

manlix:
@monkeyWie  Thank you for details.

> ["uvicorn","app.main:app","--host","0.0.0.0","--proxy-headers","--forwarded-allow-ips='*'"]

Try to replace `"--forwarded-allow-ips='*'"` to `"--forwarded-allow-ips", "*"`:

```
["uvicorn","app.main:app","--host","0.0.0.0","--proxy-headers","--forwarded-allow-ips", "*"]
```

[...]

PR #1225, opened 2021-10-20
Show warning about incorrect IP in trusted hosts

manlix:
Right now you can pass any incorrect data to `--forwarded-allow-ips` and `uvicorn` starts as expected with no warning at least.

Correct values for now:
* `--forwarded-allow-ips="1.2.3.4,777.289.357.999"`
* `--forwarded-allow-ips="1.2.3.4.5.6.7.8,foo bar"`
* `--forwarded-allow-ips="2001:db8::1X,333.333.333.333"`
* `--forwarded-allow-ips=""`

I think we should check IPs and raise up warning about existing incorrect IP in `trusted hosts`.

This **PR** implements checking IPs in `--forwarded-allow-ips` and raises warning to stdout like: 
`Incorrect IP in trusted hosts`.

Inspired by #1223

### DIFF
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -1,3 +1,4 @@
+from logging import WARNING
 from typing import List, Union
 
 import httpx
@@ -5,6 +6,8 @@ import pytest
 from asgiref.typing import ASGIReceiveCallable, ASGISendCallable, Scope
 
 from tests.response import Response
+from tests.utils import run_server
+from uvicorn import Config
 from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
 
 
@@ -18,6 +21,28 @@ async def app(
     addr = "%s://%s:%d" % (scheme, host, port)
     response = Response("Remote: " + addr, media_type="text/plain")
     await response(scope, receive, send)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("ips"),
+    [
+        ("1.22.333.4444"),
+        ("192.168.0.1,1234"),
+        ("2001:db8::1X"),
+        (""),
+    ],
+)
+async def test_incorrect_ip_in_trusted_hosts(
+    ips: str, caplog: pytest.LogCaptureFixture
+) -> None:
+    config = Config(app=app, forwarded_allow_ips=ips)
+
+    async with run_server(config):
+        pass
+    warnings = [record for record in caplog.records if record.levelno == WARNING]
+
+    assert warnings[0].message == "Incorrect IP in trusted hosts"
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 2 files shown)


uvicorn/middleware/proxy_headers.py
```diff
@@ -31,6 +33,12 @@ def __init__(
             self.trusted_hosts = set(trusted_hosts)
         self.always_trust = "*" in self.trusted_hosts
 
+        try:
+            list(map(ipaddress.ip_address, self.trusted_hosts))
+        except ValueError:
+            logger = logging.getLogger("uvicorn.warning")
```
Kludex, line 39: We don't have this logger.

```suggestion
            logger = logging.getLogger("uvicorn.error")
```

tests/middleware/test_proxy_headers.py
```diff
@@ -20,6 +23,28 @@ async def app(
...
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("ips"),
+    [
+        ("1.22.333.4444"),
+        ("192.168.0.1,1234"),
+        ("2001:db8::1X"),
+        (""),
```
Kludex, line 33:
```suggestion
        (""),
        ("*"),
```

tests/middleware/test_proxy_headers.py
```diff
@@ -20,6 +23,28 @@ async def app(
...
+async def test_incorrect_ip_in_trusted_hosts(
+    ips: str, caplog: pytest.LogCaptureFixture
+) -> None:
+    config = Config(app=app, forwarded_allow_ips=ips)
+
+    async with run_server(config):
+        pass
+    warnings = [record for record in caplog.records if record.levelno == WARNING]
+
+    assert warnings[0].message == "Incorrect IP in trusted hosts"
```
Kludex, line 45: Is there a reason for not following the same test structure as the other tests in this file? i.e. instantiating the middleware directly.

tests/middleware/test_proxy_headers.py
```diff
@@ -20,6 +23,28 @@ async def app(
...
+        (""),
+    ],
+)
+async def test_incorrect_ip_in_trusted_hosts(
+    ips: str, caplog: pytest.LogCaptureFixture
+) -> None:
+    config = Config(app=app, forwarded_allow_ips=ips)
+
+    async with run_server(config):
+        pass
```
Kludex, line 42:
```suggestion
    config.load()
```

Kludex (commented):
I'm not sure if we should handle the field validation on the middleware.

I'm also not sure if it should be a warning. It looks like an error. :thinking:

Kludex:
We don't perform validation on parameters. If we wish to do that in the future, let's do it in an organized manner.

That being said, and considering that this PR has become stale, I'll be closing it.

PR #1235, opened 2021-11-06
Deprecate WSGIMiddleware

abersheeran:
Related links: https://github.com/encode/uvicorn/pull/1049#issuecomment-870520562

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -64,6 +64,7 @@ filterwarnings=
     error
     # Workaround for Python 3.9.7 (see https://bugs.python.org/issue45097)
     ignore:The loop argument is deprecated since Python 3\.8, and scheduled for removal in Python 3\.10\.:DeprecationWarning:asyncio
+    ignore:The built-in `WSGIMiddleware` will be deprecated in future versions, please use `a2wsgi.WSGIMiddleware` instead.:DeprecationWarning
 
 [coverage:run]
 omit = venv/*
--- a/uvicorn/middleware/wsgi.py
+++ b/uvicorn/middleware/wsgi.py
@@ -2,6 +2,7 @@ import asyncio
 import concurrent.futures
 import io
 import sys
+import warnings
 from collections import deque
 from typing import Deque, Iterable, Optional, Tuple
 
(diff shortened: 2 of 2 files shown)


abersheeran:
It seems that using DeprecationWarning directly will cause the test to fail, how can I fix it?

euri10:
add it in setup.cfg ?

```
    ignore:The built-in `WSGIMiddleware` will be deprecated in future versions, please use `a2wsgi.WSGIMiddleware` instead.:DeprecationWarning
```

Kludex (commented):
Maybe we should deprecate the module instead?

Like, we are not strict when it comes to private/public objects, so I guess the middleware and `build_environ`/`WSGIResponder` are in the same privacy level. Does it make sense?

abersheeran:
> Maybe we should deprecate the module instead?
> 
> Like, we are not strict when it comes to private/public objects, so I guess the middleware and `build_environ`/`WSGIResponder` are in the same privacy level. Does it make sense?

Agree with your ideas. But I am not sure how to do it...I have no experience in this area. Is there a module that has already done this for me?

Kludex:
I guess like this: https://github.com/encode/uvicorn/pull/1205/files#diff-7c8a9d71c98f0ed83c32a88a1b7e777ef85776c98869045d24c74e3f7714494a

Yeah, we never deprecated modules on uvicorn before, this would be the first time. We have done it for starlette: https://github.com/encode/starlette/blob/8a3e41a5442239bb2da5f2df6cfaaf4683e49f96/starlette/graphql.py#L12-L17

Kludex:
What's the plan with `--interface wsgi`?

The idea is to import the middleware from `a2wsgi`?

abersheeran:
> What's the plan with `--interface wsgi`?
> 
> The idea is to import the middleware from `a2wsgi`?

Do I need to submit it in this PR? I took a quick look at the code and it shouldn't be difficult to implement.

Kludex:
> > What's the plan with `--interface wsgi`?
> > 
> > The idea is to import the middleware from `a2wsgi`?
> 
> Do I need to submit it in this PR? I took a quick look at the code and it shouldn't be difficult to implement.

No, I just want to confirm the best steps.

Kludex:
@abersheeran What's our plan for it? :thinking:

abersheeran:
> @abersheeran What's our plan for it? 🤔

I don't quite understand what you mean. 👀

euri10:
let's favor https://github.com/encode/uvicorn/pull/1303

PR #1237, opened 2021-11-07
Make use of the unused_tcp_port fixture to spawn the server on a random port

euri10:
small attempt to see if this could solve debian packaging connection issues, see https://github.com/encode/uvicorn/discussions/1231#discussioncomment-1600482

### DIFF
--- a/tests/middleware/test_logging.py
+++ b/tests/middleware/test_logging.py
@@ -28,12 +28,12 @@ async def app(scope, receive, send):
 
 
 @pytest.mark.asyncio
-async def test_trace_logging(caplog):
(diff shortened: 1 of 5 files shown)


Kludex:
What's bad about this? What is missing? @euri10

euri10:
> What's bad about this? What is missing? @euri10

nothing bad, I originally quickly wrote this to help the aforementioned debian packaging issue, but since this seem to be not useful for it, I'm not sure we want it, there was an old PR that wanted to do this a while back, I cant find it anymore, I accepted it but the author rage closed it because I failed to see the interest in having it, so if someone wants this,  happy for it to land,

Issue #1234, opened 2021-11-06
Dealing with large requests: possibility to set max_incomplete_event_size for h11


alexiri:
### Checklist

- [x] There are no similar issues or pull requests for this yet.
- [ ] I discussed this idea on the [community chat](https://gitter.im/encode/community) and feedback is positive.
(Asked in the chat, but no response received)

### Is your feature related to a problem? Please describe.

I'm working on an application that needs to deal with large http headers, but I'm getting the following exception:

```
WARNING:  Invalid HTTP request received.
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/uvicorn/protocols/http/h11_impl.py", line 136, in handle_events
    event = self.conn.next_event()
  File "/usr/local/lib/python3.9/site-packages/h11/_connection.py", line 432, in next_event
    raise RemoteProtocolError(
h11._util.RemoteProtocolError: Receive buffer too long
```

## Describe the solution you would like.

This error could be fixed if I was able to configure the `max_incomplete_event_size` of the h11 connection:

https://github.com/python-hyper/h11/blob/master/h11/_connection.py#L119

## Describe alternatives you considered

For the time being, I've moved to Hypercorn which does allow me to change that parameter.

Kludex:
Yeah, some time ago I was reviewing the `h11` implementation, and I've noticed that we just use the default.

The default value is [explained on `h11`](https://h11.readthedocs.io/en/latest/_modules/h11/_connection.html#Connection), but I guess it does make sense to allow the user to change its value if needed.

PR is welcome, but let's use the same strategy that we use with other CLI arguments: if the argument is not used, just ignore it.

Also, the "describe alternatives you considered" is meant to be an alternative solution that we can use to solve the problem in hands. Have you also tried the `httptools` implementation?

alexiri:
I confess I haven't tried the `httptools` implementation, I was in a bit of a rush with the deployment of my application and didn't think of testing it. I'll see if I find some time this week to do so.

johnraz:
Doesn't this also applies to large query string?
If so I think the issue should be renamed to avoid that people discard reading the issue while the explanation is actually there ;-)

lovelydinosaur:
> I'm working on an application that needs to deal with large http headers

So, there is a trade-off with https://github.com/encode/uvicorn/pull/1240 - which is that although it's a very nicely worked on PR. It also adds more configuration controls, which *really* we ought to try to avoid if at all possible.

[...]

PR #1240, opened 2021-11-07
Allow configuring max_incomplete_event_size for h11 implementation

alexiri:
Solves #1234.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -50,6 +50,7 @@ By default Uvicorn uses simple changes detection strategy that compares python f
 * `--ws-ping-interval <float>` - Set the WebSockets ping interval, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-timeout <float>` - Set the WebSockets ping timeout, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--lifespan <str>` - Set the Lifespan protocol implementation. **Options:** *'auto', 'on', 'off'.* **Default:** *'auto'*.
+* `--h11-max-incomplete-event-size <int>` - For h11 protocol implementation, set the maximum number of bytes to buffer of an incomplete event.
 
 ## Application Interface
 
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -309,7 +309,7 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     help="Specify custom default HTTP response headers as a Name:Value pair",
 )
 @click.option(
-    "--h11_max_incomplete_event_size",
+    "--h11-max-incomplete-event-size",
     "h11_max_incomplete_event_size",
     type=int,
     default=None,
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -309,6 +309,13 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     help="Specify custom default HTTP response headers as a Name:Value pair",
 )
 @click.option(
+    "--h11_max_incomplete_event_size",
+    "h11_max_incomplete_event_size",
+    type=int,
+    default=None,
+    help="For h11, the maximum number of bytes to buffer of an incomplete event",
+)
+@click.option(
     "--version",
     is_flag=True,
     callback=print_version,
@@ -373,6 +380,7 @@ def main(
     ssl_ca_certs: str,
     ssl_ciphers: str,
     headers: typing.List[str],
+    h11_max_incomplete_event_size: int,
     use_colors: bool,
     app_dir: str,
     factory: bool,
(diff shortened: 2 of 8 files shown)


uvicorn/main.py
```diff
@@ -308,6 +308,13 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     multiple=True,
     help="Specify custom default HTTP response headers as a Name:Value pair",
 )
+@click.option(
+    "--h11_max_incomplete_event_size",
```
```suggestion
    "--h11-max-incomplete-event-size",
```

Kludex (changes requested):

alexiri (commented):
Changes implemented in ee25dfcce52887db5400abe5276a305071ade569. @Kludex could you please take a look?

Kludex (commented):
We probably want to add a test on the `test_http.py` for h11, using the new `h11_max_incomplete_event_size`.

alexiri:
@Kludex I've added a test to demonstrate how h11 fails and another to show how the parameter fixes it. I wanted to actually assert the right exception, but the mock doesn't let it bubble up. Can you take a look and let me know what you think of this approach?

uvicorn/protocols/http/h11_impl.py
```diff
@@ -48,7 +48,13 @@ def __init__(
...
         self.access_logger = logging.getLogger("uvicorn.access")
         self.access_log = self.access_logger.hasHandlers()
-        self.conn = h11.Connection(h11.SERVER)
+        if config.h11_max_incomplete_event_size:
+            self.conn = h11.Connection(
+                h11.SERVER,
+                max_incomplete_event_size=config.h11_max_incomplete_event_size,
+            )
+        else:
+            self.conn = h11.Connection(h11.SERVER)
```
Kludex, line 57: Maybe we should set the default ourselves instead of using the one h11 defines? Even if it's with the same value?

Just a thought.

Kludex (commented):
Maybe we should also add some information on the `settings.md`?

uvicorn/protocols/http/h11_impl.py
alexiri: I think it's better to respect the h11 default in case they change it for some reason.

Kludex: I think the reason you mentioned is exactly why we should set it here :sweat_smile:

Maybe we should check what hypercorn is doing here :thinking:

Kludex:
btw, thanks for the PR! :tada:

uvicorn/protocols/http/h11_impl.py
alexiri: Hypercorn sets it, but I think it's the wrong approach. If you're using h11, you should trust them to set their defaults correctly. If they change it, they probably have a good reason for doing so, like fixing a bug or implementing a new feature. If you're still using the old default, now you're affected by the bug they already fixed and you have to update your default as well.

Anyway, if you feel strongly about this, feel free to set the default directly. The code will certainly be cleaner.

alexiri:
> Maybe we should also add some information on the `settings.md`?

Ok, done.

uvicorn/protocols/http/h11_impl.py
Kludex: > If they change it, they probably have a good reason for doing so, like fixing a bug or implementing a new feature.

I think we need to be in charge here. `h11` can change their values as they want, they are on pre-stable release (I doubt they would, because this value is actually based on several other implementations).

Also, the code becomes cleaner anyway.

Would you mind changing it?

Kludex:
> @Kludex I've added a test to demonstrate how h11 fails and another to show how the parameter fixes it. I wanted to actually assert the right exception, but the mock doesn't let it bubble up. Can you take a look and let me know what you think of this approach?

Can't you do something like:
```
    with pytest.raises(AssertionError) as exc:
        assert exc.msg(?) == ??
```

docs/settings.md
```diff
@@ -50,6 +50,7 @@ By default Uvicorn uses simple changes detection strategy that compares python f
 * `--ws-ping-interval <float>` - Set the WebSockets ping interval, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-timeout <float>` - Set the WebSockets ping timeout, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--lifespan <str>` - Set the Lifespan protocol implementation. **Options:** *'auto', 'on', 'off'.* **Default:** *'auto'*.
+* `--h11-max-incomplete-event-size <int>` - For h11 protocol implementation, set the maximum number of bytes to buffer of an incomplete event.
```
Kludex, line 53:
```suggestion
* `--h11-max-incomplete-event-size <int>` - Set the maximum number of bytes to buffer of an incomplete event. Only available for `h11` HTTP protocol implementation.
```

Kludex (commented):
@alexiri Are you still interested in this PR?

alexiri:
Not really, no. Feel free to close it.

doiko:
Any plans to merge this one?

Kludex:
Let's reevaluate over the issue.

PR #1241, opened 2021-11-08
Avoid error messages from logging when port is str

naslundx:
Currently, starting uvicorn with port argument as a string would work - but the logging message at startup

> Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)

would crash because it expected an int. This fixes that.

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -188,7 +188,7 @@ class Config:
         self,
         app: Union[ASGIApplication, Callable, str],
         host: str = "127.0.0.1",
-        port: int = 8000,
+        port: Union[int, str] = 8000,
         uds: Optional[str] = None,
         fd: Optional[int] = None,
         loop: LoopSetupType = "auto",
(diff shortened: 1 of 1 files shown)


euri10:
I dont understand what it solves, care to explain @naslundx what the message would crash means.
Will reopen in case there is indeed something

naslundx:
Sorry for the hasty description @euri10 . Here's an example (especially the last two lines):

```py
import uvicorn
from fastapi import FastAPI
app = FastAPI()

@app.get("/")
def index():
    return "Hello world"

if __name__ == "__main__":
    PORT = "5000"  # NOTE This is a string
    uvicorn.run(app, host="0.0.0.0", port=PORT)
```

This throws an exception in the logging module "TypeError: %d format: a number is required, not str" caused by uvicorn\server.py", line 217, in _log_started_message. The exception is caught and uvicorn runs normally, so this is entirely due to the port not passed properly to the logging message. This PR would fix that.

PR #1243, opened 2021-11-08
Bump pinning httptools 0.2.0

### LINKED ISSUE #1242: Remove pinning of httptools==0.2.* - this blocks wheel usage of httptools on Python 3.10 for uvicorn[standard]
In summer, httptools upgraded to version `0.3.*` as a maintainace release (no functional changes), just changing the packaging and wheel building to support Python 3.10, dropping Python 3.5 support, see here: https://github.com/MagicStack/httptools/releases/tag/v0.3.0

Currently, the `uvicorn[standard]` dependencies pin `httptools` at version `0.2.*`:
https://github.com/encode/uvicorn/blob/a7d26bb76c29a9bd12f53b3a73ae5614592cdfb4/setup.py#L59

This prevents `uvicorn[standard]` to use wheels for `httptools `on Python 3.10 (cp310), **often preventing the simple usage of uvicorn[standard] with Python 3.10**, as they are only available on `httptools>=0.3.0`.

**Proposed fix:**
Either use `httptools=0.3.x` or drop the pin (initial reason for pinning?)

**Explanation** 
As the `httptools` upgrade is just a packaging release and Python 3.5 is not supported by `uvicorn `anyways, upgrading the pinning to 0.3.x (or dropping it entirely, after careful consideration) should not introduce any side-effects or breaking changes for the supported user base, but instead enable the simple usage of Python 3.10 without the need to manually build wheels for `httptools` on Python 3.10 during `pip install`

I did not open a pull request yet as I don't know if the pin is intentional and wanted to discuss this first.

quantumtec:
resolves https://github.com/encode/uvicorn/issues/1242

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ minimal_requirements = [
 extra_requirements = [
     "websockets>=9.1; " + env_marker_below_37,
     "websockets>=10.0; " + env_marker_gte_37,
-    "httptools>=0.2.0",
+    "httptools>=0.2.0,<0.4.0",
     "uvloop>=0.14.0,!=0.15.0,!=0.15.1; " + env_marker_cpython,
     "colorama>=0.4;" + env_marker_win,
     "watchgod>=0.6",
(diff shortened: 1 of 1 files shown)


Kludex (approved):
Thanks @quantumtec! :tada:

PR #1244, opened 2021-11-08
Ensure HTTP connections are closed in case of data-less TCP pings

### LINKED ISSUE #1226: Using UvicornWorkers in Gunicorn cause OOM on K8s
### Checklist

### Describe the bug

I'm developing a FastApi application deployed on a Kubernetes cluster using gunicorn as process manager.

I'm also using UvicornWorkers for sure, because of the async nature of fastapi.

After the application deployment I can see the memory growing up at rest, until OOM.

This happen just when I use UvicornWorker. 

Tests made by me:

- Comment all my code to ensure is not a my application mem leak (leak present);
- Start the application using uvicorn instead of gunicorn (no leak present);
- Start application using gunicorn sync workers (no leak present);
- Start application using gunicorn + UvicornWorker (leak present);
- Start application using gunicorn + UvicornWorker + max_requests (leak present);

Plus, this happens just on the Kubernetes cluster, when I run my application locally (MacBook pro 16) (is the same docker image used on k8s) the leak is not present.

Anyone else had a similar problem?

> [!IMPORTANT]

<a href="https://polar.sh/encode/uvicorn/issues/1226">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/1226/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

florimondmanca:
Closes https://github.com/encode/uvicorn/issues/1226

Completes #869 (released in 0.14.0) with a missing branch: it _is_ possible that `data` be `b""`, e.g. if the client is a health checker that uses empty TCP pings to verify a service is up. As shown in #1226, this may be the case in k8s environments for example. In that case, we should be mindful of closing the `transport`, i.e. closing the connection (hence the connection/memory leak reported in #1226).

To reproduce before/after, consider this server:

```python
    if scope["type"] == "lifespan":
        await lifespan(scope, receive, send)
        return

    await http(scope, receive, send)
    await send(
        {
            "status": 200,
            "headers": [[b"content-type", b"text/plain"]],
        }
    )
```

And this "ping as fast as you can" client:

```python
import socket

while True:
    with socket.create_connection(("localhost", 8000)) as sock:
        pass
```

0. Fire up a resource monitor, e.g. Activity Monitor on macOS.
1. Launch the server.
2. Lookup the Python process activity.
2. Launch the client.

Before the change, the Python process memory would go up indefinitely, at a pace that varies from system to system. Eg I start from 16MB and get to >20MB and up in just a few seconds.

After the change, memory remains stable. Eg I see a stable 17.1 MB despite the client hitting the server with pings.

### DIFF
--- a/uvicorn/_handlers/http.py
+++ b/uvicorn/_handlers/http.py
@@ -81,6 +81,11 @@ async def handle_http(
     data = reader._buffer  # type: ignore
     if data:
         protocol.data_received(data)
+    else:
+        # Client already disconnected.
+        # E.g. they sent a TCP ping health check.
+        # Fine, let's disconnect too.
+        transport.close()
 
     # Let the transport run in the background. When closed, this future will complete
     # and we'll exit here.
--- a/uvicorn/_handlers/http.py
+++ b/uvicorn/_handlers/http.py
@@ -79,13 +79,7 @@ async def handle_http(
     # yet: all data that the client might have already sent since the connection has
     # been established is in the `_buffer`.
     data = reader._buffer  # type: ignore
-    if data:
-        protocol.data_received(data)
-    else:
-        # Client already disconnected.
-        # E.g. they sent a TCP ping health check.
-        # Fine, let's disconnect too.
-        transport.close()
+    protocol.data_received(data)
 
     # Let the transport run in the background. When closed, this future will complete
     # and we'll exit here.
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -223,6 +223,11 @@ class H11Protocol(asyncio.Protocol):
                 self.cycle.more_body = False
                 self.cycle.message_event.set()
 
+            elif event_type is h11.ConnectionClosed:
+                if not self.transport.is_closing():
+                    self.transport.close()
+                break
+
     def handle_upgrade(self, event):
         upgrade_value = None
         for name, value in self.headers:
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -137,6 +137,9 @@ class HttpToolsProtocol(asyncio.Protocol):
         except httptools.HttpParserUpgrade:
             self.handle_upgrade()
 
+        if data == b"" and not self.transport.is_closing():
+            self.transport.close()
+
     def handle_upgrade(self):
         upgrade_value = None
         for name, value in self.headers:


Kludex (dismissed):

florimondmanca:
Seems like there's more subtlety to it than I thought.

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -137,6 +137,9 @@ def data_received(self, data):
         except httptools.HttpParserUpgrade:
             self.handle_upgrade()
 
+        if data == b"" and not self.transport.is_closing():
+            self.transport.close()
+
```
euri10, line 142:
```suggestion
        if data == b"":
            self.on_response_complete()
```

euri10: seems like this pass tests and fix the issue if I'm correct

uvicorn/_handlers/http.py
```diff
@@ -79,8 +79,7 @@ def retrieve_exception(task: asyncio.Task) -> None:
     # yet: all data that the client might have already sent since the connection has
     # been established is in the `_buffer`.
     data = reader._buffer  # type: ignore
-    if data:
-        protocol.data_received(data)
+    protocol.data_received(data)
```
euri10, line 82:
```suggestion
    if isinstance(protocol, H11Protocol):
        if data:
            protocol.data_received(data)
    elif isinstance(protocol, HttpToolsProtocol):
        protocol.data_received(data)
```

uvicorn/protocols/http/h11_impl.py
```diff
@@ -223,6 +223,11 @@ def handle_events(self):
                 self.cycle.more_body = False
                 self.cycle.message_event.set()
 
+            elif event_type is h11.ConnectionClosed:
+                if not self.transport.is_closing():
+                    self.transport.close()
+                break
```
euri10, line 229:
```suggestion

```

uvicorn/_handlers/http.py
florimondmanca: I guess we should be moving that `if data:` check to `H11Protocol` instead, then.

euri10: the thing is that this `b""` case is already handled by h11 in it's `receive_data` we use in our protocol's `data_received`

see https://gitlab.com/kalilinux/packages/python-h11/-/blob/kali/master/h11/_connection.py#L303-348

so I dont know

Kludex:
Replaced by #1332

PR #1248, opened 2021-11-12
Allow dependency injection Config and Server

abersheeran:
Related links:
- https://github.com/encode/uvicorn/issues/451
- https://github.com/encode/uvicorn/pull/1091
- https://github.com/encode/uvicorn/pull/1012
- https://github.com/encode/uvicorn/issues/742
- https://github.com/encode/uvicorn/issues/1151

To solve #742 , a short and simple code:

```python
config = Config("example:app", host="127.0.0.1", port=5000, log_level="info")

thread = threading.Thread(target=uvicorn.run_server, args=(server,))
thread.start()
```

To solve #1151 

```python
async with app_factory() as app:
    config = Config(app)
    uvicorn.run_server(server)
```

To solve #451 and #1091 

```python
should_exit = False

class CustomServer(uvicorn.Server):
    def handle_exit(self, sig, frame):
        global should_exit
        should_exit = True
        return super().handle_exit(sig, frame)

class App:
    ...

        """Handle http messages"""

        message = await receive()

        while message['more_body']:
            message = await receive()

        await send({
            'status': 200,
            'headers': [
            ]
        })

        while not should_exit:
            ............ # send some message

if __name__ == '__main__':
    config = Config(App())
    server = CustomServer(config=config)
    uvicorn.run_server(server)
```

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -444,6 +444,13 @@ class Config:
 
         self.lifespan_class = import_from_string(LIFESPAN[self.lifespan])
 
+        if (self.reload or self.workers > 1) and not isinstance(self.app, str):
+            logger.warning(
+                "You must pass the application as an import string to enable"
+                " 'reload' or 'workers'."
+            )
+            sys.exit(1)
+
         try:
             self.loaded_app = import_from_string(self.app)
         except ImportFromStringError as exc:
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -429,7 +429,7 @@ def main(
 def run(app: typing.Union[ASGIApplication, str], **kwargs: typing.Any) -> None:
     config = Config(app, **kwargs)
     server = Server(config=config)
-    serve(config, server)
+    run_server(config, server)
 
 
 def run_server(config: Config, server: Server) -> None:
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -428,20 +428,11 @@ def main(
 
 def run(app: typing.Union[ASGIApplication, str], **kwargs: typing.Any) -> None:
     config = Config(app, **kwargs)
-
-    if (config.reload or config.workers > 1) and not isinstance(app, str):
-        logger = logging.getLogger("uvicorn.error")
-        logger.warning(
-            "You must pass the application as an import string to enable 'reload' or "
-            "'workers'."
-        )
-        sys.exit(1)
-
     server = Server(config=config)
     serve(config, server)
 
 
-def serve(config: Config, server: Server) -> None:
+def run_server(config: Config, server: Server) -> None:
     if config.should_reload:
         sock = config.bind_socket()
         ChangeReload(config, target=server.run, sockets=[sock]).run()
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -429,10 +429,11 @@ def main(
 def run(app: typing.Union[ASGIApplication, str], **kwargs: typing.Any) -> None:
     config = Config(app, **kwargs)
     server = Server(config=config)
-    run_server(config, server)
+    run_server(server)
 
 
-def run_server(config: Config, server: Server) -> None:
+def run_server(server: Server) -> None:
+    config = server.config
     if config.should_reload:
         sock = config.bind_socket()
         ChangeReload(config, target=server.run, sockets=[sock]).run()
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -429,7 +429,7 @@ def main(
 def run(app: typing.Union[ASGIApplication, str], **kwargs: typing.Any) -> None:
     config = Config(app, **kwargs)
     server = Server(config=config)
-    serve(config, Server)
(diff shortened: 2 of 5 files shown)


Kludex (changes requested):
The examples on the description shows `uvicorn.serve`, but that is not available yet. We need to add `from uvicorn.main import serve` on the `__init__.py`.

Kludex:
@euri10 Do you have something against this or can we we merge it?

euri10:
> @euri10 Do you have something against this or can we we merge it?

nothing against, that may be a breaking change though for some people, @florimondmanca will know for sure, so maybe hold it onto a minor bump if patches need to be released in between

uvicorn/main.py
```diff
@@ -438,6 +437,11 @@ def run(app: typing.Union[ASGIApplication, str], **kwargs: typing.Any) -> None:
         )
         sys.exit(1)
 
+    server = Server(config=config)
+    serve(config, server)
+
+
+def serve(config: Config, server: Server) -> None:
```
HansBrende: Out of curiosity, why not simplify the signature here to:
```python
def serve(server: Server) -> None:
    config = server.config
    # ... carry on as before
```

One other question: why is the `isinstance(app, str)` check happening in `run` rather than `serve`? Seems like this check is now missing if `serve` is called independently of `run`. Perhaps the check should simply be moved to the `Config` constructor?

One final observation: it seems like the method-naming conventions here are getting a bit convoluted & confusing, with: `uvicorn.run` CALLS `uvicorn.serve` CALLS `Server.run` CALLS `Server.serve` (where `Server.serve` and `uvicorn.serve` don't share any particular distinguishing commonality!) If the only reason for calling this method `serve` is because the word `run` has already been used... perhaps it might make sense to keep it more consistent with `uvicorn.run` and call it something like `run_server` instead? (Esp. since the method is ultimately calling `Server.run`, _not_ `Server.serve`!)

HansBrende: After further research, it looks like #1012 has the right idea, as it resolves all 3 of the above concerns. I wouldn't change anything from that PR... why not just use that one?

florimondmanca:
Well I think the general idea is quite nice.

Tho as @HansBrende pointed out, seems there’s a distinction and decision to take about whether the `serve()` utility should include the multiprocessing/reload code branches or not.

k don’t have much more background to tell which would best address the issues this aims at, though. Just thought I’d raise that distinction as something that should be discussed before merging.

In a all cases, one of this or #1012 should end up merged and the other one closed, right?

florimondmanca:
I also agree with @HansBrende that the `uvicorn.serve` naming would be a bit confusing, as eg `Server.serve()` is while this is not, and this is actually calling into `server.run()`. I’d also be in favor of something like `uvicorn.run_server()` like in #1012.

uvicorn/main.py
abersheeran: I just want `Config` to be passed independently as a parameter. I think both PRs are fine. If yours is merged, I will close this one.

abersheeran: > One other question: why is the isinstance(app, str) check happening in run rather than serve? Seems like this check is now missing if serve is called independently of run. Perhaps the check should simply be moved to the Config constructor?

It is a good idea to move this check to Config. I will do it as soon as possible.

HansBrende: > I just want Config to be passed independently as a parameter.

Right, that's my question: _why_ do you want `Config` passed independently? Shouldn't you (always) use the same config object that initialized the`Server` to run the server? If so, you could simply do `config = server.config`.

abersheeran: There is no deeper reason, I just intuitively think this might be better. 😀

florimondmanca: I agree we should probably go for `run_server(server)`, as it would be awkward and bug-prone to pass a `config` that doesn't match the server's, right?

abersheeran: After thinking about it carefully, you are right. This time my intuition went wrong.

uvicorn/config.py
```diff
@@ -444,6 +444,13 @@ def load(self) -> None:
 
         self.lifespan_class = import_from_string(LIFESPAN[self.lifespan])
 
+        if (self.reload or self.workers > 1) and not isinstance(self.app, str):
```
HansBrende, line 447: Shouldn't this check go in the constructor? By the time we get here, we are already in the subprocess I believe...

abersheeran: I think this check should be combined with `import_from_string(self.app)`.

HansBrende: Why? Isn't it preferable to "fail fast" in the main process rather than failing later in possibly N subprocesses?

abersheeran: When the user provides a string, checks the type in one place, but tries to import it in another far away, don't you think this is unreasonable?

If it is for this purpose, then `import_from_string(self.app)` should also "fail fast".

Kludex:
Let's continue on #1012. It's easier, yep.

PR #1201, opened 2021-09-29
Version 0.16.0

Kludex:
This release will complete the milestone [Version 0.16.0](https://github.com/encode/uvicorn/milestone/5). 

## Missing

- https://github.com/encode/uvicorn/pull/1173
- https://github.com/encode/uvicorn/pull/1194
- https://github.com/encode/uvicorn/pull/1159
- https://github.com/encode/uvicorn/pull/1110
- https://github.com/encode/uvicorn/pull/1114
- https://github.com/encode/uvicorn/pull/1177
- https://github.com/encode/uvicorn/pull/1128
- https://github.com/encode/uvicorn/pull/1165

## Anything else?

Let me know if there are PRs missing.

## Draft release?

Still not created. I'll create later on.

PR #1252, opened 2021-11-15
Fix 1230; don't call ensure_open in WebSocketProtocol.asgi_receive

### LINKED ISSUE #1230: Bug: calling `WebSocketProtocol.asgi_receive` returns close frame even if there are data messages before close frame in read queue
### Checklist

### Describe the bug

Once a client sends a close frame, calling `WebSocketProtocol.asgi_receive` returns `{"type": "websocket.disconnect", "code": exc.code}`, even if there are unread messages in the server's read queue that we sent **before** the close frame.

### To reproduce

The easiest way for me to repro is to use an ASGI application framework written on top of `uvicorn`, e.g. FastAPI.

- Install FastAPI and create a module `main.py`
- Run the websocket server: `uvicorn main:app --reload --host 0.0.0.0 --port 8001`
- Open the browser and create a websocket connection to this test endpoint

`main.py`

```py
import asyncio
from fastapi import FastAPI


app = FastAPI()


@app.websocket("/echo")
async def echo_ws(ws: WebSocket):
    await ws.accept()
    await asyncio.sleep(1)

    data = await ws.receive_bytes()
    print(data)
```

Paste the following into the browser console:

```js
var socket = new WebSocket('ws://localhost:8001/atom_ws/echo')
socket.onopen = () => {
  console.log('socket opened')
  socket.send('first')
  socket.send('second')
  socket.close()
}
```

### Expected behavior

The call to `print(data)` should print `first`, and not raise a `starlette.websockets.WebSocketDisconnect` exception.

### Actual behavior

`starlette.websockets.WebSocketDisconnect` is raised on the first read, even though messages were successfully sent to the server before the close frame, and these messages are in the connection's read queue.

### Debugging material

Logs when running the server code from above:

Screen Shot 2021-11-01 at 9 24 46 PM

If you instead run a simple websocket server written with code directly from the `websockets` library, [as suggested in their docs](https://websockets.readthedocs.io/en/3.0/intro.html#basic-example), you don't have this problem:

```py
import asyncio
import websockets


async def echo(ws, path):
    print(f"accepted connection to path {path}")
    await asyncio.sleep(1)
    # By this time client has already closed connection

    # In uvicorn, `await ws.ensure_open()` is called before recv; this is the bug
    data = await ws.recv()
    print(data)  # Prints first
    data = await ws.recv()
    print(data)  # Prints second

    # The next `recv` call raises `websockets.exceptions.ConnectionClosedError`, because it reads the close frame
    data = await ws.recv()

start_server = websockets.serve(echo, 'localhost', 8001)

asyncio.get_event_loop().run_until_complete(start_server)
asyncio.get_event_loop().run_forever()
```

We can see that the `first` and `second` messages are received, even though the client sent the close frame before the server tried to read any messages:

Screen Shot 2021-11-01 at 9 30 37 PM

### Environment

- MacOS 10.14.6 / Python 3.8.5 / uvicorn 0.15.0 (this bug is still present in 0.16.0, and in the latest commit in `master`)
- Can also repro on Ubuntu 20.04

- The exact command you're running uvicorn with, all flags you passed included: `uvicorn main:app --reload --host 0.0.0.0 --port 8001`

### Additional context

The bug is caused by this line of code: https://github.com/encode/uvicorn/blob/48edc940522a3d0d7529922a23ac019eeb53f629/uvicorn/protocols/websockets/websockets_impl.py#L286-L286

```py
await self.ensure_open()
data = await self.recv()
```

This change was made in this commit: https://github.com/encode/uvicorn/commit/9a3040c9cd56844631b28631acd5862b5a4eafdd

`uvicorn` depends on `websockets` under the hood, but it shouldn't be calling `ensure_open` before calling `recv`, because `ensure_open` raises an exception if a close frame has been sent by the client **even if there are earlier unread messages in the read queue**.

I'm not sure what the intent of that line of code is, but the server shouldn't raise a "connection closed" exception on read until the actual close frame is read. Otherwise, neither the client nor the server has any way of knowing that data that was successfully sent from the client to the server was ignored by the server.

kylebebak:
Closes https://github.com/encode/uvicorn/issues/1230. Reverts most of https://github.com/encode/uvicorn/commit/9a3040c9cd56844631b28631acd5862b5a4eafdd

We can't call `WebSocketCommonProtocol.ensure_open` before calling `WebSocketCommonProtocol.recv`, because there may be unread frames in the server's read queue that were sent by the client **before** the client sent the close frame. [See this comment from the websockets source code](https://github.com/aaugustin/websockets/blob/ed9a7b446c7147f6f88dbeb1d86546ad754e435e/src/websockets/legacy/protocol.py#L521-L525)

`WebSocketCommonProtocol.ensure_open` raises an exception and prevents the server from reading these unread frames. Neither the client, nor application code using uvicorn, has any way of knowing that frames successfully sent by the client to the server's read queue, before any close frame was sent, will never be read by the server.

The spec is clear on this: https://datatracker.ietf.org/doc/html/rfc6455#section-1.4

> After sending a control frame indicating the connection should be closed, a peer does not send any further data; after receiving a control frame indicating the connection should be closed, a peer discards any further data received.

The current behavior in uvicorn is rather "a peer discards any further data received, and any data previously received"

Also worth noting, https://github.com/aaugustin/websockets/pull/670 was eventually closed without any changes being made; in other words, there was no bug in `websockets`, and the maintainer agrees that calling `await self.ensure_open()` in `recv` is incorrect

-----

I tested `uvicorn` after making these changes, and the behavior is as expected; the server can read frames in the read queue after a close frame is sent by the client, as long as those frames were sent before the close frame

### DIFF
--- a/uvicorn/protocols/websockets/websockets_impl.py
+++ b/uvicorn/protocols/websockets/websockets_impl.py
@@ -261,7 +261,6 @@ class WebSocketProtocol(_LoggerMixin, websockets.WebSocketServerProtocol):
 
             elif message_type == "websocket.close":
                 code = message.get("code", 1000)
-                self.close_code = code  # for WebSocketServerProtocol
                 reason = message.get("reason", "")
                 await self.close(code, reason)
                 self.closed_event.set()
@@ -285,11 +284,10 @@ class WebSocketProtocol(_LoggerMixin, websockets.WebSocketServerProtocol):
         await self.handshake_completed_event.wait()
 
         if self.closed_event.is_set():
-            # If the client disconnected: WebSocketServerProtocol set self.close_code.
+            # If client disconnected, use WebSocketServerProtocol.close_code property.
             # If the handshake failed or the app closed before handshake completion,
             # use 1006 Abnormal Closure.
-            code = getattr(self, "close_code", 1006)
-            return {"type": "websocket.disconnect", "code": code}
+            return {"type": "websocket.disconnect", "code": self.close_code or 1006}
 
         try:
             data = await self.recv()
--- a/uvicorn/protocols/websockets/websockets_impl.py
+++ b/uvicorn/protocols/websockets/websockets_impl.py
@@ -261,6 +261,7 @@ class WebSocketProtocol(_LoggerMixin, websockets.WebSocketServerProtocol):
 
             elif message_type == "websocket.close":
                 code = message.get("code", 1000)
+                self.close_code = code  # for WebSocketServerProtocol
                 reason = message.get("reason", "")
                 await self.close(code, reason)
                 self.closed_event.set()
@@ -282,10 +283,18 @@ class WebSocketProtocol(_LoggerMixin, websockets.WebSocketServerProtocol):
             return {"type": "websocket.connect"}
 
         await self.handshake_completed_event.wait()
+
+        if self.closed_event.is_set():
+            # If the client disconnected: WebSocketServerProtocol set self.close_code.
+            # If the handshake failed or the app closed before handshake completion,
+            # use 1006 Abnormal Closure.
+            code = getattr(self, "close_code", 1006)
+            return {"type": "websocket.disconnect", "code": code}
+
         try:
-            await self.ensure_open()
             data = await self.recv()
         except websockets.ConnectionClosed as exc:
+            self.closed_event.set()
             return {"type": "websocket.disconnect", "code": exc.code}
 
         msg = {"type": "websocket.receive"}
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -539,3 +539,36 @@ async def test_send_binary_data_to_server_bigger_than_default(
             with pytest.raises(websockets.ConnectionClosedError) as e:
                 data = await send_text("ws://127.0.0.1:8000")
             assert e.value.code == expected_result
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
+@pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
+async def test_server_reject_connection(ws_protocol_cls, http_protocol_cls):
+    async def app(scope, receive, send):
+        assert scope["type"] == "websocket"
+
+        # Pull up first recv message.
+        message = await receive()
+        assert message["type"] == "websocket.connect"
+
+        # Reject the connection.
+        await send({"type": "websocket.close"})
+        # -- At this point websockets' recv() is unusable. --
+
+        # This doesn't raise `TypeError`:
+        # See https://github.com/encode/uvicorn/issues/244
+        message = await receive()
+        print("message", message)
+        assert message["type"] == "websocket.disconnect"
+
+    async def websocket_session(url):
+        try:
+            async with websockets.connect(url):
+                pass
+        except Exception:
+            pass
+
+    config = Config(app=app, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off")
+    async with run_server(config):
+        await websocket_session("ws://127.0.0.1:8000")
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -1,3 +1,5 @@
+import asyncio
+
 import httpx
 import pytest
 
@@ -559,7 +561,6 @@ async def test_server_reject_connection(ws_protocol_cls, http_protocol_cls):
         # This doesn't raise `TypeError`:
         # See https://github.com/encode/uvicorn/issues/244
         message = await receive()
-        print("message", message)
         assert message["type"] == "websocket.disconnect"
 
     async def websocket_session(url):
@@ -572,3 +573,35 @@ async def test_server_reject_connection(ws_protocol_cls, http_protocol_cls):
     config = Config(app=app, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off")
     async with run_server(config):
         await websocket_session("ws://127.0.0.1:8000")
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
+@pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
+async def test_server_can_read_messages_in_buffer_after_close(
+    ws_protocol_cls, http_protocol_cls
+):
+    frames = []
+
+    class App(WebSocketResponse):
+        async def websocket_connect(self, message):
+            await self.send({"type": "websocket.accept"})
+            # Ensure server doesn't start reading frames from read buffer until
+            # after client has sent close frame, but server is still able to
+            # read these frames
+            await asyncio.sleep(0.2)
+
+        async def websocket_receive(self, message):
+            frames.append(message.get("bytes"))
+
+    async def send_text(url):
+        async with websockets.connect(url) as websocket:
+            await websocket.send(b"abc")
+            await websocket.send(b"abc")
+            await websocket.send(b"abc")
+
+    config = Config(app=App, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off")
+    async with run_server(config):
+        await send_text("ws://127.0.0.1:8000")
+
+    assert frames == [b"abc", b"abc", b"abc"]


euri10:
ok, cant say I'm following everything here as it's been quite a while, 
question is does this PR keeps intact the original intended fix for #244 ?

euri10:
second thing, this may be a good idea to come up with tests for this

kylebebak:
Hey @euri10 

Indeed, this PR keeps the original fix to #244 , written by @adamhooper and outlined in this PR: https://github.com/encode/uvicorn/pull/704

Small changes were made Adam Hooper's code because the websockets library has changed since then; `WebSocketServerProtocol.close_code` is now a property, not an attribute, so we can't assign to it.

The `TypeError` bug that @florimondmanca  was able to reproduce, [described here](https://github.com/encode/uvicorn/issues/244#issuecomment-740882314), can't be reproduced with this code. **I added a test to ensure this bug doesn't reappear**.

I also added a test that ensures the bug I found in #1230 doesn't come back. It's worth noting that this test immediately passed for the `WSProtocol` class without making any changes. It's good to know that `WSProtocol` doesn't have the bug in `WebSocketProtocol`.

kylebebak:
Hey @euri10 @Kludex 

I think this is ready to go, but please let me know if you think there's anything else we should add to it.

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -282,10 +282,17 @@ async def asgi_receive(self):
             return {"type": "websocket.connect"}
 
         await self.handshake_completed_event.wait()
+
+        if self.closed_event.is_set():
+            # If client disconnected, use WebSocketServerProtocol.close_code property.
+            # If the handshake failed or the app closed before handshake completion,
+            # use 1006 Abnormal Closure.
+            return {"type": "websocket.disconnect", "code": self.close_code or 1006}
```
euri10, line 290: looks good to me overall @kylebebak  thanks a lot for that, one small comment.

I noticed that if in `tests.protocols.test_websocket.test_server_reject_connection`, should you add `config = Config(app=app, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off", log_level="trace")` to get fine-grained output, then the close code is None for the `WSProtocol` and 1006 for the `WebsocketProtocol`, why this is the case, shouldn't both be aligned ?

`WebsocketProtocol`
```
PASSED [100%]INFO:     Started server process [346460]
TRACE:    127.0.0.1:49858 - HTTP connection made
TRACE:    127.0.0.1:49858 - Upgrading to WebSocket
DEBUG:    = connection is CONNECTING
TRACE:    127.0.0.1:49858 - WebSocket connection made
DEBUG:    < GET / HTTP/1.1
DEBUG:    < host: 127.0.0.1:8000
DEBUG:    < upgrade: websocket
DEBUG:    < connection: Upgrade
DEBUG:    < sec-websocket-key: DYANNbdkc+2NIoMN3KZIkA==
DEBUG:    < sec-websocket-version: 13
DEBUG:    < sec-websocket-extensions: permessage-deflate; server_max_window_bits=12; client_max_window_bits=12
DEBUG:    < user-agent: Python/3.9 websockets/10.0
TRACE:    127.0.0.1:49858 - ASGI [1] Started scope={'type': 'websocket', 'asgi': {'version': '3.0', 'spec_version': '2.1'}, 'scheme': 'ws', 'server': ('127.0.0.1', 8000), 'client': ('127.0.0.1', 49858), 'root_path': '', 'path': '/', 'raw_path': '/', 'query_string': b'', 'headers': '<...>', 'subprotocols': []}
TRACE:    127.0.0.1:49858 - ASGI [1] Receive {'type': 'websocket.connect'}
TRACE:    127.0.0.1:49858 - ASGI [1] Send {'type': 'websocket.close'}
INFO:     ('127.0.0.1', 49858) - "WebSocket /" 403
DEBUG:    > HTTP/1.1 403 Forbidden
DEBUG:    > Date: Wed, 24 Nov 2021 07:36:20 GMT
DEBUG:    > Server: Python/3.9 websockets/10.0
DEBUG:    > Content-Length: 0
DEBUG:    > Content-Type: text/plain
DEBUG:    > Connection: close
DEBUG:    > [body] (0 bytes)
INFO:     connection failed (403 Forbidden)
DEBUG:    x closing TCP connection
TRACE:    127.0.0.1:49858 - WebSocket connection lost
DEBUG:    = connection is CLOSED
TRACE:    127.0.0.1:49858 - ASGI [1] Receive {'type': 'websocket.disconnect', 'code': 1006}
TRACE:    127.0.0.1:49858 - ASGI [1] Completed
INFO:     connection closed
INFO:     Shutting down
```

 `WSProtocol`
```
PASSED [ 75%]INFO:     Started server process [346460]
TRACE:    127.0.0.1:49842 - HTTP connection made
TRACE:    127.0.0.1:49842 - Upgrading to WebSocket
TRACE:    127.0.0.1:49842 - WebSocket connection made
TRACE:    127.0.0.1:49842 - ASGI [1] Started scope={'type': 'websocket', 'asgi': {'version': '3.0', 'spec_version': '2.1'}, 'http_version': '1.1', 'scheme': 'ws', 'server': ('127.0.0.1', 8000), 'client': ('127.0.0.1', 49842), 'root_path': '', 'path': '/', 'raw_path': '/', 'query_string': b'', 'headers': '<...>', 'subprotocols': []}
TRACE:    127.0.0.1:49842 - ASGI [1] Receive {'type': 'websocket.connect'}
TRACE:    127.0.0.1:49842 - ASGI [1] Send {'type': 'websocket.close'}
INFO:     ('127.0.0.1', 49842) - "WebSocket /" 403
TRACE:    127.0.0.1:49842 - ASGI [1] Receive {'type': 'websocket.disconnect', 'code': None}
TRACE:    127.0.0.1:49842 - ASGI [1] Completed
TRACE:    127.0.0.1:49842 - WebSocket connection lost
INFO:     Shutting down
```

euri10 (changes requested):

kylebebak:
Hey @euri10 

I just rolled back the code in `websockets_impl.py` on my machine to how it was in [commit `99acac5`](https://github.com/encode/uvicorn/commits/99acac5), from 2 months ago, added a print to `test_server_reject_connection`, and ran it again.

In `test_server_reject_connection`, I observed the same behavior you just did, a close code of `None` for `WSProtocol` and `1006` for `WebsocketProtocol`.

In other words, this is preexisting behavior, and is not related to calling `ensure_open` in `WebSocketProtocol.asgi_receive`.

I don't know the websocket spec well enough to know whether the correct close code in this situation is `None` or `1006`, or whether the spec even cares, but in any case we shouldn't gate the fix in this PR on unrelated, preexisting behavior which may or may not be an issue.

Here's what the tests printed out:

```
tests/protocols/test_websocket.py::test_server_reject_connection[H11Protocol-WSProtocol] INFO:     Started server process [15640]
INFO:     ('127.0.0.1', 52429) - "WebSocket /" 403
{'type': 'websocket.disconnect', 'code': None}
INFO:     Shutting down
PASSED
tests/protocols/test_websocket.py::test_server_reject_connection[H11Protocol-WebSocketProtocol] INFO:     Started server process [15640]
INFO:     ('127.0.0.1', 52430) - "WebSocket /" 403
INFO:     connection failed (403 Forbidden)
{'type': 'websocket.disconnect', 'code': 1006}
INFO:     connection closed
INFO:     Shutting down
PASSED
tests/protocols/test_websocket.py::test_server_reject_connection[HttpToolsProtocol-WSProtocol] INFO:     Started server process [15640]
INFO:     ('127.0.0.1', 52431) - "WebSocket /" 403
{'type': 'websocket.disconnect', 'code': None}
INFO:     Shutting down
PASSED
tests/protocols/test_websocket.py::test_server_reject_connection[HttpToolsProtocol-WebSocketProtocol] INFO:     Started server process [15640]
INFO:     ('127.0.0.1', 52432) - "WebSocket /" 403
INFO:     connection failed (403 Forbidden)
{'type': 'websocket.disconnect', 'code': 1006}
```

euri10 (commented):
ok let's roll with this, thanks for the digging into finding the root cause of this @kylebebak
let's keep in mind that websocket close code discrepancy, not sure I wont forget since I'm not using that part of the library but never knows !

euri10 (approved):

kylebebak:
Thank you @euri10 for all the feedback and helping to get this merged!

I agree that it could be good to open an issue pointing out the discrepancy between the close codes when the server rejects the connection using `WSProtocol` vs `WebsocketProtocol`

I think this issue here has all the context we need. If you want to open an issue, we can just point to this PR

kylebebak:
@euri10 @Kludex 

One last thing... I think it would be good to release `0.15.1`, so libraries that depend on `uvicorn`, e.g. FastAPI, can take advantage of this fix

The latest release of FastAPI depends on `uvicorn[standard] (>=0.12.0,<0.16.0)`, so it would be able to pull in a fix from `0.15.1` immediately

Kludex:
@euri10 I don't see any reason to bump a minor considering what we merged, do you? 

We can close https://github.com/encode/uvicorn/pull/1201, and aim for the patch **if you think it's a patch**. If it's a minor, I think we can continue on that PR.

But... FYK, FastAPI doesn't depend on uvicorn. `fastapi[all]` extra requirements do.

euri10:
in all honesty I dont know, 0.15.0 was mid august iirc and I've been afk most of the time since so really cant say without looking more closely at what has been pushed

Kludex:
I just confirmed, it's a minor. We bumped websockets and httptools, and there are also other commits that makes the next release a minor.

I'm going to reevaluate my wish list for 0.16.0 tonight on #1201 . If you have a PR that you wish as well @euri10 , feel free to add there.

kylebebak:
Thanks for clarifying guys!

PR #1255, opened 2021-11-19
Load more than one ca_cert

brussee:
`ca_certs` is plural, but currently does not accept a list of CA certificate paths.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import socket
+import ssl
 import sys
 import typing
 from copy import deepcopy
@@ -345,7 +346,9 @@ def test_ssl_config_ca_cert(
     )
     config.load()
 
-    assert len(config.ssl.get_ca_certs()) == 1
+    assert isinstance(config.ssl, ssl.SSLContext)
+    config_ssl: ssl.SSLContext = config.ssl
+    assert len(config_ssl.get_ca_certs()) == 1
 
 
 def test_ssl_config_ca_certs(
@@ -361,7 +364,9 @@ def test_ssl_config_ca_certs(
     )
     config.load()
 
-    assert len(config.ssl.get_ca_certs()) == 1
+    assert isinstance(config.ssl, ssl.SSLContext)
+    config_ssl: ssl.SSLContext = config.ssl
+    assert len(config_ssl.get_ca_certs()) == 1
 
 
 def asgi2_app(scope: Scope) -> typing.Callable:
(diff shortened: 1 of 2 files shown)


euri10:
I'm dubious about this PR, I'm no SSL expert though but this seems like this link explains it quite well : pass a directory containing your CAs : https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations
Feel free to reopen if that's not the case

brussee:
You are referring to the 2nd argument of `load_verify_locations` but the current implementation gives only the 1st positional argument:
> The cafile string, if present, is the path to a file of concatenated CA certificates in PEM format.

My PR proposes to allow a list, such that instead of concatenating all CA certs beforehand, you can provide a list of certificate files in different locations. This is arguably more flexible than requiring all of the files to be in a single directory.

@euri10 please re-open if you agree/want to re-consider

euri10:
Happy to reconsider then effectively, but: 
1. add tests
2. change the positional only arg we currently have to kwargs so that we're clear on what location we use

brussee:
@euri10 done

brussee:
@euri10 updated formatting to make the linter happy

brussee:
@euri10 Would you like me to squash before merging?

euri10:
No need it's automatic on the repo iirc

tests/test_config.py
```diff
@@ -332,6 +333,42 @@ def test_ssl_config_combined(tls_certificate_key_and_chain_path: str) -> None:
     assert config.is_ssl is True
 
 
+def test_ssl_config_ca_cert(
```
euri10, line 336: appart from the name what is the difference between those 2 tests ?

tests/test_config.py
```diff
@@ -332,6 +333,42 @@ def test_ssl_config_combined(tls_certificate_key_and_chain_path: str) -> None:
...
+        app=asgi_app,
+        ssl_certfile=tls_ca_certificate_pem_path,
+        ssl_keyfile=tls_ca_certificate_private_key_path,
+        ssl_ca_certs=tls_ca_certificate_pem_path,
+    )
+    config.load()
+
+    assert isinstance(config.ssl, ssl.SSLContext)
+    config_ssl: ssl.SSLContext = config.ssl
+    assert len(config_ssl.get_ca_certs()) == 1
```
euri10, line 351: This doesn't test the added feature ie more than one path/str or I'm missing something ?
Adding a cert in `test_ssl` would certainly be the place you'd do that

euri10 (changes requested):

Kludex:
@brussee Are you still interested on this PR? If so, would you mind checking @euri10 's comments? 🙏

brussee:
I switched from the approach in this PR to creating and injecting a fully configured SSLContext directly.
Instead of maintaining code that wraps the SSLContext class like now, the alternative seems like a more maintainable approach.

Issue #1220, opened 2021-10-18
Is there a reason for overriding the default asyncio policy on Windows?


polys:
### Checklist

- [ ] There are no similar issues or pull requests for this yet.
- [ ] I discussed this idea on the [community chat](https://gitter.im/encode/community) and feedback is positive.

### Is your feature related to a problem? Please describe.

While developing on Windows, my app being served by `uvicorn` keeps raising `ValueError: too many file descriptors in select()`, which I understand is related to the suboptimal `asyncio.WindowsSelectorEventLoopPolicy`.

[Python 3.8 changed the default to `ProactorEventLoop` on Windows](https://docs.python.org/3/library/asyncio-policy.html#asyncio.DefaultEventLoopPolicy) but [your code seems to override this](https://github.com/encode/uvicorn/blob/master/uvicorn/loops/asyncio.py#L6-L7).

Is there a reason for this?

## Describe the solution you would like.

Ideally, all the `too many file descriptors in select()` exceptions go away.

## Additional context

This isn't an issue at all when running my app in a Linux container.

sekrause:
Looking at the commit history the `ProactorEventLoop` was disabled to fix issue #529.

It probably wasn't understood that using `SelectorEventLoop` basically makes uvicorn completely useless for scalable applications on Windows because you will be limited to only 512 socket connections: https://docs.python.org/3.10/library/asyncio-platforms.html#windows

So simply disabling the `ProactorEventLoop` (the only scalable event loop on Windows) is generally not a good idea.

sekrause:
@polys For now you can work around this issue by starting uvicorn programmatically in your own loop, for example this will give you a `ProactorEventLoop` in Python >= 3.8:

```python
from asyncio import get_event_loop, new_event_loop
from fastapi import FastAPI
from uvicorn import Config, Server

app = FastAPI()

@app.get("/")
async def read_main():
    return f'Running loop: {get_event_loop()}'

loop = new_event_loop()
config = Config(app=app)
server = Server(config)
loop.run_until_complete(server.serve())
```

However I still think that `uvicorn` should use the `ProactorEventLoop` whenever possible. For example if the `--reload` option is the problem (issue #529), only switch to the `SelectorEventLoop` when `--reload` is enabled on Windows, and issue a warning in the logs that `--reload` should never be used in production on Windows.

polys:
Thanks for the suggestion 😀

[...]

PR #535, opened 2019-12-20
Iocp drop

euri10:
Closes #529

Removes iocp entirely and uses SelectorEventLoop for the asyncio auto default

PR #1257, opened 2021-11-20
Override default asyncio event loop with reload only

jammer87:
#535 changed the default event loop in windows to the selector event loop to avoid an issue with the --reload flag (see #529).

This can cause the issue described in #1220 as well as make the asyncio.subprocess.create_subprocess_* calls fail.

This PR overrides the event loop only with reload configured on windows and logs a warning as suggested in this [comment](https://github.com/encode/uvicorn/issues/1220#issuecomment-974027158).

### DIFF
--- a/uvicorn/loops/asyncio.py
+++ b/uvicorn/loops/asyncio.py
@@ -1,12 +1,11 @@
 import asyncio
 import logging
 import sys
-from typing import Any
 
 logger = logging.getLogger("uvicorn.error")
 
 
-def asyncio_setup(reload: bool = False, **_: Any) -> None:  # pragma: no cover
+def asyncio_setup(reload: bool = False) -> None:  # pragma: no cover
     if sys.version_info >= (3, 8) and sys.platform == "win32" and reload:
         logger.warning("The --reload flag should not be used in production on Windows.")
         asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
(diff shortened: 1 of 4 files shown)


uvicorn/loops/asyncio.py
```diff
@@ -1,7 +1,12 @@
...
+import logging
 import sys
+from typing import Any
 
+logger = logging.getLogger("uvicorn.error")
 
-def asyncio_setup() -> None:  # pragma: no cover
-    if sys.version_info >= (3, 8) and sys.platform == "win32":
+
+def asyncio_setup(reload: bool = False, **_: Any) -> None:  # pragma: no cover
```
euri10: is there really a need to declare unused kwargs here ?

uvicorn/loops/auto.py
```diff
@@ -1,11 +1,14 @@
-def auto_loop_setup() -> None:
+from typing import Any
+
+
+def auto_loop_setup(**kwdargs: Any) -> None:
```
euri10: can't we use an explicit reload argument here and below or there is something I dont get ?

euri10 (changes requested):

uvicorn/loops/asyncio.py
jammer87: @euri10 No. I was unnecessarily avoiding an unused argument linter error.

Updated to explicit reload keyword argument in the next revision.

euri10 (approved):
thnkas @jammer87

PR #1267, opened 2021-11-24
fix reload_dirs

### LINKED ISSUE #1266: Passing `reload_dirs` as a string instead of a list
@Roang-zero1 @euri10 @Kludex this PR introduced a regression and deleted the test for it (`test_reload_dir_is_set()`). Previously you could pass `reload_dirs` a string instead of a list. Now:
```python
Config('myapp:app', reload=True, reload_dirs='src')
```
prints:
```
WARNING:  Provided reload directories ['r', 'c', 's'] did not contain valid directories, watching current working directory.
```
https://github.com/encode/uvicorn/pull/820/files#diff-6982a5e270879281f9182303a9da9776148857b46325afb230e50ef8eb3264c9R294
(Of course, `set` expects an iterable... and strings are iterables of chars.)

_Originally posted by @HansBrende in https://github.com/encode/uvicorn/issues/820#issuecomment-977957988_

HansBrende:
Fixes #1266

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -180,7 +180,15 @@ def resolve_reload_patterns(
 
     directories = list(set(directories).difference(set(children)))
 
-    return (list(set(patterns)), directories)
+    return list(set(patterns)), directories
+
+
+def _normalize_dirs(dirs: Union[List[str], str, None]) -> List[str]:
+    if dirs is None:
+        return []
+    if isinstance(dirs, str):
+        return [dirs]
+    return list(set(dirs))
 
 
 class Config:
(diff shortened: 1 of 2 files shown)


uvicorn/config.py
```diff
@@ -180,7 +180,15 @@ def resolve_reload_patterns(
 
     directories = list(set(directories).difference(set(children)))
 
-    return (list(set(patterns)), directories)
+    return list(set(patterns)), directories
```
euri10, line 183: if the return changes (I dont get why by the way) then why mypy doesnt catch this if you've not changed the signature ?

HansBrende: @euri10 the return doesn't change. The two statements are exactly the same... the parentheses are simply redundant so I removed them (PyCharm had a warning about it saying something like "remove redundant parentheses" so I did). It's like saying `if (True):` in python... it's exactly the same as `if True:` so there's no reason to write the parentheses.

euri10: wahoo, my bad, totally misread that :face_with_spiral_eyes:

euri10 (approved):

PR #1271, opened 2021-11-27
Allow app-dir parameter on the run() function

### LINKED ISSUE #1126: allow uvicorn.run to also accept app-dir kwarg

Kludex:
Closes #1126


Kludex:
@HansBrende You recommended the solution, so feel free to review as well. :pray:

HansBrende (approved):
LGTM!

Kludex:
@euri10 ping

euri10 (approved):
I think this looks good, the only small thing that may make it better is adding a test, I realize original PR that introduced the feature doesn't have one either so we're naked here and this will hit when working again on coverage anyway.
So approved, but feel free to add tests ;)

Kludex:
I will add, thanks. ✌️

Kludex:
How would you test this?

I started with this thought:
```python
def test_app_dir(tmp_path: Path, caplog: pytest.LogCaptureFixture) -> None:
    app_dir = tmp_path / "dir" / "app_dir"
    app_file = app_dir / "main.py"
    app_dir.mkdir(parents=True)
    app_file.touch()
    app_file.write_text(
        dedent(
            """
                ...
            """
        )
    )
    run("main:app", app_dir=str(app_dir))
```
But the server will run, and we don't want that. We use `run_server` in similar cases.

After that I thought: "we'll, let's mock the `server.run`. But if I do that, we'll not test that we reached the `import_string` successfully... 

Ideas?

humrochagf:
Hi!

You can try to use the way I did for the cli tools https://github.com/encode/uvicorn/blob/master/tests/test_cli.py#L69-L78
Mock just the `Config.bind_socket` and `Server.run` to prevent the actual server from running and also from locking the socket.

Then you can check `sys.path` to validate the `app_dir` path was added to it

Kludex:
I'm going to add the test suggested by @humrochagf (thanks).

But that's not fully what I want. I want to be able to know that `app` was imported. But to do that, I'll have to complicate things.

PR #1272, opened 2021-11-28
Fix bug in `WebSocketProtocol.asgi_receive`

kylebebak:
@euri10 @Kludex 

There were 2 things to fix in https://github.com/encode/uvicorn/issues/1230:

- ensure the server can read frames in the read queue after **the client** has sent the close frame
- ensure the server can read frames in the read queue after **the server** has sent the close frame

The first item was fixed was fixed by this PR: https://github.com/encode/uvicorn/pull/1252. But I realized after this PR was merged that the second item hadn't been fixed.

This PR fixes the second item. It ensures that the server can "drain" unread frames in the read queue even after the server has sent the close frame, as long as those frames were sent before the server sent the close frame.

As with the first item, the second item [is part of the websockets spec](https://datatracker.ietf.org/doc/html/rfc6455#section-1.4). The following code, using just the `websockets` library, makes it clear that this is the intended behavior of this library.

I did not include code in this PR that fixes this bug in `wsproto_impl.py`, because I'm not familiar with `wsproto`. That said, I don't think we should gate this fix on the corresponding fix in `wsproto`.

`client.py`

```py
import asyncio

import websockets


async def hello():
    async with websockets.connect("ws://localhost:8765") as websocket:
        await websocket.send("message 1")
        await websocket.send("message 2")
        await websocket.send("message 3")

        # Sleep until after server has sent close frame
        await asyncio.sleep(0.3)

        try:
            await websocket.send("message 4")
        except websockets.exceptions.ConnectionClosed as e:
            print("message 4 not sent, connection closed:", e)


asyncio.run(hello())
```

`server.py`

```py
import asyncio

import websockets


async def handler(websocket, *args):
    # Give client time to send frames that go into read queue before server sends close frame
    await asyncio.sleep(0.2)
    await websocket.close(1000)

    # Server can still read these messages from read queue after sending close frame
    print(await websocket.recv())
    print(await websocket.recv())
    print(await websocket.recv())


async def main():
    async with websockets.serve(handler, "localhost", 8765):
        await asyncio.Future()  # Run forever


asyncio.run(main())
```


kylebebak:
@euri10 @Kludex , just pinging you guys in case you have time to look at this. It's small and pretty well-tested

PR #1273, opened 2021-11-28
Fix h11 event conditional

Kludex:
I've found this while studying hypercorn, and I noticed that we don't do the same as: https://gitlab.com/pgjones/hypercorn/-/blob/main/src/hypercorn/protocol/h11.py#L138-172

Ideas on how to test this are welcome.


synodriver:
According to the [doc](https://h11.readthedocs.io/en/latest/api.html?highlight=NEED#h11.PAUSED) , both of them should work, I think.

Kludex:
Perfect! Thanks @synodriver !

PR #1278, opened 2021-12-04
Ensure non-zero exit code when startup fails

Kludex:
This PR proposes that running uvicorn without `--workers` and `--reload` at startup failure will exit with code 3.

The reasoning for the number 3 is to differentiate from the general exit code, and to match gunicorn's decision (consensus).

*What about `--workers` and `--reload`?*

Well, using `--workers` you actually don't exit at startup failure yet. See https://github.com/encode/uvicorn/issues/1115

As for the `--reload`, what do you recommend @euri10 ? `server.started` is always `False` on the `main.run()` because `target=server.run` runs in another process. Until #1177 (not released yet), on startup failure it didn't exit as well.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -57,7 +57,7 @@ def test_cli_call_server_run() -> None:
     with mock.patch.object(Server, "run") as mock_run:
         result = runner.invoke(cli, ["tests.test_cli:App"])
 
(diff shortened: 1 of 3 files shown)


Kludex:
@euri10 ping

euri10 (approved):

PR #1282, opened 2021-12-06
Added app_dir cli test

euri10:
see https://github.com/encode/uvicorn/pull/1271#issuecomment-986709111

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -2,6 +2,7 @@ import importlib
 import platform
 import sys
 from pathlib import Path
+from textwrap import dedent
 from unittest import mock
 
 import pytest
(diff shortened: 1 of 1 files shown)


tests/test_cli.py
```diff
@@ -131,5 +132,26 @@ def test_cli_reloader_incomplete_app_parameter(
...
+                ...
+            """
+        )
+    )
+    runner = CliRunner()
+    with mock.patch.object(main, "run") as mock_run:
+        result = runner.invoke(cli, ["main:app", "--app-dir", f"{str(app_dir)}"])
+
+    assert result.exit_code == 0
+    mock_run.assert_called_once()
```
Kludex, line 153: I don't think this is a valid test for us. I can replace the value for `--app-dir` with anything and it will still pass the assertions. See below: 

```python
def test_app_dir(tmp_path: Path, caplog: pytest.LogCaptureFixture) -> None:
    runner = CliRunner()
    with mock.patch.object(main, "run") as mock_run:
        result = runner.invoke(cli, ["main:app", "--app-dir", "potato"])

    assert result.exit_code == 0
    mock_run.assert_called_once()
```

PR #1288, opened 2021-12-09
Set `propagate` to `False` on "uvicorn" logger

### LINKED ISSUE #1285: Duplicate logs when use logging.basicConfig in application
### Checklist

### Describe the bug

When use logging.basicConfig to config logging in application, uvicorn logs will duplicate in console.

app.py:

```python
import logging
from fastapi import FastAPI

logging.basicConfig(level=logging.INFO)

app = FastAPI(debug=True)
```

run it:
```
uvicorn app:app
```

### Expected behavior

expect no duplicate logs.

### Actual behavior

uvicorn.error logs duplicated:

```
INFO:     Started server process [85610]
INFO:uvicorn.error:Started server process [85610]
INFO:uvicorn.error:Waiting for application startup.
INFO:uvicorn.error:Application startup complete.
INFO:uvicorn.error:Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
```

### Debugging material

I think the uvicorn logger should not propagate, because it already handled by uvicorn handler. And the behavior of gunicorn is no duplicate logs.

Add `"propagate": False` to https://github.com/encode/uvicorn/blob/master/uvicorn/config.py#L107 can fix the issue.
```python
"uvicorn": {"handlers": ["default"], "level": "INFO", "propagate": False},
```

### Environment

Running uvicorn 0.15.0 with CPython 3.9.7 on Darwin

### Additional context

Maybe related: https://github.com/encode/uvicorn/issues/614

guyskk:
Fix https://github.com/encode/uvicorn/issues/1285


euri10:
this seems to break quite a lots of our tests @guyskk

guyskk:
@euri10 I see, will check on it.

guyskk:
Hi @euri10, the failures is because pytest caplog not able to capture `propagate=False` logger, see also: https://github.com/pytest-dev/pytest/issues/3697

And the [caplog_for_logger](https://github.com/encode/uvicorn/blob/master/tests/middleware/test_logging.py#L13) helper also not work on test config cases, because when create Config object, `Config.configure_logging` will remove caplog.handler.

A simple work around is set `propagate=True` before execute tests. Otherwise we need refactor Config class to make it friendly to unit tests.

euri10:
> And the [caplog_for_logger](https://github.com/encode/uvicorn/blob/master/tests/middleware/test_logging.py#L13) helper also not work on test config cases, because when create Config object, `Config.configure_logging` will remove caplog.handler.
> 

mm this is where I'm not too sure, I remeber that pytest issue and I think the fixture was written just for that, and if you look at the way we use it, we do Config first then only we use the fixture so I dont think that explanation is correct, but I may be off

guyskk:
Yes, the caplog_for_logger works when test logs after create Config object, but it can not capture logs inside `Config.__init__`.
eg: `test_should_warn_on_invalid_reload_configuration`, `test_reload_dir_is_set` and `test_non_existant_reload_dir_is_not_set`

tests/conftest.py
```diff
@@ -5,6 +5,12 @@
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 
+from uvicorn.config import LOGGING_CONFIG
+
+# Workaround: pytest caplog not able to capture no-propagate loggers
```
euri10: I'd leave a more detailed explanation here as to why we explicitly turn the propagate on just for tests, mainly for the maintainers looking at that part because it's really not obvious, something from your comment here would be cool :
https://github.com/encode/uvicorn/pull/1288#issuecomment-991585560

euri10 (changes requested):
with some tweaks in the explanation and a rebase that solves conflicts it's ready to be merged for me

tests/conftest.py
guyskk: @euri10 resolved

euri10 (approved):

euri10:
thanks a lot @guyskk 
@Kludex since you tagged this on the 0.20.0 release I'll let you merge it, or maybe we can point it to this PR release but I'm not sure it exists yet

Kludex:
It does. Changed the milestone.

The milestones are mainly so I can organize myself. It's a self promise that I'm going to look for the current milestone before the release. :)

tests/conftest.py
```diff
@@ -15,6 +15,17 @@
...
+# Note: We explicitly turn the propagate on just for tests, because pytest
+# caplog not able to capture no-propagate loggers.
+#
+# And the caplog_for_logger helper also not work on test config cases, because
+# when create Config object, Config.configure_logging will remove caplog.handler.
+#
+# The simple solution is set propagate=True before execute tests.
+#
+# See also: https://github.com/pytest-dev/pytest/issues/3697
+LOGGING_CONFIG["loggers"]["uvicorn"]["propagate"] = True
```
Kludex, line 27: But then... all our users needs to do this if capturing the logs on their setup, right? 🤔 

Should we mention this somewhere, or it's fine?

guyskk: > all our users needs to do this if capturing the logs on their setup, right?

I think if users want to config uvicorn's logs, they should provide full logging config, no needs to set propagate. https://www.uvicorn.org/settings/#logging

And if users only want to config application logs, call `logging.basicConfig` works as expected, uvicorn's logs will not affected.

Kludex: I mean, if people want to capture those logs on their test, for some reason (?) they need to do the same thing.

euri10: > But then... all our users needs to do this if capturing the logs on their setup, right? thinking
> 
> Should we mention this somewhere, or it's fine?

It introduces a change only for the users who in their tests, use a Config object and are looking for logs if I get it correctly, logging is complex so I'll admit I'm not 100% sure.

I'd be tempted to say it's a rather slim part of people in that case but I may be proven wrong, but there's no harm in saying something like:

> Should you use the caplog pytest fixture in your test suite to capture uvicorn logs, you might want to use (the line above) if you use a Config object, please see (test line that explains it)

euri10: thoughts on this @guyskk ?

Kludex: (I just wanted to point this out, I don't have any strong opinion about this, jfyk)

guyskk: > I mean, if people want to capture those logs on their test, for some reason (?) they need to do the same thing.

Yes. Maybe mention it in changelog? if those user's tests broken(rare usage), they can easily know why.

euri10: ok let's add a note then !

uvicorn/config.py
```diff
@@ -119,7 +119,7 @@
         },
     },
     "loggers": {
-        "uvicorn": {"handlers": ["default"], "level": "INFO"},
+        "uvicorn": {"handlers": ["default"], "level": "INFO", "propagate": False},
         "uvicorn.error": {"level": "INFO"},
         "uvicorn.access": {"handlers": ["access"], "level": "INFO", "propagate": False},
```
Kludex, line 124: @guyskk Is this `"propagate": False` still needed?

Kludex: It is.

PR #1293, opened 2021-12-15
Support `extra_headers` for WS `accept` message

matiuszka:
The `extra_headers` support for WS `accept` message.
According to ASGI spec: https://asgi.readthedocs.io/en/latest/specs/www.html#accept-send-event

### DIFF
--- a/uvicorn/protocols/websockets/websockets_impl.py
+++ b/uvicorn/protocols/websockets/websockets_impl.py
@@ -236,7 +236,7 @@ class WebSocketProtocol(_LoggerMixin, websockets.WebSocketServerProtocol):
                     self.extra_headers.extend(
                         # ASGI spec requires bytes
                         # But for compability we need to convert it to strings
-                        (name.decode(), value.decode())
+                        (name.decode("latin-1"), value.decode("latin-1"))
                         for name, value in message.get("headers")
                     )
                 self.handshake_started_event.set()
--- a/uvicorn/protocols/websockets/websockets_impl.py
+++ b/uvicorn/protocols/websockets/websockets_impl.py
@@ -82,6 +82,7 @@ class WebSocketProtocol(_LoggerMixin, websockets.WebSocketServerProtocol):
             ping_timeout=self.config.ws_ping_timeout,
             extensions=[ServerPerMessageDeflateFactory()],
             logger=logging.getLogger("uvicorn.error"),
+            extra_headers=[],
         )
 
     def connection_made(self, transport):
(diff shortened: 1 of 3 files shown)


matiuszka:
I need to fix the pipeline.

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -231,6 +232,13 @@ async def asgi_send(self, message):
                 )
                 self.initial_response = None
                 self.accepted_subprotocol = message.get("subprotocol")
+                if "headers" in message:
+                    self.extra_headers.extend(
+                        # ASGI spec requires bytes
+                        # But for compability we need to convert it to strings
+                        (name.decode(), value.decode())
```
Kludex: It cannot be UTF-8. Most of the places we use latin-1, so I guess we'll go for it, but I'm confirming here: https://gitter.im/encode/community?at=61ba0a7cfd9244223e47b122

```suggestion
                        (name.decode("latin-1"), value.decode("latin-1"))
```

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -231,6 +232,13 @@ async def asgi_send(self, message):
                 )
                 self.initial_response = None
                 self.accepted_subprotocol = message.get("subprotocol")
+                if "headers" in message:
+                    self.extra_headers.extend(
```
Kludex, line 241: Is there a need to `extend` here? Can't we just make an attribution and on the `init` not send the empty list?

Kludex:
Thanks for the PR @matiuszka ! :tada:

uvicorn/protocols/websockets/websockets_impl.py
matiuszka: We need something mutable. The `handshake` is invoked before we are receiving headers via `send`.
So if we set `.extra_headers` to the proper value, it's too late because the `handshake` method was called with `None` value.

There is also the possibility to set `.extra_headers` as callable and that will be called during the handshake. What do you think about that?

matiuszka: @Kludex , can you look at the proposal above?

Kludex: I'm fine with the current one. Thanks for the clarifications!

Kludex (approved):
Thanks @matiuszka ! :)

Let's wait for https://github.com/encode/starlette/pull/1361 to be fixed, and then we can merge this one :)

NOTE: Starlette PR is actually not a requisite to have this merged, but I'd like to merge them together :pray:

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -236,6 +237,13 @@ async def asgi_send(self, message):
...
+                        # ASGI spec requires bytes
+                        # But for compability we need to convert it to strings
+                        (name.decode("latin-1"), value.decode("latin-1"))
+                        for name, value in message.get("headers")
```
t1waz, line 245: just small syntax comment:
for name, value in message["headers"]
due to already checking if "headers" in message

Kludex: true, PR welcome :P

matiuszka: @t1waz are you going to open PR for that? I can do that also.

PR #1294, opened 2021-12-15
Remove `root_path` from logs

### LINKED ISSUE #1290: Remove superfluous `root_path` from logging messages
### Checklist

### Describe the bug

As per https://github.com/encode/starlette/issues/1336, `scope["root_path"]` is never supposed to be joined to `scope["path"]`, as the latter is supposed to be already prefixed with the former. Luckily uvicorn is never doing this _except_ in logging. It is logging incorrect paths (according to ASGI spec) in the following places:

https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/websockets/wsproto_impl.py#L256
https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/websockets/wsproto_impl.py#L272
https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/websockets/websockets_impl.py#L230
https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/websockets/websockets_impl.py#L240
https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/utils.py#L50

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Debugging material

_No response_

### Environment

Running uvicorn 0.16.0 with CPython 3.8.10 on Darwin

### Additional context

_No response_

aminalaee:
Closes #1290 and removes `scope["root_path"]` from logs and only logs `scope["path"]`.

I could not find any other occurrences.

### DIFF
--- a/uvicorn/protocols/utils.py
+++ b/uvicorn/protocols/utils.py
@@ -46,9 +46,7 @@ def get_client_addr(scope: WWWScope) -> str:
 
 
 def get_path_with_query_string(scope: WWWScope) -> str:
-    path_with_query_string = urllib.parse.quote(
-        scope.get("root_path", "") + scope["path"]
-    )
+    path_with_query_string = urllib.parse.quote(scope["path"])
     if scope["query_string"]:
         path_with_query_string = "{}?{}".format(
             path_with_query_string, scope["query_string"].decode("ascii")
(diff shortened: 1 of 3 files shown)


euri10 (approved):

euri10:
thanks @aminalaee

levrik:
This change broke our monitoring a bit. We use FastAPI and mount a ASGI-app to `/graphql` route.
Since access logs are not really useful for GraphQL we filter them out in our monitoring to avoid a lot of noise.
Since this change access logs are logging `/` instead of `/graphql` which breaks this and makes our logs very noisy again.

PR #1304, opened 2022-01-03
Add Python 3.11 alpha to the pipeline

Kludex:
Jan 3rd: Issues with uvloop found.


Kludex:
I'll close this for now. I need `uvloop` to act. I'll get notified when there's a new release there, and bump here.

PR #1307, opened 2022-01-06
Revert stream interface

### LINKED ISSUE #1226: Using UvicornWorkers in Gunicorn cause OOM on K8s

Kludex:
Closes #1226
Reverts #869

#869 was created as a first step to a large set of changes, but we didn't go further, so I'm not sure if we should take a step back and proceed with a set of PRs later on, or be satisfied with the protocol/transport API. But considering the issue on #1226, I guess we need to do something.

This PR proposes the reversion of #869, considering that we lack a pleasant solution for the problem introduced on the mentioned issue.

@florimondmanca @euri10 What do you think?


abersheeran:
If this problem cannot be resolved within a short period of time, then I think recovery is a good idea. We have many services that depend on Uvicorn and are deployed in K8s. In order to solve the memory leak problem, I had to downgrade Uvicorn to version 0.13.4.

Kludex:
I still think this should be done, but I'll close the PR for now.

PR #1064, opened 2021-06-01
🚨 Cover protocols/websockets/websocket_impl.py on mypy

Kludex:

PR #1308, opened 2022-01-08
Add type hint to `websockets_impl.py`

Kludex:
Replaces #1064


lovelydinosaur:
I'm a little lost on the approach to typing in uvicorn, so I don't feel able to adequately review this.

First questions...

Where are we / aren't we testing with mypy at the moment?
What's the overall plan?

setup.cfg
```diff
@@ -41,7 +41,8 @@ files =
     uvicorn/middleware/__init__.py,
     uvicorn/protocols/__init__.py,
     uvicorn/protocols/http/__init__.py,
-    uvicorn/protocols/websockets/__init__.py
+    uvicorn/protocols/websockets/__init__.py,
+    uvicorn/protocols/websockets/websockets_impl.py
```
Kludex: > Where are we / aren't we testing with mypy at the moment?

We add the files here, whenever a PR for a file is added.

Kludex:
The overall plan is to do it incrementally, file by file. Check https://github.com/encode/uvicorn/issues/998 for more details.

We're almost at the end of the road on this. You can also see every PR that was done on our _mypy project_: https://github.com/encode/uvicorn/projects/2

lovelydinosaur:
Impressive! Thanks for the pointers.

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -171,7 +205,7 @@ def send_500_response(self):
         # itself (see https://github.com/encode/uvicorn/issues/920)
         self.handshake_started_event.set()
 
-    async def ws_handler(self, protocol, path):
+    async def ws_handler(self, protocol: WebSocketServerProtocol) -> Any:
```
lovelydinosaur: Curious that the signature on this method has changed.
Is that intentional?

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -1,32 +1,61 @@
...
+from asgiref.typing import (
+    ASGISendEvent,
+    WebSocketAcceptEvent,
+    WebSocketCloseEvent,
+    WebSocketConnectEvent,
+    WebSocketDisconnectEvent,
+    WebSocketReceiveEvent,
+    WebSocketScope,
+    WebSocketSendEvent,
+)
```
lovelydinosaur: I'm actually a bit lukewarm on using `aspiref.typing` vs. just messages are annotated as `dict`.

There are some cases where using very precise typing feels like a port trade-off vs. readability and comprehensibility, and this case is one where I'm a bit "meh" about if it's a good trade-off or not.

I don't have any *absolute* feelings about that, but one option we *could* consider would be to type ASGI messages as `dict` as a first pass, and then have a look at a PR switch *all cases* over to `aspiref.typing` in one go.

As I say, I'm not absolute one way or the other, so if folks feel strongly here, then sure - let's go with the consensus. But just voicing that I sometimes find there are places where overly-typed annotations can start to become less comprehensible and useful than more loosely typed annotations.

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -209,28 +243,32 @@ async def run_asgi(self):
...
                 self.logger.info(
                     '%s - "WebSocket %s" [accepted]',
                     self.scope["client"],
                     self.scope["path"],
                 )
                 self.initial_response = None
-                self.accepted_subprotocol = message.get("subprotocol")
+                self.accepted_subprotocol = message.get(  # type: ignore[assignment]
+                    "subprotocol"
+                )
```
lovelydinosaur, line 262: Just curious - why does this need an ignore? What error do we get without it?

uvicorn/protocols/websockets/websockets_impl.py
Kludex: Yes, it is. The old signature was deprecated on `websockets`, and it causes mypy issues. I can bring it back, but I'll have to `# type: ignore[<error>]` the line.

lovelydinosaur: Whichever `websockets` *actually* calls makes sense.

What version of websockets does this change in? Are we at any risk of introducing a breakage to our users here?
How come our existing method isn't *already* broken, if it's got the wrong call signature?

Kludex: You're right. It was introduced in 10.1: https://websockets.readthedocs.io/en/stable/project/changelog.html#id2

I need to ignore this or change the version on the requirements, but I'll go with just ignore.

lovelydinosaur: I suppose switching to `protocol, path=None` would help ensure that we're okay again any of `10.0`, `10.1`, and (whenever it lands) `11.0` (?)

uvicorn/protocols/websockets/websockets_impl.py
tony: > I'm actually a bit lukewarm on using `aspiref.typing` vs. just messages are annotated as `dict`.

Opened at #1373

uvicorn/protocols/websockets/websockets_impl.py
Kludex: Because `accepted_subprotocol` was a `websockets.Subprotocol` type, but I've replaced it by simple `str` and remove the ignore.

uvicorn/protocols/websockets/websockets_impl.py
Kludex: I've readded it.

uvicorn/protocols/websockets/websockets_impl.py
Kludex: I had to readd the `websockets.Subprotocol` because the parent class uses it.

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -1,30 +1,64 @@
 import asyncio
 import http
 import logging
+import sys
+from typing import Any, List, Optional, Sequence, Tuple, Union, cast
 from urllib.parse import unquote
 
 import websockets
+from trustme import TYPE_CHECKING
```
euri10: is it the right library you want to import that from ?

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -1,30 +1,64 @@
...
+        WebSocketScope,
+        WebSocketSendEvent,
+    )
 
 
 class Server:
     closing = False
 
-    def register(self, ws):
+    def register(self, ws: WebSocketServerProtocol) -> None:
```
euri10, line 42: I guess it's needed since our `WebSocketProtocol` inherits from `WebSocketServerProtocol` ?

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -70,39 +104,42 @@ def __init__(self, config, server_state, _loop=None):
...
+        transport = cast(asyncio.Transport, transport)
         self.connections.add(self)
         self.transport = transport
         self.server = get_local_addr(transport)
         self.client = get_remote_addr(transport)
         self.scheme = "wss" if is_ssl(transport) else "ws"
 
         if self.logger.isEnabledFor(TRACE_LOG_LEVEL):
-            prefix = "%s:%d - " % tuple(self.client) if self.client else ""
+            prefix = "{}:{} - ".format(*self.client) if self.client else ""
```
euri10: would that be for an unrelated PR ?

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -283,11 +334,13 @@ async def asgi_receive(self):
 
         try:
             data = await self.recv()
-        except websockets.ConnectionClosed as exc:
+        except ConnectionClosed as exc:
             self.closed_event.set()
             return {"type": "websocket.disconnect", "code": exc.code}
 
-        msg = {"type": "websocket.receive"}
+        msg: WebSocketReceiveEvent = {  # type: ignore[typeddict-item]
```
euri10, line 341: ignore here because we set subsequent keys after right ?

euri10 (commented):
Added some minor feedback and questions just to make sure I get all the ignores correctly here @Kludex 

Alos, after checking asgiref
https://github.com/django/asgiref/blob/8b615134bd481c4262356ad3194fbeff9240d0d3/asgiref/typing.py#L142-L151
it seems to me we're not using the above 2 types, is there something we're missing ?

uvicorn/protocols/websockets/websockets_impl.py
Kludex: Exactly.

uvicorn/protocols/websockets/websockets_impl.py
Kludex: No, it's part of the PR, but it was a good observation, because I actually intended to match what the HTTP implementations are doing.

uvicorn/protocols/websockets/websockets_impl.py
Kludex: Yes! The `WebSocketRecieveEvent` on asgiref should be using `NoRequired` IMHO. But as it's not, then this is needed, or initializing the dictionary with all keys.

uvicorn/protocols/websockets/websockets_impl.py
euri10: happens to me all the time with Pycharm :)

uvicorn/protocols/websockets/websockets_impl.py
euri10: ok no big deal if we keep it in here

euri10:
lgtm @Kludex , do you know why asgiref has `WebSocketResponseStartEvent` and `WebSocketResponseBodyEvent` by the way and we never ever touch those ?

euri10 (approved):

Kludex:
@euri10 We don't support that yet. But we have this on Starlette: https://github.com/encode/starlette/pull/1535, and for it to land, we'd have to comply on `uvicorn`.

See: https://asgi.readthedocs.io/en/latest/extensions.html?highlight=websocket.http.response.start#websocket-denial-response

euri10:
might be worth opening an issue to track that

Kludex:
Yes. I'll do it in 30 minutes. 🙏

PR #1310, opened 2022-01-08
Remove old comment about wsproto version

Kludex:


abersheeran (approved):
https://github.com/python-hyper/wsproto/releases/tag/1.0.0 The wsproto API should be stable.

Issue #1305, opened 2022-01-03
`asgiref` backward compatibility


tony:
### Checklist

- [X] The bug is reproducible against the latest release or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

Hi, thank you for the project, this is a shot in the dark since it's not a `channels` bug!

https://github.com/encode/uvicorn/pull/1100 pinned the `asgiref` dependency to `>=3.4.0` for mypy purposes.  It may be worth loosening it if it's only `mypy` as the restriction requires the user to use `asgiref` `3.4.0`, which triggers a well-known but unsolved channels (a different package)

- https://github.com/django/channels/issues/1722
- https://stackoverflow.com/q/68208931
- https://stackoverflow.com/q/68343003

Alternatives to pulling in `asgiref` for typings:

1. Dropping the `asgiref` dependency and copying `asgiref.typings` to `_types`

   `_types` can have compat-like functionality depending on the `asgiref` version installed so things so uvicorn can be independent of django and channels. 
2. Loosening the constraint
    
    Would mypy work if `try`/`catch` with `ImportError` was used? Would be willing to consider it as a compatibility workaround until https://github.com/django/channels/issues/1722 is resolved?
    
    ```python
    try:
        from asgiref.typing import WebSocketScope
    except ImportError:
       from asgiref.typing import WebsocketScope
       WebSocketScope = WebsocketScope
    ```

A word of note:

This is for posterity, of course it's never needed for an independent project such as `uvicorn` to change it's pinning due to _another_ package's issues, but noteworthy since `channels` is often used alongside `uvicorn` by users. 

The reasons why I post this here is channel's bug at https://github.com/django/channels/issues/1722 is prominent, its easiest apparent workaround is to decrease `asgiref`'s version below `<3.4.0`, and the only reason I can see `3.4.0` was pinned in https://github.com/encode/uvicorn/pull/1100 is `mypy` and there may be a way to satisfy mypy and relax the constraint.

I think it's up to the maintainer and I wanted to document the issue.

### Steps to reproduce the bug

I believe this happens on sites where static files are served via the asgi server. This happens when there's a heavy load (such as selenium tests).

### Expected behavior

Static files via dev servers should serve without raising deadlock errors

### Debugging material

Assuming uvicode `0.16.0`, it requires we use `asgiref` `3.4.0`+

```
Error: 1-03 17:27:52 +0000] [9] [ERROR] Exception in ASGI application
[...]

PR #1312, opened 2022-01-08
Refactor/vendorize typings

tony:
In re: #1305

I'm having difficulty with if making switches that work between `ASGIREF_33`/`ASGIREF_34`. Apparently type annotations can't handle that since it's more of a runtime thing?

@Kludex Is this possible to do without watering the typings down too much? 

- What's done `_types.py` being backported to 3.3.2 and removed `asgiref` as a dependency
- version checking without needing `distutils`/`packaging`

It can do branching like a `_compat.py` module would, but mypy considers things like:

```python
if ASGIREF_33:

    class HTTPScope(TypedDict):
        type: Literal["http"]
        asgi: ASGISpecInfo
        http_version: str
        method: str
        scheme: str
        path: str
        raw_path: bytes
        query_string: bytes
        root_path: str
        headers: Iterable[Tuple[bytes, bytes]]
        client: Optional[Tuple[str, int]]
        server: Optional[Tuple[str, Optional[int]]]
        extensions: Dict[str, Dict[object, object]]


elif ASGIREF_34:

    class HTTPScope(TypedDict):
        type: Literal["http"]
        asgi: ASGISpecInfo
        http_version: str
        method: str
        scheme: str
        path: str
        raw_path: bytes
        query_string: bytes
        root_path: str
        headers: Iterable[Tuple[bytes, bytes]]
        client: Optional[Tuple[str, int]]
        server: Optional[Tuple[str, Optional[int]]]
        extensions: Optional[Dict[str, Dict[object, object]]]
```

to be duplication.

And it's also not possible to override or extend:

```python

  class HTTPScope(TypedDict):
      type: Literal["http"]
      asgi: ASGISpecInfo
      http_version: str
      method: str
      scheme: str
      path: str
      raw_path: bytes
      query_string: bytes
      root_path: str
      headers: Iterable[Tuple[bytes, bytes]]
      client: Optional[Tuple[str, int]]
      server: Optional[Tuple[str, Optional[int]]]
      extensions: Dict[str, Dict[object, object]]

if ASGIREF_34:

    class HTTPScope(HTTPScope):
        extensions: Optional[Dict[str, Dict[object, object]]]

```

I suppose it's because that requires runtime that `mypy` isn't meant to analyze

Thoughts on this? I've allowed edits and you're welcome to fork [my branch `refactor/vendorize-typings`](https://github.com/tony/uvicorn/tree/refactor/vendorize-typings)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,6 @@ env_marker_win = "sys_platform == 'win32'"
 env_marker_below_38 = "python_version < '3.8'"
 
 minimal_requirements = [
-    "asgiref>=3.3.4",
     "click>=7.0",
     "h11>=0.8",
     "typing-extensions;" + env_marker_below_38,
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ env_marker_win = "sys_platform == 'win32'"
 env_marker_below_38 = "python_version < '3.8'"
 
 minimal_requirements = [
-    "asgiref>=3.4.0",
+    "asgiref>=3.3.4",
     "click>=7.0",
     "h11>=0.8",
     "typing-extensions;" + env_marker_below_38,
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -2,9 +2,9 @@ from typing import List, Union
 
 import httpx
 import pytest
-from asgiref.typing import ASGIReceiveCallable, ASGISendCallable, Scope
 
 from tests.response import Response
+from uvicorn._types import ASGIReceiveCallable, ASGISendCallable, Scope
 from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
 
 
--- a/tests/middleware/test_wsgi.py
+++ b/tests/middleware/test_wsgi.py
@@ -3,9 +3,8 @@ from typing import List
 
 import httpx
 import pytest
-from asgiref.typing import HTTPRequestEvent, HTTPScope
 
-from uvicorn._types import Environ, StartResponse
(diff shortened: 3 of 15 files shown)


Kludex:
I was thinking more about ignoring the `asgiref` version, vendoring the current `asgiref.typing`, and testing comparing the types from the latest `asigref.typing` to the vendored file.

EDIT: Remember to include the copyright notice.

tony:
Some `asgiref.typing` history (for posterity):

- `2.3.10` and `3.3.0` - `3.3.1` don't have this
- [`3.3.2`](https://github.com/django/asgiref/blob/3.3.2/asgiref/typing.py) added
- [`3.4.1`](https://github.com/django/asgiref/blob/3.4.1/asgiref/typing.py) - clarification of `{HTTP,Websocket}Scope`'s extension attribute, `Websocket` -> WebSocket` (I didn't realize today you actually made those contributions, nice!)

> testing comparing the types from the latest `asigref.typing` to the vendored file.

Sorry, not 100% sure what this would look like. Can you clarify?

P.S. If we were to vendorize the typings, they wouldn't necessarily hold true for `3.2.10` as `asgiref.typing` doesn't exist. In `3.3.4` the are different but perhaps this can be worked around for ones where the class name is different (`Websocket*` -> `WebSocket*`) Would the typings only be expected to work with the latest asgiref release?

I should also clarify: Is the use of the typing purely internal, or is it also intended to be used by people implementing the package/using it as a library in some fashion?

> EDIT: Remember to include the copyright notice.

For `asgiref` right? In the pydoc for the file?

Kludex:
Thanks for the efforts here @tony! 🙇

Sorry for taking so long to act.

I'll be taking this over on #1476.

PR #1315, opened 2022-01-09
Fixed keyword args due to plurality mis-match

wshayes:
Changed the keyword arguments:  reload-dirs, reload_includes and reload_excludes in the help documents to match what is in config.py.


Kludex:
This is actually not valid, the current names are matching what we have in the `main.py`.

Thanks for the interest tho! :)

PR #1261, opened 2021-11-22
Remove Python 3.6

Kludex:
Python 3.6 EOL is 23 Dec 2021, and this PR removes it. 

References: 
- https://endoflife.date/python
- https://pypistats.org/packages/uvicorn

PR #1323, opened 2022-01-14
Update black target version

sondrelg:
Since Python 3.6 was deprecated in #1261, this seemed like an oversight 🙂  I could be wrong though


Kludex (approved):

Kludex:
Thanks @sondrelg ! :)

PR #1324, opened 2022-01-14
Exclude type-checking blocks from coverage reports

sondrelg:
I don't think there's an issue for this, but I noticed these `# pragma: no cover` comments around a few type-checking blocks, and thought this might be a better way to do things. 

I don't think it ever makes sense to worry about the coverage of the contents of these if-statements, since they're not evaluated at runtime.


sondrelg:
Looks like ignoring all type-checking blocks has decreased the overall coverage by ~1% 🙃

Kludex (approved):
The `exclude_lines` were not being applied because they were between quotes. 😅

Now it works just fine. I've also added `pragma: nocover` to the list.

Thanks for the PR @sondrelg ! :)

sondrelg:
Ah yes quotes are for pyproject.toml configuration 😅

PR #1325, opened 2022-01-14
Guard imports not needed at runtime

sondrelg:
Please ignore. This is just a proof-of-concept to discuss in the gitter chat.

## Motivation

Imports that are used for annotations are not required at runtime (assuming you don't evaluate the annotations yourself, or use a library that does - see pydantic). We don't want to remove them, as they're used by type checkers, but we can remove them from the runtime of the application by guarding them with `if TYPE_CHECKING:` which equates to `if False:`.

As far as I can tell, there are two primary benefits to doing this:

- Organisation: it's simpler to reason about your code, when you have a clear separation of runtime and non-runtime imports
- Efficiency: only importing things we need creates less of an overhead at runtime

And a secondary benefit that's more relevant in larger projects is that guarding imports will remove a lot of import circularity issues for you.

## Changes

This PR moves all imports that satisfy this condition into type-checking blocks.

It also adds a flake8-plugin to tell you:
- When an import can be moved **into* a type-checking block (it is unused, except for type hinting), and perhaps more importantly
- When an import need to be moved **out** of a type-checking block, because the remaining file might have changed in a way that no longer satisfies the above
 
But I can remove the addition of the flake8-hook if you prefer. I should say that it's my plugin, but I don't mind if you use it or not 👍 

### Guarding imports - Our options

We have two ways of managing forwards references:

1. Stringified references

```python
if TYPE_CHECKING:
    from typing import Any

x: "Any"
```

2. [PEP 563](https://www.python.org/dev/peps/pep-0563/) - Postponed evaluation of annotations

```python
from __futures__ import annotations

if TYPE_CHECKING:
    from typing import Any

x: Any
```

I've gone with number two here, but I could revert to number 1 if that is preferred. As a quick disclaimer, the `annotations` futures import also enables other back-ported language features, which you may not want.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,6 +11,7 @@ wheel==0.37.0
 autoflake==1.4
 black==21.9b0
 flake8==3.9.2
+flake8-type-checking==1.2.0
 isort==5.9.3
 pytest==6.2.5
 pytest-mock==3.6.1
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,7 @@
 [flake8]
 ignore = W503, E203, B305
 max-line-length = 88
+enable-extensions = TC,TC1
 
 [mypy]
 disallow_untyped_defs = True
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -1,11 +1,17 @@
-from typing import List, Union
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
 
 import httpx
 import pytest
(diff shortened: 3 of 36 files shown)


Kludex:
Does this PR makes asgiref optional?

sondrelg:
AFAICT it removes asgiref from the package runtime completely, with the exception of this one use:

```python
# uvicorn/middleware/proxy_headers.py
scope = cast(Union[HTTPScope, WebSocketScope], scope)
```

Not sure whether this strictly makes it optional or not though, assuming the last usage was removed 🤔

sondrelg:
I'll close this for now, but happy to reopen if you want to look more at it later

PR #1329, opened 2022-01-16
Fix WSGI middleware not to explode quadratically in the case of a larger body

### LINKED ISSUE #708: WSGI interface: reading request body is O(n^2) wrt payload size
Hello!
This is related to https://github.com/encode/uvicorn/issues/371 and potentially a duplicate/complementary information, feel free to close if it is deemed to be the case.

I was fooling around with different WSGI and ASGI servers while hacking on a web framework that would support both in the future.

As noted in https://github.com/encode/uvicorn/issues/371, this is indeed a very inefficient scenario, however, what is worse, I found out the time required to turn around the request is at least O(n^2) or worse wrt the payload size.

`POST`ing some files and multipart forms with `curl`:
```
Payload size: 94171344
real	0m5.763s
user	0m0.033s
sys	0m0.012s
```
```
Payload size: 186054864
real	0m24.183s
user	0m0.018s
sys	0m0.061s
```
```
Payload size: 372178858
real	1m37.760s
user	0m0.044s
sys	0m0.152s
```

So it is not just inefficient loading everything into memory, but something else is going on.

Test case used:
```python
def app(environ, start_response):
    status = '200 OK'  # HTTP Status
    headers = [('Content-type', 'text/plain')]  # HTTP Headers
    start_response(status, headers)

    stream = environ['wsgi.input']
    length = 0
    while True:
        chunk = stream.read(8192)
        if not chunk:
            break
        length += len(chunk)

    resp = "Payload size: {}".format(length)
    return [resp.encode()]
```

Running `uvicorn==0.11.5` as `uvicorn --interface wsgi simple_reader:app`.

vytas7:
Fixes #708

NB: the proposed patch changes the type signature of a public utility function (`build_environ()`), however the method in question doesn't seem to be documented as part of any externally available interface. If that's unacceptable, I could add a compatibility check/wrapper for it.


vytas7:
I've just realized this might get superseded by https://github.com/encode/uvicorn/pull/1303, so judge accordingly.

lovelydinosaur:
Great - thanks!

Given that this clearly resolves an open bug, I'd suggest that we get this in and fixed. I can't see any downside to that.
It's possible that it'll be superseded by https://github.com/encode/uvicorn/pull/1303 - but taking a call on that one is more complex than this.

The test suite currently has a listing failure. I expect it's unrelated to this pull request(?) but it'll needs resolving first, either way.

vytas7:
Hi @tomchristie , and thanks for looking at this.
I think these were typing issues and other glitches possibly introduced by _this_ pull request, but I just didn't dig too deep since I didn't know what the plan was. If there is a chance to get this merged, I can try to rectify these.

lovelydinosaur:
Rather than using `io.BytesIO` could you use list appends instead, eg...

Change from the existing:

```python
         body = message.get("body", b"")
         more_body = message.get("more_body", False)
         while more_body:
             body_message: HTTPRequestEvent = await receive()  # type: ignore[assignment]
             body += body_message.get("body", b"")
             more_body = body_message.get("more_body", False)
```

to this...

```python
        body_list = [message.get("body", b"")]
        more_body = message.get("more_body", False)
        while more_body:
             body_message: HTTPRequestEvent = await receive()  # type: ignore[assignment]
             body_list.append(body_message.get("body", b""))
             more_body = body_message.get("more_body", False)
        body = b"".join(body_list)
```

That'd probably result in a pull request with a smaller change footprint, right?

vytas7:
I've actually considered this; but I think it would in general result in higher overhead.

Writing to a `BytesIO` only very marginally loses against appending, and may even come close or surpass other approaches for many large chunks of data. Furthermore, it is more memory-efficient since a `BytesIO` instance only needs to keep an internal memory buffer that it can reallocate on demand; appending to a list would need twice the memory right after concatenation.

lovelydinosaur:
Thanks for you time on this @vytas7.
I think we need a more comprehensive fix, so let's close this given #1303.

(Either we'll rely on `a2wsgi` or we need a fix that doesn't load all the body into memory at once.)

lovelydinosaur:
Re-opening given https://github.com/encode/uvicorn/pull/1303#issuecomment-1036102453

Currently blocked on the CI type checking failures.

vytas7:
Thanks Tom; if this is still of interest, I'll rectify 'em typing problems later in the evening.

lovelydinosaur:
Seems like a good plan to me.
It's a fairly low impact PR, and it'll close an outstanding bug, so... 👍

(Thanks for your time on this)

vytas7:
@tomchristie done :heavy_check_mark: (The CI now passes ~~with flying colours~~.)

lovelydinosaur (approved):

lovelydinosaur:
Thanks @vytas7!

PR #1331, opened 2022-01-16
Fix misleading docs around `--reload` behaviors

### LINKED ISSUE #1326: Interactions between --reload-dir, --reload-include, and --reload-exclude are not very intuitive or well explained (and may even be wrong)
### Checklist

### Describe the bug

I am trying to get `uvicorn` to automatically reload in a dev environment based on source file changes and config file changes. I have two Python src directories (`/path/to/first_dir` and `/path/to/second_dir`) and a config directory (`/path/to/env_dir`) which contains `*.env` files that my application loads.

It is _**really**_ hard to understand what `uvicorn` intends to do, much less what it actually does with regard to tuned reload behavior. Consider:

``` sh
% uvicorn \
     --reload \
     --reload-exclude '*_flymake.py' \
     --reload-exclude 'flycheck_*.py' \
     --reload-dir /path/to/first_dir \
     --reload-dir /path/to/second_dir \
     mm.txns.main:app
```

~~The above will _**only**_ ignore files matching `*_flymake.py` and `flycheck_*.py` in `/path/to/first_dir`, _**not**_ `/path/to/second_dir`. So it looks like one needs to repeat the exclude patterns prior to each `--reload-dir`:~~ _UPDATE: This wasn't correct. It turns out_ nothing _was being excluded in this configuration because watchgod was not installed. This confusion is part of the bug._

``` sh
% uvicorn \
     --reload \
     --reload-exclude '*_flymake.py' \
     --reload-exclude 'flycheck_*.py' \
     --reload-dir /path/to/first_dir \
     --reload-exclude '*_flymake.py' \
     --reload-exclude 'flycheck_*.py' \
     --reload-dir /path/to/second_dir \
     mm.txns.main:app
```

~~There is no mention of the need for repeated uses of `--reload-exclude` _**in advance of**_ any `--reload-dir` options that I can find.~~

Further, (in what I am perceiving as a contradiction of the docs), this won't work:

``` sh
% uvicorn \
     --reload \
     --reload-exclude '*_flymake.py' \
     --reload-exclude 'flycheck_*.py' \
     --reload-dir /path/to/first_dir \
     --reload-exclude '*_flymake.py' \
     --reload-exclude 'flycheck_*.py' \
     --reload-dir /path/to/second_dir \
     --reload-include '*.env' \
     --reload-dir /path/to/env_dir \
     mm.txns.main:app
```

Adding, deleting, or modifying `/path/to/env_dir/foo.env` will _**not**_ trigger a reload.

The [documentation for ``--reload-exclude``](https://www.uvicorn.org/settings/#development) makes it clear (or at least strongly suggests) I should be able to expand the patterns matched for reload detection (**_emphasis_** mine):

> `--reload-exclude <glob-pattern>` - Specify a glob pattern to match files or directories which will excluded from watching. May be used multiple times. By default the following patterns are excluded: `.*, .py[cod], .sw.*, ~*`. **_These defaults can be overwritten by including them in `--reload-include`._**

That's not what I'm seeing by including `*.env` files. Then there's this, which seems to contradict the inclusion docs (**_emphasis_** mine):

> By default Uvicorn uses simple changes detection strategy that compares **_python files_** modification times few times a second. If … or you need watching of **_non python files_** you can install [watchgod](https://pypi.org/project/watchgod/) or install uvicorn with `uvicorn[standard]`, which will include watchgod.

Does that mean there's some "upper bound" of patterns uvicorn is capable of considering? Is it `*.py`? That can't be correct, because the counterexample explicitly states that `*.py[cod]` is excluded by default, but could be included via `--reload-include`. It's not clear where the limitations are, even after spending dozens of minutes reading, re-reading, re-re-reading, experimenting, etc.

If the answer is, "Use watchgod," I can't find any documentation of how to do that or what additional configuration steps I would need to meet my use case.

### Environment

[root@localhost vagrant]# uname -a
Linux localhost.localdomain 5.4.17-2136.300.7.el8uek.x86_64 #2 SMP Fri Oct 8 16:23:01 PDT 2021 x86_64 x86_64 x86_64 GNU/Linux
[root@localhost vagrant]# /opt/local/app/bin/uvicorn --version
Running uvicorn 0.16.0 with CPython 3.9.6 on Linux

### Additional context

I'm not the only person to be confused by this. [This StackOverflow question](https://stackoverflow.com/questions/64118680/reload-flag-with-uvicorn-can-we-exclude-certain-code#comment123408591_64118929) poses something very similar, but remains unanswered:

> Could you advise on how to exclude [\*.json] using this method while preserving the reload functionality? I tried: [uvicorn main:app --reload-exclude '\*.json'] however it generated this warning: WARNING: Current configuration will not reload as not all conditions are met, please refer to documentation.

When I [misfiled this against gunicorn](https://github.com/benoitc/gunicorn/issues/2722) in a fit of blindess, I noted that [this comment](https://github.com/benoitc/gunicorn/issues/2339#issuecomment-651270632) from @Andrew-Chen-Wang suggests there is some magic around merely making watchgod available to uvicorn, but fails to go into additional detail, so I have no way of verifying it (at least not without digging into a lot of code), much less figuring out how I can take advantage:

> … [I]f you have watchgod installed, uvicorn should automatically pick that up to make the reloading process better. …

He responded with:

> … In terms of resolving the issue, I believe you're not supposed to reuse reload-include several times. Instead, wrap all your values in double quotes and make them csv

If that's true, it's also not clear from the docs. It also doesn't help me, since I'm only using `--reload-include` once. (The above use of `--reload-exclude` multiple times does appear to get the behavior I want with respect to excluding various files/directories.)

_UPDATE: I'm pretty sure the single, comma-separated argument advice is wrong. [This strongly suggests](https://github.com/encode/uvicorn/blob/27f76476a14dac68a62dc2998717997607a36197/uvicorn/main.py#L85-L99) one should be able to provide `--reload-include` and `--reload-exclude` multiple times via the command line._

posita:
The existing docs regarding `--reload` (and adjacent options) were frustratingly vague
and too often led to misunderstandings about specific behaviors. This attempts to
address those documentation deficiencies. Further, this adds a logging warning to
surface to users that `--reload-include` and `--reload-exclude` are ignored unless
Uvicorn is using watchgod to monitor filesystem changes.

Fixes #1326.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -28,12 +28,21 @@ For example, in case you want to run the app on port `5000`, just set the enviro
 
 ## Development
 
-* `--reload` - Enable auto-reload.
+* `--reload` - Enable auto-reload. Uvicorn supports two versions of auto-reloading behavior enabled by this option. See below for important differences between them.
 * `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default the whole current directory will be watched. If you are running programmatically use `reload_dirs=[]` and pass a list of strings.
+
+### Naive Reloading
+
+If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to any `*.py` file inside of its monitored directories. See the `--reload-dir` option above. Note: _all_ `*.py` files, and _only_ `*.py` files will be monitored for changes. This is not configurable under naive reloading.
+
+### watchgod-Enabled Reloading
+
+For more nuanced control over which file modifications trigger reloads, install [watchgod](https://pypi.org/project/watchgod/) where Unvicorn can see it. Alternatively, install `uvicorn[standard]`, which includes watchgod as a dependency. This will enable the following options (which are otherwise ignored).
+
 * `--reload-include <glob-pattern>` - Specify a glob pattern to match files or directories which will be watched. May be used multiple times. By default the following patterns are included: `*.py`. These defaults can be overwritten by including them in `--reload-exclude`.
 * `--reload-exclude <glob-pattern>` - Specify a glob pattern to match files or directories which will excluded from watching. May be used multiple times. By default the following patterns are excluded: `.*, .py[cod], .sw.*, ~*`. These defaults can be overwritten by including them in `--reload-include`.
 
-By default Uvicorn uses simple changes detection strategy that compares python files modification times few times a second. If this approach doesn't work for your project (eg. because of its complexity), or you need watching of non python files you can install [watchgod](https://pypi.org/project/watchgod/) or install uvicorn with `uvicorn[standard]`, which will include watchgod.
+_Note: As of this writing, watchgod relies on poll-based OS machinery to surface modifications to Uvicorn. This means that monitored directories must live on filesystems that support that operation. That_ should _accommodate most use cases. Be aware that for locations where such mechanisms are unavailable, modifications will be silently overlooked._
 
 ## Production
 
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -33,7 +33,7 @@ For example, in case you want to run the app on port `5000`, just set the enviro
 
 ### Naive Reloading
 
-If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to all `*.py` files (and only `*.py` files) inside of its monitored directories. See the `--reload-dir` option. Which files are monitored is not configurable unless watchgod in installed. See the `--reload-include` and `--reload-exclude` options for details.
+If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to all `*.py` files (and only `*.py` files) inside of its monitored directories. See the `--reload-dir` option. Specifying other file extensions is not supported unless watchgod is installed. See the `--reload-include` and `--reload-exclude` options for details.
 
 ### watchgod-Enabled Reloading
 
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -33,7 +33,7 @@ For example, in case you want to run the app on port `5000`, just set the enviro
 
 ### Naive Reloading
 
-If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to any `*.py` file inside of its monitored directories. See the `--reload-dir` option. Note: _all_ `*.py` files, and _only_ `*.py` files will be monitored for changes. This is not configurable unless watchgod in installed.
+If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to all `*.py` files (and only `*.py` files) inside of its monitored directories. See the `--reload-dir` option. Which files are monitored is not configurable unless watchgod in installed. See the `--reload-include` and `--reload-exclude` options for details.
 
 ### watchgod-Enabled Reloading
 
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -28,12 +28,12 @@ For example, in case you want to run the app on port `5000`, just set the enviro
 
 ## Development
 
-* `--reload` - Enable auto-reload. Uvicorn supports two versions of auto-reloading behavior enabled by this option. See below for important differences between them.
+* `--reload` - Enable auto-reload. Uvicorn supports two versions of auto-reloading behavior enabled by this option. There are important differences between them.
 * `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default the whole current directory will be watched. If you are running programmatically use `reload_dirs=[]` and pass a list of strings.
 
 ### Naive Reloading
 
-If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to any `*.py` file inside of its monitored directories. See the `--reload-dir` option above. Note: _all_ `*.py` files, and _only_ `*.py` files will be monitored for changes. This is not configurable under naive reloading.
+If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to any `*.py` file inside of its monitored directories. See the `--reload-dir` option. Note: _all_ `*.py` files, and _only_ `*.py` files will be monitored for changes. This is not configurable unless watchgod in installed.
 
 ### watchgod-Enabled Reloading
 
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -31,18 +31,16 @@ For example, in case you want to run the app on port `5000`, just set the enviro
 * `--reload` - Enable auto-reload. Uvicorn supports two versions of auto-reloading behavior enabled by this option. There are important differences between them.
 * `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default the whole current directory will be watched. If you are running programmatically use `reload_dirs=[]` and pass a list of strings.
 
-### Naive Reloading
+### Reloading without watchgod
 
 If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to all `*.py` files (and only `*.py` files) inside of its monitored directories. See the `--reload-dir` option. Specifying other file extensions is not supported unless watchgod is installed. See the `--reload-include` and `--reload-exclude` options for details.
 
-### watchgod-Enabled Reloading
+### Reloading with watchgod
 
-For more nuanced control over which file modifications trigger reloads, install [watchgod](https://pypi.org/project/watchgod/) where Unvicorn can see it. Alternatively, install `uvicorn[standard]`, which includes watchgod as a dependency. This will enable the following options (which are otherwise ignored).
+For more nuanced control over which file modifications trigger reloads, install `uvicorn[standard]`, which includes watchgod as a dependency. Alternatively, install [watchgod](https://pypi.org/project/watchgod/) where Unvicorn can see it. This will enable the following options (which are otherwise ignored).
 
 * `--reload-include <glob-pattern>` - Specify a glob pattern to match files or directories which will be watched. May be used multiple times. By default the following patterns are included: `*.py`. These defaults can be overwritten by including them in `--reload-exclude`.
 * `--reload-exclude <glob-pattern>` - Specify a glob pattern to match files or directories which will excluded from watching. May be used multiple times. By default the following patterns are excluded: `.*, .py[cod], .sw.*, ~*`. These defaults can be overwritten by including them in `--reload-include`.
-
-_Note: As of this writing, watchgod relies on poll-based OS machinery to surface modifications to Uvicorn. This means that monitored directories must live on filesystems that support that operation. That_ should _accommodate most use cases. Be aware that for locations where such mechanisms are unavailable, modifications will be silently overlooked._
 
 ## Production
 
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -39,12 +39,15 @@ Options:
                                   of using the current working directory.
   --reload-include TEXT           Set glob patterns to include while watching
                                   for files. Includes '*.py' by default; these
-                                  defaults can be overridden in `--reload-
-                                  exclude`.
+                                  defaults can be overridden with `--reload-
+                                  exclude`. This option has no effect unless
+                                  watchgod is installed.
   --reload-exclude TEXT           Set glob patterns to exclude while watching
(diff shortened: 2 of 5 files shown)


docs/settings.md
```diff
@@ -28,12 +28,21 @@ For example, in case you want to run the app on port `5000`, just set the enviro
 
 ## Development
 
-* `--reload` - Enable auto-reload.
+* `--reload` - Enable auto-reload. Uvicorn supports two versions of auto-reloading behavior enabled by this option. See below for important differences between them.
 * `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default the whole current directory will be watched. If you are running programmatically use `reload_dirs=[]` and pass a list of strings.
+
+### Naive Reloading
+
+If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to any `*.py` file inside of its monitored directories. See the `--reload-dir` option above. Note: _all_ `*.py` files, and _only_ `*.py` files will be monitored for changes. This is not configurable under naive reloading.
```
Andrew-Chen-Wang: I think it's better to be explicit by mentioning in the last sentence with " unless watchgod is utilized"

```suggestion
If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to any `*.py` file inside of its monitored directories. See the `--reload-dir` option above. Note: _all_ `*.py` files, and _only_ `*.py` files will be monitored for changes. This is not configurable under naive reloading unless watchgod is utilized.
```

docs/settings.md
```diff
@@ -28,12 +28,21 @@ For example, in case you want to run the app on port `5000`, just set the enviro
 
 ## Development
 
-* `--reload` - Enable auto-reload.
+* `--reload` - Enable auto-reload. Uvicorn supports two versions of auto-reloading behavior enabled by this option. See below for important differences between them.
```
Andrew-Chen-Wang: I can't remember if it was this project that had auto creation of CLI flags using docs, so I'd be weary with the "See below" and replace the last sentence with an explicit explanation. But Kludex would know.

Andrew-Chen-Wang (changes requested):
Thanks for this!

docs/settings.md
posita: Good call. I think I can eliminate even more words in adopting yours.

docs/settings.md
posita: I think I can eliminate the relative reference anyway. Let me try….

docs/settings.md
```diff
@@ -28,12 +28,21 @@ For example, in case you want to run the app on port `5000`, just set the enviro
 
 ## Development
 
-* `--reload` - Enable auto-reload.
+* `--reload` - Enable auto-reload. Uvicorn supports two versions of auto-reloading behavior enabled by this option. There are important differences between them.
 * `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default the whole current directory will be watched. If you are running programmatically use `reload_dirs=[]` and pass a list of strings.
+
+### Naive Reloading
+
+If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to any `*.py` file inside of its monitored directories. See the `--reload-dir` option. Note: _all_ `*.py` files, and _only_ `*.py` files will be monitored for changes. This is not configurable unless watchgod in installed.
```
posita: @Andrew-Chen-Wang, this adopts your suggested edit and removes what I perceive as extraneous words.

Andrew-Chen-Wang:
```suggestion
If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to any `*.py` file inside of its monitored directories (see `--reload-dir`). Note: _all_ `*.py` files, and _only_ `*.py` files will be monitored for changes. This is not configurable unless watchgod in installed.
```

LGTM.

One concern is this quote: "Note: _all_ `*.py` files, and _only_ `*.py`". I thought only `.py` files which were imported would be subject to a reload. This wording implies any `.py` file will be reloaded, regardless of its location in the entire file system.

posita: I might be confused, but unless I'm misreading [the code](https://github.com/encode/uvicorn/blob/6885e9c6ec4d881ec5cf1464a07c619706ad8240/uvicorn/supervisors/statreload.py#L49-L52), I'm pretty sure it's _any_ `*.py` file inside the monitored directories that has a change to its `mtime`. In any event, I word-smithed this a little bit to make clearer that it was bound by the monitored directories.

Either way, we should definitely surface a clearer understanding of what actually happens, so I'm very grateful for the (continued) attention.

docs/settings.md
```diff
@@ -28,12 +28,21 @@ For example, in case you want to run the app on port `5000`, just set the enviro
 
 ## Development
 
-* `--reload` - Enable auto-reload.
+* `--reload` - Enable auto-reload. Uvicorn supports two versions of auto-reloading behavior enabled by this option. There are important differences between them.
 * `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default the whole current directory will be watched. If you are running programmatically use `reload_dirs=[]` and pass a list of strings.
+
+### Naive Reloading
+
+If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to all `*.py` files (and only `*.py` files) inside of its monitored directories. See the `--reload-dir` option. Which files are monitored is not configurable unless watchgod in installed. See the `--reload-include` and `--reload-exclude` options for details.
```
posita: @Andrew-Chen-Wang, here's my proposal which is hopefully responsive to [your recent observations](https://github.com/encode/uvicorn/pull/1331#discussion_r786256431) about clarity regarding which `*.py` files would trigger reloads in this context.

Andrew-Chen-Wang: Just misspelling (is vs. in installed) and clarity. Otherwise, LGTM!:

```suggestion
If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to all `*.py` files (and only `*.py` files) inside of its monitored directories. See the `--reload-dir` option. Specifying other file extensions is not supported unless watchgod is installed. See the `--reload-include` and `--reload-exclude` options for details.
```

posita: Awesome catch! Thanks!

docs/settings.md
posita: I'm going to resolve this for now, since I believe subsequent commits resolve this, but please revive if you think I've missed something.

docs/settings.md
posita: Resolved by subsequent commit.

Andrew-Chen-Wang:
LGTM. I'm sure Kludex will follow up since I'm not a maintainer.

docs/settings.md
```diff
@@ -28,12 +28,21 @@ For example, in case you want to run the app on port `5000`, just set the enviro
...
 * `--reload-exclude <glob-pattern>` - Specify a glob pattern to match files or directories which will excluded from watching. May be used multiple times. By default the following patterns are excluded: `.*, .py[cod], .sw.*, ~*`. These defaults can be overwritten by including them in `--reload-include`.
 
-By default Uvicorn uses simple changes detection strategy that compares python files modification times few times a second. If this approach doesn't work for your project (eg. because of its complexity), or you need watching of non python files you can install [watchgod](https://pypi.org/project/watchgod/) or install uvicorn with `uvicorn[standard]`, which will include watchgod.
+_Note: As of this writing, watchgod relies on poll-based OS machinery to surface modifications to Uvicorn. This means that monitored directories must live on filesystems that support that operation. That_ should _accommodate most use cases. Be aware that for locations where such mechanisms are unavailable, modifications will be silently overlooked._
```
euri10: what locations are you thinking of when saying that ?

docs/settings.md
```diff
@@ -28,12 +28,21 @@ For example, in case you want to run the app on port `5000`, just set the enviro
...
 
-* `--reload` - Enable auto-reload.
+* `--reload` - Enable auto-reload. Uvicorn supports two versions of auto-reloading behavior enabled by this option. There are important differences between them.
 * `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default the whole current directory will be watched. If you are running programmatically use `reload_dirs=[]` and pass a list of strings.
+
+### Naive Reloading
+
+If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to all `*.py` files (and only `*.py` files) inside of its monitored directories. See the `--reload-dir` option. Specifying other file extensions is not supported unless watchgod is installed. See the `--reload-include` and `--reload-exclude` options for details.
+
+### watchgod-Enabled Reloading
```
euri10:
```suggestion
### Reloading with watchgod
```

docs/settings.md
```diff
@@ -28,12 +28,21 @@ For example, in case you want to run the app on port `5000`, just set the enviro
 
 ## Development
 
-* `--reload` - Enable auto-reload.
+* `--reload` - Enable auto-reload. Uvicorn supports two versions of auto-reloading behavior enabled by this option. There are important differences between them.
 * `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default the whole current directory will be watched. If you are running programmatically use `reload_dirs=[]` and pass a list of strings.
+
+### Naive Reloading
```
euri10:
```suggestion
### Reloading without watchgod
```

docs/settings.md
```diff
@@ -28,12 +28,21 @@ For example, in case you want to run the app on port `5000`, just set the enviro
...
+
+If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to all `*.py` files (and only `*.py` files) inside of its monitored directories. See the `--reload-dir` option. Specifying other file extensions is not supported unless watchgod is installed. See the `--reload-include` and `--reload-exclude` options for details.
+
+### watchgod-Enabled Reloading
+
+For more nuanced control over which file modifications trigger reloads, install [watchgod](https://pypi.org/project/watchgod/) where Unvicorn can see it. Alternatively, install `uvicorn[standard]`, which includes watchgod as a dependency. This will enable the following options (which are otherwise ignored).
```
euri10: I'd suggest the opposite, 1st install `uvicorn[standard]` and then watchgod manually

docs/settings.md
mmmbogosian: I don't have a handle on all the weird things that masquerade as filesystems where `poll` or `select` don't maintain appropriate semantics. Whatever those are, they're likely very esoteric setups. I can remove the note, if you think it's not helpful.

posita: I don't have a handle on all the weird things that masquerade as filesystems where `poll` or `select` don't maintain appropriate semantics. Whatever those are, they're likely very esoteric setups. I can remove the note, if you think it's not helpful.

euri10: yes let's get rid of it, no need to go into too much details here we just want users to know how we handle things and this PR clarifies it  quite nicely without that note imho

posita: Removed by latest commit.

docs/settings.md
posita: Adopted in lastest commit.

docs/settings.md
posita: Adopted in latest commit.

docs/settings.md
```diff
@@ -28,13 +28,20 @@ For example, in case you want to run the app on port `5000`, just set the enviro
...
 * `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default the whole current directory will be watched. If you are running programmatically use `reload_dirs=[]` and pass a list of strings.
+
+### Reloading without watchgod
+
+If Uvicorn _cannot_ load [watchgod](https://pypi.org/project/watchgod/) at runtime, it will periodically look for changes in modification times to all `*.py` files (and only `*.py` files) inside of its monitored directories. See the `--reload-dir` option. Specifying other file extensions is not supported unless watchgod is installed. See the `--reload-include` and `--reload-exclude` options for details.
+
+### Reloading with watchgod
+
+For more nuanced control over which file modifications trigger reloads, install `uvicorn[standard]`, which includes watchgod as a dependency. Alternatively, install [watchgod](https://pypi.org/project/watchgod/) where Unvicorn can see it. This will enable the following options (which are otherwise ignored).
```
posita, line 40: @euri10, does this capture [what you meant](https://github.com/encode/uvicorn/pull/1331#discussion_r803498434)?

docs/settings.md
posita: I swapped the phrases in the latest commit. (See https://github.com/encode/uvicorn/pull/1331/files#r803758196. Please let me know there if I missed the mark.)

docs/settings.md
euri10: yep

euri10 (approved):
thanks @posita , I'll let it sit a little to see if others have further comments but this looks like a great addition

Issue #1199, opened 2021-09-28
uvicorn not closing sockets


johnrtipton:
We are using the latest uvicorn 0.15.  It seems that after we upgraded to uvicorn 0.15 that our AWS load balancer health checks fail after a certain amount of time.  Upon further investigation we found that all of our sockets are being used up because they are staying open in a CLOSE_WAIT status.

```
# netstat -ant | awk '{print $6}' | sort | uniq -c | sort -n
   1 Foreign
   1 LISTEN
   1 established)
   7 TIME_WAIT
  13 ESTABLISHED
7047 CLOSE_WAIT
```

### Workaround
For now we have switched our worker from UvicornWorker to UvicornH11Worker and it has resolved our issue.

### Environment
Running uvicorn 0.15.0 with CPython 3.7.12 on Linux

Kludex:
I don't see how the change of worker class would solve this issue :thinking:

pacoyang:
> I don't see how the change of worker class would solve this issue 🤔

It maybe because h11 will close the transport when if the connection is invalid HTTP request, this issue similarity to #1192 
https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/http/h11_impl.py#L140

johnrtipton:
Yes @pacoyang this seems to be exactly the problem we are seeing.   Thanks!

Kludex:
@johnrtipton from which version did you guys upgrade? 0.13.4?

johnrtipton:
@Kludex We upgraded from `0.14.0`

shahar-medigate:
We encountered the issue in our systems as well, a simple fastapi service behind AWS EKS load-balancer
we also managed to reproduce the issue to a single commit
https://github.com/encode/uvicorn/commit/960d4650db0259b64bc41f69bc7cdcdb1fdbcbf3

A simple program to reproduce
```python
import uvicorn

async def app(scope, receive, send):
    assert scope["type"] == "http"

    await send(
        {
            "type": "http.response.start",
            "status": 200,
            "headers": [
                [b"content-type", b"text/plain"],
            ],
        }
    )
    await send(
        {
            "type": "http.response.body",
            "body": b"Hello, world!",
        }
    )


if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", reload=False, log_level=5)
```

and run `nc -z localhost:8000`
now you will have a new CLOSE_WAIT connection with netstat

[...]

PR #1332, opened 2022-01-17
Fix ensure http connection is closed

pacoyang:
Ref Issues: #1199 #1226
Ref PRs: #1192 #1244 #1307
## Description
As #869 (released in 0.14.0) is missing handle the scenario of `b""`, it is possible to make the HTTP connection not close correctly.
Here is the memory usage plot with the current main branch, running a single `tcping` process for 5 minutes.
p1

The memory usage is increasing, and the `netstat` shows many `CLOSE_WAIT` connections.

p2

I read the code in #869 , I have found two issues:
1. As we handle the connection close or not in the protocol, I think we should pass the scenario of `b""`  to [protocol.data_received](https://github.com/encode/uvicorn/blob/master/uvicorn/_handlers/http.py#L84) too. It is up to the protocol to determine if the connection should be closed or not.
2. Reading the data from [`_buffer`](https://github.com/encode/uvicorn/blob/master/uvicorn/_handlers/http.py#L82) seems not a good way. In some scenarios, the client will initially send empty TCP pings, then  `_buffer` will receive `b""`, determine if the connections should be closed or not through the `_buffer` data is confusing (refs from the [read](https://github.com/python/cpython/blob/main/Lib/asyncio/streams.py#L669) function).

Here is current PR code running a single `tcping` process for 5 minutes in the same testing environment:
p3

Test codes:
```python
import uvicorn

    await send(
        {
            "status": 200,
            "headers": [
                [b"content-type", b"text/plain"],
            ],
        }
    )
    await send(
        {
            "type": "http.response.body",
            "body": b"Hello, world!",
        }
    )


if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", reload=False, log_level=5)

```

### DIFF
--- a/uvicorn/_handlers/http.py
+++ b/uvicorn/_handlers/http.py
@@ -7,6 +7,9 @@ if TYPE_CHECKING:  # pragma: no cover
     from uvicorn.server import ServerState
 
 
+MAX_RECV = 2 ** 16
+
+
 async def handle_http(
     reader: asyncio.StreamReader,
     writer: asyncio.StreamWriter,
@@ -33,13 +36,13 @@ async def handle_http(
     # Use a future to coordinate between the protocol and this handler task.
     # https://docs.python.org/3/library/asyncio-protocol.html#connecting-existing-sockets
     loop = asyncio.get_event_loop()
-    connection_lost = loop.create_future()
+    reader_read = asyncio.create_task(reader.read(MAX_RECV))
 
     # Switch the protocol from the stream reader to our own HTTP protocol class.
     protocol = config.http_protocol_class(  # type: ignore[call-arg, operator]
         config=config,
         server_state=server_state,
-        on_connection_lost=lambda: connection_lost.set_result(True),
+        on_connection_lost=reader_read.cancel,
     )
     transport = writer.transport
     transport.set_protocol(protocol)
@@ -56,7 +59,7 @@ async def handle_http(
 
     @task.add_done_callback
     def retrieve_exception(task: asyncio.Task) -> None:
-        exc = task.exception()
+        exc = task.exception() if not task.cancelled() else None
 
         if exc is None:
             return
@@ -74,15 +77,5 @@ async def handle_http(
 
     # Kick off the HTTP protocol.
     protocol.connection_made(transport)
-
-    # Pass any data already in the read buffer.
-    # The assumption here is that we haven't read any data off the stream reader
-    # yet: all data that the client might have already sent since the connection has
-    # been established is in the `_buffer`.
-    data = reader._buffer  # type: ignore
-    if data:
-        protocol.data_received(data)
-
-    # Let the transport run in the background. When closed, this future will complete
-    # and we'll exit here.
-    await connection_lost
+    data = await reader_read
+    protocol.data_received(data)
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -222,6 +222,10 @@ class H11Protocol(asyncio.Protocol):
                     continue
                 self.cycle.more_body = False
                 self.cycle.message_event.set()
+            elif event_type is h11.ConnectionClosed:
+                break
+        if self.conn.our_state is h11.MUST_CLOSE and not self.transport.is_closing():
+            self.transport.close()
 
     def handle_upgrade(self, event):
         upgrade_value = None
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -136,6 +136,8 @@ class HttpToolsProtocol(asyncio.Protocol):
             self.transport.close()
         except httptools.HttpParserUpgrade:
             self.handle_upgrade()
+        if data == b"" and not self.transport.is_closing():
+            self.transport.close()
 
     def handle_upgrade(self):
         upgrade_value = None


uvicorn/protocols/http/h11_impl.py
```diff
@@ -222,6 +222,10 @@ def handle_events(self):
                     continue
                 self.cycle.more_body = False
                 self.cycle.message_event.set()
+            elif event_type is h11.ConnectionClosed:
+                break
+        if self.conn.our_state is h11.MUST_CLOSE and not self.transport.is_closing():
+            self.transport.close()
```
Kludex, line 228: Fine :+1:

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -136,6 +136,8 @@ def data_received(self, data):
             self.transport.close()
         except httptools.HttpParserUpgrade:
             self.handle_upgrade()
+        if data == b"" and not self.transport.is_closing():
+            self.transport.close()
```
Kludex, line 140: Fine :+1:

uvicorn/_handlers/http.py
```diff
@@ -56,7 +59,7 @@ async def handle_http(
 
     @task.add_done_callback
     def retrieve_exception(task: asyncio.Task) -> None:
-        exc = task.exception()
+        exc = task.exception() if not task.cancelled() else None
 
         if exc is None:
             return
```
Kludex, line 65: Can we do the following instead? It makes it easier to understand. 
```python
exc = task.exception()

if task.cancelled() or exc is None:
```

Kludex (commented):
Thanks for the PR @pacoyang!

I've confirmed that this PR solves our issues: memory leak and CLOSE_WAIT. The logic here also makes sense to me, and I've specially liked the usage of the `MUST_CLOSE` state. :+1:

That being said... I still think `uvicorn` should revert #869 . It's been almost a year and no one took the initiative to continue the work into trio compatibility.

In any case, I'll wait 10 days. If no other member gets involved on this PR, I'll approve, merge and release.

Kludex:
@tomchristie @abersheeran @euri10 @florimondmanca friendly ping if someone wants to check this as well

uvicorn/_handlers/http.py
pacoyang: If the task has been cancelled, `task.exception()` will raise a CancelledError exception.
https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.exception

pacoyang:
> That being said... I still think uvicorn should revert #869 . It's been almost a year and no one took the initiative to continue the work into trio compatibility.

@Kludex Sure, I just want to solve this issue as soon as possible, any way is acceptable. I love `uvicorn`, but this serious problem has forced me to choose another ASGI server. It's been a while since the discovery. 

Using a streams handler can do more things like move the keep-alive timeout task out of the protocol, so we don't need to write multiple times in the protocol implementation. The keep-alive timeout task trigger still has issue, when a TCP connection is established and the client does not send any bytes, the server does not close this connection when keep-alive timeout. It can be reproduced by `telnet`.

uvicorn/_handlers/http.py
Kludex: Yeah, I'm aware. But the conditional will work as intended anyway. 🤔

abersheeran: > Can we do the following instead? It makes it easier to understand.
> 
> ```python
> exc = task.exception()
> 
> if task.cancelled() or exc is None:
> ```

Calling `task.exception()` after the task is canceled will directly throw a `CanceledError`, which will cause subsequent code to fail to execute.

pacoyang: It will throw exception in `task.exception() `
```bash
INFO:     Started server process [2935281]
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
INFO:     127.0.0.1:35362 - "GET /1 HTTP/1.1" 200 OK
Exception in callback <function handle_http.<locals>.retrieve_exception at 0x7f24f7ab7310>
handle: <Handle handle_http.<locals>.retrieve_exception>
  File "uvloop/cbhandles.pyx", line 63, in uvloop.loop.Handle._run
  File "/home/ubuntu/uvicorn/uvicorn/_handlers/http.py", line 63, in retrieve_exception
    exc = task.exception()
```

uvicorn/_handlers/http.py
```diff
@@ -7,6 +7,9 @@
     from uvicorn.server import ServerState
 
 
+MAX_RECV = 2 ** 16
```
abersheeran, line 10: Are we considering adding this value to a config option?

abersheeran (approved):

uvicorn/_handlers/http.py
pacoyang: I don't see users have the need to customize the chunk size. The stream buffer limit in asyncio is `2 ** 16` 64 KiB [by default](https://github.com/python/cpython/blob/main/Lib/asyncio/streams.py#L22), `read` will never return a chunk larger than the buffer limit.

bittermandel:
Just stumbled upon this when running on fly.io with TCP health checks. Luckily I could switch over to HTTP checks to get around this, but following this to be sure it's resolved.

Kludex (approved):
Thanks for the PR @pacoyang !

PR #1333, opened 2022-01-17
performance read large file

gshmu:


t1waz:
@gshmu it's interesting, can You describe how You tested performance? thanks

Kludex:
@gshmu Thanks for the PR!

It's important to notice that we also care about readebility, specially because it's on the docs.

For that reason, and for the fact that there's not much information. I'll be closing this PR.

PR #1336, opened 2022-01-23
Fix `TypeError: __init__() got an unexpected keyword argument 'logger…

arqvt:
…'` error.

I got a TypeError: __init__() got an unexpected keyword argument 'logger' error when i made a websocket and made a request, i came here to fix it.


Kludex:
What's the version of `uvicorn` and `websockets`?

Kludex:
The issue here is that you have an old `websockets` version.

We set the minimal version for `websockets` to work with `uvicorn` on the [`setup.py`](https://github.com/encode/uvicorn/blob/master/setup.py). The thing is that we don't have a conditional to verify if the version installed is enough for `uvicorn`.

That being said, the current PR doesn't solve the issue. Hence, I'll be closing it.

PR #1339, opened 2022-01-24
Require operations approval for package deploys.

lovelydinosaur:
Switch the "publish" workflow to run on the "deploy" environment, so that we can require reviews from @encode/operations  before deploying to PyPI.

I've also:

* Setup a deploy environment in this repo's settings.
* Configured it to have "required reviews" from the @encode/operations team
* Created a new PyPI API token for `uvicorn`.
* Saved that token in the deploy environment's secrets as PYPI_TOKEN.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,4 +1,3 @@
----
 name: Publish
 
 on:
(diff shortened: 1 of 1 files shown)


Kludex (approved):

lovelydinosaur:
Right, let's give this a go.

PR #1341, opened 2022-01-27
Check extra requirement versions at runtime

Kludex:
Currently, if a user installs `uvicorn` without the `[standard]` extra requirements, the user is able to install packages with incompatible versions. Like, they can do `pip install uvicorn websockets==0.9.0`, and this will break at runtime, because `uvicorn` only works with `websockets>=10.0`.

Alternatives, and discussion, are more than welcome.

 We can also not act here, but the error pointed out is completely misleading.

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -22,6 +22,8 @@ from asgiref.typing import ASGIApplication
 
 try:
     import yaml
+
+    assert yaml.__version__ >= "5.1", "Uvicorn requires PyYAML version 5.1 or higher"
 except ImportError:
     # If the code below that depends on yaml is exercised, it will raise a NameError.
     # Install the PyYAML package or the uvicorn[standard] optional dependencies to
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ minimal_requirements = [
 
 extra_requirements = [
     "websockets>=10.0",
-    "httptools>=0.2.0,<0.4.0",
+    "httptools>=0.2.0",
     "uvloop>=0.14.0,!=0.15.0,!=0.15.1; " + env_marker_cpython,
     "colorama>=0.4;" + env_marker_win,
     "watchgod>=0.6",
--- a/uvicorn/loops/auto.py
+++ b/uvicorn/loops/auto.py
@@ -1,6 +1,10 @@
 def auto_loop_setup(reload: bool = False) -> None:
     try:
         import uvloop  # noqa
+
+        assert (
+            uvloop.__version__ >= "0.14.0"
+        ), "Uvicorn requires uvloop version 0.14.0 or higher"
     except ImportError:  # pragma: no cover
         from uvicorn.loops.asyncio import asyncio_setup as loop_setup
 
--- a/uvicorn/protocols/http/auto.py
+++ b/uvicorn/protocols/http/auto.py
@@ -4,6 +4,10 @@ from typing import Type
 AutoHTTPProtocol: Type[asyncio.Protocol]
 try:
     import httptools  # noqa
+
+    assert (
+        httptools.__version__ >= "0.2.0"
+    ), "Uvicorn requires httptools version 0.2.0 or higher"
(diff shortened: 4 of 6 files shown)


lovelydinosaur:
⚠️ **warning** ⚠️

```python
>>> "0.10" > "0.9"
False
```

uvicorn/config.py
```diff
@@ -22,6 +22,8 @@
 
 try:
     import yaml
+
+    assert yaml.__version__ >= "5.1", "Uvicorn requires PyYAML version 5.1 or higher"
```
adriangb, line 26: I would use https://github.com/pypa/packaging to do the version comparisons

Kludex:
> warning **warning** warning
> 
> ```python
> >>> "0.10" > "0.9"
> False
> ```

Ah... Well, that's what happens when you copy code without thinking :sunglasses: :
https://github.com/encode/uvicorn/blob/c84979a92a5d1a95fc02a6ef77af016d4064c22a/uvicorn/protocols/websockets/wsproto_impl.py#L16

In any case, opinions about this?

adriangb (commented):
So if I understand the situation correctly, the issue is that internally Uvicorn says "if I have package XYZ installed, then enable this feature" but if package XYZ was installed by the user or some other dependency (not `uvicorn[standard]`) then the version of package XYZ could be anything. I see how that is problematic.

A couple options:
1. Whenever possible, don't enable features at import time, postpone them until the user actually uses them (not sure what this would look like in each case, obviously not possible in many cases, and always looks ugly). This means that if they don't use the feature, they can have whatever version of the package they want installed.
2. Create an empty `uvicorn-standard` package and publish that on PyPi. Then instead of checking for these actual dependencies we need, we just check for `uvicorn-standard` being installed. But a user could just install `uvicorn-standard` directly. So I guess we could move all of the dependencies to be dependencies of `uvicorn-standard`, but that's ugly and at that point it's not a footgun, it's a user intentionally doing wonky stuff.

adriangb:
> this will break at runtime

could you clarify a bit what the failure mode is? like do they get an error when they try to establish a web socket connection, or do they get an error when they try to start up uvicorn because we try to access some class/function that was renamed?

Kludex:
> > this will break at runtime
> 
> could you clarify a bit what the failure mode is? like do they get an error when they try to establish a web socket connection, or do they get an error when they try to start up uvicorn because we try to access some class/function that was renamed?

Real example: https://github.com/encode/uvicorn/pull/1336

adriangb:
> Real example: #1336

gotcha. so this is breaking when they actually try to use the feature, not at import time (which would have been harder to fix I think).

with the current approach in 105265457ae0d7edb179e0a83c89cebde9eb3cc3, what happens if the user has the wrong websockets version but doesn't use websockets in uvicorn? do they still get an assertion error?

Kludex:
> with the current approach in 1052654, what happens if the user has the wrong websockets version but doesn't use websockets in uvicorn? do they still get an assertion error?

No.

Kludex:
> with the current approach in 1052654, what happens if the user has the wrong websockets version but doesn't use websockets in uvicorn? do they still get an assertion error?

Uvicorn by default, selects the WebSocket package it will use automatically i.e. if `websockets` is installed, it will use it. If the user specify the WebSocket implementation it will use, and the user selects `wsproto`, then no error will happen. Not sure if it's a realistic thing to happen.

adriangb:
> Uvicorn by default, selects the WebSocket package it will use automatically i.e. if `websockets` is installed, it will use it. If the user specify the WebSocket implementation it will use, and the user selects `wsproto`, then no error will happen. Not sure if it's a realistic thing to happen.

makes sense. what I'm more concerned about is some sort of mono repo scenario where there's 2 services, one using websockets without uvicorn and one using uvicorn without websockets. with the current approach the uvicorn one would fail to import because of the websocket version, even if it doesn't use websockets. so yeah I think we should push this check into runtime if possible.

PR #1342, opened 2022-01-28
Fix wsproto version check expression

Kludex:


uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -13,7 +14,9 @@
 from uvicorn.logging import TRACE_LOG_LEVEL
 from uvicorn.protocols.utils import get_local_addr, get_remote_addr, is_ssl
 
-assert wsproto.__version__ > "0.13", "Need wsproto version 0.13"
+assert version.parse(wsproto.__version__) > version.parse(
+    "0.13"
+), "Need wsproto version 0.13"
```
lovelydinosaur: I guess you probably mean `>=` here?

Also the "Need wsproto version 0.13" probably ought to be "Need wsproto version >= 0.13".

```suggestion
assert version.parse(wsproto.__version__) >= version.parse(
    "0.13"
), "Need wsproto version >= 0.13"
```

Kludex: To be honest, I just wanted to compare the version right, but yeah, I can also fix it hahaha

Fix applied. I also took the opportunity to improve a little the message. What do you think?

lovelydinosaur (approved):

PR #1351, opened 2022-02-01
Fix coverage precision to 96.77%

Kludex:
"Aren't you being too pedantic?"

Not really... It's really bothersome having to regain the same coverage over and over again because PRs get merged without covering all of it. The idea here is to not regress anymore.

`coverage` doesn't support the precision argument on the CLI. I've created a PR to fix it: https://github.com/nedbat/coveragepy/pull/1317. Until then, we should use the `setup.cfg`.

### DIFF
--- a/scripts/coverage
+++ b/scripts/coverage
@@ -8,4 +8,4 @@ export SOURCE_FILES="uvicorn tests"
 
 set -x
 
-${PREFIX}coverage report --show-missing --skip-covered --fail-under=97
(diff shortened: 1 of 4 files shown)


lovelydinosaur (approved):
Sure thing.

Kludex:
I may have a few rounds of commits to add pragma coverage conditional, so I can match the Ubuntu Python 3.8 coverage.

PR #1352, opened 2022-02-01
Send HTTP 400 response for invalid request

bertramr:
Given an invalid request, respond with an HTTP 400 error instead of
closing the connection without a response.

Please find here a rebased version of #205

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -135,9 +135,9 @@ class H11Protocol(asyncio.Protocol):
             try:
                 event = self.conn.next_event()
             except h11.RemoteProtocolError as exc:
-                msg = "Invalid HTTP request received."
+                msg = b"Invalid HTTP request received."
                 self.logger.warning(msg, exc_info=exc)
-                self.transport.close()
+                self.send_400_response(msg)
                 return
             event_type = type(event)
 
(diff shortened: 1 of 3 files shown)


euri10 (commented):
would it be possible for both implementations of `send_400_response` to have the same signature ? one take str the other bytes.

or maybe there's a good reason

bertramr:
Yes, would be possible. What would be the preferred datatype?

euri10 (approved):
thanks @bertramr

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -190,6 +171,29 @@ def handle_upgrade(self):
         protocol.data_received(b"".join(output))
         self.transport.set_protocol(protocol)
 
+    def send_400_response(self, msg: str):
+
+        from uvicorn.protocols.websockets.auto import AutoWebSocketsProtocol
+
+        if AutoWebSocketsProtocol is None:
+            msg = "No supported WebSocket library detected. Please use 'pip install uvicorn[standard]', or install 'websockets' or 'wsproto' manually."  # noqa: E501
+            self.logger.warning(msg)
```
lovelydinosaur, line 180: This looks like it needs to be deleted.

uvicorn/protocols/http/h11_impl.py
```diff
@@ -278,6 +257,30 @@ def handle_upgrade(self, event):
         protocol.data_received(b"".join(output))
         self.transport.set_protocol(protocol)
 
+    def send_400_response(self, msg: str):
+
+        from uvicorn.protocols.websockets.auto import AutoWebSocketsProtocol
+
+        if AutoWebSocketsProtocol is None:
+            msg = "No supported WebSocket library detected. Please use 'pip install uvicorn[standard]', or install 'websockets' or 'wsproto' manually."  # noqa: E501
+            self.logger.warning(msg)
```
lovelydinosaur, line 266: This looks like it needs to be deleted.

uvicorn/protocols/http/httptools_impl.py
euri10: this was part of improving user experience in https://github.com/encode/uvicorn/pull/926 @tomchristie 
Do you want we get rid of it ?

lovelydinosaur: Perhaps it just needs to stay within the `if upgrade_value != b"websocket" or self.ws_protocol_class is None:` block. It doesn't make sense for this to be within `send_400_response` right? (Eg. it shouldn't log if it's called because of `except httptools.HttpParserError as exc:`)

euri10: ha yes you're right I didn't notice if was sent also in that case, I misread thinking it was just for the upgrade case

PR #1353, opened 2022-02-02
Handle ip networks in trusted host list of forwarded_allow_ips param

yacir:
Hi folks 

## Context

Currently we can not set ip networks as forwarded allow ips.

```python
Config(
    forwarded_allow_ips="10.0.0.0/8",
)
```

This PR adds the this possibility by taking advantage of `ipaddress` library in plus of string matching as fallback.

## Changes
* Middleware
    * [x] ProxyHeadersMiddleware

### DIFF
--- a/uvicorn/middleware/proxy_headers.py
+++ b/uvicorn/middleware/proxy_headers.py
@@ -8,6 +8,7 @@ the connecting client, rather that the connecting proxy.
 
 https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#Proxies
 """
+from ipaddress import ip_address, ip_network
 from typing import List, Optional, Tuple, Union, cast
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Hiya @yacir - thanks for jumping in.

We're currently trying to push for Uvicorn to be in a feature-pause, while we focus on letting it bed in, and making sure that the issue backlog is under control.

This is a neat bit of functionality, but for now you'll need to install it as an ASGI middleware, rather updating uvicorn itself.

Thanks for your understanding.

@encode/maintainers:  *We ought to look at getting an updated `CONTRIBUTING.md` in place, and making sure to point contributors towards discussions as a first point of contact, similar to how `httpx` is currently setup.*

PR #1354, opened 2022-02-03
Revert "Fix ensure http connection is closed"

Kludex:
Reverts encode/uvicorn#1332

PR #1355, opened 2022-02-03
Revert stream interface

Kludex:
Supersedes #1307
On top of #1354


lovelydinosaur (approved):
Given https://github.com/encode/uvicorn/issues/1226#issuecomment-949445052 I'm just going to defer to your judgement on this for now, @kludex.

Kludex:
I think this is a good thing. If we really wish to have a stream interface, then we can go back to this and study the plan of action in the future.

wlaub:
https://github.com/encode/uvicorn/blob/1c20f546574cc905a8b28a5390c3f815e27a84ab/uvicorn/server.py#L137
`create_server` doesn't seem too happy about receiving `path=config.uds`

```
    File "/usr/lib/python3.8/asyncio/runners.py", line 44, in run
      return loop.run_until_complete(main)
    File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
      return future.result()
    File "/snap/standalone-dev-env/x1/lib/python3.8/site-packages/uvicorn/server.py", line 77, in serve
      await self.startup(sockets=sockets)
    File "/snap/standalone-dev-env/x1/lib/python3.8/site-packages/uvicorn/server.py", line 137, in startup
      server = await loop.create_server(  # type: ignore[call-overload]
  TypeError: create_server() got an unexpected keyword argument 'path'
```

PR #1360, opened 2022-02-03
improve handling of transfer-encoding

mkiesel:
httptools_impl added a second transfer-encoding header
in case the app set transfer-encoding: gzip or similar
without giving a content length. Also, it used chunked
for responses to HTTP/1.0 requests (those clients do
not support chunked encoding). Finally, this adds a way
to force non-chunked transfer as there are quite a few
music streaming devices that issue HTTP/1.1 requests
but do not understand chunked encoding.

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -283,6 +283,63 @@ def test_close(protocol_cls, event_loop):
         assert protocol.transport.is_closing()
 
 
+@pytest.mark.parametrize("protocol_cls", [HttpToolsProtocol])
+def test_get_streaming_gzip_request(protocol_cls, event_loop):
+    async def app(scope, receive, send):
+        await send({"type": "http.response.start", "status": 200, "headers": [[b"transfer-encoding", b"gzip"]]})
+        await send({"type": "http.response.body", "body": b"some zipped stream data", "more_body": True})
+
+    with get_connected_protocol(app, protocol_cls, event_loop) as protocol:
+        protocol.data_received(SIMPLE_GET_REQUEST)
+        protocol.loop.run_one()
+        assert b"HTTP/1.1 200 OK" in protocol.transport.buffer
+        assert b"transfer-encoding: gzip" in protocol.transport.buffer
+        assert b"transfer-encoding: chunked" not in protocol.transport.buffer
+        assert b'some zipped stream data' in protocol.transport.buffer
+
+
+@pytest.mark.parametrize("protocol_cls", [HttpToolsProtocol])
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Could you provide a link to give some more context here?

Nothing here: https://github.com/MagicStack/httptools/issues?q=just_stream
Or here: https://www.google.com/search?q=httptools+just_stream

lovelydinosaur:
Going to close this off in the meantime, as we really need to start pushing for suggestions to come through discussions first.

mkiesel:
Where/how am I to give more context? These problems are all Uvicorn related and have nothing to do with upstream httptools.

This PR includes four test cases, did you see those?
* **test_get_streaming_gzip_request** makes sure Uvicorn doesn't garble responses made by the app with Transfer-Encoding: gzip or similar. Currently, for those responses, it injects a second Transfer-Encoding: chunked header.
* **test_get_streaming_http10_request** makes sure Uvicorn doesn't reply with Transfer-Encoding: chunked for a HTTP/1.0 request. HTTP 1.0 clients don't support chunked encoding.
* **test_get_streaming_nochunked_request** checks the Transfer-Encoding: just_stream header handling. This is a bit of a hack to somehow allow the app to make sure Uvicorn just passes through any response stream data without "chunking" it; the code will not send that header to the client but rather treat it as a flag only. As explained above, some hardware streaming devices cannot process chunked encoding, but as it is,Uvicorn forces chunked encoding for all streams. If you have a better idea of how to implement this workaround, let me know.
* **test_get_streaming_http11_request** is the only test that passes on current Uvicorn as well; I just added that since there appeared to be no actual tests for that case in the existing tests.

I can also create a couple of issues if you think that's better than one PR.

mkiesel:
Discussion see https://github.com/encode/uvicorn/discussions/2645

lovelydinosaur:
It's not obvious to me that #2645 is correct - does it ensure connection closing?
It looks like the h11 implementation correctly supports this... https://h11.readthedocs.io/en/latest/api.html

mkiesel:
Do you mean https://github.com/encode/uvicorn/pull/1360 ? https://github.com/encode/uvicorn/discussions/2645 is just a discussion.

I'm not sure the code in https://github.com/encode/uvicorn/pull/1360 (that also contains the `just_stream` hack that I need for my scenarios in which clients connect that say they can manage HTTP/1.1 but actually they don't support chunked encoding...) is 100% complete and correct, but I'm fairly sure that the issue raised in https://github.com/encode/uvicorn/discussions/2645 is correct. :)

Issue #1368, opened 2022-02-10
Uvicorn logging error when port is a string although the server is running


uripeled2:
### Checklist

- [X] The bug is reproducible against the latest release or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

If you use uvicorn.run and enter the port as string:
`uvicorn.run(app=app, host="localhost", port="3000")`

the server will still run but the logger will raise: 

> TypeError: %d format: a number is required, not str

### Steps to reproduce the bug

```
import uvicorn


async def app(scope, receive, send):
    assert scope['type'] == 'http'

    await send({
        'type': 'http.response.start',
        'status': 200,
        'headers': [
            [b'content-type', b'text/plain'],
        ],
    })
    await send({
        'type': 'http.response.body',
        'body': b'Hello, world!',
    })

uvicorn.run(app=app, host="localhost", port="3000")

```

### Expected behavior

> INFO:     Uvicorn running on http://localhost:3000 (Press CTRL+C to quit)

### Actual behavior

> TypeError: %d format: a number is required, not str

### Debugging material

INFO:     Started server process [9761]
INFO:     Waiting for application startup.
INFO:     ASGI 'lifespan' protocol appears unsupported.
INFO:     Application startup complete.
--- Logging error ---
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 1083, in emit
    msg = self.format(record)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 927, in format
    return fmt.format(record)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 663, in format
    record.message = record.getMessage()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 367, in getMessage
    msg = msg % self.args
TypeError: %d format: a number is required, not str
Call stack:
  File "/Users/uri/PycharmProjects/PokemonFetcher/try.py", line 18, in <module>
    uvicorn.run(app=app, host="localhost", port="3000")
  File "/Users/uri/PycharmProjects/PokemonFetcher/venv/lib/python3.9/site-packages/uvicorn/main.py", line 461, in run
    server.run()
[...]

PR #1369, opened 2022-02-11
Uvicorn logging error when port is a string although the server is running, see issue #1368

uripeled2:
See issue https://github.com/encode/uvicorn/issues/1368


Kludex:
In theory, this is not needed, as the Config class is properly annotated.

I think we should go with https://github.com/encode/uvicorn/pull/1108.

uripeled2:
Yes I guess this will be https://github.com/encode/uvicorn/pull/1108 a better solution... 
Do you need any help?

PR #1370, opened 2022-02-11
Moved warning about missing ws library only in upgrade case

euri10:
Follow up to #1352 
Solves the issue raised in https://github.com/encode/uvicorn/pull/1352#discussion_r804502418

### DIFF
--- a/scripts/coverage
+++ b/scripts/coverage
@@ -8,4 +8,4 @@ export SOURCE_FILES="uvicorn tests"
 
 set -x
 
-${PREFIX}coverage report --show-missing --skip-covered --fail-under=96.93
+${PREFIX}coverage report --show-missing --skip-covered --fail-under=96.87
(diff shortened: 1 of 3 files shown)


Kludex:
Why is there a need for coverage downgrade if there's a _no pragma_?

euri10:
I added it in this PR to try fix the coverage downgrade but it wasn't enough so I fixed it like that, which is bad, but the pragma wasnt in the previous commit and the coverage should have dropped there, so really I dont know, seems flacky

lovelydinosaur (approved):
Similarly to @Kludex I also don't get the `nocover`/coverage change.
But also I don't care enough about that for it to block the PR.

(Because correctness wins as a priority)

euri10:
Ok I merge and will dig into the coverage stuff later

PR #1371, opened 2022-02-11
Warn about not supported websocket protocol only on handle_upgrade

bertramr:
Follow up to https://github.com/encode/uvicorn/pull/1352
Solves the issue raised in https://github.com/encode/uvicorn/pull/1352#discussion_r804502418

Same as https://github.com/encode/uvicorn/pull/1370, but with different order of warnings.


bertramr:
Seems the discussion in #926 was explicitly about sending the required install steps also within the HTTP 400 response and not only logging it. Therefore I close this pull request in favor of #1370

PR #1373, opened 2022-02-11
chore(mypy): Loosen scope typing to `dict`

tony:
Per @tomchristie https://github.com/encode/uvicorn/issues/1305#issuecomment-1036157361 and https://github.com/encode/uvicorn/pull/1308/files#r781146416

- Does this seem like a correct understanding of what you mean? Is it okay?
- If so, any other typings here that can have similar done?


tony:
Closed per https://github.com/encode/uvicorn/issues/1305#issuecomment-1036285154

PR #1385, opened 2022-02-22
Use path with query string on WebSockets logs

### LINKED ISSUE #1384: Access logs for mounted apps doesnt not contain the root path
This change broke our monitoring a bit. We use FastAPI and mount a ASGI-app to `/graphql` route.
Since access logs are not really useful for GraphQL we filter them out in our monitoring to avoid a lot of noise.
Since this change access logs are logging `/` instead of `/graphql` which breaks this and makes our logs very noisy again.

_Originally posted by @levrik in https://github.com/encode/uvicorn/issues/1294#issuecomment-1046761532_

levrik:
I've started to use `get_path_with_query_string` for websockets logging as well to have a single place responsible for creating the path for logging.
I don't know why this method wasn't used before so in case I'm missing something here, please let me know.

Fixes #1384


euri10 (commented):
this looks good, 
could you please @aminalaee take a look at it since you fixed https://github.com/encode/uvicorn/issues/1290 with https://github.com/encode/uvicorn/issues/1294
I'm wondering here if #1290 was legit in fact, but not sure

levrik:
I was looking into the linked issue https://github.com/encode/starlette/issues/1336 and since FastAPI uses Starlette under the hood, this is probably where the issue originates from. From my point of view the fix done in #1290 is legit but causes some issues due to https://github.com/encode/starlette/issues/1336.
The change to using `raw_path` might not be correct as well as `path` should be completed by spec but is broken in combination with Starlette. Maybe a revert to combining `root_path` and `path` should be done until Starlette is fixed? I'm feeling unsure about the change to `raw_path`.

aminalaee:
@levrik  Yes I think you're right and also changing back to `root_path` + `path` would make more sense as according the ASGI the `raw_path` can be optional.

The change to use `get_path_with_query_string ` for websocket looks reasonable though.

Kludex (approved):
Thanks @levrik :)

PR #1076, opened 2021-06-09
Cover h11_impl.py on mypy

Kludex:
From https://github.com/encode/uvicorn/pull/1017.

PR #1397, opened 2022-03-05
Add type annotation to `h11_impl.py`

Kludex:
Supersedes #1076


uvicorn/protocols/http/h11_impl.py
```diff
@@ -385,25 +420,29 @@ async def run_asgi(self, app):
...
-            }
+            )
         )
         await self.send(
-            {"type": "http.response.body", "body": b"Internal Server Error"}
+            HTTPResponseBodyEvent(
+                type="http.response.body",
+                body=b"Internal Server Error",
+                more_body=False,
+            )
```
Kludex: We were missing a `more_body`.

uvicorn/protocols/http/h11_impl.py
```diff
@@ -417,12 +456,16 @@ async def send(self, message):
...
                 raise RuntimeError(msg % message_type)
+            message = cast(HTTPResponseStartEvent, message)
 
             self.response_started = True
             self.waiting_for_100_continue = False
 
             status_code = message["status"]
-            headers = self.default_headers + message.get("headers", [])
+            message_headers = cast(
+                List[Tuple[bytes, bytes]], message.get("headers", [])
```
Kludex, line 464: As `headers` field in `message` is an `Iterable`, the operation `+` below is not right.

I'm unsure if we should be doing this or make sure that the operation works with iterables that are not list.

uvicorn/protocols/http/h11_impl.py
```diff
@@ -385,25 +420,29 @@ async def run_asgi(self, app):
...
+                headers=[
                     (b"content-type", b"text/plain; charset=utf-8"),
                     (b"connection", b"close"),
                 ],
-            }
+            )
         )
         await self.send(
-            {"type": "http.response.body", "body": b"Internal Server Error"}
+            HTTPResponseBodyEvent(
```
Kludex: Should we just annotate, or is it fine to instantiate the dictionary with this TypedDic?

adriangb: If you can get it to work with a plain dict, I'd prefer that, I imagine there is _some_ overhead to using the TypedDict

uvicorn/protocols/http/h11_impl.py
adriangb: I would just change this to `headers: TheHeaderType = [*self.default_headers, *message.get("headers", ())]]  # type: ignore[whatever]` or without the #type: ignore` if the type of message["headers"] is inferred.

uvicorn/protocols/http/h11_impl.py
```diff
@@ -82,14 +117,14 @@ def connection_made(self, transport):
         self.scheme = "https" if is_ssl(transport) else "http"
 
         if self.logger.level <= TRACE_LOG_LEVEL:
-            prefix = "%s:%d - " % tuple(self.client) if self.client else ""
+            prefix = "%s:%d - " % self.client if self.client else ""
```
euri10, line 120: was that highlighted by mypy ?

uvicorn/protocols/http/h11_impl.py
```diff
@@ -385,25 +420,27 @@ async def run_asgi(self, app):
...
+            "headers": [
+                (b"content-type", b"text/plain; charset=utf-8"),
+                (b"connection", b"close"),
+            ],
+        }
+        await self.send(response_start_event)
+        response_body_event: HTTPResponseBodyEvent = {
+            "type": "http.response.body",
+            "body": b"Internal Server Error",
+            "more_body": False,
```
euri10, line 438: thank type hints we were missing more_body

uvicorn/protocols/http/h11_impl.py
```diff
@@ -417,12 +454,16 @@ async def send(self, message):
             if message_type != "http.response.start":
                 msg = "Expected ASGI message 'http.response.start', but got '%s'."
                 raise RuntimeError(msg % message_type)
+            message = cast(HTTPResponseStartEvent, message)
```
euri10, line 457: here we can cast because we're sure that is is a HTTPResponseStartEvent ?

uvicorn/protocols/http/h11_impl.py
euri10: I like the casting more but either way is fine I think asgiref is too permissive here

euri10 (approved):

uvicorn/protocols/http/h11_impl.py
Kludex: Exactly! I guess at some point, mypy will be able to infer that based on the conditional...

uvicorn/protocols/http/h11_impl.py
Kludex: yep 🙏

uvicorn/protocols/http/h11_impl.py
Kludex: Yes:

```
uvicorn/protocols/http/h11_impl.py:120: error: Incompatible types in string interpolation (expression has type "object", placeholder has type "Union[int, float, SupportsInt]")
```

Kludex:
I've first annotated this file on June last year: https://github.com/encode/uvicorn/pull/1076.

I'm happy we made it! Thanks! ❤️

PR #1400, opened 2022-03-11
Bump `httptools` to `>=0.4.0`

Kludex:
As there are 2 CVE that affects `httptools`, they've bumped the package to 0.4.0.

Do we remove the previous versions to force users to bump `httptools` or do we just increase the range (i.e. `<0.5.0`)?

Reference: https://github.com/MagicStack/httptools/issues/76

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ minimal_requirements = [
 
 extra_requirements = [
     "websockets>=10.0",
-    "httptools>=0.2.0,<0.4.0",
+    "httptools>=0.4.0",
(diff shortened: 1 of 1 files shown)


Kludex:
I'll create a release after this (or the alternative) is merged.

euri10:
> Do we remove the previous versions to force users to bump `httptools` or do we just increase the range (i.e. `<0.5.0`)?

force bump imho, unless the fix is backported which I dont think it is looking briefly at it

euri10 (approved):

Kludex:
It's not backported. 👍 

Thanks for the approval, and good morning Benoit! 👋

PR #1412, opened 2022-03-20
Add macos to the CI

euri10:
Slighlty different than https://github.com/encode/uvicorn/pull/1395 @Kludex , let me know what you think.
In particular it shrink the socket name in macos CI and solves the issues you had on the `test_fragmentation` for macos

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,10 @@
+import os
 import ssl
 from copy import deepcopy
+from hashlib import md5
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from uuid import uuid4
 
 import pytest
 import trustme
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -136,3 +141,38 @@ def reload_directory_structure(tmp_path_factory: pytest.TempPathFactory):
 @pytest.fixture(scope="function")
 def logging_config() -> dict:
     return deepcopy(LOGGING_CONFIG)
+
+
+@pytest.fixture
+def short_socket_name(tmp_path, tmp_path_factory):  # pragma: py-win32
+    max_sock_len = 100
```
Kludex, line 148: I'm usually against comments, but a comment explaining the reason for the number 100 would be cool.

Kludex (approved):
Just a comment about the number 100, but other than that, this is great.

Thanks! :)

euri10:
> Just a comment about the number 100, but other than that, this is great.
> 
> Thanks! :)

I've taken it from here https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars/367012#367012

we could have taken 104, but some new new bsd implementations go as low as 92 bytes.
100 seemed a good compromise :)

Kludex:
Yep, I'm aware, but I meant a comment because it will be easier to grasp in code what the number represents without having to open the PR.

euri10:
oups I didnt understand sorry :)

PR #1419, opened 2022-03-24
Add `watchfiles` based reload class

Kludex:
First thoughts:
- Not sure if we want to deprecate `WatchGodReload`, and add `WatchFilesReload`, or if we want to just replace. As `watchfiles` is not stable, I'd opt for the first option.
- I have the feeling that our reload options are excessive. :disappointed:

### DIFF
--- a/uvicorn/supervisors/watchfilesreload.py
+++ b/uvicorn/supervisors/watchfilesreload.py
@@ -1,0 +1,53 @@
+import logging
+from pathlib import Path
+from socket import socket
+from typing import TYPE_CHECKING
+
+from watchfiles import PythonFilter, watch
+
+from uvicorn.supervisors.basereload import BaseReload
+
+if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


uvicorn/supervisors/watchfilesreload.py
```diff
@@ -0,0 +1,53 @@
...
+        self.reloader_name = "watchfiles"
+        self.reload_dirs = []
+
+        for directory in config.reload_dirs:
+            if Path.cwd() not in directory.parents:
+                self.reload_dirs.append(directory)
+        if Path.cwd() not in self.reload_dirs:
+            self.reload_dirs.append(Path.cwd())
+
+        self.watch_filter = CustomFilter()
```
Kludex, line 45: Just need to add the filter for this to work. Need to implement the filter tho.

Kludex:
Close this in favor of #1437

PR #1421, opened 2022-03-27
Annotate `run` function

Kludex:
The way I see it, we have two options here:
1. Annotate the way I did.
2. Unpack the `kwargs`, and add them to the config manually.

This PR is to improve the UX, as many users use `uvicorn.run` (me included).

This PR actually ignores the following mypy error:
- `uvicorn/main.py:445: error: Single overload definition, multiple required  [misc]`

Which means the ideal solution would be `2`. So, any solution satisfies me, I'll let the reviewer decide the path to go here.


euri10:
hey @Kludex 
I just tested it in pycharm and I dont see how it improves UX, care to explain ?
appart from that looks good to me, the CI just fails because of previous merge I think

Kludex:
@euri10 autocomplete and docstring basically

https://user-images.githubusercontent.com/7353520/160294895-3527ca18-db05-4a95-882f-9cdb239f7d5c.mp4

uvicorn/main.py
```diff
@@ -388,53 +393,162 @@ def main(
...
+    ssl_ca_certs: str,
+    ssl_ciphers: str,
+    headers: typing.Sequence[typing.Sequence[str]],
+    use_colors: bool,
+    app_dir: str,
+    factory: bool,
+) -> None:
+    """Run uvicorn via Python code.
+
+    Args:
```
euri10, line 496: is the docstring necessary here or the type hints are enough to get the completion ?

euri10 (approved):

euri10:
ok no comments appart the one above, this does not make the autocomplete work on pycharm though

Kludex:
If it doesn't, then it's not a valid solution :/

uvicorn/main.py
Kludex: Annotation is enough for that, the docstring are a plus

euri10:
ok let's close this #1423 is working as intended

PR #1423, opened 2022-03-28
Annotate `run` function

Kludex:
Alternative to #1421.

How does it look? :thinking:


euri10 (approved):

Kludex:
@PrettyWood can we do better?

adriangb (approved):

adriangb:
I didn't test this with MyPy or anything, but I don't see any issues just looking at the code. As long as tests pass and you checked it against MyPy and Pylance in strict mode, I'm happy with this @Kludex

PR #1424, opened 2022-03-28
Add SIGQUIT handler on UvicornWorker

Kludex:
I've implemented this... But it's actually for discussion, as I'm not convinced we should do this. I think Gunicorn should not send SIGQUIT.

This PR is related to:
- https://github.com/benoitc/gunicorn/issues/2604

Note: I haven't tested yet, really just for discussion.

### DIFF
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -2,6 +2,7 @@ import asyncio
 import logging
 import signal
 import sys
+import threading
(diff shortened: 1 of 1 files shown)


Kludex:
I'm going to create a separate package with a `UvicornWorker`, and do proper integration tests with `gunicorn`.

PR #1426, opened 2022-03-28
Modified logging module name to avoid clash with stdlib

euri10:
Most of the time it's bad usage but it's good practice anyway to avoid that


Kludex:
Should we actually call it `_logging.py`? Also, should we think about changing our module names to show their privacy? 🤔

euri10:
> Should we actually call it `_logging.py`? Also, should we think about changing our module names to show their privacy? thinking

yup good call, implemented

Kludex (approved):

PR #1427, opened 2022-03-29
Upgraded black to avoid clash with newest click

euri10:
see https://github.com/psf/black/issues/2964

an example of our CI broken : https://github.com/encode/uvicorn/runs/5735122920?check_suite_focus=true

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -119,7 +119,7 @@ Options:
                                   adding this to the PYTHONPATH. Defaults to
                                   the current working directory.  [default: .]
   --factory                       Treat APP as an application factory, i.e. a
-                                  () -> <ASGI app> callable.  [default: False]
+                                  () -> <ASGI app> callable.
   --help                          Show this message and exit.
 ```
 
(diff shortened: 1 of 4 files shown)


Kludex (dismissed):

euri10:
yup realized that too late :) updated

euri10:
ok we got an issue click is 7.18 in CI so should be bunp it as well ?

Kludex:
How is it `click` that old there? 🤔 Which package is holding it down?

euri10:
> How is it `click` that old there? thinking Which package is holding it down?

my bad it's types-click which is on that version, ci on linux install 8.1 fine, it seems like it's windows specifics, again...

euri10:
seems like windows ci install 8.0.4 and linux 8.1 so we're doomed

Kludex:
Ah... Then we should run the cli tool only on linux... ?

euri10:
> Ah... Then we should run the cli tool only on linux... ?

or drop windows support ! :imp:

Kludex:
Replaced by #1428

PR #1441, opened 2022-04-05
Add default code for `websocket.close`

Kludex:
I need to add tests for it.

Inspired by https://github.com/encode/starlette/issues/1560#issuecomment-1086855500

- [x] ~tests~ this is actually being tested... against none and 1000... maybe we should be more restrictive on the test...

### DIFF
--- a/uvicorn/protocols/websockets/wsproto_impl.py
+++ b/uvicorn/protocols/websockets/wsproto_impl.py
@@ -65,7 +65,7 @@ class WSProtocol(asyncio.Protocol):
 
     def connection_lost(self, exc):
         if exc is not None:
-            self.queue.put_nowait({"type": "websocket.disconnect"})
+            self.queue.put_nowait({"type": "websocket.disconnect", "code": exc.code})
         self.connections.remove(self)
 
         if self.logger.level <= TRACE_LOG_LEVEL:
@@ -262,7 +262,7 @@ class WSProtocol(asyncio.Protocol):
                 self.transport.write(output)
 
             elif message_type == "websocket.close":
-                self.queue.put_nowait({"type": "websocket.disconnect", "code": None})
+                self.queue.put_nowait({"type": "websocket.disconnect", "code": message.get("code", 1000)})
                 self.logger.info(
                     '%s - "WebSocket %s" 403',
                     self.scope["client"],
--- a/uvicorn/protocols/websockets/wsproto_impl.py
+++ b/uvicorn/protocols/websockets/wsproto_impl.py
@@ -262,7 +262,9 @@ class WSProtocol(asyncio.Protocol):
                 self.transport.write(output)
 
             elif message_type == "websocket.close":
-                self.queue.put_nowait({"type": "websocket.disconnect", "code": message.get("code", 1000)})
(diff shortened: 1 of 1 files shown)


uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -262,7 +262,7 @@ async def send(self, message):
                 self.transport.write(output)
 
             elif message_type == "websocket.close":
-                self.queue.put_nowait({"type": "websocket.disconnect", "code": None})
+                self.queue.put_nowait({"type": "websocket.disconnect", "code": message.get("code", 1000)})
```
Kludex:
```suggestion
                self.queue.put_nowait(
                    {"type": "websocket.disconnect", "code": message.get("code", 1000)}
                )
```

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -65,7 +65,7 @@ def connection_made(self, transport):
 
     def connection_lost(self, exc):
         if exc is not None:
-            self.queue.put_nowait({"type": "websocket.disconnect"})
+            self.queue.put_nowait({"type": "websocket.disconnect", "code": exc.code})
```
Kludex: It's not guaranteed that the exception will have a code. I don't know from where I got that from... It works with Starlette because the exception has a `code` field...
```suggestion
            self.queue.put_nowait({"type": "websocket.disconnect", "code": 1000})
```

adriangb:
> this is actually being tested... against none and 1000... maybe we should be more restrictive on the test...

Would you mind highlighting the relevant test?

euri10:
are you referring to those tests @Kludex  ?
https://github.com/encode/uvicorn/blob/9c57549e054d8c58b53fa25217155c2f1c8ee20f/tests/protocols/test_websocket.py#L453-L492

if this is the case how come the test was not failing for wsproto ?

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -65,7 +65,7 @@ def connection_made(self, transport):
 
     def connection_lost(self, exc):
         if exc is not None:
-            self.queue.put_nowait({"type": "websocket.disconnect"})
+            self.queue.put_nowait({"type": "websocket.disconnect", "code": 1000})
```
Kludex:
```suggestion
            self.queue.put_nowait({"type": "websocket.disconnect", "code": 1005})
```

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -262,7 +262,9 @@ async def send(self, message):
                 self.transport.write(output)
 
             elif message_type == "websocket.close":
-                self.queue.put_nowait({"type": "websocket.disconnect", "code": None})
+                self.queue.put_nowait(
+                    {"type": "websocket.disconnect", "code": message.get("code", 1000)}
```
Kludex:
```suggestion
                    {"type": "websocket.disconnect", "code": message.get("code", 1005)}
```

Kludex:
Working on it on another PR.

Issue #1317, opened 2022-01-10
OSError occurs when workers > 1


mnghiap:
### Checklist

- [X] The bug is reproducible against the latest release or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

This bug outputs an error similar to [#529](https://github.com/encode/uvicorn/issues/529) but occurs on different parameters of uvicorn. It occurs when I run uvicorn with the following settings:
`uvicorn.run("main:app", host="0.0.0.0", port=8000, workers=2, reload=False, loop="asyncio")`
I received the following exceptions:

INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
INFO:     Started parent process [16448]
INFO:     Started server process [17848]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Started server process [8104]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
Accept failed on a socket
socket: <asyncio.TransportSocket fd=3552, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('0.0.0.0', 8000)>
Traceback (most recent call last):
  File "C:\Users\phanm\AppData\Local\Programs\Python\Python38\lib\asyncio\proactor_events.py", line 834, in loop
    f = self._proactor.accept(sock)
  File "C:\Users\phanm\AppData\Local\Programs\Python\Python38\lib\asyncio\windows_events.py", line 549, in accept
    self._register_with_iocp(listener)
  File "C:\Users\phanm\AppData\Local\Programs\Python\Python38\lib\asyncio\windows_events.py", line 718, in _register_with_iocp
    _overlapped.CreateIoCompletionPort(obj.fileno(), self._iocp, 0, 0)
OSError: [WinError 87] The parameter is incorrect

Note that the app is still working completely fine. Setting either `workers=1` or `reload=True` makes the exceptions go away, while the app still works perfectly fine. Setting `workers=n` to any number `n>1` makes the exception appear `n-1` times. For example, with `uvicorn.run("main:app", host="0.0.0.0", port=8000, workers=4, reload=False, loop="asyncio")`:

INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
INFO:     Started parent process [14188]
INFO:     Started server process [18744]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Started server process [25288]
[...]

PR #1443, opened 2022-04-06
Fix OSError occurs when workers > 1 #1317

smitchell6879:
Should fix OSError

### DIFF
--- a/uvicorn/loops/asyncio.py
+++ b/uvicorn/loops/asyncio.py
@@ -12,4 +12,12 @@ def asyncio_setup(reload: bool = False) -> None:  # pragma: no cover
                 "The --reload flag should not be \
                 used in production on Windows."
             )
-        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+        try:
+            from asyncio import WindowsSelectorEventLoopPolicy
+        except ImportError:
+            logger.error("Can't assign a policy which doesn't exist.")
+        else:
+            if not isinstance(
+                asyncio.get_event_loop_policy(), WindowsSelectorEventLoopPolicy
+            ):
+                asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())
--- a/uvicorn/loops/asyncio.py
+++ b/uvicorn/loops/asyncio.py
@@ -6,6 +6,7 @@ logger = logging.getLogger("uvicorn.error")
 
 
 def asyncio_setup(reload: bool = False) -> None:  # pragma: no cover
-    if sys.version_info >= (3, 8) and sys.platform == "win32" and reload:
-        logger.warning("The --reload flag should not be used in production on Windows.")
+    if sys.version_info >= (3, 8) and sys.platform == "win32":
+        if reload:
+            logger.warning("The --reload flag should not be used in production on Windows.")
         asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
--- a/uvicorn/loops/asyncio.py
+++ b/uvicorn/loops/asyncio.py
@@ -8,5 +8,6 @@ logger = logging.getLogger("uvicorn.error")
 def asyncio_setup(reload: bool = False) -> None:  # pragma: no cover
     if sys.version_info >= (3, 8) and sys.platform == "win32":
         if reload:
-            logger.warning("The --reload flag should not be used in production on Windows.")
+            logger.warning("The --reload flag should not be \
+                used in production on Windows.")
         asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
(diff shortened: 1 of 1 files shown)


smitchell6879:
I am using python 3.8 with 3 workers on windows 10. Is there any more info on why it failed the test?

Kludex:
linter https://github.com/encode/uvicorn/runs/5849365812?check_suite_focus=true

smitchell6879:
> linter https://github.com/encode/uvicorn/runs/5849365812?check_suite_focus=true

No log available. So whats my next step?

Kludex:
Logs are available...
image

euri10:
> > linter https://github.com/encode/uvicorn/runs/5849365812?check_suite_focus=true
> 
> No log available. So whats my next step?

you can run the lint script in the scripts folder

smitchell6879:
> > > linter https://github.com/encode/uvicorn/runs/5849365812?check_suite_focus=true
> > 
> > 
> > No log available. So whats my next step?
> 
> you can run the lint script in the scripts folder

So I finally figured out how to run the lint script, so it should be good now. Sorry for the my ignorance.

uvicorn/loops/asyncio.py
```diff
@@ -6,6 +6,10 @@
 
 
 def asyncio_setup(reload: bool = False) -> None:  # pragma: no cover
-    if sys.version_info >= (3, 8) and sys.platform == "win32" and reload:
-        logger.warning("The --reload flag should not be used in production on Windows.")
+    if sys.version_info >= (3, 8) and sys.platform == "win32":
+        if reload:
+            logger.warning(
+                "The --reload flag should not be \
+                used in production on Windows."
```
Kludex, line 13:
```suggestion
                "The --reload flag should not be "
                "used in production on Windows."
```

smitchell6879:
Is there something else I need to fix or does this look good?

Kludex:
I'm going to check this tonight.

Kludex:
This is basically a revert of #1257, which was in place to solve other issues...

euri10:
ok this one I dont get it, what is the purpose of the PR, please describe the issue

smitchell6879:
> ok this one I dont get it, what is the purpose of the PR, please describe the issue

The issue is #1317. On windows when running multiple workers without reload true you would get a oserror. By writing it this way you still provide the warning if reload is true but you no longer get the oserror.

Kludex:
> This is basically a revert of #1257, which was in place to solve other issues...

Is my statement false?

smitchell6879:
I am assuming  no your statement is not false but I am unsure of what issues the change fixed when it also created a new issue. I am currently home sick when I feel better I will dig into the code more to try to figure out the issue that this change fixed and see if I can recreate it.

Kludex:
Hope you get better soon. 😟

smitchell6879:
So I have linked the log to show the errors I am getting without the above change. I am using a custom logger so [Backend] is actually the uvicorn channel. https://pastebin.com/cXQuG3SH
Reviewing the code it seems to me without this change you are not setting the `asyncio.set_event_loop_policy` at all unless you have the reload parameter set to True in which case it shouldn't matter as much because you go back you only 1 worker. As you see in the paste bin I have reload set to False and workers set to 5 I forgot to add the python version which is 3.8 .
I am not going to act like I full understand `asyncio.set_event_loop_policy`  But after digging around the net I did find "https://github.com/tornadoweb/tornado/issues/2608" so I have updated to their teams fix. This does not give me any errors.

euri10:
I think https://github.com/encode/uvicorn/pull/1454 is a better alternative, ymmv

smitchell6879:
Thats cool too as long as it fixes the issue.

smitchell6879:
Has anyone had a chance to review either of these and decide which fix you prefer?

smitchell6879:
Any chance for a update on this issue?

Kludex:
This solution reverts previous changes. As mentioned on https://github.com/encode/uvicorn/issues/1220#issuecomment-947920138, we should use the `ProactorEventLoop` whenever is possible. The case on which we don't want to use it is when we have either `reload` being `True` or the value of `workers` being higher than 1.

The ideal solution here would be to just make that check. I think we can even check if we're running on a child process instead of passing the `reload`/`workers` around.

As a note, the warning saying that reload should not be used in production is misplaced. It's not because we got to that point that we are in production. The warning should be removed.

In any case, I'll be closing this because of what I said on the first paragraph. Let's continue this on https://github.com/encode/uvicorn/pull/1454/ . Thanks for the efforts here @smitchell6879 ! 🙇 

References:
- https://github.com/encode/uvicorn/issues/529
- https://github.com/encode/uvicorn/issues/1220
- https://github.com/encode/uvicorn/issues/1317
- https://github.com/encode/uvicorn/pull/1454/
- https://github.com/encode/uvicorn/pull/535/

PR #1445, opened 2022-04-13
format the host when x_forwarded_for contains ip:port

zffocussss:
1. pass the right format of host when its host:port,especially behind a multi reverse proxy


euri10:
@zffocussss please open a discussion first on what you think may be an issue, we'll deal with a PR later on

PR #1446, opened 2022-04-13
refactoring the files inside supervisors folder

yernazarov:
Found unused parameters that cause warnings and removed them.


Kludex:
Those values are being received by those functions.

Issue #1358, opened 2022-02-03
`packaging` module missing?


dantownsend:
### Checklist

- [X] The bug is reproducible against the latest release or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

When installing the latest version from PyPI (0.17.2) I get an error that `packaging` doesn't exist:

https://github.com/encode/uvicorn/blob/41e339dd953a00d071c8bb5704f4671564f9f6e1/uvicorn/protocols/websockets/wsproto_impl.py#L7

I looked at setup.py, and couldn't find any reference to `packaging`. Presumably it's meant to be this module:

https://pypi.org/project/packaging/

It could just be me - but thought I'd report just in case.

### Steps to reproduce the bug

`pip install uvicorn` - try and run any ASGI app

### Environment

Mac OS, Python 3.10

lovelydinosaur:
Thanks for reporting this @dantownsend - fix incoming in #1359

lovelydinosaur:
Now resolved in 0.17.3 - #1361

PR #1456, opened 2022-04-23
Check extra package version

### LINKED ISSUE #1453: Add extra package version verification before running it
Check the version of packages that are not mandatory.

This causes issues that can have an infinity range of possible exceptions.

Kludex:
- Closes #1453

Changes:
- Vendorized `packaging.version` and `packaging._structures` modules.
- Use `packaging.version.parse` to get the version of the packages, and be able to compare them.
- Ignore vendorized package, and comparison conditionals on `coverage`.

Questions:
- Do we really want this?
- Isn't it cleaner to just include `setuptools` as dependency?
- Do we have alternatives here? Not checking the version is really an option?

### DIFF
--- a/uvicorn/_vendor/packaging/_structures.py
+++ b/uvicorn/_vendor/packaging/_structures.py
@@ -1,0 +1,61 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+
+class InfinityType:
+    def __repr__(self) -> str:
+        return "Infinity"
+
+    def __hash__(self) -> int:
+        return hash(repr(self))
+
+    def __lt__(self, other: object) -> bool:
+        return False
+
+    def __le__(self, other: object) -> bool:
+        return False
+
+    def __eq__(self, other: object) -> bool:
+        return isinstance(other, self.__class__)
+
+    def __gt__(self, other: object) -> bool:
+        return True
+
+    def __ge__(self, other: object) -> bool:
+        return True
+
+    def __neg__(self: object) -> "NegativeInfinityType":
+        return NegativeInfinity
+
+
+Infinity = InfinityType()
+
+
+class NegativeInfinityType:
+    def __repr__(self) -> str:
+        return "-Infinity"
+
+    def __hash__(self) -> int:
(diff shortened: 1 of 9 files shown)


uvicorn/_vendor/packaging/_structures.py
```diff
@@ -0,0 +1,85 @@
...
+        return True
+
+    def __ge__(self, other: object) -> bool:
+        return True
+
+    def __neg__(self: object) -> "NegativeInfinityType":
+        return NegativeInfinity
+
+
+Infinity = InfinityType()
```
ahopkins, line 56: `float("inf")`?

adriangb:
I think vendoring makes sense when it's unlikely that users would otherwise have the dependency installed. I don't think that's the case here, there's a good chance they already have packaging as a dependency. Is there another reason we'd want to vendor it?

If we did vendor, could we use the knowledge that we only need to support checking versions on certain packages to reduce the amount of code needed? Versioning is a tricky thing, and I think `packaging` tries to handle it in it's full complexity, but don't need that right?

Kludex:
> I think vendoring makes sense when it's unlikely that users would otherwise have the dependency installed. I don't think that's the case here, there's a good chance they already have packaging as a dependency. Is there another reason we'd want to vendor it?

Well... https://github.com/encode/uvicorn/issues/1358

adriangb:
I don't understand. Making a mistake and forgetting to include a dependency is a reason to vendor it? Sorry if I'm missing context

Kludex:
This PR is vendorizing because of project history mainly. We tried to maintain the core `uvicorn` with minimal dependencies. I know it's a different case, but today, I think adding `asgiref` was a mistake, for example.

Should we add `packaging` as dependency instead?

Seems pretty uncool to add a mandatory dependency to check the version of the dependencies that are not even mandatory... Right?

adriangb:
You have a good point. The situation is a bit different though, I think it's likely that `asgiref` would already exist in the dependency graph.

That said, can we solve the problem without vendoring of `packaging`? In other words, would a simple `__version__.split(".")[:3]` work for us? I realize this would fail for a _lot_ of version parsing, but if we're talking about 2 packages with known versioning schemes, maybe it works?

Kludex:
Let's proceed with #1461

PR #1457, opened 2022-04-24
Add support for max requests jitter

adnaanbheda:
Also, log when max requests is reached so that we can be sure about max requests being respected by Gunicorn


euri10:
Gonna close this since you can achieve it I think by subclassing the UvicornWorker, look for the docs on the topic

adnaanbheda:
@euri10 It would be nice to provide a link to the docs on the topic before preemptively closing the PR.

```
  class MyUvicornWorker(UvicornWorker):
      CONFIG_KWARGS = {"loop": "asyncio", "http": "h11", "lifespan": "off"}
```

Are you saying that I can pass `max_requests_jitter` to the config dict above to achieve the same result ?

euri10:
Yes that's the way I think you can do this, we'll reopen if needed, please raise a discussion or issue if that's the case

Kludex:
We can open an issue on how to document properly on our docs, if that's the case.

PR #1458, opened 2022-04-26
Match coverage between systems

Kludex:
Changes:
- Move `coverage report` settings to `setup.cfg`.
- Match coverage between Python versions and platforms.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -72,7 +72,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 97.37
+fail_under = 97.34
 show_missing = true
 skip_covered = true
 exclude_lines =
--- a/setup.cfg
+++ b/setup.cfg
@@ -72,6 +72,9 @@ plugins =
 
 [coverage:report]
 precision = 2
+fail_under = 97.37
+show_missing = true
+skip_covered = true
 exclude_lines =
     pragma: no cover
     pragma: nocover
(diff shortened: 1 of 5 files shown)


setup.cfg
```diff
@@ -72,6 +72,9 @@ plugins =
 
 [coverage:report]
 precision = 2
+fail_under = 97.33
```
br3ndonland, line 75:
```suggestion
fail_under = 97
```

Just a small suggestion here - could we just round down to the nearest integer? I know our eventual goal is 100% (encode/uvicorn#102), but in the meantime, do we need the decimal places?

I'm thinking about new contributors I guess. It's a little intimidating to see all the CI checks fail just because coverage decreased by 0.05%.

Related: encode/uvicorn#1062

br3ndonland (approved):
Just one small suggestion, otherwise LGTM

setup.cfg
Kludex: There was an issue on coverage. I didn't notice at the time. I've fixed it on https://github.com/nedbat/coveragepy/pull/1317. It only took me 9 months to realize the issue. 😎

Kludex: The thing is that over the year, there were countless PRs dropping the percentage... It feels kind tiring have to fix those. 🙁

Kludex: And one last thing... This will be useful as self motivation. I'll get this to 100% fast enough! 👨‍🏭🔨

PR #1460, opened 2022-04-26
Remove `asgiref` from minimal requirements

Kludex:
This PR removes `asgiref` from the minimal requirements.

Changes:
- Leverage `from __future__ import annotations` to get a small diff - besides the [_collateral benefits_](https://lukasz.langa.pl/61df599c-d9d8-4938-868b-36b67fdb4448/).
- Move `asgiref` from `minimal_requirements` to `extra_requirements`.

### DIFF
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -1,11 +1,15 @@
-from typing import List, Union
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, List, Union
 
 import httpx
 import pytest
-from asgiref.typing import ASGIReceiveCallable, ASGISendCallable, Scope
 
 from tests.response import Response
 from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
+
+if TYPE_CHECKING:
+    from asgiref.typing import ASGIReceiveCallable, ASGISendCallable, Scope
 
 
 async def app(
--- a/uvicorn/middleware/asgi2.py
+++ b/uvicorn/middleware/asgi2.py
@@ -1,9 +1,14 @@
-from asgiref.typing import (
-    ASGI2Application,
-    ASGIReceiveCallable,
-    ASGISendCallable,
-    Scope,
-)
+from __future__ import annotations
+
+import typing
+
+if typing.TYPE_CHECKING:
+    from asgiref.typing import (
+        ASGI2Application,
+        ASGIReceiveCallable,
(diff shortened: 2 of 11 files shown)


br3ndonland:
In general I think this looks fine, and I'm glad to see Uvicorn entering the `__future__`.

There's been a lot of discussion about lumping all the optional dependencies into `uvicorn[standard]` though, and asgiref isn't directly related to the other ones. Do we want to make people install all the optional dependencies just to do type checking? I'm not sure.

Related:

- encode/uvicorn#219
- encode/uvicorn#1274

Kludex:
You have a good point. Then, are we against having it as optional but not on the standard extras? 🤔

uvicorn/middleware/asgi2.py
```diff
@@ -1,9 +1,14 @@
...
+
+import typing
+
+if typing.TYPE_CHECKING:
+    from asgiref.typing import (
+        ASGI2Application,
+        ASGIReceiveCallable,
+        ASGISendCallable,
+        Scope,
+    )
```
adriangb, line 11: This is quite a bit of boilerplate that we'll have to have in every module

tests/middleware/test_proxy_headers.py
```diff
@@ -1,12 +1,16 @@
-from typing import List, Union
+from __future__ import annotations
```
adriangb, line 1: Given that this `__future__` is in somewhat of a limbo, I would be reluctant to merge it.

adriangb (commented):
Is this worth it? I understand that Uvicorn wants to be minimal on dependencies, but I'm not sure the internal complexity with `TYPE_CHECKING` is worth it. I also agree that due to how the extras are currently specified it's problematic to ask users to install a bunch of dependencies just to get type checking, and in fact that is contrary to the goal of not forcing dependencies on users (granted, users could look into the extra and cherry pick what they want but that's quite a bit to ask and defeats the point of extras in the first place). Somewhat tangential, but I think this would be a good fit for a `default-extras` if it existed.

Kludex:
> Is this worth it? I understand that Uvicorn wants to be minimal on dependencies, but I'm not sure the internal complexity with TYPE_CHECKING is worth it.

This is more for `django-channel` users than to be minimal (but also): https://github.com/encode/uvicorn/issues/1305.

Kludex:
- Let's discuss it first on https://github.com/encode/uvicorn/discussions/1527

PR #1461, opened 2022-04-27
Check extra package version

### LINKED ISSUE #1453: Add extra package version verification before running it

Kludex:
- Alternative to #1456
- Closes #1453 

Changes:
- Create the `_utils.py` module - this modules has the `version_parse()` function.
- Add tests to make sure `version_parse()` works in the exact same way as `packaging.version.parse` considering the extra packages we have.

Notes: 
- Should I change the tests to not get all possible versions, but actually get the cases we are covering, so we don't need to get them from pypi each time we run the tests? 🤔 Looks reasonable, right?


adriangb (dismissed):

tests/test__utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def versions(pkg_name: str):
+    """Retrieve all the versions from a given pkg_name.
+
+    Extracted from https://stackoverflow.com/a/40745656/13087720.
+    """
+    url = f"https://pypi.python.org/pypi/{pkg_name}/json"
+    releases = json.loads(request.urlopen(url).read())["releases"]
+    return sorted(releases, key=parse_version, reverse=True)
```
Kludex, line 36: I'm calling pypi on every test here. This should be removed.

euri10 (commented):
the only thing that worries me with this PR is the amount  of "manual" checks we'll have to do when bumping versions, I'm not too familiar with the setup.py or packaging generally speaking, but isn't there a more automated way to do that ?

This is a runtime check right, so the issue we're trying to deal with in this PR is a given user installing uvicorn with the right set of packages, then manually updating some parts and messing up the versions ?

Kludex:
> the only thing that worries me with this PR is the amount of "manual" checks we'll have to do when bumping versions, I'm not too familiar with the setup.py or packaging generally speaking, but isn't there a more automated way to do that ?

If we add `setuptools` as dependency, we can use `packaging`... But then... It's really worth having the package for this?

> This is a runtime check right, so the issue we're trying to deal with in this PR is a given user installing uvicorn with the right set of packages, then manually updating some parts and messing up the versions ?

I'm not sure if I got what you said. We're trying to avoid old the usage of old versions of extra packages.

Kludex:
Let's not do this... Too _workaroundy_.

PR #1476, opened 2022-05-06
Replace asgiref for native types

### LINKED ISSUE #1305: `asgiref` backward compatibility

Kludex:
- Alternative to #1460
- Closes #1305

## Questions
- Where should I add the `asgiref` copyright?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,6 @@ env_marker_win = "sys_platform == 'win32'"
 env_marker_below_38 = "python_version < '3.8'"
 
 minimal_requirements = [
-    "asgiref>=3.4.0",
     "click>=7.0",
     "h11>=0.8",
     "typing-extensions;" + env_marker_below_38,
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -2,9 +2,9 @@ from typing import List, Union
 
 import httpx
 import pytest
-from uvicorn._types import ASGIReceiveCallable, ASGISendCallable, Scope
 
 from tests.response import Response
+from uvicorn._types import ASGIReceiveCallable, ASGISendCallable, Scope
 from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
 
 
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -2,7 +2,7 @@ from typing import List, Union
 
 import httpx
 import pytest
-from asgiref.typing import ASGIReceiveCallable, ASGISendCallable, Scope
+from uvicorn._types import ASGIReceiveCallable, ASGISendCallable, Scope
 
 from tests.response import Response
 from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
--- a/tests/middleware/test_wsgi.py
+++ b/tests/middleware/test_wsgi.py
@@ -4,7 +4,7 @@ from typing import AsyncGenerator, List
 
 import httpx
 import pytest
-from asgiref.typing import HTTPRequestEvent, HTTPScope
+from uvicorn._types import HTTPRequestEvent, HTTPScope
 
 from uvicorn._types import Environ, StartResponse
 from uvicorn.middleware.wsgi import WSGIMiddleware, build_environ
--- a/tests/middleware/test_wsgi.py
+++ b/tests/middleware/test_wsgi.py
@@ -4,9 +4,8 @@ from typing import AsyncGenerator, List
 
 import httpx
 import pytest
-from uvicorn._types import HTTPRequestEvent, HTTPScope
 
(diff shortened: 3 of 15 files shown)


adriangb:
First off thank you for all of these great PRs. I can tell you're putting a great deal of work into trying out all of the alternatives.

I like the idea of having our own types vs. #1460. Two questions:

1. Will this break integrations with other packages, e.g. a middleware that uses the `asgiref` types? Or will it still work as long as they are "equivalent"?
1. What do you think would be the pros/cons of having simpler types, like we have in Starlette? I think we're mainly using these internally within the codebase, so if it's just for us, do we need the more complex types to catch our mistakes, or would the simpler types be enough?

Kludex:
> Will this break integrations with other packages, e.g. a middleware that uses the asgiref types? Or will it still work as long as they are "equivalent"?

It will work as long as the types are equivalent.

```python
from typing import Protocol, cast


class AsgirefType(Protocol):
    def __call__(self, a: int) -> None:
        ...


class UvicornType(Protocol):
    def __call__(self, a: int) -> None:
        ...


def uvicorn_code(callable: UvicornType) -> None:
    ...


def lamb(a: int) -> None:
    ...


asgiref_callable = cast(AsgirefType, lamb)
uvicorn_code(asgiref_callable)  # uvicorn_code expects UvicornType, but we are giving AsgirefType
```

The above works - no issues with mypy or pyright.

> What do you think would be the pros/cons of having simpler types, like we have in Starlette? I think we're mainly using these internally within the codebase, so if it's just for us, do we need the more complex types to catch our mistakes, or would the simpler types be enough?

I can see a path from the less strict types to the more strict ones, but I don't see any reason to follow this now.

Kludex:
- Let's discuss it first on https://github.com/encode/uvicorn/discussions/1527

PR #1484, opened 2022-05-09
Annotate `httptools_impl.py`

Kludex:
- Replaces #1102
- Similar to #1397

### DIFF
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -2,12 +2,26 @@ import asyncio
 import http
 import logging
 import re
+import sys
 import urllib
+from asyncio.events import TimerHandle
 from collections import deque
+from typing import Callable, Deque, List, Optional, Tuple, Union, cast
 
 import httptools
+from asgiref.typing import (
+    ASGI3Application,
+    ASGIReceiveEvent,
+    ASGISendEvent,
+    HTTPDisconnectEvent,
+    HTTPRequestEvent,
+    HTTPResponseBodyEvent,
+    HTTPResponseStartEvent,
+    HTTPScope,
+)
 
 from uvicorn._logging import TRACE_LOG_LEVEL
+from uvicorn.config import Config
 from uvicorn.protocols.http.flow_control import (
     CLOSE_HEADER,
     HIGH_WATER_LIMIT,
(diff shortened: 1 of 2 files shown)


uvicorn/protocols/http/httptools_impl.py
```diff
@@ -66,21 +91,24 @@ def __init__(self, config, server_state, _loop=None):
         self.default_headers = server_state.default_headers
 
         # Per-connection state
-        self.transport = None
-        self.flow = None
-        self.server = None
-        self.client = None
-        self.scheme = None
-        self.pipeline = deque()
+        # Per-connection state
```
Kludex:
```suggestion
```

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -66,21 +91,23 @@ def __init__(self, config, server_state, _loop=None):
...
 
         # Per-connection state
-        self.transport = None
-        self.flow = None
-        self.server = None
-        self.client = None
-        self.scheme = None
-        self.pipeline = deque()
+        self.transport: asyncio.Transport = None  # type: ignore[assignment]
+        self.flow: FlowControl = None  # type: ignore[assignment]
```
adriangb, line 95: Are you using the `# type: ignore[assignment]` just to avoid some `assert self.transport is not None`s later?

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -21,12 +35,18 @@
     get_remote_addr,
     is_ssl,
 )
+from uvicorn.server import ServerState
+
+if sys.version_info < (3, 8):  # pragma: py-gte-38
+    from typing_extensions import Literal
+else:  # pragma: py-lt-38
+    from typing import Literal
```
adriangb, line 43: do we have a `_compat.py` module or something where this sorta boilerplate can go?

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -391,25 +418,27 @@ async def run_asgi(self, app):
...
-                "headers": [
-                    (b"content-type", b"text/plain; charset=utf-8"),
-                    (b"connection", b"close"),
-                ],
-            }
-        )
-        await self.send(
-            {"type": "http.response.body", "body": b"Internal Server Error"}
-        )
+            self.on_response = lambda: None
```
adriangb, line 421: This went from `self.on_response = None` to self.on_response = lambda: None`. The new version sounds more correct. Was this a bug?

adriangb (approved):

uvicorn/protocols/http/httptools_impl.py
Kludex: Good question. That was introduced on https://github.com/encode/uvicorn/pull/189. I don't know if the line itself is really necessary (?)...

uvicorn/protocols/http/httptools_impl.py
Kludex: No

uvicorn/protocols/http/httptools_impl.py
Kludex: Yes

Kludex:
Thanks for the review @adriangb 🙏

PR #1486, opened 2022-05-10
Remove conditional on `RemoteProtocolError.event_hint`

Kludex:
I'm not sure if this is the right `code`, but `events.CloseConnection` receives `code` as a mandatory field, so without it, you'll have an exception on missing parameters. 

List of codes: https://www.iana.org/assignments/websocket/websocket.xhtml
RFC: https://www.rfc-editor.org/rfc/rfc6455.html

I've noticed this issue while annotating `wsproto_impl`.

*EDIT*: Quite funny... `wsproto` doesn't raise a `RemoteProtocolError` without an `event_hint`. Check their source code: https://github.com/python-hyper/wsproto/.

### DIFF
--- a/uvicorn/protocols/websockets/wsproto_impl.py
+++ b/uvicorn/protocols/websockets/wsproto_impl.py
@@ -91,7 +91,7 @@ class WSProtocol(asyncio.Protocol):
                 self.transport.write(self.conn.send(err.event_hint))
                 self.transport.close()
             else:
-                self.handle_no_connect(events.CloseConnection())
+                self.handle_no_connect(events.CloseConnection(code=1007))
         else:
             self.handle_events()
 
--- a/uvicorn/protocols/websockets/wsproto_impl.py
+++ b/uvicorn/protocols/websockets/wsproto_impl.py
@@ -87,13 +87,8 @@ class WSProtocol(asyncio.Protocol):
         try:
             self.conn.receive_data(data)
         except RemoteProtocolError as err:
-            if err.event_hint is not None:
-                self.transport.write(self.conn.send(err.event_hint))
-                self.transport.close()
-            else:
-                # Response with the "1002 Protocol Error" code.
-                # See https://www.iana.org/assignments/websocket/websocket.xhtml#close-code-number
-                self.handle_no_connect(events.CloseConnection(code=1002))
+            self.transport.write(self.conn.send(err.event_hint))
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Well spotted.

How about we use 1002 "Protocol Error"? That seems like it'd fit quite well here.

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -86,7 +86,7 @@ def data_received(self, data):
                 self.transport.write(self.conn.send(err.event_hint))
                 self.transport.close()
             else:
-                self.handle_no_connect(events.CloseConnection())
+                self.handle_no_connect(events.CloseConnection(code=1007))
```
lovelydinosaur:
```suggestion
                # Response with the "1002 Protocol Error" code.
                # See https://www.iana.org/assignments/websocket/websocket.xhtml#close-code-number
                self.handle_no_connect(events.CloseConnection(code=1002))
```

Kludex: That makes sense. Thanks! 🙇

I'll update later on.

lovelydinosaur: Cool.

Lazy option if you're happy with the suggestion. Just whack the big 'ol button.

Screen Shot 2022-05-11 at 10 58 42

(You probs already know this, but mentioning it in case it's a bit of the GitHub UI you've not noticed yet. It's such a super helpful tool.)

Kludex: Yeah, but when I saw your suggestion, I thought it would be better if I add a test here. 😅

Kludex:
The `handle_no_connect()` method ignores the `code` from the event anyway. I guess there needs to be some change there.

Kludex:
I've replaced the logic here. :sweat_smile:

lovelydinosaur (approved):
Do we pause this until there's a resolution to https://github.com/python-hyper/wsproto/issues/178, or are we okay with knowing that `event_hint` *isn't* ever `None` in this context, although it's annotated as `Optional`?

Kludex:
I'm fine with this being merged. Thanks @tomchristie 🙏 

As a note, `wsproto_impl.py` is the only file that is missing to add type annotation on `uvicorn` folder. I'll work on that this week.

PR #1487, opened 2022-05-11
Turn `raw_path` into bytes on both websockets implementations

Kludex:
`raw_path` should be `bytes`, not `string`.

See: https://asgi.readthedocs.io/en/latest/specs/www.html#websocket-connection-scope

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -220,7 +220,7 @@ async def test_path_and_raw_path(ws_protocol_cls, http_protocol_cls):
             path = self.scope.get("path")
             raw_path = self.scope.get("raw_path")
             assert path == "/one/two"
-            assert raw_path == "/one%2Ftwo"
+            assert raw_path == b"/one%2Ftwo"
             await self.send({"type": "websocket.accept"})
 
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
Looks like some easy to resolve test failures...

* https://github.com/encode/uvicorn/runs/6382208111?check_suite_focus=true#step:7:364
* https://github.com/encode/uvicorn/runs/6382208111?check_suite_focus=true#step:7:110

Assertions against `raw_path` that need updating.

Kludex:
> Looks like some easy to resolve test failures...
> 
> * https://github.com/encode/uvicorn/runs/6382208111?check_suite_focus=true#step:7:364
> * https://github.com/encode/uvicorn/runs/6382208111?check_suite_focus=true#step:7:110
> 
> Assertions against `raw_path` that need updating.

Right. The other implementation had the same issue. 👍

Kludex:
I can also use `partition(b"?")` instead, to be align with the HTTP implementations. I'll change this.

```python
        path_portion, _, query_string = path.partition("?")
```

abersheeran (approved):

PR #1488, opened 2022-05-15
Import application before running the event loop

Kludex:
- Replaces #1110
- Solves #346

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -57,6 +57,8 @@ class Server:
 
     def run(self, sockets: Optional[List[socket.socket]] = None) -> None:
         self.config.setup_event_loop()
+        if not self.config.loaded:
+            self.config.load()
         return asyncio.run(self.serve(sockets=sockets))
 
     async def serve(self, sockets: Optional[List[socket.socket]] = None) -> None:
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -2,6 +2,7 @@ import importlib
 import os
 import platform
 import sys
+import textwrap
 from pathlib import Path
 from textwrap import dedent
 from unittest import mock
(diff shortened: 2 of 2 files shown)


graingert:
how does this interact with reload?

Kludex:
@graingert With the simple application:
```python
import anyio


async def example():
    await anyio.sleep(0)
    print("All good!")


anyio.run(example)


    pass
```
Running `uvicorn` with `reload`:

```bash
(uvicorn3.8.12) ➜  uvicorn git:(feat/import-app-before-event-loop) ✗ uvicorn main:app --reload --port 8005
INFO:     Will watch for changes in these directories: ['/Users/marcelotryle/Development/encode/uvicorn']
INFO:     Uvicorn running on http://127.0.0.1:8005 (Press CTRL+C to quit)
INFO:     Started reloader process [6724] using watchgod
All good!
INFO:     Started server process [6738]
WARNING:  WatchGodReload detected file change in '['/Users/marcelotryle/Development/encode/uvicorn/main.py']'. Reloading...
INFO:     Shutting down
INFO:     Finished server process [6738]
All good!
INFO:     Started server process [6774]
```

Kludex:
Question for my future self... Does this work if `MongoClient` is instantiated? The event loop will differ, right? If that's the case, do we really want this change?

euri10 (approved):
lgtm thanks @Kludex

uvicorn/server.py
```diff
@@ -57,6 +57,8 @@ def __init__(self, config: Config) -> None:
 
     def run(self, sockets: Optional[List[socket.socket]] = None) -> None:
         self.config.setup_event_loop()
```
graingert, line 59: I think you should call setup_event_loop as close to asyncio.run as possible

graingert (changes requested):

uvicorn/server.py
Kludex: In other words... 🤔

Kludex:
I'm not able to reproduce the issue that motivated this PR.

I'm going to close this. If someone wants to try to reproduce, and come back with the proof again, we take a decision then.

PR #1489, opened 2022-05-15
Refactor uvicorn logger setup

### LINKED ISSUE #562: unfavorable logger name "uvicorn.error"
https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/websockets/websockets_impl.py

Every time a new websocket request comes in something like this appears in the logs:

    INFO  [uvicorn.error] ('127.0.0.1', 47014) - "WebSocket /api" [accepted]

This is slightly confusing since it is in fact not an error. What's the purpose of naming it uvicorn.error ? We noticed this when ossec kept triggering on every incoming request spamming us with false alarms.

There are other questionable occurences like

    INFO  [uvicorn.error] Application startup complete.

> [!IMPORTANT]

<a href="https://polar.sh/encode/uvicorn/issues/562">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/562/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

Kludex:
- Closes #562

## Changes

- Add `uvicorn.http` and `uvicorn.websockets` loggers.
- Replace `uvicorn.error` by `uvicorn.server` logger.

## Missing

- [ ] Docs update.
- [ ] Check  with `UvicornWorker`.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -205,8 +205,7 @@ There are several ways to run uvicorn directly from your application.
 
 If you're looking for a programmatic equivalent of the `uvicorn` command line interface, use `uvicorn.run()`:
 
-```python
-# main.py
+```py title="main.py"
 import uvicorn
 
 async def app(scope, receive, send):
@@ -275,9 +274,7 @@ For more information, see the [deployment documentation](deployment.md).
 
 The `--factory` flag allows loading the application from a factory function, rather than an application instance directly. The factory will be called with no arguments and should return an ASGI application.
 
-**example.py**:
-
-```python
+```py title="example.py"
 def create_app():
     app = ...
     return app
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -1,0 +1,1 @@
+# Logging
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -1,1 +1,40 @@
 # Logging
+
+Logging in Python is pretty complex, and there are many different ways to do it.
+This document is intended to provide an overview about the logger objects available in **Uvicorn**.
+
+### Loggers
+
+**Uvicorn** has many loggers available, each one with a different purpose.
+
+It contains an ancestor logger called `uvicorn`, and then multiple children loggers.
+
+#### Server Logger (`uvicorn.server`)
+
+This logger is used to log information about the core server functionality, such as startup and shutdown.
+
+#### WebSocket Logger (`uvicorn.websocket`)
+
+This logger is used to log server-side information about WebSocket protocol messages.
+
+#### HTTP Logger (`uvicorn.http`)
+
+This logger is used to log server-side information about HTTP protocol messages.
+
+#### Access Logger (`uvicorn.access`)
+
+This logger is used to log client-side information about each request/response cycle.
+
+#### ASGI Logger (`uvicorn.asgi`)
+
+This logger is used to log low-level server-side ASGI messages. Only available when `--log-level` is set to `trace`.
+
+### Configuration
+
+**Uvicorn** uses the standard Python `logging` module to handle logging.
+
+This section will be used to teach people how to configure uvicorn loggers.
+
+### Tutorials
+
+This section will be used to link tutorials.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -24,12 +24,18 @@ edit_uri: ""
 nav:
   - Introduction: "index.md"
   - Settings: "settings.md"
+  - Logging: "logging.md"
   - Deployment: "deployment.md"
   - Server Behavior: "server-behavior.md"
   - Contributing: "contributing.md"
 
 markdown_extensions:
   - admonition
+  - pymdownx.highlight:
+      anchor_linenums: true
+  - pymdownx.inlinehilite
+  - pymdownx.snippets
+  - pymdownx.superfences
   - codehilite:
       css_class: highlight
   - toc:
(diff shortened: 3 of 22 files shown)


br3ndonland (commented):
Thanks @Kludex, generally looks good.

Some users might be confused about the difference between `"uvicorn.access"` vs. `"uvicorn.http"`.

- `"uvicorn.access"` logs client-side HTTP access requests.
- `"uvicorn.http"` logs server-side HTTP protocol messages.

Did I get that right?

Do you think it might help to add some docs explaining what all the Uvicorn loggers do? Maybe on the [server behavior page](https://www.uvicorn.org/server-behavior/), cross-referencing it with the [logging settings section](https://www.uvicorn.org/settings/#logging)? I'm happy to update the docs in a separate PR if that makes more sense.

Looking at the loggers we currently have:

https://github.com/encode/uvicorn/blob/b566f4781b4904658b3c92ac4699d896d297026a/uvicorn/config.py#L411-L415

- `"uvicorn"`
  - Description: [Parent/ancestor logger](https://docs.python.org/3/library/logging.html#logger-objects) to which the child loggers (like `"uvicorn.server"`) will propagate.
  - File descriptor: `stderr`
- `"uvicorn.server"`
  - Description: Logs messages related to core Uvicorn server functionality. _(is there a better way to describe this?)_
  - File descriptor: `stderr`
- `"uvicorn.websockets"`
  - Description: Logs server-side websocket messages.
  - File descriptor: `stderr`
- `"uvicorn.http"`
  - Description: Logs server-side HTTP protocol messages.
  - File descriptor: `stderr`
- `"uvicorn.access"`
  - Description: Logs client-side HTTP requests.
  - File descriptor: `stdout`
- `"uvicorn.asgi"`
  - Description: Used with `MessageLoggerMiddleware` to log low-level server-side ASGI messages, only for trace-level logging (encode/uvicorn#474)
  - File descriptor: `stderr` _(??? There's no handler configured, so I guess it defaults to `stderr`?)_

Kludex:
> Did I get that right?

You did.

> Do you think it might help to add some docs explaining what all the Uvicorn loggers do? Maybe on the [server behavior page](https://www.uvicorn.org/server-behavior/), cross-referencing it with the [logging settings section](https://www.uvicorn.org/settings/#logging)? I'm happy to update the docs in a separate PR if that makes more sense.

I'd be happy to receive help. You can use this branch if you want, so we can merge the functionality with the docs together (I'm not possessive with my branches 👀 ).

But do we agree with those changes? I'm not sure about it myself... @br3ndonland @euri10 opinions?

br3ndonland:
Ok cool! I will get some docs pushed up here soon.

And definitely open to suggestions on the descriptions and such. I will defer to @euri10 for the direction we want to go overall.

euri10:
well I'd be in favor of NOT changing the name `uvicorn.error`

Not because I think the name is not confusing, it is.

But because it mimics what gunicorn does with gunicorn.error` and dozens of users, not to say hundreds or even thousands, are on a daily basis using, seeing this same exact name, so really I don't get the fuss on this (last commit on https://github.com/benoitc/gunicorn/blob/master/examples/logging.conf is 20 feb 2014) : people literally use this name for almost ten years now....

I don't think we gain anything in changing something that people are most likely used to now for a long time.

This said, I won't oppose it should people feel strongly about it, but to me there's almost no gain in changing, so I'll let you choose :)

Kludex:
I'm totally fine with not changing it. :)

Kludex:
Let's not do this then. I've closed #562. If someone has a good argument, and/or strategy to achieve it, feel free to comment on the issue.

br3ndonland:
That sounds good. I'm also fine with keeping the current logger name.

doncatnip:
> This said, I won't oppose it should people feel strongly about it, but to me there's almost no gain in changing, so I'll let you choose :)

I don't feel very strongly about it, as I don't work with uvicorn anymore - but I do find the reasoning to keep the current scheme questionable at best. Gunicorn using a very confusing and in fact problem-inducing (i.e. log monitoring) naming scheme is not a good reason to propagate that anti-pattern further. I also wonder if the extend of the confusion can be compared. Uvicorn greets the user with "[gunicorn.error] Application startup complete" for example.

alexsantos:
For me as a newcomer it was really confusing to see a server starting and greeting me with a bunch of errors saying things like
 `INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)`.
For me, stderr has been always to log errors and not informational messages. For instance, starting uvicorn like this:
`uvicorn main:app 2>error.log 1>out.log`
will result in a file error.log with the start information.
I'm using Google Cloud Run and Cloud Functions and this might trigger alerts that aren't real errors.

Kludex:
I've changed my mind. I'm strong about this change now. I'm reopening this PR.

I'll update the documentation.

Kludex:
I'm closing this as stale.

If someone wants to help me, and take this, I'd appreciate.

zanieb:
I'd consider taking a poke at this; what critical parts remain?

Kludex:
I think the "missing" section on the description. :eyes:

ltbd78:
Would push for this change, as I dug into a rabbit hole of confusion seeing the logger named as `uvicorn.error`

Like a user said earlier, what if if we rename `uvicorn.access` to`uvicorn.client` and `uvicorn.error` to `uvicorn.client`?

MinasA1:
Hi, is this planned to be merged? Can I help? @Kludex ?

PR #1437, opened 2022-04-01
Switch from `watchgod` to `watchfiles`

samuelcolvin:
See https://watchfiles.helpmanual.io/migrating/.

TL;DR; I've renamed `watchgod` to `watchfiles` and rewritten it to use the rust "notify" crate for all the heavy lifting. This PR is a WIP to see if you're willing to move ahead with a change like this.

Still TODO/to answer:
* I've changed the logic of `BaseReload` to try and make `StatReload` and watchfiles share an interface, are you happy with this?
* tests are a fair bit more complicated, are you happy with how I'm implemented/changed them?
* ~~I don't see an obvious way to deal with `--reload-include` and `--reload-exclude` being paths rather than extensions. Are you happy to drop this functionality? It's much less important now there's little overhead to scanning big directories, if not, we can implement something but it'll be complicated.~~ I think I've dealt with this a287d5c

PR #1490, opened 2022-05-16
Replaced watchgod with watchfiles

timkofu:
As it's [deprecated](https://pypi.org/project/watchgod/).

> The package under the name watchgod will not be further developed and will only receive critical security fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## 0.17.7 - 2022-05-18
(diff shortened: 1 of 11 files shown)


Kludex:
https://github.com/encode/uvicorn/pull/1437

timkofu:
Ah, alright.

Kludex:
Thanks for the PR @timkofu 😁

But as you see, that's already in progress 🙏

PR #1485, opened 2022-05-10
Annotate `wsproto_impl.py`

Kludex:

PR #1496, opened 2022-05-18
typing: add type hints for wsproto_impl module

jkklapp:
part of #998


Kludex:
Thanks @jkklapp for the interest! 

I have already started this file on #1485, so I'll proceed there.

Issue #1375, opened 2022-02-11
Add support for passing a queue to the run method to know when the server is ready.


tchaton:
### Checklist

- [X] There are no similar issues or pull requests for this yet.
- [X] I discussed this idea on the [community chat](https://gitter.im/encode/community) and feedback is positive.

### Is your feature related to a problem? Please describe.

I would like to know if the server is ready without hitting the endpoint. I would like to simply use a queue to know if the server is ready. 

### Describe the solution you would like.

Here is my current hack to know when the server is ready outside the main thread. The problem is the run method is changing on the uvicorn side and I can't patch it, otherwise I would be missing on latest features.

```py
class PatchServer(Server):
    def __init__(self, config, has_started_queue):
        super().__init__(config)
        self.has_started_queue = has_started_queue

    def run(self, sockets=None):
        self.config.setup_event_loop()
        loop = asyncio.get_event_loop()
        asyncio.ensure_future(self.serve(sockets=sockets))
        asyncio.ensure_future(self.check_is_started(self.has_started_queue))
        loop.run_forever()

    async def check_is_started(self, queue):
        while not self.started:
            await asyncio.sleep(1)
        queue.put(True)


# Code taken from https://github.com/encode/uvicorn/blob/master/uvicorn/main.py#L438
# TODO: Keep this updated to latest.
def uvicorn_run_fn(app, has_started_queue=None, **kwargs):
    config = Config(app, **kwargs)
    server = PatchServer(config=config, has_started_queue=has_started_queue)

    if (config.reload or config.workers > 1) and not isinstance(app, str):
        logger = logging.getLogger("uvicorn.error")
        logger.warning("You must pass the application as an import string to enable 'reload' or " "'workers'.")
        sys.exit(1)

    if config.should_reload:
        sock = config.bind_socket()
        supervisor = ChangeReload(config, target=server.run, sockets=[sock])
        supervisor.run()
    elif config.workers > 1:
        sock = config.bind_socket()
        supervisor = Multiprocess(config, target=server.run, sockets=[sock])
        supervisor.run()
    else:
        server.run()
```

### Describe alternatives you considered

I didn't consider any other.

### Additional context

None

Issue #1301, opened 2021-12-29
CTRL^C doesn't work while startup in progress


levsh:
### Checklist

- [X] The bug is reproducible against the latest release or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

Hi!

ctrl^c doesn't work until on_startup function completes. This happens because [install_signal_handlers](https://github.com/encode/uvicorn/blob/0.16.0/uvicorn/server.py#L80) call before [startup](https://github.com/encode/uvicorn/blob/0.16.0/uvicorn/server.py#L86)

### Steps to reproduce the bug

```python
import asyncio

import uvicorn
from fastapi import FastAPI

async def on_startup():
    for _ in range(10):
        await asyncio.sleep(1)

app = FastAPI(on_startup=[on_startup])

if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=5000, log_level="info")
```

### Environment

uvicorn - 0.16.0
python - 3.8.6

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/uvicorn/issues/1301">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/1301/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/uvicorn/issues/1301/pledge.svg">
</picture>
</a>

msenol86:
This happens to me too. This is my solution:
```python
def receive_signal(signalNumber, frame):
    print('Received:', signalNumber)
    sys.exit()


@app.on_event("startup")
async def startup_event():
    import signal
    signal.signal(signal.SIGINT, receive_signal)
    # startup tasks
```

lovelydinosaur:
This is probably a reasonable behaviour?

I'd assume that a sensible approach would be that we always allow the startup to run to completion, but then exit (and run shutdown) if we've been interrupted.

jmwielandt:
> This is probably a reasonable behaviour?
> 
> I'd assume that a sensible approach would be that we always allow the startup to run to completion, but then exit (and run shutdown) if we've been interrupted.

[...]

PR #1497, opened 2022-05-18
Server lifecycle

bwhmather:
Adds `close`, `wait_closed` and `start_serving` methods that match the semantics of `asyncio.Server`.
Adds `sockets` argument constructor in order to enable new methods to be used with existing bindings.

This makes it easier to use uvicorn as part of a larger application, by making it possible to block until a service has started, and easier to shut down cleanly.

For background, we use `uvicorn` extensively and are very happy with its behaviour in production, but have had a lot of difficulty with server lifecycle when trying to use it as part of our test suite.  In particular:
 1. It's not currently possible to block until the service has started without polling.
 2. Shutting a background server down and blocking until completion requires keeping a reference to both the server and an `asyncio.Task` wrapping `serve`.
 3. Waiting for the main loop to poll while shutting down currently dominates our test run time.
 4. Signal handler installation interferes with signal handlers from other instances, and with test shutdown.

It also took a fair bit of digging in the source code to figure out that calling `Server.shutdown` from outside does not do what one might expect.

I'd like to propose a few changes, the first of which I have attempted in this PR, which I think will resolve the issues we've encountered:

 1. Add `close`, `wait_closed` and `start_serving` methods that match the semantics of `asyncio.Server` (resolves issues 1 and 2).
 2. Use an `asyncio.Event` to make it so that `close` can pre-empt the polling loops (resolves issue 3).
 3. Add a `register_event_handlers` keyword argument to `serve` which can be used to disable registering of event handlers (resolves issue 4).
 4. Make any methods which would result in breakage if called from outside the class private.

If you are happy with this approach then I would be glad to work on 2, 3 and 4 as follow ups.

Resolves:
  - https://github.com/encode/uvicorn/issues/1301, with behaviour matching what @tomchristie suggests.
  - https://github.com/encode/uvicorn/issues/1375, although with a different API.
  - https://github.com/encode/uvicorn/issues/742

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,4 +1,3 @@
-import asyncio
 import os
 from contextlib import asynccontextmanager, contextmanager
 from pathlib import Path
@@ -8,14 +7,12 @@ from uvicorn import Config, Server
 
 @asynccontextmanager
 async def run_server(config: Config, sockets=None):
-    server = Server(config=config)
-    cancel_handle = asyncio.ensure_future(server.serve(sockets=sockets))
-    await asyncio.sleep(0.1)
+    server = Server(config=config, sockets=sockets)
+    await server.start_serving()
     try:
         yield server
     finally:
         await server.shutdown()
-        cancel_handle.cancel()
 
 
 @contextmanager
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -61,9 +61,9 @@ class Server:
 
         self._main_task: Optional[asyncio.Task] = None
 
-        #: Set once immediately after startup has completed and the server has
-        #: started listening for requests.
-        self._startup_event = asyncio.Event()
+        #: Created on demand and set once immediately after startup has
+        #: completed and the server has started listening for requests.
+        self._startup_event: Optional[asyncio.Event] = None
 
     def run(self, sockets: Optional[List[socket.socket]] = None) -> None:
         self.config.setup_event_loop()
@@ -100,6 +100,12 @@ class Server:
         Idempotent.  Can be called multiple times without creating multiple
         instances.
         """
+        if self._startup_event is None:
+            # We defer creating the startup event until start serving is called
+            # because there is no guarantee that the constructor will be called
+            # with an active loop.
+            self._startup_event = asyncio.Event()
+
         if self._main_task is None:
             self._main_task = asyncio.create_task(self._main())
 
@@ -214,6 +220,8 @@ class Server:
             pass
 
         self.started = True
+
+        assert self._startup_event is not None
         self._startup_event.set()
 
     def _log_started_message(self, listeners: Sequence[socket.SocketType]) -> None:
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -120,10 +120,10 @@ class Server:
         if self._main_task is not None:
             raise RuntimeError("cannot call serve on running server")
 
-        if sockets is not None and self._sockets is not None:
-            raise RuntimeError("cannot override already provided sockets list")
-
-        self._sockets = sockets
+        if sockets is not None:
+            if self._sockets is not None:
+                raise RuntimeError("cannot override already provided sockets list")
+            self._sockets = sockets
 
         self.install_signal_handlers()
         try:
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -46,7 +46,9 @@ class ServerState:
 
 
 class Server:
-    def __init__(self, config: Config) -> None:
+    def __init__(
+        self, config: Config, *, sockets: Optional[List[socket.socket]] = None
+    ) -> None:
         self.config = config
         self.server_state = ServerState()
 
@@ -55,7 +57,7 @@ class Server:
         self.force_exit = False
         self.last_notified = 0.0
 
-        self._sockets: Optional[List[socket.socket]] = None
+        self._sockets: Optional[List[socket.socket]] = sockets
 
         self._main_task: Optional[asyncio.Task] = None
 
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -55,6 +55,14 @@ class Server:
         self.force_exit = False
         self.last_notified = 0.0
 
+        self._sockets: Optional[List[socket.socket]] = None
+
+        self._main_task: Optional[asyncio.Task] = None
+
+        #: Set once immediately after startup has completed and the server has
+        #: started listening for requests.
+        self._startup_event = asyncio.Event()
+
         #: Set once immediately after all requests have been closed and shutdown
         #: has completed.
         self._shutdown_event = asyncio.Event()
@@ -63,7 +71,7 @@ class Server:
         self.config.setup_event_loop()
         return asyncio.run(self.serve(sockets=sockets))
 
-    async def serve(self, sockets: Optional[List[socket.socket]] = None) -> None:
+    async def _main(self) -> None:
         process_id = os.getpid()
 
         config = self.config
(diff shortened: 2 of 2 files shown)


tests/utils.py
```diff
@@ -8,14 +7,12 @@
...
-    server = Server(config=config)
-    cancel_handle = asyncio.ensure_future(server.serve(sockets=sockets))
-    await asyncio.sleep(0.1)
+    server = Server(config=config, sockets=sockets)
+    await server.start_serving()
     try:
         yield server
     finally:
         await server.shutdown()
-        cancel_handle.cancel()
```
bwhmather, line 18: I think it's worth pointing this bit out as it suffers from all of the same problems that our test fixtures have suffered from:
  - `asyncio.sleep(0.1)` instead of checking for startup properly.
  - Two handles required to be able to shutdown the service fully.
  - `server.shutdown()` not actually killing the server main loop.
  
I think it's a good demonstration of my motivation for proposing these changes.

euri10: yep pretty interesting, unfortunate I cant dedicate more time on this, the only real "interrogation" I have it the interaction it may have with --reload  and --workers

bwhmather: As far as I can tell, there shouldn't be any interaction.  Both just call `server.run` in a subprocess and kill the process by sending the appropriate signal.  There doesn't appear to be much difference from running without multiple workers or a watcher, which is as it should be.

euri10: going to play with this but re-reading it I like it,
@florimondmanca what are your thoughts on this, you extensively played with this a while ago, your input should be very interesting

florimondmanca:
As far as I can see, this PR introduces public API changes to `Server`, which means ideally there should be some documentation changes. Eg. does it affect the "Running programmatically" section? (Note there were some recent changes: #1525.)

The reason is, right now my main question to assess these changes would be: how would you use the introduced changes to solve 1) and 2) _in your situation_? What's your current programmatic usage of Uvicorn like, and what does it become with these changes?

There are several ways we could go about "Manage server lifecycle programmatically".

For example another way would be taking inspiration from structured concurrency, and promoting an async context manager usage. It would do the equivalent of `start_serving()` on aenter, and `.shutdown()` on aclose (but without exposing those internals):

```python
async with Server(...):
    # Server has started (no need to poll)
    ...

# Server has shut down
```

We'd allow users to use this instead of having to write an equivalent of our `run_server()` test utility, which the current "expose methods" proposal would require.

Non-block usage would be possible with `AsyncExitStack`:

```python
from contextlib import AsyncExittack

exit_stack = AsyncExitStack()

await exit_stack.enter_async_context(Server(...))  # Starts up
...
await exit_stack.aclose()  # Shuts down
```

Also, I'm not sure what the use case for a `.close()` or `.wait_closed()` couple of methods would be. Could you also illustrate how you'd expect to use those in your setup?

bwhmather:
Hi @florimondmanca - Regarding our situation:  Our test fixtures currently look a bit like this:

```python
@pytest.fixture
async def server() -> AsyncIterator[str]:
    host = "127.0.0.1"
    port = choose_port(host)

    app = create_app()

    # Uvicorn will always try to override the log config with something.
    # These settings should cause this to be a no-op.
    log_config = {"version": 1, "disable_existing_loggers": False, "incremental": True}

    uvicorn_config = uvicorn.config.Config(
        app, host=host, port=port, reload=False, workers=1, log_config=log_config
    )

    server = uvicorn.server.Server(uvicorn_config)

    # TODO uvicorn unconditionally overrides signal handlers if called from
    # the main thread.  This completely breaks `oaknorth_processes`' soft
    # shutdown mechanism if we don't do something about it.  A proper fix is
    # needed upstream.
    server.install_signal_handlers = lambda: None  # type: ignore

    while not server.started:
        await asyncio.sleep(0.01)

    yield f"http://{host}:{port}"

    server.should_exit = True
    try:
        await asyncio.wait_for(task, 1.0)
    except asyncio.exceptions.TimeoutError:
        logger.exception("Could not shutdown server cleanly")     
```
It all works most of the time, but it's brittle (for example, if the server fails to start it will block forever) and the polling, both in the fixture and the `shutdown` method) makes it either slow or inneficient.  It has also taken us a couple of years of patching and rewriting as we encounter bugs to get it shaken down even this far.


Regarding the choice of interface:  I picked this exact set of methods because it matches the interface of `asyncio.Server` in the standard library (https://docs.python.org/3/library/asyncio-eventloop.html?highlight=wait_closed#asyncio.Server).  Give that it supports my use case (as well as pretty much everything I could think of doing including blocking and non-blocking mode, triggering shutdown from inside a request handler, signal handler, or from another thread, waiting for startup and shutdown)  I thought it best not to try to be too clever.

Strongly agree on making server a context manager, but I would prefer to do what the stdlib does and implement `__aenter__` and `__aexit__` as conveneince wrappers.  Given that, I think it makes more makes sense to leave them out of this PR.



Going back to our situation:  in the ideal case, I would like for our test fixtures to look something like this:

```python
@pytest.fixture
async def server() -> AsyncIterator[str]:
    host = "127.0.0.1"
    port = choose_port(host)

    app = create_app()

    uvicorn_config = uvicorn.config.Config(
        app,
        host=host,
        port=port,
        reload=False,
        workers=1,
        register_signal_handlers=False,
        log_config=None,
    )

    async with uvicorn.server.Server(uvicorn_config):
        yield f"http://{host}:{port}"
```

With only this PR merged, we could go to something like this:
```python
@pytest.fixture
async def server() -> AsyncIterator[str]:
    host = "127.0.0.1"
    port = choose_port(host)

    app = create_app()

    # Uvicorn will always try to override the log config with something.
    # These settings should cause this to be a no-op.
    log_config = {"version": 1, "disable_existing_loggers": False, "incremental": True}

    uvicorn_config = uvicorn.config.Config(
        app, host=host, port=port, reload=False, workers=1, log_config=log_config
    )

    server = uvicorn.server.Server(uvicorn_config)

    await server.start_serving()
    
    yield f"http://{host}:{port}"

    server.close()
    await server.wait_closed()
```    
which would already be a substantial improvement.

florimondmanca:
@bwhmather Thanks.

I think I understand the idea of having aenter/aexit be straight-up calls to other public methods, basically making the async context manager usage a shorthand more than anything else.

My comment on documentation still holds, I believe. I think the idea of a publicly documented way to run a server programmatically has been around for a while. It'd be great if we could clear that out with a fully documented and refined solution. Please note the current documentation: https://www.uvicorn.org/#config-and-server-instances

For reference, we use a Uvicorn server in the HTTPX test suite, and what we do is run Uvicorn in a thread instead. Looks like this:

```python
class TestServer(Server):
    def install_signal_handlers(self) -> None:
        pass

def serve_in_thread(server: Server):
    thread = threading.Thread(target=server.run)
    thread.start()
    try:
        while not server.started:
            time.sleep(1e-3)
        yield server
    finally:
        server.should_exit = True
        thread.join()

@pytest.fixture(scope="session")
def server():
    config = Config(app=app, lifespan="off", loop="asyncio")
    server = TestServer(config=config)
    yield from serve_in_thread(server)
```

I believe it suffers from some of the same limitations, esp. regarding hanging on startup failures, as well as asyncio/threading interactions. I'm curious what we'd have to change in HTTPX to use the methods proposed here.

uvicorn/server.py
```diff
@@ -68,22 +78,62 @@ async def serve(self, sockets: Optional[List[socket.socket]] = None) -> None:
...
+        )
+
+    async def serve(self, sockets: Optional[List[socket.socket]] = None) -> None:
+        if self._main_task is not None:
+            raise RuntimeError("cannot call serve on running server")
+
+        if sockets is not None and self._sockets is not None:
+            raise RuntimeError("cannot override already provided sockets list")
+
+        self._sockets = sockets
```
florimondmanca: I think we need to override `self._sockets` only in the case when `sockets=...` was passed here? Currently we would be overriding with `self._sockets = None` in most cases.

```suggestion
        if sockets is not None:
            if self._sockets is not None:
                raise RuntimeError("cannot override already provided sockets list")
            self._sockets = sockets
```

florimondmanca (commented):
I tested this locally and I think overall this is looking _great_. Thanks a lot.

I did a thorough review, leaving some comments.

To finalize things, we might want to drop / extract anything that's not strictly related to this new start_serving/close/wait_closed/shutdown API.

Eg. I'm not sure I understood what needed fixing in #1301, or how this PR fixes that in particular. Is there a way to fix it without having this PR in? Is it possible to fix it _after_ merging this PR?

uvicorn/server.py
bwhmather: Good catch.

bwhmather:
> For reference, we use a Uvicorn server in the HTTPX test suite, and what we do is run Uvicorn in a thread instead. Looks like this:

Yup, that looks very familiar!  We were doing the same, but then tried to apply technique to a different project which expected to be create the app with already created resources which referenced the main thread event loop.

Regarding docs:  Sorry, yes, of course.

Regarding #1301:  You are right.  This doesn't actually resolve it.  I think I misread the `if self.should_exit` check after `self.startup`, but this was inherited from before and doesn't fix `_shutdown` not being called.  Best fixed as a follow up.

bwhmather:
One other thing that is concerning me on re-reading this PR is the question of what to do about exceptions?  Should they be re-raised by `wait_closed()`, or simply swallowed?

[asyncio.Server](https://github.com/python/cpython/blob/f882d33778ee2625ab32d90e28edb6878fb8af93/Lib/asyncio/base_events.py#L274) appears to opt for swallowing any errors.

adriangb:
This seems like a great idea. I’m in favor of promoting and developing better APIs to run Uvicorn programmatically

PR #1514, opened 2022-06-10
Allow configuring max_incomplete_event_size for h11 implementation

gauravojha:
PR to help solve #1234

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -65,6 +65,7 @@ For more nuanced control over which file modifications trigger reloads, install 
 * `--ws-ping-interval <float>` - Set the WebSockets ping interval, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-timeout <float>` - Set the WebSockets ping timeout, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--lifespan <str>` - Set the Lifespan protocol implementation. **Options:** *'auto', 'on', 'off'.* **Default:** *'auto'*.
+* `--h11-max-incomplete-event-size <int>` - Set the maximum number of bytes to buffer of an incomplete event. Only available for `h11` HTTP protocol implementation. **Default:** *'16384'*
 
 ## Application Interface
 
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -65,7 +65,7 @@ For more nuanced control over which file modifications trigger reloads, install 
 * `--ws-ping-interval <float>` - Set the WebSockets ping interval, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-timeout <float>` - Set the WebSockets ping timeout, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--lifespan <str>` - Set the Lifespan protocol implementation. **Options:** *'auto', 'on', 'off'.* **Default:** *'auto'*.
-* `--h11-max-incomplete-event-size <int>` - Set the maximum number of bytes to buffer of an incomplete event. Only available for `h11` HTTP protocol implementation. **Default:** *'16384'*
+* `--h11-max-incomplete-event-size <int>` - Set the maximum number of bytes to buffer of an incomplete event. Only available for `h11` HTTP protocol implementation. **Default:** *'16384'* (16 KB).
 
 ## Application Interface
 
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -809,46 +809,60 @@ def test_fragmentation():
     t.join()
 
 
-@pytest.mark.parametrize("protocol_cls", HTTP_PROTOCOLS)
-def test_huge_headers_h11_fail(protocol_cls, event_loop):
+@pytest.mark.parametrize("protocol_cls", [H11Protocol])
+def test_huge_headers_h11protocol_will_fail(protocol_cls, event_loop):
     app = Response("Hello, world", media_type="text/plain")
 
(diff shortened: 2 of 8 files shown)


uvicorn/main.py
```diff
@@ -339,6 +340,13 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     help="Look for APP in the specified directory, by adding this to the PYTHONPATH."
     " Defaults to the current working directory.",
 )
+@click.option(
+    "--h11-max-incomplete-event-size",
+    "h11_max_incomplete_event_size",
+    type=int,
+    default=None,
+    help="For h11, the maximum number of bytes to buffer of an incomplete event",
```
```suggestion
    help="For h11, the maximum number of bytes to buffer of an incomplete event.",
```

docs/settings.md
```diff
@@ -65,6 +65,7 @@ For more nuanced control over which file modifications trigger reloads, install
 * `--ws-ping-interval <float>` - Set the WebSockets ping interval, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-timeout <float>` - Set the WebSockets ping timeout, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--lifespan <str>` - Set the Lifespan protocol implementation. **Options:** *'auto', 'on', 'off'.* **Default:** *'auto'*.
+* `--h11-max-incomplete-event-size <int>` - Set the maximum number of bytes to buffer of an incomplete event. Only available for `h11` HTTP protocol implementation. **Default:** *'16384'*
```
```suggestion
* `--h11-max-incomplete-event-size <int>` - Set the maximum number of bytes to buffer of an incomplete event. Only available for `h11` HTTP protocol implementation. **Default:** *'16384'* (16 KB).
```

Kludex (commented):
I've studied the issue again. I have questions:

- Instead of accepting this PR, can we just work with `httptools` when there's a need for larger headers?
- Can we improve the tests that have conditionals in the body of the test? 🤔

If using `httptools` is an alternative, then we can also document this.

euri10:
> * Instead of accepting this PR, can we just work with `httptools` when there's a need for larger headers?

I think it's fair to have this flag for people not able to run uvicorn with C extensions

Kludex:
> I think it's fair to have this flag for people not able to run uvicorn with C extensions

Perfect. 👍

gauravojha:
thanks @Kludex  @euri10 

Sorry for the delayed response. Let me take a stab at fixing the conditionals in the test

gauravojha:
@Kludex  let me know if these changes are closer to what you had in mind with the conditionals change in tests.

Kludex (approved):
As reference: https://h11.readthedocs.io/en/latest/api.html?highlight=max-incomplete-event-size#the-connection-object

After seeing the new tests, I think we either should remove the `parametrize`, or go back to the previous: https://github.com/encode/uvicorn/pull/1514/files/bf74201a9887f2b2de5d5948f8f4a0a4aaf46fd3... I'll let @euri10 decide what he likes more.

euri10 (approved):

euri10:
> After seeing the new tests, I think we either should remove the `parametrize`, or go back to the previous: https://github.com/encode/uvicorn/pull/1514/files/bf74201a9887f2b2de5d5948f8f4a0a4aaf46fd3... I'll let @euri10 decide what he likes more.

it feels ok to me this way

Kludex:
I've tweaked the tests a bit to be more clear... Can you check what you think @euri10 ? If you're fine, please merge it.

tests/protocols/test_http.py
```diff
@@ -801,3 +812,75 @@ def send_fragmented_req(path):
...
+    # Huge headers make h11 fail in it's default config
+    # h11 sends back a 400 in this case
+    protocol.data_received(GET_REQUEST_HUGE_HEADERS[0])
+    assert b"HTTP/1.1 400 Bad Request" in protocol.transport.buffer
+    assert b"Connection: close" in protocol.transport.buffer
+    assert b"Invalid HTTP request received." in protocol.transport.buffer
+
+
+@pytest.mark.anyio
+@pytest.mark.skipif(HttpToolsProtocol is None, reason="httptools is not installed")
```
euri10, line 831: is it really necessary to have this skipif ?

tests/protocols/test_http.py
```diff
@@ -801,3 +812,75 @@ def send_fragmented_req(path):
...
```
euri10, line 861: same as above

tests/protocols/test_http.py
Kludex: Hmmm... No? I'll check in an hour.

I've added so the test logic doesn't run when httptools not installed.

euri10: but the test suite has httptools installed no ? or we should skip on windows ?

Kludex: We usually take into consideration the setup of ppl that run our test suite locally (there are some conditionals related to this idea somewhere I think... I'll check soon).

Kludex: We have this on the `test_websockets.py` for example:
https://github.com/encode/uvicorn/blob/39621c20b44187f9d31d5348d34aeab400436170/tests/protocols/test_websocket.py#L24-L26

Kludex: @euri10 Let me know if it's fine. 👍

euri10: ok ok , one of these days where I was off :) fine by me

Kludex: I think it was introduced 4 years ago 👀

PR #886, opened 2020-12-12
Log exception traceback in case of invalid HTTP request when using httptools

florimondmanca:
Refs #344 

To help with moving #344 forward, let's have Uvicorn log the full exception traceback in case an invalid HTTP request is detected by HttpTools.

Here's what we can now expect:

App:

```python
async def app(scope, receive, send):
    await send(
        {
            "type": "http.response.start",
            "status": 200,
            "headers": [(b"content-type", b"text/plain")]
        }
    )
    await send({"type": "http.response.body", "body": b"Hello, world!"})
```

Server:

```
uvicorn app:app
```

Nasty client:

```python
import socket

sock = socket.create_connection(("localhost", 8000))

# Send a broken HTTP request
sock.send(b"did you expect an HTTP method?")
```

Before:

```console
WARNING:  Invalid HTTP request received.
```

After:

```console
WARNING:  Invalid HTTP request received.
Traceback (most recent call last):
  File "/Users/florimond/Developer/python-projects/uvicorn/uvicorn/protocols/http/httptools_impl.py", line 165, in data_received
    self.parser.feed_data(data)
  File "httptools/parser/parser.pyx", line 193, in httptools.parser.parser.HttpParser.feed_data
httptools.parser.errors.HttpParserInvalidMethodError: invalid HTTP method
```

PR #1518, opened 2022-06-13
Revert log exception traceback in case of invalid HTTP request

### LINKED ISSUE #1296: Handle exception when receive request with custom method
### Checklist

### Describe the bug

I'm developing fastapi application which is using starlette and run with uvicorn.

Currently I developing locally on my PC which I've done for a long time, so I frequently get some external random requests from the internet from random IPs.

While it is fine if I get some request that lead to some php endpoint and server log it and show nice 404 error in the logs, sometimes I get weird custom requests that break uvicorn and it throw errors, it look like this in the console:
image

The main issue in `/uvicorn/protocols/http/httptools_impl.py`:
https://github.com/encode/uvicorn/blob/a9ea657339e64c2236f3fb49616e3e0a133273c1/uvicorn/protocols/http/httptools_impl.py#L131-L138

* Use curl to make custom request, the custom method is `-X` flag where you in theory should use GET/POST/... but instead in my case I get some weird methods, that make uvicorn to throw exception
```sh
curl -X MUX -i -H "Custom-Header: Test" http://127.0.0.1/
```

### Expected behavior

I should get minimal and useful log, this may be need to pass on underlying libraries (to handle it there) or being able to control behavior from them. Right now when you get this as it shows you don't understand what is happening, who is sending to you (IP) and what (what method?).

### Actual behavior

Logs get full of this exception spam, it essentially work as DDoS for a server. I have to patch library manually to be able to fix this.

### Debugging material

I attach screenshot above, there is also example from one request:
```
WARNING:  Invalid HTTP request received.
  File "/usr/local/lib/python3.9/site-packages/uvicorn/protocols/http/httptools_impl.py", line 131, in data_received
    self.parser.feed_data(data)
httptools.parser.errors.HttpParserInvalidMethodError: Invalid method encountered
```
First line is come, probably from fastapi and other are exception.

### Environment

- Windows, WSL, Docker
- `Running uvicorn 0.15.0 with CPython 3.8.6 on Windows`
- Running fastapi application like that: `uvicorn main:app --host 0.0.0.0 --port 80`

### Additional context

This is similar issues:
- https://github.com/encode/uvicorn/issues/75
- https://stackoverflow.com/questions/65878595/invalid-http-method-in-traceback-uvicorn
- https://stackoverflow.com/questions/69961123/host-fastapi-backend-api-and-react-app-frontend-locally
- https://github.com/encode/uvicorn/issues/1364

euri10:
Fixes https://github.com/encode/uvicorn/issues/1296
Since https://github.com/encode/uvicorn/issues/344#issuecomment-1142391856 seems to be fixed (we should close it) there's no reason anymore for having those

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -162,9 +162,9 @@ class H11Protocol(asyncio.Protocol):
         while True:
             try:
                 event = self.conn.next_event()
-            except h11.RemoteProtocolError as exc:
+            except h11.RemoteProtocolError:
                 msg = "Invalid HTTP request received."
-                self.logger.warning(msg, exc_info=exc)
+                self.logger.warning(msg)
                 self.send_400_response(msg)
                 return
             event_type = type(event)
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -149,9 +149,9 @@ class HttpToolsProtocol(asyncio.Protocol):
 
         try:
             self.parser.feed_data(data)
-        except httptools.HttpParserError as exc:
+        except httptools.HttpParserError:
             msg = "Invalid HTTP request received."
-            self.logger.warning(msg, exc_info=exc)
+            self.logger.warning(msg)
             self.send_400_response(msg)
             return
         except httptools.HttpParserUpgrade:


Kludex:
Do we test this already?

euri10:
this just reverts the 2 PR #889 and #886 and they didn't have tests and iirc we dont test that

Kludex (approved):
I meant "are those lines covered by tests"?

But anyway, it's not a blocker.

euri10:
thanks @Kludex 
I'll add tests if time permits, not a huge deal not having them for now

PR #1525, opened 2022-06-18
Document how to run uvicorn programatically

### LINKED ISSUE #1505: Documentation for `uvicorn.{Server, Config}`
### Discussed in https://github.com/encode/uvicorn/discussions/1504

<sup>Originally posted by **Bibo-Joshi** May 30, 2022</sup>
Hi. I noticed that the classes `uvicorn.Server` and `uvicorn.Config` are not mentioned in the documentation over at uvicorn.org. In fact I only learned about them when I wanted to run uvicorn in the same process as other functionality and read the source of `uvicorn.run` to understand how that method works.

However, both classes seem to be intended for pulic usage, as they're both [declared in `__all__`](https://github.com/encode/uvicorn/blob/1a56da93744bd0b5a273960c0b7541759d5280e9/uvicorn/__init__.py#L5) and https://github.com/encode/uvicorn/pull/302 even mentiones the use cases.

IMO it would be very beneficial to have both classes documented along with the available parameters and accepted argument types :)</div>

Kludex:
- Closes #1505
- Inspired by #302

Accepting suggestions on wording.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -196,9 +196,11 @@ For more information, see the [settings documentation](settings.md).
 
 ### Running programmatically
 
-To run uvicorn directly from your application...
+There are many ways to run uvicorn directly from your application.
 
-**example.py**:
+#### `uvicorn.run` function
+
+The easiest way is to run via `uvicorn.run` function:
 
 ```python
 import uvicorn
@@ -207,7 +209,43 @@ async def app(scope, receive, send):
     ...
 
 if __name__ == "__main__":
-    uvicorn.run("example:app", host="127.0.0.1", port=5000, log_level="info")
+    uvicorn.run("main:app", host="127.0.0.1", port=5000, log_level="info")
+```
+
+#### Explicitly setup `Config` and `Server` instances
+
+You can create the `uvicorn.Config` and `uvicorn.Server` instances manually, and then use `uvicorn.Server.run()`:
+
+```python
+import uvicorn
+
+async def app(scope, receive, send):
+    ...
+
+if __name__ == "__main__":
+    config = uvicorn.Config("main:app", host="127.0.0.1", post=5000, log_level="info")
+    server = uvicorn.Server(config)
+    server.run()
+```
+
+#### Run inside the event loop
+
+There are cases that `async` logic needs to be performed, and it's interesting to run `uvicorn` inside the event loop:
+
+```python
+import asyncio
+import uvicorn
+
+async def app(scope, receive, send):
+    ...
+
+async def main():
+    config = uvicorn.Config("main:app", host="127.0.0.1", post=5000, log_level="info")
+    server = uvicorn.Server(config)
+    await server.serve()
+
+if __name__ == "__main__":
+    asyncio.run(main())
 ```
 
 ### Running with Gunicorn
--- a/docs/index.md
+++ b/docs/index.md
@@ -209,7 +209,7 @@ async def app(scope, receive, send):
     ...
 
 if __name__ == "__main__":
-    uvicorn.run("main:app", host="127.0.0.1", port=5000, log_level="info")
+    uvicorn.run("main:app", port=5000, log_level="info")
 ```
 
 #### `Config` and `Server` instances
@@ -223,7 +223,7 @@ async def app(scope, receive, send):
     ...
 
 if __name__ == "__main__":
-    config = uvicorn.Config("main:app", host="127.0.0.1", post=5000, log_level="info")
+    config = uvicorn.Config("main:app", port=5000, log_level="info")
     server = uvicorn.Server(config)
     server.run()
 ```
@@ -238,7 +238,7 @@ async def app(scope, receive, send):
     ...
 
 async def main():
-    config = uvicorn.Config("main:app", host="127.0.0.1", post=5000, log_level="info")
+    config = uvicorn.Config("main:app", port=5000, log_level="info")
     server = uvicorn.Server(config)
     await server.serve()
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -196,11 +196,11 @@ For more information, see the [settings documentation](settings.md).
 
 ### Running programmatically
 
-There are many ways to run uvicorn directly from your application.
+There are several ways to run uvicorn directly from your application.
 
-#### `uvicorn.run` function
+#### `uvicorn.run`
 
-The easiest way is to run via `uvicorn.run` function:
+If you're looking for a programmatic equivalent of the `uvicorn` command line interface, use `uvicorn.run()`:
 
 ```python
 import uvicorn
@@ -214,7 +214,7 @@ if __name__ == "__main__":
 
 #### Explicitly setup `Config` and `Server` instances
 
-You can create the `uvicorn.Config` and `uvicorn.Server` instances manually, and then use `uvicorn.Server.run()`:
+For more control over configuration and server lifecycle, use `uvicorn.Config` and `uvicorn.Server`:
 
 ```python
 import uvicorn
@@ -228,9 +228,7 @@ if __name__ == "__main__":
     server.run()
 ```
 
-#### Run inside the event loop
-
-There are cases that `async` logic needs to be performed, and it's interesting to run `uvicorn` inside the event loop:
+If you'd like to run Uvicorn from an already running async environment, use `uvicorn.Server.serve()` instead:
 
 ```python
 import asyncio
--- a/docs/index.md
+++ b/docs/index.md
@@ -203,6 +203,7 @@ There are several ways to run uvicorn directly from your application.
 If you're looking for a programmatic equivalent of the `uvicorn` command line interface, use `uvicorn.run()`:
 
 ```python
+# main.py
 import uvicorn
 
 async def app(scope, receive, send):
--- a/docs/index.md
+++ b/docs/index.md
@@ -212,7 +212,7 @@ if __name__ == "__main__":
     uvicorn.run("main:app", host="127.0.0.1", port=5000, log_level="info")
 ```
 
-#### Explicitly setup `Config` and `Server` instances
+#### `Config` and `Server` instances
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -196,9 +196,25 @@ For more information, see the [settings documentation](settings.md).
...
+async def app(scope, receive, send):
+    ...
+
+if __name__ == "__main__":
+    uvicorn.run("main:app", host="127.0.0.1", port=5000, log_level="info")
+```
+
+#### Explicitly setup `Config` and `Server` instances
+
+You can create the `uvicorn.Config` and `uvicorn.Server` instances manually, and then use `uvicorn.Server.run()`:
```
florimondmanca:
```suggestion
For more control over configuration and server lifecycle, use `uvicorn.Config` and `uvicorn.Server`:
```

docs/index.md
```diff
@@ -196,9 +196,25 @@ For more information, see the [settings documentation](settings.md).
 
 ### Running programmatically
 
-To run uvicorn directly from your application...
+There are many ways to run uvicorn directly from your application.
 
-**example.py**:
+#### `uvicorn.run` function
```
florimondmanca: Aiming at concise headings

```suggestion
#### `uvicorn.run`
```

docs/index.md
```diff
@@ -196,9 +196,25 @@ For more information, see the [settings documentation](settings.md).
...
+import uvicorn
+
+async def app(scope, receive, send):
+    ...
+
+if __name__ == "__main__":
+    uvicorn.run("main:app", host="127.0.0.1", port=5000, log_level="info")
+```
+
+#### Explicitly setup `Config` and `Server` instances
```
florimondmanca:
```suggestion
#### `Config` and `Server` instances
```

docs/index.md
```diff
@@ -196,9 +196,25 @@ For more information, see the [settings documentation](settings.md).
 
 ### Running programmatically
 
-To run uvicorn directly from your application...
+There are many ways to run uvicorn directly from your application.
 
-**example.py**:
+#### `uvicorn.run` function
+
+The easiest way is to run via `uvicorn.run` function:
```
florimondmanca: "Easiest" is subjective unfortunately; here's a wording trying to more explicitly convey what use cases this usage could address:

```suggestion
If you're looking for a programmatic equivalent of the `uvicorn` command line interface, use `uvicorn.run()`:
```

(It does accept all options that `$ uvicorn [OPTIONS]` does, correct?)

docs/index.md
```diff
@@ -207,7 +223,29 @@ async def app(scope, receive, send):
...
+async def app(scope, receive, send):
+    ...
+
+async def main():
+    config = uvicorn.Config("main:app", host="127.0.0.1", post=5000, log_level="info")
+    server = uvicorn.Server(config)
+    await server.serve()
+
+if __name__ == "__main__":
+    asyncio.run(main())
```
florimondmanca: I'd be tempted to merge this section with "Config and Server instances" since it describes an alternate way of using these -- .run() for launching an async environment, .serve() if you've already got one running.

It might also be that we don't need any h4 here at all (I'm looking at the table of contents sidebar of this home page which doesn't contain any h4 yet).

```suggestion
If you'd like to run Uvicorn from an already running async environment, use `uvicorn.Server.serve()` instead:

```python
import asyncio
import uvicorn

    ...

async def main():
    config = uvicorn.Config("main:app", host="127.0.0.1", post=5000, log_level="info")
    await server.serve()

if __name__ == "__main__":
    asyncio.run(main())
```

docs/index.md
```diff
@@ -196,9 +196,25 @@ For more information, see the [settings documentation](settings.md).
...
+
+```python
+import uvicorn
+
+async def app(scope, receive, send):
+    ...
+
+if __name__ == "__main__":
+    uvicorn.run("main:app", host="127.0.0.1", port=5000, log_level="info")
+```
```
florimondmanca, line 214:
```markdown
`uvicorn.run` accepts the following arguments:

* `app`: Either an ASGI application, or (as shown above) an import string to the target ASGI application. The latter is required if using auto-reload or multiple workers.
* `**kwargs`: Keyword arguments that mirror command line options (see [Settings](/settings.md)): `--host <str>` becomes `host=<str>`, `--log-level <str>` becomes `log_level=<str>`, etc.
```

(I think this kind of info is valuable in docs, but maybe we might want to kick off an `API Reference` page for this kind of thing. I'm OK if we defer this bit from this PR.)

docs/index.md
```diff
@@ -196,9 +196,25 @@ For more information, see the [settings documentation](settings.md).
 
 ### Running programmatically
 
-To run uvicorn directly from your application...
+There are many ways to run uvicorn directly from your application.
```
florimondmanca: "Many" feels a bit overwhelming — as we document here, there are only 3 ways (run / config-serve.run() / config-serve.serve()).

```suggestion
There are several ways to run uvicorn directly from your application.
```

florimondmanca (commented):
Nice addition!

I think we also need to change the introduction paragraphs in [Settings](https://www.uvicorn.org/settings/):

```markdown
Use the following options to configure Uvicorn, when running from the command line.

If you're running programmatically, using uvicorn.run(...), then use equivalent keyword arguments, eg. uvicorn.run("example:app", port=5000, reload=True, access_log=False). Please note that in this case, if you use reload=True or workers=NUM, you should put uvicorn.run into if __name__ == '__main__' clause in the main module.

You can also configure Uvicorn using environment variables with the prefix UVICORN_. For example, in case you want to run the app on port 5000, just set the environment variable UVICORN_PORT to 5000.

!!! note
    CLI options and the arguments for uvicorn.run() take precedence over environment variables.

    Also note that UVICORN_* prefixed settings cannot be used from within an environment configuration file. Using an environment configuration file with the --env-file flag is intended for configuring the ASGI application that uvicorn runs, rather than configuring uvicorn itself.
```

Suggestion ⬇️

```markdown
Use the following options to configure Uvicorn, when running from the command line.

These options are mirrored as keyword arguments when [running programmatically](/index.md#running-programmatically).

Default settings can also be passed as environment variables using the `UVICORN_` prefix. For example, setting `UVICORN_PORT` to `5000` would make Uvicorn serve on port 5000 if `--port` (when using the CLI) or `port` (when running programmatically) have not been passed.
```

Kludex:
I knew you would provide a lot of value here. Much appreciated @florimondmanca 🙏

I'm going to apply/reply those tonight. :)

docs/index.md
Kludex: It does. 👍

docs/index.md
Kludex: We _improved_ the `uvicorn.run` type annotation, so we can have autocomplete on it: https://github.com/encode/uvicorn/blob/b1a45821a2ab916b4428bdcb9eba4a4a0cc8fb74/uvicorn/main.py#L445-L492

I'll not apply this for now. An API reference would be fine by me.

Kludex:
I've removed the `host` argument, as the visualization on the docs becomes more clear without it (either that or I need to format the python code...).

Applied the changes. 👍

Kludex:
> Nice addition!
> 
> I think we also need to change the introduction paragraphs in [Settings](https://www.uvicorn.org/settings/):
> 
> ```gfm
> Use the following options to configure Uvicorn, when running from the command line.
> 
> If you're running programmatically, using uvicorn.run(...), then use equivalent keyword arguments, eg. uvicorn.run("example:app", port=5000, reload=True, access_log=False). Please note that in this case, if you use reload=True or workers=NUM, you should put uvicorn.run into if __name__ == '__main__' clause in the main module.
> 
> You can also configure Uvicorn using environment variables with the prefix UVICORN_. For example, in case you want to run the app on port 5000, just set the environment variable UVICORN_PORT to 5000.
> 
> !!! note
>     CLI options and the arguments for uvicorn.run() take precedence over environment variables.
> 
>     Also note that UVICORN_* prefixed settings cannot be used from within an environment configuration file. Using an environment configuration file with the --env-file flag is intended for configuring the ASGI application that uvicorn runs, rather than configuring uvicorn itself.
> ```
> 
> Suggestion ⬇️
> 
> ```gfm
> Use the following options to configure Uvicorn, when running from the command line.
> 
> These options are mirrored as keyword arguments when [running programmatically](/index.md#running-programmatically).
> 
> Default settings can also be passed as environment variables using the `UVICORN_` prefix. For example, setting `UVICORN_PORT` to `5000` would make Uvicorn serve on port 5000 if `--port` (when using the CLI) or `port` (when running programmatically) have not been passed.
> ```

I think we should mention who takes precedence as well 🤔

docs/index.md
```diff
@@ -207,7 +223,27 @@ async def app(scope, receive, send):
...
 if __name__ == "__main__":
-    uvicorn.run("example:app", host="127.0.0.1", port=5000, log_level="info")
+    config = uvicorn.Config("main:app", port=5000, log_level="info")
+    server = uvicorn.Server(config)
+    server.run()
+```
+
+If you'd like to run Uvicorn from an already running async environment, use `uvicorn.Server.serve()` instead:
+
+```python
```
euri10, line 234:
```suggestion
main.py
```python
```

euri10: just to say this is the name of the file, I'll approve it without this change but we have too many people not understanding how the app string works

euri10 (approved):
lgtm, with or without a minor comment

docs/index.md
```diff
@@ -196,9 +196,25 @@ For more information, see the [settings documentation](settings.md).
...
 
-To run uvicorn directly from your application...
+There are several ways to run uvicorn directly from your application.
 
-**example.py**:
+#### `uvicorn.run`
+
+If you're looking for a programmatic equivalent of the `uvicorn` command line interface, use `uvicorn.run()`:
+
+```python
```
Kludex, line 205:
```suggestion
```python
# main.py
```

docs/index.md
Kludex: Added on the first file here.

PR #1537, opened 2022-06-24
Handle missing watchfiles and watchgod gracefully in tests

mgorny:
Skip the tests relying on watchfiles and watchgod reloaders when
the required dependencies are not available rather than causing
an ImportError.  Since these packages are optional and watchfiles
does not support all the platforms that uvicorn used to work on so far,
it is useful to be able to run at least part of the test suite there.

### DIFF
--- a/tests/supervisors/test_reload.py
+++ b/tests/supervisors/test_reload.py
@@ -2,7 +2,7 @@ import logging
 import signal
 from pathlib import Path
 from time import sleep
-from typing import Type
+from typing import Optional, Type
 
 import pytest
 
@@ -10,8 +10,16 @@ from tests.utils import as_cwd
 from uvicorn.config import Config
 from uvicorn.supervisors.basereload import BaseReload, _display_path
 from uvicorn.supervisors.statreload import StatReload
-from uvicorn.supervisors.watchfilesreload import WatchFilesReload
-from uvicorn.supervisors.watchgodreload import WatchGodReload
+
+try:
+    from uvicorn.supervisors.watchfilesreload import WatchFilesReload
+except ImportError:  # pragma: no cover
+    WatchFilesReload = None
+
+try:
+    from uvicorn.supervisors.watchgodreload import WatchGodReload
+except ImportError:  # pragma: no cover
+    WatchGodReload = None
 
 
 def run(sockets):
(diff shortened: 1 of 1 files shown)


Kludex:
Maybe we can just skip the tests on the file if the package is not installed instead or not...? 😅 

```python
pytestmark = pytest.mark.skipif("watchfiles" not in sys.module)
```
Thoughts @euri10 ?

euri10:
I don't understand why you'd want to run the testsuite on a machine that can't run it in the first place, I mean it's a maintainer /developper thing,

ok we have the case of the ci, where some stuff is or is not available, and we cover that case.

we could have the case of distribution packagers, who have their own sets of constraints (like no network with the debian guys) also, and it's legitimate.

But what is this new use case ? running it on linux and not having rust installed ?

to me a better approach if we'd like to take that way, would be to create custom markers (https://docs.pytest.org/en/7.1.x/example/markers.html?highlight=marker) so that people with specific needs could enable / disable what they want / don't want to run.

This said I'd still like to know if this new case seems valid or if it's just theoretical like it seems to be to me at the moment without more precisions. I thought I'd just ask the above because in the end I don't want to go down this path where we'd have to add those changes for every package that is not available on a set of os / distribution

mgorny:
Yes, it's about targets that aren't supported at all by the Rust compiler yet used to have working uvicorn installations.

mgorny:
(this applies to watchfiles, I've covered watchgod as well because it was easy to do so but there's no strict need for that right now — though I suppose the package will grow stale sooner than later and may start failing e.g. with new Python versions)

euri10:
> Yes, it's about targets that aren't supported at all by the Rust compiler yet used to have working uvicorn installations.

ok I get that, but what is the usecase for those targets, why do you need to run tests on them ?

mgorny:
Because there are still people using them, and we want to check whether things still work on them. Given that Python has very little ability to detect problems statically, running test suites is pretty much the only way to catch issues.

euri10:
Understood the theory and I'm not dismissing it, but in very concrete terms, what os or target for instance are you having trouble with ? In what circumstances, is it when you run your CI that you're running tests for every library you're using, are you packaging uvicorn for that target ?

mgorny:
Gentoo. We're running tests for all Python packages we're packaging, for all architectures they're supported on. In case of uvicorn, this means amd64, arm (including older ARM boards that aren't supported by Rust), arm64, hppa, ia64, loong, mips, ppc, ppc64, riscv, s390, sparc, x86. In general, we also provide support for alpha and m68k but not for this particular package.

euri10 (approved):
> `pytestmark = pytest.mark.skipif("watchfiles" not in sys.module)`

I like it more that way @Kludex 

Ok sorry about all those clarifications on my side @mgorny , just wanted to be extra sure we weren't doing this and the other httpstools for obscure reasons, and thanks for packaging it on Gentoo :)

mgorny:
Thanks a lot!

thesamesam:
Really appreciate the flexibility here, thanks guys!

PR #1538, opened 2022-06-24
Fix skipping tests requiring httptools

mgorny:
The test suite currently skips some tests when httptools are missing.
However, there are more tests requiring it and they currently fail
if that is the case.  Cover them with necessary skips as well.

### DIFF
--- a/tests/middleware/test_logging.py
+++ b/tests/middleware/test_logging.py
@@ -7,6 +7,15 @@ import websockets
 
 from tests.utils import run_server
 from uvicorn import Config
+from uvicorn.protocols.http.h11_impl import H11Protocol
+
+try:
+    from uvicorn.protocols.http.httptools_impl import HttpToolsProtocol
+except ImportError:  # pragma: nocover
+    HttpToolsProtocol = None
+
+
+HTTP_PROTOCOLS = [p for p in [H11Protocol, HttpToolsProtocol] if p is not None]
 
 
 @contextlib.contextmanager
@@ -49,7 +58,7 @@ async def test_trace_logging(caplog, logging_config):
 
 
 @pytest.mark.anyio
-@pytest.mark.parametrize("http_protocol", [("h11"), ("httptools")])
+@pytest.mark.parametrize("http_protocol", HTTP_PROTOCOLS)
 async def test_trace_logging_on_http_protocol(http_protocol, caplog, logging_config):
     config = Config(
         app=app,
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -766,6 +766,7 @@ async def test_invalid_http_request(request_line, protocol_cls, caplog):
     assert b"Invalid HTTP request received." in protocol.transport.buffer
 
 
+@pytest.mark.skipif(HttpToolsProtocol is None, reason="httptools is not installed")
 def test_fragmentation():
     def receive_all(sock):
         chunks = []


Kludex (changes requested):
Can we follow the same style as `test_http.py` on `test_logging.py`?

```python
HTTP_PROTOCOLS = [p for p in [H11Protocol, HttpToolsProtocol] if p is not None]
```

mgorny:
This is much nicer. Thanks, done that!

tests/protocols/test_http.py
```diff
@@ -766,6 +766,7 @@ async def test_invalid_http_request(request_line, protocol_cls, caplog):
     assert b"Invalid HTTP request received." in protocol.transport.buffer
 
 
+@pytest.mark.skipif(HttpToolsProtocol is None, reason="httptools is not installed")
 def test_fragmentation():
```
Kludex, line 770: Do we know why this test is only running for `httptools`? 😰

tests/middleware/test_logging.py
```diff
@@ -7,6 +7,15 @@
...
 from uvicorn import Config
+from uvicorn.protocols.http.h11_impl import H11Protocol
+
+try:
+    from uvicorn.protocols.http.httptools_impl import HttpToolsProtocol
+except ImportError:  # pragma: nocover
+    HttpToolsProtocol = None
+
+
+HTTP_PROTOCOLS = [p for p in [H11Protocol, HttpToolsProtocol] if p is not None]
```
euri10, line 18:
```suggestion
HTTP_PROTOCOLS = [p for p in [H11Protocol, HttpToolsProtocol]]
```

tests/protocols/test_http.py
euri10: because it never happens with h11, https://github.com/encode/uvicorn/pull/1263

tests/middleware/test_logging.py
```diff
@@ -49,7 +58,7 @@ async def test_trace_logging(caplog, logging_config):
 
 
 @pytest.mark.anyio
-@pytest.mark.parametrize("http_protocol", [("h11"), ("httptools")])
+@pytest.mark.parametrize("http_protocol", HTTP_PROTOCOLS)
```
euri10, line 61: I would not change this, if we were using it elsewhere why not

tests/middleware/test_logging.py
euri10: if we go this way I see no reason for p to be None, see comment below, I'd be in favor for not refgactoring that for just this instance

euri10 (changes requested):

tests/protocols/test_http.py
Kludex: Thanks 🙇

tests/middleware/test_logging.py
Kludex: I didn't check, but if the `p is not None` is not there, and `httptools` is not installed, for example, then the list will be `[H11Protocol, None]`, and `None` in the tests will get to `auto`, and then `auto` will choose `h11`. Is that right? 🤔

euri10: ok your right, so one step back for me,  what is the purpose of those changes ? it's for dev doing stuff on windows and not being able to install httptools (cant remember now if it is possible) or for the CI ? or somnething else

mgorny: It's for users who want to run the test suite against their production environment that doesn't have (or need) `httptools`.

Kludex: > ok your right, so one step back for me, what is the purpose of those changes ? it's for dev doing stuff on windows and not being able to install httptools (cant remember now if it is possible) or for the CI ? or somnething else

The thing is that we were already doing this on the test suite:

https://github.com/encode/uvicorn/blob/b21ecabc5bf911f571e0629438315a1e5472065c/tests/protocols/test_http.py#L20

I guess this was in place so ppl that don't have `httptools` installed could also run the test suite. I think the idea of this PR is to comply with that idea.

Kludex (approved):
@euri10 I'm fine with this already.

I'll let this in your hand. 🙏

tests/middleware/test_logging.py
euri10: you're right !

tests/middleware/test_logging.py
euri10: no big deal

euri10 (approved):

euri10:
ok let's roll with this, thanks @mgorny and thanks for clarifications in the other PR, and thanks for packaging uvicorn :)

mgorny:
Thanks a lot!

PR #1539, opened 2022-06-24
Fix type hints for `log_config`

wch:
This is a fix for the issue reported here: https://github.com/encode/uvicorn/discussions/1536

Previously, if external code called `uvicorn.run()` and was type-checked with pyright, it reported `error: Type of "run" is partially unknown`. This is because it wants the generic type `dict` to have parameters specifying the type, as in `dict[str, Any]`.

I'm actually a bit surprised that mypy doesn't also report this as a problem, since all the examples I saw in the mypy docs provide a parameter for generics like `list` and `dict`.

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -478,7 +478,7 @@ def run(
     reload_delay: float = 0.25,
     workers: typing.Optional[int] = None,
     env_file: typing.Optional[str] = None,
-    log_config: typing.Optional[typing.Union[dict, str]] = None,
+    log_config: typing.Optional[typing.Union[typing.Dict[str, typing.Any], str]] = None,
     log_level: typing.Optional[str] = None,
     access_log: bool = True,
     proxy_headers: bool = True,
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -478,7 +478,9 @@ def run(
     reload_delay: float = 0.25,
     workers: typing.Optional[int] = None,
     env_file: typing.Optional[str] = None,
-    log_config: typing.Optional[typing.Union[typing.Dict[str, typing.Any], str]] = LOGGING_CONFIG,
+    log_config: typing.Optional[
+        typing.Union[typing.Dict[str, typing.Any], str]
+    ] = LOGGING_CONFIG,
     log_level: typing.Optional[str] = None,
     access_log: bool = True,
     proxy_headers: bool = True,
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -10,6 +10,7 @@ import sys
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
+    Any,
     Awaitable,
     Callable,
     Dict,
@@ -91,7 +92,7 @@ INTERFACES: List[InterfaceType] = ["auto", "asgi3", "asgi2", "wsgi"]
 SSL_PROTOCOL_VERSION: int = ssl.PROTOCOL_TLS_SERVER
 
 
-LOGGING_CONFIG: dict = {
+LOGGING_CONFIG: Dict[str, Any] = {
     "version": 1,
     "disable_existing_loggers": False,
     "formatters": {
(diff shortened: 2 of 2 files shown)


Kludex (approved):
Thanks @wch ! 🎉 

I think it's normal for mypy not to warn here, as `dict == Dict[str, Any]`... It's just that Pyright on strict mode (I think?) see it as a problem... Or maybe mypy with more strict setup see it as a problem as well? 🤔

wch:
Ah, I didn't know that `dict` is equivalent to `Dict[str, Any]`.

I did find this flag for mypy:
https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-type-arguments-exist-type-arg

> If you use [--disallow-any-generics](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-any-generics), mypy requires that each generic type has values for each type argument. For example, the types list or dict would be rejected. You should instead use types like list[int] or dict[str, int]. Any omitted generic type arguments get implicit Any values. The type list is equivalent to list[Any], and so on.

I tried using mypy with that option:

```
$ scripts/check
+ venv/bin/black --check --diff --target-version=py37 uvicorn tests
All done! ✨ 🍰 ✨
65 files would be left unchanged.
+ venv/bin/flake8 uvicorn tests
+ venv/bin/mypy --show-error-codes --disallow-any-generics
uvicorn/config.py:132: error: Missing type parameters for generic type "PathLike"  [type-arg]
uvicorn/config.py:133: error: Missing type parameters for generic type "PathLike"  [type-arg]
uvicorn/config.py:137: error: Missing type parameters for generic type "PathLike"  [type-arg]
uvicorn/config.py:212: error: Missing type parameters for generic type "Callable"  [type-arg]
uvicorn/config.py:225: error: Missing type parameters for generic type "PathLike"  [type-arg]
uvicorn/config.py:250: error: Missing type parameters for generic type "PathLike"  [type-arg]
uvicorn/config.py:525: error: Missing type parameters for generic type "Callable"  [type-arg]
uvicorn/server.py:44: error: Missing type parameters for generic type "Task"  [type-arg]
uvicorn/server.py:87: error: Missing type parameters for generic type "list"  [type-arg]
tests/conftest.py:149: error: Missing type parameters for generic type "dict"  [type-arg]
uvicorn/workers.py:35: error: Missing type parameters for generic type "dict"  [type-arg]
tests/test_config.py:42: error: Missing type parameters for generic type "dict"  [type-arg]
tests/test_config.py:47: error: Missing type parameters for generic type "dict"  [type-arg]
tests/test_config.py:340: error: Missing type parameters for generic type "typing.Callable"  [type-arg]
tests/test_config.py:390: error: Missing type parameters for generic type "dict"  [type-arg]
tests/test_config.py:411: error: Missing type parameters for generic type "dict"  [type-arg]
Found 16 errors in 5 files (checked 53 source files)
```

adriangb:
I think we should use `Mapping[str, Any]`: `run` should not mutate the dict it is passed in.

Kludex:
> I think we should use `Mapping[str, Any]`: `run` should not mutate the dict it is passed in.

The `log_config` does mutate internally.

https://github.com/encode/uvicorn/blob/0954e5633a68874af00fe1f455a4a017c7b279f9/uvicorn/config.py#L399-L407

uvicorn/main.py
```diff
@@ -478,7 +478,7 @@ def run(
     reload_delay: float = 0.25,
     workers: typing.Optional[int] = None,
     env_file: typing.Optional[str] = None,
-    log_config: typing.Optional[typing.Union[dict, str]] = LOGGING_CONFIG,
+    log_config: typing.Optional[typing.Union[typing.Dict[str, typing.Any], str]] = LOGGING_CONFIG,
```
```suggestion
    log_config: typing.Optional[
        typing.Union[typing.Dict[str, typing.Any], str]
    ] = LOGGING_CONFIG,
```

adriangb:
> The `log_config` does mutate internally.

It should make a copy then (maybe another PR)

PR #1541, opened 2022-06-24
Add default `log_config` on `uvicorn.run()`

Kludex:
I broke the logs because of this. 😞

I don't know how I didn't catch this... It's too simple...

Anyway, moment of shame here. This application doesn't show any logs:

```python
import uvicorn
from fastapi import FastAPI

app = FastAPI() 
@app.get("/")
def index(): 
    return {"home": "index"}

if __name__==  "__main__": 
    uvicorn.run("main:app", port=8100)
```

The reason is that I've set the default of `log_config` to `None`, instead of `LOGGING_CONFIG`.

EDIT: The funny thing is that I noticed this behavior, and I thought the problem was the computer (mac)... 🤦

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -478,7 +478,7 @@ def run(
     reload_delay: float = 0.25,
     workers: typing.Optional[int] = None,
     env_file: typing.Optional[str] = None,
-    log_config: typing.Optional[typing.Union[dict, str]] = None,
+    log_config: typing.Optional[typing.Union[dict, str]] = LOGGING_CONFIG,
     log_level: typing.Optional[str] = None,
     access_log: bool = True,
     proxy_headers: bool = True,


adriangb:
Was this the previous default? Where (in what commit) was it changed?

uvicorn/main.py
```diff
@@ -478,7 +478,7 @@ def run(
     reload_delay: float = 0.25,
     workers: typing.Optional[int] = None,
     env_file: typing.Optional[str] = None,
-    log_config: typing.Optional[typing.Union[dict, str]] = None,
+    log_config: typing.Optional[typing.Union[dict, str]] = LOGGING_CONFIG,
```
adriangb, line 481:
```suggestion
    log_config: typing.Optional[typing.Union[typing.Mapping[str, Any], str]] = LOGGING_CONFIG,
```

At least this way we'll get an error from the type checker if we try to mutate the default argument.

Kludex:
Default values were added on: https://github.com/encode/uvicorn/pull/1423

uvicorn/main.py
Kludex: I'm going to merge https://github.com/encode/uvicorn/pull/1539 after this PR.

Kludex: But it wouldn't make any difference for here...

adriangb:
I see, so before we were just setting it to this if it wasn't in kwargs: https://github.com/encode/uvicorn/blob/4e8781c4350fc26091cdf44e333e1ba7a12e965e/uvicorn/main.py#L405

So I guess we just got it wrong in #1423 

This is an easy 👍 then.

adriangb:
One last thing: should we remove `Optional` then? You either use the default mapping or you pass in a mapping. Or does `None` have a meaning now?

Kludex:
`None` always had a meaning. Which is to not set what is in the `LOGGING_CONFIG` i.e. just not set anything for loggers...

The mistake here was to set `None` as default on the `run()`.

adriangb:
But before if you passed `None` you got `LOGGING_CONFIG` so `None` had not meaning: https://github.com/encode/uvicorn/blob/11d28e208504436e7e06603a941c01b76f54346f/uvicorn/main.py#L422

Kludex:
Via CLI it was never supposed to accept `None`. But via code, it was always allowed:

https://github.com/encode/uvicorn/blob/11d28e208504436e7e06603a941c01b76f54346f/uvicorn/config.py#L399

adriangb:
Hmm okay ✅

adriangb (approved):

PR #774, opened 2020-09-05
Make reload delay configurable

rafalp:
As I've mentioned in #338, hitting filesystem for differences four times a second is too much for larger codebases. I've contemplated bumping this to 1s (which is how often Django's default file watcher checks for changes) but I've decided to make this configurable to let end users decide what makes sense for their projects and setups. Defaults to 0.25s which is original value of the setting.

PR #1545, opened 2022-06-26
Add test to check if `run()` matches `Config.__init__()`

Kludex:
After breaking the logs on 0.18.0 (see the fix on https://github.com/encode/uvicorn/pull/1541) because of https://github.com/encode/uvicorn/pull/1423 I've been thinking of a way to have avoided that...

I came up with this. This PR adds a test that matches the type annotation and default values of `run()` and `Config.__init__()`.

This PR also fixes the mismatches.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,3 +1,4 @@
+import inspect
 import socket
 from logging import WARNING
 
@@ -107,3 +108,17 @@ def test_run_startup_failure(caplog: pytest.LogCaptureFixture) -> None:
     with pytest.raises(SystemExit) as exit_exception:
         run(app, lifespan="on")
     assert exit_exception.value.code == 3
+
+
+def test_run_match_config_params() -> None:
+    config_params = {
+        key: repr(value)
+        for key, value in inspect.signature(Config.__init__).parameters.items()
+        if key not in ("self", "timeout_notify", "callback_notify")
+    }
+    run_params = {
+        key: repr(value)
+        for key, value in inspect.signature(run).parameters.items()
+        if key not in ("app_dir",)
+    }
+    assert config_params == run_params
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -218,9 +218,9 @@ class Config:
         http: Union[Type[asyncio.Protocol], HTTPProtocolType] = "auto",
         ws: Union[Type[asyncio.Protocol], WSProtocolType] = "auto",
         ws_max_size: int = 16 * 1024 * 1024,
-        ws_ping_interval: Optional[float] = 20,
-        ws_ping_timeout: Optional[float] = 20,
-        ws_per_message_deflate: Optional[bool] = True,
+        ws_ping_interval: Optional[float] = 20.0,
+        ws_ping_timeout: Optional[float] = 20.0,
+        ws_per_message_deflate: bool = True,
         lifespan: LifespanType = "auto",
         env_file: Optional[Union[str, os.PathLike]] = None,
         log_config: Optional[Union[Dict[str, Any], str]] = LOGGING_CONFIG,
@@ -231,7 +231,7 @@ class Config:
         debug: bool = False,
         reload: bool = False,
         reload_dirs: Optional[Union[List[str], str]] = None,
-        reload_delay: Optional[float] = None,
+        reload_delay: float = 0.25,
         reload_includes: Optional[Union[List[str], str]] = None,
         reload_excludes: Optional[Union[List[str], str]] = None,
         workers: Optional[int] = None,
@@ -245,7 +245,7 @@ class Config:
         backlog: int = 2048,
         timeout_keep_alive: int = 5,
         timeout_notify: int = 30,
-        callback_notify: Callable[..., Awaitable[None]] = None,
+        callback_notify: Optional[Callable[..., Awaitable[None]]] = None,
         ssl_keyfile: Optional[str] = None,
         ssl_certfile: Optional[Union[str, os.PathLike]] = None,
         ssl_keyfile_password: Optional[str] = None,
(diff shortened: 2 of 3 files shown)


tests/test_main.py
```diff
@@ -107,3 +108,17 @@ async def app(scope, receive, send):
     with pytest.raises(SystemExit) as exit_exception:
         run(app, lifespan="on")
     assert exit_exception.value.code == 3
+
+
+def test_run_match_config_params() -> None:
+    config_params = {
+        key: repr(value)
+        for key, value in inspect.signature(Config.__init__).parameters.items()
+        if key not in ("self", "timeout_notify", "callback_notify")
```
Kludex, line 117: `timeout_notify` and `callback_notify` are not available via `run()`. Should `run() support it?

uvicorn/config.py
```diff
@@ -277,7 +277,7 @@ def __init__(
         self.interface = interface
         self.debug = debug
         self.reload = reload
-        self.reload_delay = reload_delay or 0.25
+        self.reload_delay = reload_delay
```
Kludex, line 280: There's no need to accept `None` for this. The behavior is not changed with this change.

PrettyWood: I agree it's cleaner but what if someone explicitly called it with `reload_delay=None`? The default value won't override it. Might be worth a mention in the changelog

tests/test_main.py
PrettyWood: Probably good to make a constant with those two values `CONFIG_ONLY_PARAMS`

uvicorn/config.py
Kludex: AFAIC, we shouldn't check a behavior that it's not allowed by the input, considering the type hint. I mean, `reload_delay` annotation changed from `Optional[float]` to `float` on this PR.

But... As you've pointed out, this is a breaking change, and I don't want to add breaking changes on this PR. 😅

I'll revert this change, and adapt the `run()`. 🙏

tests/test_main.py
euri10: > `timeout_notify` and `callback_notify` are not available via `run()`. Should `run() support it?

certainly not imho  `callback_notify` it is just there to let the gunicorn know the worker is still alive iirc every `timeout_notify`
those can be overridden if you are subclassing UvicornWorker, I wouldn't know how to do that with the CLI nor if it makes sense


> Probably good to make a constant with those two values `CONFIG_ONLY_PARAMS`

why not yes, not opposed having them that way too

uvicorn/config.py
```diff
@@ -231,7 +231,7 @@ def __init__(
         debug: bool = False,
         reload: bool = False,
         reload_dirs: Optional[Union[List[str], str]] = None,
-        reload_delay: Optional[float] = None,
+        reload_delay: float = 0.25,
```
euri10, line 234: I think None is valid here, it will be used in https://github.com/encode/uvicorn/blob/b1a45821a2ab916b4428bdcb9eba4a4a0cc8fb74/uvicorn/supervisors/basereload.py#L57 and https://docs.python.org/3/library/threading.html#threading.Event.wait can take None as a parameter

uvicorn/config.py
```diff
@@ -218,9 +218,9 @@ def __init__(
         http: Union[Type[asyncio.Protocol], HTTPProtocolType] = "auto",
         ws: Union[Type[asyncio.Protocol], WSProtocolType] = "auto",
         ws_max_size: int = 16 * 1024 * 1024,
-        ws_ping_interval: Optional[float] = 20,
-        ws_ping_timeout: Optional[float] = 20,
-        ws_per_message_deflate: Optional[bool] = True,
+        ws_ping_interval: Optional[float] = 20.0,
+        ws_ping_timeout: Optional[float] = 20.0,
+        ws_per_message_deflate: bool = True,
```
euri10, line 223: agreed

uvicorn/main.py
```diff
@@ -1,3 +1,4 @@
+import asyncio
```
euri10, line 1: I suppose all those types changes in run are necessary but honestly I have a hard time figuring as to why

Kludex: On this file, I'm matching the ones from `Config`. Is that an answer for what you're asking?

uvicorn/config.py
Kludex: Currently, `None` is converted to 0.25 on the `Config.__init__`. I think there's more reasoning for this value on the PR that introduced it...

Kludex: What I mean is that, is not because `wait()` accepts `None`, that we want it to ever have `None` there.

euri10: I suppose it is from b1a45821a2ab916b4428bdcb9eba4a4a0cc8fb74
I think it's better not to allow None to be passed as it would hang the app, wouldn't it ?

uvicorn/main.py
euri10: mmm ok got iy

uvicorn/config.py
Kludex: No. It's been there even after this: https://github.com/encode/uvicorn/pull/774

Kludex: `0.25` was hard coded, and it allowed modification on #774

euri10: ok I'm convinced, since click doenst allow to use None and that it would I think hang the reload if someone used it with the `Config(...)` style then we're on the safe side not allowing it

euri10 (approved):

tests/test_main.py
Kludex: > certainly not imho callback_notify it is just there to let the gunicorn know the worker is still alive iirc every timeout_notify
those can be overridden if you are subclassing UvicornWorker, I wouldn't know how to do that with the CLI nor if it makes sense

Right. Thanks 🙏

Kludex: I don't see much benefit on the `CONFIG_ONLY_PARAMS` constant for a single test function. Is there? 

I'll merge this, but I don't mind creating a PR to add that. 🙏

uvicorn/config.py
Kludex: Ok. I thought about it... Yep, you're right. I'll follow the path you suggested and add a note on the changelog about this.

PR #1555, opened 2022-07-01
use more pythonic syntax

imba-tjd:


LICENSE.md
```diff
@@ -1,4 +1,4 @@
-Copyright © 2017-present, [Encode OSS Ltd](http://www.encode.io/).
+Copyright © 2017-present, [Encode OSS Ltd](https://www.encode.io/).
```
Kludex: I think there's no problem on updating this in the license, but just to be sure... Do you know if this is a problem @tomchristie ?

uvicorn/server.py
```diff
@@ -217,7 +217,7 @@ async def main_loop(self) -> None:
         should_exit = await self.on_tick(counter)
         while not should_exit:
             counter += 1
-            counter = counter % 864000
+            counter %= 864000
```
Kludex: Can we revert this one? 🙏 

I always think that the below is previous is easier to read.
```suggestion
            counter = counter % 864000
```

Kludex (changes requested):
Thanks @imba-tjd 🙏 

Can you check my comment below?

imba-tjd:
Test fails 🤔 but the first commit passed.

Kludex:
I've rerun the pipeline. :)

LICENSE.md
lovelydinosaur: It's fiiiiiiiiiiiiiiiinnnnnneeeeeeeeeeeeeeee

lovelydinosaur: Tho unrelated to the rest of the PR. Which. Hrm.

lovelydinosaur: If we update that link, let's have it as a separate PR, from someone on the team.

Kludex (approved):

Kludex:
Thanks @imba-tjd 🙇

PR #1563, opened 2022-07-10
Drop httpx version to 0.23.0

Kludex:
0.23.0 is actually the latest... And dependabot is not able to go back from 1.0 to 0.x.


jhominal (approved):

Kludex:
Thanks @jhominal :pray:

PR #1567, opened 2022-07-12
Use skipif marker when reload package not installed

Kludex:
As we are going to remove `watchgod` soon (we should create a deprecation policy...) we can just replace the line that I'm introducing here by:

```python
pytestmark = pytest.mark.skipif("watchfiles" not in sys.modules)
```
Whenever we remove `watchgod`.

- This is an alternative for #1537.

### DIFF
--- a/tests/supervisors/test_reload.py
+++ b/tests/supervisors/test_reload.py
@@ -1,5 +1,6 @@
 import logging
 import signal
+import sys
 from pathlib import Path
 from time import sleep
 from typing import Type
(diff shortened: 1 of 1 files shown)


Kludex:
@euri10 Feel free to choose between this and #1537.

The difference is that here we just don't test at all if any of those packages is not found, on the other we, we do test on the packages that are installed.

euri10:
I think the latter is easier for packagers so I merge it

bnavigator:
FYI there is a [`pytest.importorskip()`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip)

PR #1575, opened 2022-07-15
Add note about `--no-server-header` limitations when using WebSockets implementations

iudeen:


Kludex:
I'm not sure if we should document only for `websockets`, because `wsproto` also doesn't support it.

Should we add support for `--no-server-header` on `wsproto`?

iudeen:
@Kludex i would say let's add. You asked me to check the original PR in which server header was added to understand why Websockets were left out. But I don't seem to find that one!

Currently, for wsproto server headers are not present even if `no-server-header` is not passed. 
In case we are adding support for wsproto, I can try to get my hands on it.

Kludex:
> You asked me to check the original PR in which server header was added to understand why Websockets were left out. But I don't seem to find that one!

Using:
```
git log -Sdefault_headers
```

We see the first commit on which it appears: https://github.com/encode/uvicorn/commit/07c6c18945ab4f4167118e717e197adce3aee3da.

But anyway... For now, I guess we can go with only the note on the docs.

docs/settings.md
```diff
@@ -67,6 +67,9 @@ For more nuanced control over which file modifications trigger reloads, install
 * `--lifespan <str>` - Set the Lifespan protocol implementation. **Options:** *'auto', 'on', 'off'.* **Default:** *'auto'*.
 * `--h11-max-incomplete-event-size <int>` - Set the maximum number of bytes to buffer of an incomplete event. Only available for `h11` HTTP protocol implementation. **Default:** *'16384'* (16 KB).
 
+!!! note
+    On selecting `websockets` as Websocket protocol, the server headers are not disabled even if `--no-server-header` settings is passed. This is a limitation.
```
Kludex: Instead of being here, it could be where the `--no-server-header` is mentioned, right? :thinking: 
```suggestion
    The `--no-server-header` flag doesn't have effect on the WebSockets implementations.
```

iudeen: Sure, it makes sense! Accepted your suggestion! 

Thanks!!

docs/settings.md
```diff
@@ -67,6 +67,7 @@ For more nuanced control over which file modifications trigger reloads, install
 * `--lifespan <str>` - Set the Lifespan protocol implementation. **Options:** *'auto', 'on', 'off'.* **Default:** *'auto'*.
 * `--h11-max-incomplete-event-size <int>` - Set the maximum number of bytes to buffer of an incomplete event. Only available for `h11` HTTP protocol implementation. **Default:** *'16384'* (16 KB).
 
+
```
```suggestion
```

Kludex (approved):
Thanks @iudeen 🙏

PR #1576, opened 2022-07-17
Small setup.py refactor

Maroloccio:
Minor clean-up refactor of `setup.py`:

* Remove unnecessary shell quoting
* Match delimiting quotes in RegEx
* Factor our common code
* Canonicalise and sort package metadata

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -1,18 +1,18 @@
 #!/bin/sh -e
 
 # Use the Python executable provided from the `-p` option, or a default.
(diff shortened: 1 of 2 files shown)


Kludex (commented):
Thanks for the PR, and the intention to improve uvicorn @Maroloccio !


We are planing to comply with PEP 621, and for that reason I don't think a refactor on the setup.py makes sense. 🙏

PR #1577, opened 2022-07-17
Remove unnecessary shell quoting

Maroloccio:


Kludex:
Is there any benefit on doing this?

Maroloccio:
> Is there any benefit on doing this?

I suppose in the presence of something unnecessary we might ask the same question in reverse:

"is" "there" "any" "benefit" "to" "writing" "English" "prose" "like" "this"?

I'd say a concern for readability favours the omission of unnecessary quotes and setting an example to readers of the code would too, lest they copy surrounding code "and" "keep" "following" "the" "example" for... consistency.

Kludex:
I see what you mean.

Refactoring changes on this level are quite difficult to get a _buy_ from maintainers, as opinions can be different, and it creates a noise in the history, the refactor must be worthwhile.

For not believing this is a clear win, I'll be closing the PR.

I'm sorry for rejecting three PRs in a row. I'm really grateful for the intention to improve `uvicorn`, and I hope this will not demotivate you. :disappointed:

PR #1578, opened 2022-07-18
Import from typing

Maroloccio:


Kludex:
Thanks for the PR @Maroloccio !

I'd recommend to first start a discussion about this. It's not clear if we want to add `__future__.annotations` on our source code instead. I'd rather have a conclusion on that topic before refactoring the `typing` imports. 🙏

PR #1580, opened 2022-07-20
Allow force polling when using reload with watchfiles

austinorr:
Recent release prioritizes the excellent `watchfiles` file-system notifications based library over the now-deprecated `watchgod` polling based library. This is overall an excellent development, but it leaves some windows developers who rely on containers out in the cold, since filesystem notifications aren't reliably passed through from windows to linux (see [docker for windows issue](https://github.com/docker/for-win/issues/8479), and [workaround](https://levelup.gitconnected.com/docker-desktop-on-wsl2-the-problem-with-mixing-file-systems-a8b5dcd79b22) tldr: "don't do that, just use WSL's filesystem and everything works")

This PR offers a bridge and escape hatch to allow for some transition time while we migrate our projects from blended file systems (windows + WSL, a natural starting point for many developers dealing with employer-provided windows hardware) to a WSL-only project filesystem format. I promise to start new projects as WSL only, but for now, allowing for polling rather than system notifications seems to work on blended setups.

The only change this PR proposes is to add a command line flag to allow a developer like me to opt in to the new-default-hotness `watchfiles` library while still running the polling variant of the watcher on existing projects. 

The proposed API follows along with conventions established by other reload kwargs, and appears in usage like so:

```shell
uvicorn main:app --reload --reload-force-polling
```

Throughout the source code diff the new variable is called `reload_force_polling` and is entirely optional (default=False).

For testing and coverage I've started this draft PR by modifying existing tests only. Thus far I've modified the test to make sure that the reloader initializes fully and the test to make sure that the polling side of the `watchfiles` api was used and successful for a changed .py file. 

The test patterns in this diff could be easily extended to other tests if needed, though this library may not need to extensively test all permutations of the `watchfiles` polling switch. It seemed more important to just ensure that the flag is honored when its passed. That said, I'm happy to add the modification to other tests if desired.

Thanks for the maintenance, docs, and frequent releases. Please consider this minor enhancement for merging - or let me know what I need to add/change to get it ready.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -53,6 +53,9 @@ Options:
   --reload-delay FLOAT            Delay between previous and next check if
                                   application needs to be. Defaults to 0.25s.
                                   [default: 0.25]
+  --reload-force-polling          Force auto-reload to use polling rather than
+                                  file system notifications. This option has
+                                  no effect unless watchfiles is installed.
   --workers INTEGER               Number of worker processes. Defaults to the
                                   $WEB_CONCURRENCY environment variable if
                                   available, or 1. Not valid with --reload.
--- a/docs/index.md
+++ b/docs/index.md
@@ -120,6 +120,9 @@ Options:
   --reload-delay FLOAT            Delay between previous and next check if
                                   application needs to be. Defaults to 0.25s.
                                   [default: 0.25]
+  --reload-force-polling          Force auto-reload to use polling rather than
+                                  file system notifications. This option has
+                                  no effect unless watchfiles is installed.
   --workers INTEGER               Number of worker processes. Defaults to the
                                   $WEB_CONCURRENCY environment variable if
                                   available, or 1. Not valid with --reload.
--- a/tests/supervisors/test_reload.py
+++ b/tests/supervisors/test_reload.py
@@ -51,9 +51,19 @@ class TestBaseReload:
         reloader.startup()
         return reloader
 
-    def _reload_tester(self, touch_soon, reloader: BaseReload, *files: Path) -> bool:
+    def _reload_tester(
+        self,
+        touch_soon,
+        reloader: BaseReload,
+        *files: Path,
+        force_polling: bool = False,
+    ) -> bool:
         reloader.restart()
-        if WatchFilesReload is not None and isinstance(reloader, WatchFilesReload):
+        if (
+            WatchFilesReload is not None
+            and isinstance(reloader, WatchFilesReload)
+            and not force_polling
+        ):
             touch_soon(*files)
         else:
             assert not next(reloader)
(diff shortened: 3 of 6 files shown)


Kludex:
Hi @austinorr 👋

Can't you just not install `watchfiles`? 🤔

austinorr:
> Hi @austinorr 👋
> 
> Can't you just not install `watchfiles`? 🤔

Well, ya, I guess I could uninstall watchfiles after installing uvicorn[standard] but I don’t know what kind of reload experience to expect. Watchgod is deprecated so I don’t want to depend on that, and if the built in statreloader is just as good then maybe we don’t even need watchfiles in [standard] at all?

Kludex:
We can add a marker on the standard extras for watchfiles on Windows (I didn't check the links you pasted above, so this is an idea, not a suggestion...).

On the same line of thought, you can also not use the standard and install the dependencies manually. This is actually what I do personally... 👀

It would be cool to understand the difference between using statreloader and this flag on watchfiles in terms of performance.

We can revisit the standard extra, but it would be outside this PR. 🙏

austinorr:
I actually do want to and need to use watchfiles with uvicorn and ‘just don’t install watchfiles’ does not leave me with a feature-complete alternative file watcher for ‘—reload’ mode during development. The behavior I need is fully supported by watchfiles, it’s just not accessible while running unicorn without this PR to pass the flag.

samuelcolvin:
I think this can be closed https://github.com/samuelcolvin/watchfiles/pull/170.

Kludex:
Thanks @samuelcolvin 🙏

@austinorr I'll be closing this then (hope is fine by you). If you want to discuss something else, feel free to create a discussion about it.

austinorr:
@kludex I still think this PR deserves at least a review and maybe even a merge, or at least a clear reason why it should be impossible to pass important arguments to `watchfiles` that are related to its core functionality and system compatibility. A lot of users likely get exposed to `watchfiles` only through `uvicorn`. Plus the `fastapi` -> `starlette` -> `uvicorn` ecosystem is pulling in new users all the time, so you'll probably get this issue again and again. This minimal,  tested, and fully covered diff nips all that in the bud and enables the workaround and documents it in the `uvicorn` docs (e.g., "having trouble with reload on wsl? Try using the --reload-force-polling flag as a workaround"). This would be much more helpful than forcing users to discover this hidden env flag in `watchfiles` (for which I am grateful, btw @samuelcolvin ) while trying to understand why their `uvicorn --reload` isn't working after updating from 0.17.x to 0.18.x.

I'm just not seeing a good reason this flag wasn't included in the 0.18.x release that dropped `watchgod`, and I see this diff as a pretty minimal patch on that release.

samuelcolvin:
Not my decision but I would argue against merging this PR:

1. More PRs to review, more time, more important changes get delayed
2. More flags, more cognitive overhead, more scrolling, uvicorn becomes
harder to use.
3. In a stable and popular package you should need a good reason to review
and merge a PR, not a good reason to close a PR.
4. The discoverability point is null, you can add the new env variable to
uvicorn docs with no code changes
5. Much better to add more configuration settings via env variables to
watchfiles so they can be used by all the libraries using watchfiles, now
and in the future, with no future code changes.

Kludex:
What @samuelcolvin said 🙏

PR welcome to add a note about the env variable with the reason for use it.

austinorr:
Sounds good, I will open a new PR pinning watchfiles>=0.16.0 and adding a brief aside about the existence of the env variable to the uvicorn docs.

@Kludex @samuelcolvin Thanks for your time and effort!

PR #1597, opened 2022-08-08
Improve typing of app factory

inyutin:
Original discussion, approved by @Kludex :
https://github.com/encode/uvicorn/discussions/1595#discussioncomment-3348105

> 
> There is a type definition for appin uvicorn.Config.__init__(...):
> app: Union["ASGIApplication", Callable, str],  
> 
> **So app receives any Callable, which is obviously incorrect. My first proposal is to set here type Callable[Any, "ASGIApplication"] instead of plain Callable. By doing this mypy will raise an error and modern code editors will give a hint of an error too.**

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -209,7 +209,7 @@ def _normalize_dirs(dirs: Union[List[str], str, None]) -> List[str]:
 class Config:
     def __init__(
         self,
-        app: Union["ASGIApplication", Callable, str],
+        app: Union["ASGIApplication", Callable[..., Optional["ASGIApplication"]], str],
         host: str = "127.0.0.1",
         port: int = 8000,
         uds: Optional[str] = None,
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -456,7 +456,9 @@ def main(
 
 
 def run(
-    app: typing.Union["ASGIApplication", typing.Callable, str],
+    app: typing.Union[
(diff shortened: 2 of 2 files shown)


Kludex:
Why `Optional`?

inyutin:
@Kludex there is a test in ```test_config.py```, which us an app, defines like that:
```python
def wsgi_app(environ: Environ, start_response: StartResponse) -> None:
    pass  # pragma: nocover
```
My idea is that if test is working that ```Optional``` should be accepted too

Kludex:
That's a WSGI application... The type annotation should be more complex to achieve correctness here.

The complexity of that type to achieve it may not be worth it, and that's maybe why it was just `Callable`.

inyutin:
@Kludex maybe raise error (or make os.exit()) when a factory returns an awaitable object then ?

Plus, yes, this annotation is not fully correct, but it's better than previous one, isn't it ?

Kludex:
Validating the values of the flags was thought before, but we don't do it elsewhere... You can create a discussion to suggest that, if you are up to.

As for this PR... I don't see much benefit if it's not to add the right type annotation, and even if we right that down, I'm not sure it will be worth it because the union will grow, and the benefit will be less.

inyutin:
I'm closing this PR then. I'm not willing to invest more time for it now.
Thank you @Kludex

PR #1600, opened 2022-08-10
Cooperative signal handling

### LINKED ISSUE #1579: uvicorn eats SIGINTs, does not propagate exceptions
The following snippet cannot be killed with a SIGINT (ctrl+c):

```python
import asyncio

async def web_ui():
    await Server(Config(Starlette())).serve()

async def task():
    await asyncio.sleep(100000000000000)

async def main():
    await asyncio.gather(web_ui(), task())

if __name__ == "__main__":
    asyncio.run(main())
```

It appears that uvicorn is eating SIGINTs and does not propagate the KeyboardInterrupt and/or Cancelled exceptions.   Thanks for having a look.

> [!IMPORTANT]

<a href="https://polar.sh/encode/uvicorn/issues/1579">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/1579/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

maxfischer2781:
* [x]  I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
* [x]  I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
* [x]  I've updated the documentation accordingly.

-------------------

This PR adjust the signal handling of `uvicorn.server.Server` to integrate with `asyncio` applications by not suppressing shutdown signals. Major changes include:

* Original signal handlers are restored before `Server.serve` finishes
* When terminated by a signal, the appropriate original signal handler is invoked if possible

In the usual case, this allows CTRL+C to end both `uvicorn` *and* the containing async application. Closes #1579.

-------------------

There is a caveat for tests in that signal handling on Windows is not as precise as on Unix (see e.g. [Stackoverflow: How to handle a signal.SIGINT on a Windows OS machine?](https://stackoverflow.com/questions/35772001/how-to-handle-a-signal-sigint-on-a-windows-os-machine)). For testing, I could not setup a situation where the tests receive a signal *without the test environment also being interrupted by it*.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -260,7 +260,6 @@ if __name__ == "__main__":
     asyncio.run(main())
 ```
 
-
 ### Running with Gunicorn
 
 [Gunicorn][gunicorn] is a mature, fully featured server and process manager.
--- a/docs/index.md
+++ b/docs/index.md
@@ -260,6 +260,10 @@ if __name__ == "__main__":
     asyncio.run(main())
 ```
 
+!!! note
+    Both `server.run` and `server.serve` install custom signal handlers while the server runs.
+    The original signal handlers are restored and, if appropriate, invoked once the server finishes.
+
 ### Running with Gunicorn
 
 [Gunicorn][gunicorn] is a mature, fully featured server and process manager.
--- a/docs/index.md
+++ b/docs/index.md
@@ -260,9 +260,6 @@ if __name__ == "__main__":
     asyncio.run(main())
 ```
 
-!!! note
-    Both `server.run` and `server.serve` install custom signal handlers while the server runs.
-    The original signal handlers are restored and, if appropriate, invoked once the server finishes.
 
 ### Running with Gunicorn
 
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -46,7 +46,6 @@ else:
 
 
 @pytest.mark.anyio
-@pytest.mark.skipif(sys.platform == "win32", reason="require unix-like signal handling")
 @pytest.mark.parametrize("exception_signal", signals)
 @pytest.mark.parametrize("capture_signal", signal_captures)
 async def test_server_interrupt(
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1,8 +1,10 @@
-from typing import Generator
+from __future__ import annotations
+
 import asyncio
 import os
 import signal
 import sys
+from typing import Generator
 
 import pytest
 
@@ -15,7 +17,7 @@ class CaughtSignal(Exception):
 
 
 @pytest.fixture(params=[signal.SIGINT, signal.SIGTERM])
-def exception_signal(request: "type[pytest.FixtureRequest]") -> Generator[signal.Signals, None, None]:  # pragma: py-win32
+def exception_signal(request: type[pytest.FixtureRequest]) -> Generator[signal.Signals, None, None]:  # pragma: py-win32
     """Fixture that replaces SIGINT/SIGTERM handling with a normal exception"""
 
     def raise_handler(*_: object) -> None:
@@ -27,12 +29,15 @@ def exception_signal(request: "type[pytest.FixtureRequest]") -> Generator[signal
 
 
 @pytest.fixture(params=[signal.SIGINT, signal.SIGTERM])
-def async_exception_signal(request: "type[pytest.FixtureRequest]") -> Generator[signal.Signals, None, None]:  # pragma: py-win32
+def async_exception_signal(  # pragma: py-win32
+    request: type[pytest.FixtureRequest]
+) -> Generator[signal.Signals, None, None]:
     """Fixture that replaces SIGINT/SIGTERM handling with a normal exception"""
 
     def raise_handler(*_: object) -> None:
         raise CaughtSignal
 
+    asyncio.get_running_loop().add_signal_handler
     original_handler = signal.signal(request.param, raise_handler)
     yield request.param
     signal.signal(request.param, original_handler)
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -15,7 +15,7 @@ from uvicorn.server import Server
 # asyncio does NOT allow raising in signal handlers, so to detect
 # raised signals raised a mutable `witness` receives the signal
 @contextlib.contextmanager
-def capture_signal_sync(sig: signal.Signals) -> Generator[list[int], None, None]:  # pragma: py-win32
+def capture_signal_sync(sig: signal.Signals) -> Generator[list[int], None, None]:
     """Replace `sig` handling with a normal exception via `signal"""
     witness: list[int] = []
     original_handler = signal.signal(sig, lambda signum, frame: witness.append(signum))
@@ -37,10 +37,18 @@ async def dummy_app(scope, receive, send):  # pragma: py-win32
     pass
 
 
+if sys.platform == "win32":
+    signals = [signal.SIGBREAK]
+    signal_captures = [capture_signal_sync]
+else:
+    signals = [signal.SIGTERM, signal.SIGINT]
+    signal_captures = [capture_signal_sync, capture_signal_async]
+
+
 @pytest.mark.anyio
 @pytest.mark.skipif(sys.platform == "win32", reason="require unix-like signal handling")
-@pytest.mark.parametrize("exception_signal", [signal.SIGTERM, signal.SIGINT])
-@pytest.mark.parametrize("capture_signal", [capture_signal_sync, capture_signal_async])
+@pytest.mark.parametrize("exception_signal", signals)
+@pytest.mark.parametrize("capture_signal", signal_captures)
 async def test_server_interrupt(
     exception_signal: signal.Signals, capture_signal: Callable[[signal.Signals], ContextManager[None]]
 ):  # pragma: py-win32
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
-import os
 import signal
 import sys
 from typing import Callable, ContextManager, Generator
@@ -50,7 +49,7 @@ async def test_server_interrupt(
     async def interrupt_running(srv: Server):
         while not srv.started:
             await asyncio.sleep(0.01)
-        os.kill(os.getpid(), exception_signal)
+        signal.raise_signal(exception_signal)
 
     server = Server(Config(app=dummy_app, loop="asyncio"))
     asyncio.create_task(interrupt_running(server))
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import os
 import signal
 import sys
-from typing import Generator
+from typing import Callable, ContextManager, Generator
 
 import pytest
 
@@ -12,35 +13,25 @@ from uvicorn.config import Config
 from uvicorn.server import Server
 
 
-class CaughtSignal(Exception):
-    pass
+# asyncio does NOT allow raising in signal handlers, so to detect
+# raised signals raised a mutable `witness` receives the signal
+@contextlib.contextmanager
+def capture_signal_sync(sig: signal.Signals) -> Generator[list[int], None, None]:  # pragma: py-win32
+    """Replace `sig` handling with a normal exception via `signal"""
+    witness: list[int] = []
+    original_handler = signal.signal(sig, lambda signum, frame: witness.append(signum))
+    yield witness
+    signal.signal(sig, original_handler)
 
 
-@pytest.fixture(params=[signal.SIGINT, signal.SIGTERM])
-def exception_signal(request: type[pytest.FixtureRequest]) -> Generator[signal.Signals, None, None]:  # pragma: py-win32
-    """Fixture that replaces SIGINT/SIGTERM handling with a normal exception"""
-
-    def raise_handler(*_: object) -> None:
-        raise CaughtSignal
-
-    original_handler = signal.signal(request.param, raise_handler)
-    yield request.param
-    signal.signal(request.param, original_handler)
-
-
-@pytest.fixture(params=[signal.SIGINT, signal.SIGTERM])
-def async_exception_signal(  # pragma: py-win32
-    request: type[pytest.FixtureRequest],
-) -> Generator[signal.Signals, None, None]:
-    """Fixture that replaces SIGINT/SIGTERM handling with a normal exception"""
-
-    def raise_handler(*_: object) -> None:
-        raise CaughtSignal
-
-    asyncio.get_running_loop().add_signal_handler
-    original_handler = signal.signal(request.param, raise_handler)
-    yield request.param
-    signal.signal(request.param, original_handler)
+@contextlib.contextmanager
+def capture_signal_async(sig: signal.Signals) -> Generator[list[int], None, None]:  # pragma: py-win32
+    """Replace `sig` handling with a normal exception via `asyncio"""
+    witness: list[int] = []
+    original_handler = signal.getsignal(sig)
+    asyncio.get_running_loop().add_signal_handler(sig, witness.append, sig)
+    yield witness
+    signal.signal(sig, original_handler)
 
 
 async def dummy_app(scope, receive, send):  # pragma: py-win32
@@ -49,7 +40,11 @@ async def dummy_app(scope, receive, send):  # pragma: py-win32
 
 @pytest.mark.anyio
 @pytest.mark.skipif(sys.platform == "win32", reason="require unix-like signal handling")
-async def test_server_interrupt(exception_signal: signal.Signals):  # pragma: py-win32
+@pytest.mark.parametrize("exception_signal", [signal.SIGTERM, signal.SIGINT])
+@pytest.mark.parametrize("capture_signal", [capture_signal_sync, capture_signal_async])
+async def test_server_interrupt(
+    exception_signal: signal.Signals, capture_signal: Callable[[signal.Signals], ContextManager[None]]
+):  # pragma: py-win32
     """Test interrupting a Server that is run explicitly inside asyncio"""
 
     async def interrupt_running(srv: Server):
@@ -59,7 +54,8 @@ async def test_server_interrupt(exception_signal: signal.Signals):  # pragma: py
 
     server = Server(Config(app=dummy_app, loop="asyncio"))
     asyncio.create_task(interrupt_running(server))
-    with pytest.raises(CaughtSignal):
+    with capture_signal(exception_signal) as witness:
         await server.serve()
+    assert witness
     # set by the server's graceful exit handler
     assert server.should_exit
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -30,7 +30,7 @@ def exception_signal(request: type[pytest.FixtureRequest]) -> Generator[signal.S
 
 @pytest.fixture(params=[signal.SIGINT, signal.SIGTERM])
 def async_exception_signal(  # pragma: py-win32
-    request: type[pytest.FixtureRequest]
+    request: type[pytest.FixtureRequest],
 ) -> Generator[signal.Signals, None, None]:
     """Fixture that replaces SIGINT/SIGTERM handling with a normal exception"""
 
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1,0 +1,60 @@
+from typing import Generator
+import asyncio
+import os
+import signal
+import sys
+
+import pytest
+
+from uvicorn.config import Config
+from uvicorn.server import Server
+
+
+class CaughtSignal(Exception):
+    pass
+
+
+@pytest.fixture(params=[signal.SIGINT, signal.SIGTERM])
+def exception_signal(request: "type[pytest.FixtureRequest]") -> Generator[signal.Signals, None, None]:  # pragma: py-win32
+    """Fixture that replaces SIGINT/SIGTERM handling with a normal exception"""
+
+    def raise_handler(*_: object) -> None:
+        raise CaughtSignal
+
+    original_handler = signal.signal(request.param, raise_handler)
+    yield request.param
+    signal.signal(request.param, original_handler)
+
+
+@pytest.fixture(params=[signal.SIGINT, signal.SIGTERM])
+def async_exception_signal(request: "type[pytest.FixtureRequest]") -> Generator[signal.Signals, None, None]:  # pragma: py-win32
+    """Fixture that replaces SIGINT/SIGTERM handling with a normal exception"""
+
+    def raise_handler(*_: object) -> None:
+        raise CaughtSignal
+
+    original_handler = signal.signal(request.param, raise_handler)
+    yield request.param
+    signal.signal(request.param, original_handler)
+
+
+async def dummy_app(scope, receive, send):  # pragma: py-win32
+    pass
+
+
+@pytest.mark.anyio
+@pytest.mark.skipif(sys.platform == "win32", reason="require unix-like signal handling")
+async def test_server_interrupt(exception_signal: signal.Signals):  # pragma: py-win32
+    """Test interrupting a Server that is run explicitly inside asyncio"""
+
+    async def interrupt_running(srv: Server):
+        while not srv.started:
+            await asyncio.sleep(0.01)
+        os.kill(os.getpid(), exception_signal)
+
+    server = Server(Config(app=dummy_app, loop="asyncio"))
+    asyncio.create_task(interrupt_running(server))
+    with pytest.raises(CaughtSignal):
+        await server.serve()
+    # set by the server's graceful exit handler
+    assert server.should_exit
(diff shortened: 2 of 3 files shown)


maxfischer2781:
Looks like I have to read up on Windows signals a bit. Not sure if the code or test is at fault here.

maxfischer2781:
I need help how to proceed at this point with regards to coverage. The changes are covered by tests, *but* the tests cannot be run on Windows and thus the coverage check fails there.

As far as I can tell, there is no way to test this on Windows without significantly more setup. A subprocess would be required to use CREATE_NEW_PROCESS_GROUP in order to contain the interrupt to just the test.

Should I just lower the coverage limit? It seems to already be set to a lower result to account for Windows.

JoanFM:
Hey @maxfischer2781 ,

I wonder if your PR would make this PR #1768 unnecessary?

JoanFM:
> Hey @maxfischer2781 ,
> 
> I wonder if your PR would make this PR #1768 unnecessary?

Yes, I think ur implementation is better and involves less magic

JoanFM:
It may be nice to add a test covering `SIGTERM` signal handling?

maxfischer2781:
> Hey @maxfischer2781 ,
> 
> I wonder if your PR would make this PR #1768 unnecessary?

@JoanFM Possibly, but the two changes have slightly different behaviour. This PR will trigger the original handler *after* the univocrn shutdown procedure. #1768 will trigger the original handler *before* the uvicorn shutdown procedure.

As a default, I think triggering the original handlers late is better – many handlers shutdown quickly (namely, the default handlers) and would skip the uvicorn shutdown entirely. It might be worth having a method to chain handlers explicitly, though.

JoanFM:
Yes, plus your PR does not use any private function and would work with more `loops`.

One question I have, will this PR respect the signals passed to an `event loop` as:

```python
loop.add_signal_handler(...)
```
instead of 

```python
signal.signal(...)
```
?

JoanFM (approved):

maxfischer2781:
> One question I have, will this PR respect the signals passed to an `event loop` as:
> 
> ```python
> loop.add_signal_handler(...)
> ```
> 
> instead of
> 
> ```python
> signal.signal(...)
> ```
> 
> ?

Sadly it won't. There is no clean way to query the loop for registered handlers – handlers are stored in a private attribute of the loop and methods allow only to set or delete handlers by signal number.

JoanFM:
> > One question I have, will this PR respect the signals passed to an `event loop` as:
> > ```python
> > loop.add_signal_handler(...)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead of
> > ```python
> > signal.signal(...)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> Sadly it won't. There is no clean way to query the loop for registered handlers – handlers are stored in a private attribute of the loop and methods allow only to set or delete handlers by signal number.

Okey, but signal.signal should work the same except if I need to interact with event loop itself.

Any idea when this might be merged?

Kludex:
> Any idea when this might be merged?

Is it me, or this question sounds a bit presumptious? 😅

Thanks for the PR @maxfischer2781 , and sorry the delay. This requires a bit of my focus, as there are many ideias related around.

I'll check the whole signal thingy for 0.23.0. I'm about to release 0.22.0.

No estimation can be provided, since I work on uvicorn on my free time.

JoanFM:
> > Any idea when this might be merged?
> 
> Is it me, or this question sounds a bit presumptious? 😅
> 
> Thanks for the PR @maxfischer2781 , and sorry the delay. This requires a bit of my focus, as there are many ideias related around.
> 
> I'll check the whole signal thingy for 0.23.0. I'm about to release 0.22.0.
> 
> No estimation can be provided, since I work on uvicorn on my free time.

Sorry @Kludex, No bad intentions in this question. I really appreciate the work you guys do and understand it is not easy to dedicate time.

I just do not know how this specific project worked.

Thanks for everything

setup.cfg
```diff
@@ -82,7 +82,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 97.69
+fail_under = 97.10
```
Kludex: I'm not willing to drop the current coverage, but we can use strategic `# pragma's`.

uvicorn/server.py
```diff
@@ -54,12 +55,23 @@ def __init__(self, config: Config) -> None:
...
 
     def run(self, sockets: Optional[List[socket.socket]] = None) -> None:
         self.config.setup_event_loop()
-        return asyncio.run(self.serve(sockets=sockets))
+        try:
+            return asyncio.run(self.serve(sockets=sockets))
+        except KeyboardInterrupt:
+            pass
 
     async def serve(self, sockets: Optional[List[socket.socket]] = None) -> None:
```
Kludex: Can we use the `_capture_exit_signals` as decorator instead? It would generate less footprint.

```suggestion
    @capture_signals
    async def serve(self, sockets: Optional[List[socket.socket]] = None) -> None:
```
In the suggestion, `capture_signals` = `_capture_exit_signals`.

uvicorn/server.py
```diff
@@ -54,12 +55,23 @@ def __init__(self, config: Config) -> None:
...
+        """
+        Serve requests asynchronously
+
+        This method should be used from an already existing `async` environment
+        to run the server alongside other `async` tasks.
+        The program's signal handlers are modified while this method is running.
+        """
```
Kludex:
```suggestion
```

uvicorn/server.py
```diff
@@ -283,13 +298,38 @@ async def shutdown(self, sockets: Optional[List[socket.socket]] = None) -> None:
...
+                if callable(handler):
+                    handler(self._delayed_signal, None)
+                # ... if the signal is to be ignored anyway we do nothing...
+                elif handler == signal.SIG_IGN:
+                    pass
+                # ... if we are dealing with an internal/C signal handler,
+                #   repeat the signal to trigger the handler directly.
+                # Windows Caveat: We have no guarantee that we can correctly send
+                #   the signal (due to CTRL+C as SIGINT) to the process (due to groups)!
+                #   For now, do not do anything in this situation.
```
Kludex: Can we have some reference instead?

uvicorn/server.py
```diff
@@ -283,13 +298,38 @@ async def shutdown(self, sockets: Optional[List[socket.socket]] = None) -> None:
...
+            yield
+        else:
+            original_handlers = {sig: signal.getsignal(sig) for sig in HANDLED_SIGNALS}
+            self.install_signal_handlers()
+            yield
+            for sig, handler in original_handlers.items():
+                signal.signal(sig, handler)
+            # if we did gracefully shut down due to a signal, try to
+            # trigger the expected behaviour for this signal now
+            if self._delayed_signal is not None:
```
Kludex: Can you help me understand a bit better why this is needed, and the scenario on which this is triggered?

If I got it right, this is triggered on `SIGINT` or `SIGTERM`, due to `handle_exit`, and it's triggered after the server is already down. But... Why do we need to call the original handlers? Why the `os.kill` below is not enough?

uvicorn/server.py
```diff
@@ -299,7 +339,7 @@ def install_signal_handlers(self) -> None:
                 signal.signal(sig, self.handle_exit)
 
     def handle_exit(self, sig: int, frame: Optional[FrameType]) -> None:
-
+        self._delayed_signal = signal.Signals(sig)
```
Kludex: What is the supposed behavior of multiple signals here? Because we only care about the last one...

Kludex (changes requested):
Does this solve https://github.com/encode/uvicorn/issues/1301 as well?

Ok, review done. Happy new year! :sunglasses:

Kludex:
Relevant: https://github.com/encode/uvicorn/discussions/1708

uvicorn/server.py
maxfischer2781: This just stores one signal here. [`signal.Signals`](https://docs.python.org/3/library/signal.html#signal.Signals) is the enum of all signals, `signal.Signals(sig)` selects the one corresponding to `sig`.

uvicorn/server.py
maxfischer2781: The scenario is as you understood.

Calling the original signal handler directly allows avoiding the non-portability of `os.kill`, i.e. it works the same on Windows and others.  
Would you prefer the code to unconditionally use `os.kill` on non-Windows? That would probably be more in line with what people are used to.

uvicorn/server.py
maxfischer2781: It's doable but would create a similar/larger footprint at another place.

In principle it is possible to refactor `_capture_exit_signals` to work as a decorator as well: a `contextlib.contextmanager` decorated *function* can be used as a decorator. However, this does not work for methods out of the box; I would have to write a new `contextlib.contextmanager`.

uvicorn/server.py
Kludex: Yeah, I'm asking why it only stores one.

maxfischer2781: Good question. I did not find anything on handling a sequence of signals, but it is important that [the "fatal" signal is propagated](https://www.cons.org/cracauer/sigint.html) – just handling the last does that.

According to [this SO resource](https://stackoverflow.com/questions/48434964/interruption-of-signal-handler-with-other-signal), at least the de-facto behaviour on Linux is for signal handlers to be interrupted by newer signals. That means if the currently handled signal does not kill the application, the previously invoked signal handler gets a chance. That suggests a LIFO handling of signals.

I will refactor this to storing all signals and replaying them in LIFO order.

setup.cfg
maxfischer2781: Coverage is up to the 98.80 from `master` again.

uvicorn/server.py
maxfischer2781: I've experimented a bit with this but ultimately decided to drop my sketch for a context manager method. Stacking a context manager on top of a descriptor (for the `self` binding) on top of `async def` is much more complexity than a regular `contextmanager`.

Let me know if you would prefer the decorator variant anyway.

uvicorn/server.py
maxfischer2781: I've added a reference to an Stack Overflow description of Windows signal handling and (re-)raising SIGINT/CTRL+C. Should I remove the other comments in the method?

uvicorn/server.py
maxfischer2781: I've changed the order so that the common UNIX case is handled first. The workarounds are only tried on Windows-like platform. This should make it clearer what is the UNIX case and what are the workarounds.

uvicorn/server.py
maxfischer2781: All captured signals are stored now and replayed in LIFO order.

maxfischer2781:
@Kludex I've addressed/implemented all suggestions now. Please take another look once you find the time.

I don't think an approach as suggested in #1708 makes sense in uvicorn: Practically, signal handling is a two-step approach of 1) handling the signal and setting the shutdown flag and 2) actually shutting down the server. That means one cannot run a "regular" signal handler before or after the uvicorn signal handler (1), since one needs to wait for the server to shut down (2) first.

setup.cfg
```diff
@@ -82,7 +82,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 97.69
+fail_under = 98.80
```
Kludex: Why I'm seeing this line? Isn't it rebased on master? :thinking:

uvicorn/server.py
Kludex: I'm thinking if we should do something like:

```python
async def serve(self, sockets):
    with self._capture_exit_signals():
        self._serve(sockets)
```
This is not a suggestion, it's a thought.

uvicorn/server.py
```diff
@@ -283,13 +292,38 @@ async def shutdown(self, sockets: Optional[List[socket.socket]] = None) -> None:
...
+        else:
+            original_handlers = {sig: signal.getsignal(sig) for sig in HANDLED_SIGNALS}
+            self.install_signal_handlers()
+            yield
+            for sig, handler in original_handlers.items():
+                signal.signal(sig, handler)
+            # If we did gracefully shut down due to a signal, try to
+            # trigger the expected behaviour now; multiple signals would be
+            # done LIFO, see https://stackoverflow.com/questions/48434964
+            for delayed_signal in self._delayed_signals[::-1]:
```
Kludex:
```suggestion
            for delayed_signal in reversed(self._delayed_signals):
```

Kludex (commented):
I'll check this better later on.

uvicorn/server.py
maxfischer2781: Good thought. The diff is much more on point like this.

setup.cfg
maxfischer2781: Not sure why this is still in the diff. It's been rebased again, so there shouldn't be any difference here. :/

maxfischer2781:
@Kludex Do you still see any required changes? As far as I can tell I should have everything covered.

DrInfiniteExplorer:
@Kludex It would be super to see this merged! 🙏

BilalAlpaslan (approved):

br3ndonland:
> @Kludex Do you still see any required changes? As far as I can tell I should have everything covered.

@maxfischer2781 thanks for your work on this. Since you opened the PR, we've added a PR template with a checklist. For your PR, I think it would currently look like this:

So it seems like what's missing is some documentation of the server's signal handling behavior. We have a [docs page on server behavior](https://www.uvicorn.org/server-behavior/) that could work for this ([it's here in the repo code](https://github.com/encode/uvicorn/blob/3924ef4d48a5778d3465666bb898245dbfbf8fe5/docs/server-behavior.md)). If you could add a section there and clearly explain how signal handling will work after this PR, I think it would help everyone understand this better.

After some docs are added to explain the signal handling behavior, I'm happy to give this PR a review as well.

maxfischer2781:
@br3ndonland Thanks for the reply, I'll prepare proper docs. Can you give me some input on things I would change for making this "official" via documentation:

* I think [Deployment > Running programmatically](https://www.uvicorn.org/deployment/#running-programmatically) would be a more suitable place. The exact behaviour is only relevant for people that may fiddle with signals, to everyone else it'll "just work" now.
    * Would this location be fine?
* If this gets documented as something programmers can rely on, I would prefer to make it portable and consistent. Right now the Windows and Unix versions behave slightly differently: the Unix version uses the [`asyncio` signal interface](https://docs.python.org/3/library/asyncio-eventloop.html#unix-signals) but does not use its capabilities. This makes recovering some signal handlers impossible, but technically some user code may rely on the `asyncio` capabilites.
     *  Would consistently using the plain `signal` package be fine? This is technically a breaking change, though not for documented behaviour.

br3ndonland:
> I think [Deployment > Running programmatically](https://www.uvicorn.org/deployment/#running-programmatically) would be a more suitable place.

Sure! This PR is about running the server programmatically, so either the server behavior page or the deployment docs on running programmatically could work. Thanks for the suggestion.

> If this gets documented as something programmers can rely on, I would prefer to make it portable and consistent. Right now the Windows and Unix versions behave slightly differently: the Unix version uses the [`asyncio` signal interface](https://docs.python.org/3/library/asyncio-eventloop.html#unix-signals) but does not use its capabilities. This makes recovering some signal handlers impossible, but technically some user code may rely on the `asyncio` capabilites.
>
> Would consistently using the plain `signal` package be fine? This is technically a breaking change, though not for documented behaviour.

You make a good point here. I agree that the implementation should be as portable and consistent as possible, but given the long discussion in this PR and in #1579, it might be most pragmatic to keep the existing behavior as-is and just document it. I'm happy to take a look at the code either way.

formbook:
Has it been decided to leave the broken behaviour?

Kludex:
> Has it been decided to leave the broken behaviour?

No.

maxfischer2781:
I'm still committed to finishing this but seeing how long this dragged on my initial time window for working on it is gone. It may take about 1-4 weeks until I have time to work on this again.

br3ndonland:
> I'm still committed to finishing this but seeing how long this dragged on my initial time window for working on it is gone. It may take about 1-4 weeks until I have time to work on this again.

No problem. Thanks for checking in.

lukasK9999:
Hello, documentation is only blocker of merging this PR?

chroniac:
Any update on this?

maxfischer2781:
To give an update here: I got delayed beyond my initial time window but am currently catching up with changes in `master` (this should not take much time but means a force push).

However, adding the changes to the docs means they are sort-of established policy, so should be bulletproof. I had to do some extra research on cross compatibility since both the windows and unix version currently behave slightly different and I need to fix that for programmers to rely on it.

At least the code/docs changes from my side should be done within about a week.

Kludex:
Thanks.

maxfischer2781:
I've recreated the branch from scratch now since the original state was heavily outdated and had a nasty git history. The changes are simpler now as Windows/Unix and Python version special cases are gone.

- simpler than original code because `signal.raise_signal` works on every supported version now
- use `signal.signal` on any platform since `asyncio…add_signal_handler` does not allow restoring handlers
    - this has no downside since the handler does not interact with the event loop in any case
- the "main" signal handler method is now called `capture_signals` instead of `install_signal_handlers` since changes are not backwards compatible
- the test setup is inverted to test third-party `asyncio…add_signal_handler` as well

The behaviour is documented on "Index" -> "Running programmatically" since this is the only place where the `server.serve` method is shown.

Kludex:
Oh! You shortened a lot of the diff here. So cool! I'll check in some hours.

docs/index.md
```diff
@@ -260,6 +260,10 @@ if __name__ == "__main__":
     asyncio.run(main())
 ```
 
+!!! note
+    Both `server.run` and `server.serve` install custom signal handlers while the server runs.
+    The original signal handlers are restored and, if appropriate, invoked once the server finishes.
```
Kludex: Without a code example here, I'm not so sure how helpful this message is.

uvicorn/server.py
```diff
@@ -302,22 +307,28 @@ async def _wait_tasks_to_complete(self) -> None:
...
         if threading.current_thread() is not threading.main_thread():
-            # Signals can only be listened to from the main thread.
+            yield
             return
-
-        loop = asyncio.get_event_loop()
-
+        # always use signal.signal, even if loop.add_signal_handler is available
+        # this allows to restore previous signal handlers later on
+        original_handlers = {sig: signal.signal(sig, self.handle_exit) for sig in HANDLED_SIGNALS}
```
Kludex, line 318: Why did we use `loop.add_signal_handler` before?

tests/test_server.py
```diff
@@ -0,0 +1,60 @@
...
+    signal.signal(sig, original_handler)
+
+
+async def dummy_app(scope, receive, send):  # pragma: py-win32
+    pass
+
+
+@pytest.mark.anyio
+@pytest.mark.skipif(sys.platform == "win32", reason="require unix-like signal handling")
+@pytest.mark.parametrize("exception_signal", [signal.SIGTERM, signal.SIGINT])
```
Kludex: What happens with `signal.SIGBREAK` on Windows?

We added this to `HANDLED_SIGNALS` some months ago.

docs/index.md
maxfischer2781: Would you prefer a code example or removing the message? As far as I can tell, adding custom signal handlers isn't all that common, and restoring the default signal handlers doesn't make sense as an example.

uvicorn/server.py
maxfischer2781: I haven't found a technical reason for it. It seems to have been added in #141 for no specific and  then copied over the years. I assume it was used because the `asyncio` docs promote it as a Linux feature that is better in unspecified ways.

On the technical side the *only* advantage of `loop.add_signal_handler` is that it allows synchronously triggering async code (e.g. `event.set()`), and the handlers don't use that. Since the code has to be compatible with Windows, handlers cannot rely on being invoked by the loop anyways.

tests/test_server.py
maxfischer2781: I've added that case now.

docs/index.md
Kludex: I would prefer for it to be removed, cause I believe your PR is the logical behavior.

Kludex:
```suggestion
```

docs/index.md
```diff
@@ -260,6 +260,7 @@ if __name__ == "__main__":
     asyncio.run(main())
 ```
 
+
```
Kludex:
```suggestion
```

Kludex (approved):

Kludex:
Thanks for not giving up @maxfischer2781 🙏

I'll make a release shortly.

vytas7:
In Falcon's test suite, we used to check the return code for functional tests of ASGI servers. Since we terminate the server via `SIGTERM`, Uvicorn now returns `-15` which is a proper way in Unix. However, when running _Actions_ on Windows, we are getting codes like `3221225786 == 2**31 + 2**30 + 314`. Is this expected on Windows, and how to interpret these values? @Kludex

maxfischer2781:
3221225786  is the unsigned interpretation of Windows' [`0xC000013A` / `STATUS_CONTROL_C_EXIT`](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55). This is a magic value just like signal `15` on UNIX.

vytas7:
I see, thanks @maxfischer2781!

shadchin:
Please tell me, is this the expected behavior after this pull request?

main.py
```python

    await send({
        'status': 200,
        'headers': [
        ],
    })
    await send({
        'type': 'http.response.body',
        'body': b'Hello, world!',
    })
```

# uvicorn 0.28.0
```
$ uvicorn main:app
...
$ kill -15 <proc_id>
$ echo $?
0
```
# uvicorn 0.29.0
```
$ uvicorn main:app
...
$ kill -15 <proc_id>
$ echo $?
143
```
non zero return code

maxfischer2781:
Yes, this is expected. Notice that 143 is 128+15, bash's way of saying the program was killed by signal 15 - i.e. exactly what you requested with `kill -15 <proc_id>`.

Kludex:
@maxfischer2781 Are you able to run `tests/supervisors/test_signal.py` successfully locally? We may have broken the `--timeout-graceful-shutdown` flag - I didn't have time to check properly yet.

shadchin:
Thanks for the answer

I was a little confused, because I was used to SIGTERM being a regular program termination mechanism and expecting a return code of 0, that everything is fine, now we get a non-zero return code, maybe this will break someone's work under supervisors

maxfischer2781:
@Kludex The tests are working for me locally, but that's MacOS. Windows might be behaving differently again *sigh*. If you have any leads that there is a problem, can you open a ticket and ping me? I'll have a look then.

Kludex:
I use MacOS as well. Did you run only that file?

maxfischer2781:
Now I did. I can confirm that running the tests in isolation makes them fail due to an unhandled `KeyboardInterrupt`.

What's weird is that the full test suite runs the `test_signal.py` *before* `test_server.py`, i.e. before I muck around with the exception handlers. I need to dig around a bit what to see exactly is failing here.

--------

Edit: Just running both `test_signal.py` and `test_server.py` isn't enough to hide the failure. There must be another test that interferes here.

--------

Edit: Running any of the other supervisor tests is sufficient to hide the failure. I'm guessing this is because they all install custom signal handlers but do not remove them.

As far as I can tell there are three related "issues" here:
1. The supervisors don't play nice to running in the same process, since they all modify the signal handlers without restoring them.
2. The various tests and utilities aren't prepared to contain signals/exceptions, as that wasn't necessary previously.
3. My change re-throwing `SIGINT` upsets both of these, since 1. means tests may silence signals but 2. means tests are not prepared for signals.

My guess is that this may be hiding a veritable error but I need to double-check that. I'll start by fixing 2 and working up from there, even if it just means we can rely on the tests again.

bekoev:
Hi Max @maxfischer2781 , I would greatly appreciate your advices on the following.

Context:
* Win 11, WSL 2, Ubuntu 22.04
* main.py
  ```py
  from fastapi import FastAPI
  import uvicorn
  
  app = FastAPI()
  
  if __name__ == "__main__":
      uvicorn.run(
          "main:app"
      )
  ```
* Command set 1: `uvicorn main:app` then CTRL+C
* Command set 2: `python main.py` then CTRL+C
* Observation: prior to v. 0.29 in both cases the output corresponds to a graceful shutdown. As of v. 0.29, the command set 2 results in a traceback with asyncio.exceptions.CancelledError and KeyboardInterrupt.

Questions:
1. Is there a (potential) issue in recent versions of uvicorn or the new behaviour is harmless?
2. Is a fix planned or underway?

Thanks in advance!

maxfischer2781:
@bekoev The behaviour is harmless and intentional, but it's been [raised as annoying and misleading](https://github.com/encode/uvicorn/discussions/2368) multiple times now. So this will likely be made less noisy in the future.

PR #1605, opened 2022-08-20
typing: Config.forwarded_allow_ips can also be a list of strings

gertvdijk:
Ultimately, the value is passed to `ProxyHeadersMiddleware` as `trusted_hosts` constructor argument, which accepts either a list of strings or a string.

This is only relevant when running Uvicorn programmatically.

https://github.com/encode/uvicorn/blob/b21ecabc5bf911f571e0629438315a1e5472065c/uvicorn/config.py#L518-L520

https://github.com/encode/uvicorn/blob/b22c9506b282e9c81591ed7f48781dc5759df0e0/uvicorn/middleware/proxy_headers.py#L28

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -370,7 +370,7 @@ class Config:
         if workers is None and "WEB_CONCURRENCY" in os.environ:
             self.workers = int(os.environ["WEB_CONCURRENCY"])
 
-        self.forwarded_allow_ips: Union[List[str], str]  # help mypy
+        self.forwarded_allow_ips: Union[List[str], str]
         if forwarded_allow_ips is None:
             self.forwarded_allow_ips = os.environ.get(
                 "FORWARDED_ALLOW_IPS", "127.0.0.1"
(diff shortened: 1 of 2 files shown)


uvicorn/config.py
```diff
@@ -370,6 +370,7 @@ def __init__(
         if workers is None and "WEB_CONCURRENCY" in os.environ:
             self.workers = int(os.environ["WEB_CONCURRENCY"])
 
+        self.forwarded_allow_ips: Union[List[str], str]  # help mypy
```
Kludex: What happens if this is removed?

Kludex (changes requested):

uvicorn/config.py
gertvdijk: then mypy gets confused about the assignments below.

```
uvicorn/config.py:378: error: Incompatible types in assignment (expression has type "Union[List[str], str]", variable has type "str")  [assignment]
```

should I clarify the comment a bit more? :)

```suggestion
        self.forwarded_allow_ips: Union[List[str], str]
```

Kludex (approved):
Thanks @gertvdijk 🙏

PR #1581, opened 2022-07-25
Add no-server-header support for Web Socket implementations

iudeen:

PR #1606, opened 2022-08-21
Add default headers to `WebSocket`s implementations

iudeen:
Refer https://github.com/encode/uvicorn/pull/1581 for detailed discussion.

Feature Description:

This feature fixes the issue that discussed in https://github.com/encode/uvicorn/discussions/1574

Currently it implements support for `no-server-header` for `wsproto` implementations.


uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -260,12 +261,19 @@ async def asgi_send(self, message: "ASGISendEvent") -> None:
...
+                    if name.lower() in _added_names:
+                        continue
+                    _added_names.append(name.lower())
+                    self.extra_headers.append(
+                        (
+                            # ASGI spec requires bytes
+                            # But for compatibility we need to convert it to strings
+                            name.decode("latin-1"),
+                            value.decode("latin-1"),
+                        )
```
Kludex: You can use the master branch of websockets on this PR, and when they release it we just need to change the version..

iudeen: I have added. Updated the dependency is `setup.py` to point to the master repo of Websockets to run the tests. 

Websockets does not allow `--no-date-header` and adding multiple `Server` headers.

iudeen:
@Kludex the new setup.py alternative seems to not allow direct references, and hence it cannot install the WIP master branch from `websockets` repo. That's why the tests are failing ...

Error:
```
ValueError: Dependency #7 of option `standard` of field `project.optional-dependencies` cannot be a direct reference unless field `tool.hatch.metadata.allow-direct-references` is set to `true`
```

Kludex:
Add that field as true on this PR... Remove it when new websockets is released. 😟

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -260,12 +274,19 @@ async def asgi_send(self, message: "ASGISendEvent") -> None:
...
```
Kludex: Are you going to remove the `_added_names` logic?

docs/settings.md
```diff
@@ -103,4 +103,4 @@ connecting IPs in the `forwarded-allow-ips` configuration.
 
 ## Timeouts
 
-* `--timeout-keep-alive <int>` - Close Keep-Alive connections if no new data is received within this timeout. **Default:** *5*.
+* `--timeout-keep-alive <int>` - Close Keep-Alive connections if no new data is received within this timeout. **Default:** *5*.
```
Kludex: This is removing a new line?

docs/settings.md
```diff
@@ -83,7 +83,7 @@ connecting IPs in the `forwarded-allow-ips` configuration.
 * `--date-header` / `--no-date-header` - Enable/Disable default `Date` header.
 
 !!! note
-    The `--no-server-header` flag doesn't have effect on the WebSockets implementations.
+    The `--no-server-header` and `--no-date-header` flags doesn't have effect on the WebSockets implementations.
```
Kludex: Don't we want to just remove this note, or just let the `--no-date-header` be here?

uvicorn/protocols/websockets/websockets_impl.py
iudeen: Let me remove and check if it causes any issue. If it doesn't, I will remove it.

docs/settings.md
iudeen: Yes, this has to change to just `--no-date-header`

pyproject.toml
```diff
@@ -60,3 +62,4 @@ path = "uvicorn/__init__.py"
 include = [
     "/uvicorn",
 ]
+
```
Kludex:
```suggestion
```

requirements.txt
```diff
@@ -4,8 +4,8 @@
 asgiref==3.5.2
 
 # Explicit optionals
-wsproto==1.1.0
-
+wsproto==1.2.0
+git+https://github.com/aaugustin/websockets.git@main
```
Kludex:
```suggestion
wsproto==1.2.0
git+https://github.com/aaugustin/websockets.git@main

```

uvicorn/protocols/websockets/websockets_impl.py
iudeen: Nope, removing the logic results in the following error

```
  File "D:\ProgramData\Anaconda3\envs\uvicorn\lib\site-packages\websockets\legacy\server.py", line 165, in handler
    await self.handshake(
  File "D:\ProgramData\Anaconda3\envs\uvicorn\lib\site-packages\websockets\legacy\server.py", line 643, in handshake
    response_headers.setdefault("Server", self.server_header)
  File "D:\ProgramData\Anaconda3\envs\uvicorn\lib\_collections_abc.py", line 1007, in setdefault
    return self[key]
  File "D:\ProgramData\Anaconda3\envs\uvicorn\lib\site-packages\websockets\datastructures.py", line 119, in __getitem__
    raise MultipleValuesError(key)
websockets.datastructures.MultipleValuesError: 'Server'
```

iudeen: ~Let me see if there is a way...~

I think its a limitation in Wesockets library. It is not accepting multiple headers with same name.

docs/settings.md
iudeen: Yes, it was. Fixed it.

uvicorn/protocols/websockets/websockets_impl.py
iudeen: I have raised an issue in Websockets repo to get a confirmation.

iudeen: I have raised an issue in Websockets repo to get a confirmation.

iudeen: Update:

As discussed here (https://github.com/aaugustin/websockets/issues/1226#issuecomment-1243022760), it seems like its a bug to be fixed on Wesockets side.

@Kludex, Aymeric asked the use-case for having multiple server headers. I don't see any benefit as such, but is there a better answer?

iudeen: I have PR https://github.com/aaugustin/websockets/pull/1227 to fix the same.

iudeen: The PR was closed as any fix to it was causing backward compatibility issues. 😔

Can we stick to `_added_names` logic?😟

iudeen: Or the other way is just check for `Server` header
```python
                for name, value in headers:
                    if name.lower() in [b"server"]:
                        continue
                    self.extra_headers.append(
                        (
                            # ASGI spec requires bytes
                            # But for compatibility we need to convert it to strings
                            name.decode("latin-1"),
                            value.decode("latin-1"),
                        )
                    )
```

The advantage here is, users can still set other arbitrary headers with same names. Restrictions will be only on Server header.

iudeen:
The default headers test seems to fail on Ubuntu.

Kludex:
Reverting...

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -19,12 +19,13 @@
...
     get_path_with_query_string,
     get_remote_addr,
+    get_server_header,
     is_ssl,
 )
 from uvicorn.server import ServerState
 
 if sys.version_info < (3, 8):
-    from typing_extensions import Literal
+    from typing_extensions import Literal  # pragma: no cover
```
iudeen: Reviewer, I hope this `pragma: no cover` is okay.

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -265,12 +274,17 @@ async def asgi_send(self, message: "ASGISendEvent") -> None:
...
-                if "headers" in message:
-                    self.extra_headers.extend(
-                        # ASGI spec requires bytes
-                        # But for compatibility we need to convert it to strings
-                        (name.decode("latin-1"), value.decode("latin-1"))
-                        for name, value in message["headers"]
+                headers = list(message.get("headers", []))
+                for name, value in headers:
+                    if name.lower() in [b"server"]:
+                        continue
```
iudeen: Websockets doesn't allow multiple `Server` headers. This was added to prevent it from failing. 

Alternatively we can raise a `RuntimeError` as well.

uvicorn/protocols/websockets/websockets_impl.py
Kludex: It's not. Use the `pragma: py-gte-38` and the analogous one, please.

docs/settings.md
```diff
@@ -90,7 +90,7 @@ connecting IPs in the `forwarded-allow-ips` configuration.
 * `--date-header` / `--no-date-header` - Enable/Disable default `Date` header.
 
 !!! note
-    The `--no-server-header` flag doesn't have effect on the WebSockets implementations.
+    The `--no-date-header` flag doesn't have effect on the WebSockets implementations.
```
Kludex: This is not accurate. It should be explicit that only the `websockets` implementation doesn't support it.

tests/protocols/test_websocket.py
```diff
@@ -658,3 +660,226 @@ async def send_text(url):
...
+    cancel_handle = asyncio.ensure_future(server.serve(sockets=None))
+    await asyncio.sleep(0.1)
+    async with websockets.connect("ws://127.0.0.1:8000") as websocket:
+        ws_conn = list(server.server_state.connections)[0]
+        ws_conn.shutdown()
+        await asyncio.sleep(0.1)
+        assert websocket.close_code == 1012
+        assert ws_conn.transport.is_closing()
+    await server.shutdown()
+    cancel_handle.cancel()
```
Kludex: Is this a non-related test?

tests/protocols/test_websocket.py
```diff
@@ -658,3 +660,226 @@ async def send_text(url):
...
+            await self.send({"type": "websocket.accept"})
+
+    config = Config(
+        app=App,
+        ws=ws_protocol_cls,
+        http=http_protocol_cls,
+        lifespan="off",
+    )
+    server = Server(config=config)
+    cancel_handle = asyncio.ensure_future(server.serve(sockets=None))
```
```suggestion
    task = asyncio.create_task(server.serve(sockets=None))
```

tests/protocols/test_websocket.py
```diff
@@ -658,3 +660,226 @@ async def send_text(url):
...
+async def test_multiple_server_header_in_websockets(ws_protocol_cls, http_protocol_cls):
+    class App(WebSocketResponse):
+        async def websocket_connect(self, message):
+            await self.send(
+                {
+                    "type": "websocket.accept",
+                    "headers": [
+                        (b"Server", b"over-ridden"),
+                        (b"Server", b"another-value"),
+                    ],
```
Kludex, line 742: If the user wants to send multiple Server headers (:man_shrugging:) we shouldn't care. As we don't care with `wsproto`.

tests/protocols/test_websocket.py
```diff
@@ -658,3 +660,226 @@ async def send_text(url):
...
+        lifespan="off",
+    )
+    server = Server(config=config)
+    cancel_handle = asyncio.ensure_future(server.serve(sockets=None))
+    await asyncio.sleep(0.1)
+    async with websockets.connect("ws://127.0.0.1:8000"):
+        ws_conn = list(server.server_state.connections)[0]
+        ws_conn.shutdown()
+        assert ws_conn.ws_server.closing is True
+        assert ws_conn.transport.is_closing()
```
Kludex: We are not sending a close code here?

tests/protocols/test_websocket.py
```diff
@@ -658,3 +660,226 @@ async def send_text(url):
...
+            await self.send({"type": "websocket.accept"})
+
+    config = Config(
+        app=App,
+        ws=ws_protocol_cls,
+        http=http_protocol_cls,
+        lifespan="off",
+    )
+    server = Server(config=config)
+    cancel_handle = asyncio.ensure_future(server.serve(sockets=None))
```
Kludex: We should use task group on Python 3.11+ here.

tests/protocols/test_websocket.py
```diff
@@ -658,3 +660,226 @@ async def send_text(url):
...
+        http=http_protocol_cls,
+        lifespan="off",
+    )
+    async with run_server(config):
+        headers = await open_connection("ws://127.0.0.1:8000")
+        assert len(headers.get_all("Server")) == 1
+        assert headers.get("Server") == "uvicorn"
+
+
+@pytest.mark.anyio
```
Kludex: Let's handle those 2 tests below in another PR, please :pray:

tests/protocols/test_websocket.py
iudeen: Yes, this needs to be a new PR.

tests/protocols/test_websocket.py
iudeen: Thanks, will do this in the new PR.

tests/protocols/test_websocket.py
iudeen: sure, I will look into that.

tests/protocols/test_websocket.py
iudeen: Not in uvicorn's code I guess. Let me verify.

tests/protocols/test_websocket.py
iudeen: This was an issue when we pass server header directly to websockets. As per your suggestions moving them to extra headers works in this case. I have merged your suggestions.

iudeen:
Merged suggestions given by Marcelo in https://github.com/encode/uvicorn/pull/1736

tests/protocols/test_websocket.py
iudeen: Close code is `None`. Will detail in new PR.

tests/protocols/test_websocket.py
```diff
@@ -658,3 +659,133 @@ async def send_text(url):
...
+    )
+    async with run_server(config):
+        headers = await open_connection("ws://127.0.0.1:8000")
+        assert "server" not in headers
+
+
+@pytest.mark.anyio
+@pytest.mark.parametrize("ws_protocol_cls", ONLY_WS_PROTOCOL)
+@pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
+async def test_no_date_header(ws_protocol_cls, http_protocol_cls):
```
Kludex, line 709: This is not 100% true. I forgot to mention... The `Date` headers from the `websockets` cannot be removed... :thinking: 

Should we do something about it on the tests? It's on purpose... :thinking: 

It's a question, I'm still thinking.

tests/protocols/test_websocket.py
```diff
@@ -658,3 +659,133 @@ async def send_text(url):
...
+
+    config = Config(
+        app=App,
+        ws=ws_protocol_cls,
+        http=http_protocol_cls,
+        lifespan="off",
+    )
+    async with run_server(config):
+        headers = await open_connection("ws://127.0.0.1:8000")
+        assert headers.get_all("Potato") == ["cool-potato", "super-cool-potato"]
```
Kludex: This is basically the same test as the above. Isn't it? :thinking: 

```suggestion
```

iudeen: In a way yes. Just with a different name. But since we are not using the `server_header` property of websockets, we can remove this test.

tests/protocols/test_websocket.py
iudeen: Do what on tests? There is no way to remove date header in `websockets` I think.

Kludex: I mean, I know, but the tests make us believe that we forgot about `Date` headers... :thinking:

Kludex (approved):
It was a long way, but we got it. :pray: 

Just a remark for myself: the `Date` headers remains on the `websockets` implementation.

tests/protocols/test_websocket.py
iudeen: Continued in https://github.com/encode/uvicorn/pull/1738

Issue #1549, opened 2022-06-28
Update the "Deployments" section on the documentation


Kludex:
That section is not up-to-date with the current most used technologies.

The idea of this issue is to gather information on what people would like to see on that page e.g. "How to deploy on Kubernetes", and also to update the documentation accordingly.

odiseo0:
I'd like to write about this, since in running FastAPI on uvicorn using Elastic Beanstalk

Kludex:
> I'd like to write about this, since in running FastAPI on uvicorn using Elastic Beanstalk

Of course! :)

We need to also think on how to reorganize the docs for those sections to fit there.

adriangb:
@Tinche and I have had some good chats about Uvicorn, Gunicorn and Kubernetes, I might write that section

Kludex:
Before starting, it would be cool to have an idea on how you plan to organize it.

Hi Tin :wave:

iudeen:
I would love to see `How to deploy on Kubernetes?` its something the company I work for is experimenting with for over 1.5 years now. We settled on following the FastAPI docs as of now.  But I am sure there can be further optimizations done.

adriangb:
I think the main question for me in these sorts of docs is how deep we dive into the other pieces involved. Are we going to explain Kubernetes from scratch? Do we assume the reader is familiar with Gunicorn and Uvicorn? Or do we do a short gloss over the high level of each of these while skipping over a lot of the technical detail?

iudeen:
I think a quick "How To" guide relevant to Uvicorn is what most people would benefit from. We can assume that intended audience would know Uvicorn, Gunicorn and Kubernetes to some extent that we can happily skip basics.

iudeen:
### Things which are essential (IMO):
1. Deployment at development environment. 
2. Deployment using systemd (behind Gunicorn/process manager).
3. Deployment using Docker and Docker compose. 
4. Deployment using Kubernetes 

### Nice to have:
5. Running behind NGINX or Traefik
6. Recommended auto scaling for Kubernetes 

### Advanced/Optional:
7. Deployment on serverless (aws, azure, etc)
8. Distributing as exe/bin 


We can debate over what goes under each section and how detailed we should get under each of them. 

Also please feel free to strike off those which might be too much for the docs.

iudeen:
@adriangb are you working on this? Can I also help, if needed?

adriangb:
[...]

PR #1607, opened 2022-08-21
Enhancing Deployment Documentation

iudeen:
Based on the discussion happened at https://github.com/encode/uvicorn/issues/1549


docs/deployment/getting-started.md
```diff
@@ -0,0 +1,180 @@
...
+As a general rule, you probably want to:
+
+* Run `uvicorn --reload` from the command line for local development.
+* Run `gunicorn -k uvicorn.workers.UvicornWorker` for production (non containerized).
+* Run behind Nginx for self-hosted deployments.
+* Finally, run everything behind a CDN for caching support, and serious DDOS protection.
+
+## Quick Start
+
+In the Quick Start section, we will cover how to start serving your application. The application mentioned here is any ASGI compatible application. For instance if an app that was built using a framework like `Starlette` or `FastAPI`.
```
adriangb:
```suggestion
This section is meant to get you started serving an ASGI application as quick as possible. ASGI applications are applications that adhere to the [ASGI Specification], which is the contract between the ASGI server (Uvicorn) and the ASGI application. Starlette  is just one ASGI framework amongst many, see [awesome-asgi] for a listing of some of the others.

[ASGI Specification]: https://asgi.readthedocs.io/en/latest/
[awesome-asgi]: https://github.com/florimondmanca/awesome-asgi
```

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,180 @@
...
+
+The following are ideal for running your application in a development environment.
```
adriangb:
```suggestion
```

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Running from the command line
+
+The simplest way to run `uvicorn` is from the command line.
+
+```bash
+$ uvicorn example:app --reload --port 5000
+```
+
+The ASGI application should be specified in the form `path.to.module:instance.path`
```
adriangb, line 29: What does `instance.path` mean here? As far as I know it's just `path.to.module:asgi_app_global`, right?

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,180 @@
...
+    `--reload` is recommended only for local use, during the development phase.
+    
+    The `--reload` and `--workers` arguments are **mutually exclusive**.
+
+
+
+
+To see the complete set of available options, use `uvicorn --help`:
+
+<!-- :cli_usage: -->
```
adriangb: Can we just link to the `Introduction` section of the docs instead of repeating the CLI options here?

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,180 @@
...
+    ```python
+    uvicorn.run(app, host="127.0.0.1", port=5000, log_level="info")
+    ```
+
+    However, this style only works if you are not using multiprocessing (`workers=NUM`)
+    or reloading (`reload=True`), so we recommend using the import string style.
+
+    Also note that in this case, you should put `uvicorn.run` into `if __name__ == '__main__'` clause in the main module.
+
+    The `reload` and `workers` parameters are **mutually exclusive**.
```
adriangb:
```suggestion
    Remember, the `reload` and `workers` parameters are **mutually exclusive**.
```

docs/deployment/run-in-prod.md
```diff
@@ -0,0 +1,212 @@
...
+## Run using Process Managers
+
+### What is a Process Manager?
+
+In modern computers and operating systems, the processor helps execute multiple programs which are arranged by the process manager. 
+In other words, the **process manager** arranges the execution of the processes or programs by making schedule for to 
+ensure efficient use of the resources which includes both software and hardware.
+
+A process manager will handle the socket setup, start-up multiple server processes, monitor process aliveness, and 
+listen for signals to provide for processes restarts, shutdowns, or dialing up and down the number of running processes.
```
adriangb:
```suggestion
Computers can run multiple programs in parallel using processes. The OS provides primitives for creating a new process and running a program in it, but it does not know what that program is doing. The OS cannot for example realize that your web app is crashing and needs to be restarted. That's where process managers come in. Within the context of a web server, a process manager is a program that monitors the health of multiple server processes and coordinates killing and spinning up of new processes to keep the application online and healthy.
```

docs/deployment/run-in-prod.md
```diff
@@ -0,0 +1,212 @@
...
+
+
+
```
adriangb:
```suggestion

```

docs/deployment/run-in-prod.md
```diff
@@ -0,0 +1,212 @@
...
+!!! question "Do you know?"
+    Uvicorn provides a lightweight way to run multiple worker processes, for example `--workers 4`, but does not provide any process monitoring.
+
+There are several tools you can use as process manager. In this section we will see how to use the following tools:
+
+- Gunicorn
+- Supervisor
+- Circus
```
adriangb, line 20: I have seen Supervisor or Circus being used to serve web apps in production, is these something people do?

I think we can cover:

1. Built in process manager
2. Gunicorn
3. Kubernetes / Cloud Run / PaaS offerings

What do you think?

docs/deployment/run-in-prod.md
```diff
@@ -0,0 +1,212 @@
...
+numprocesses = 4
+
+[socket:web]
+host = 0.0.0.0
+port = 8000
+```
+
+Then run `circusd circus.ini`.
+
+## Run using Docker Containers
```
adriangb, line 113: I don't think we need to dive into Docker here in this much depth. That goes a bit beyond an ASGI server, and there's plenty of content out there on it.

docs/settings.md
```diff
@@ -83,7 +83,7 @@ connecting IPs in the `forwarded-allow-ips` configuration.
 * `--date-header` / `--no-date-header` - Enable/Disable default `Date` header.
 
 !!! note
-    The `--no-server-header` flag doesn't have effect on the WebSockets implementations.
+    The `--no-server-header` and `--no-date-header` flags doesn't have effect on the WebSockets implementations.
```
adriangb: Could these fixes be their own PR? They don't seem directly related to deployments, I'm guessing your just noticed them as you were working on the new docs, right?

docs/deployment/deploy-in-prod.md
```diff
@@ -0,0 +1,14 @@
...
+## Kubernetes
+
+## Systemd
+
+## NGINX
+
+
+
+
+
```
adriangb:
```suggestion

```

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,180 @@
...
+                                  buffer of an incomplete event.
+  --factory                       Treat APP as an application factory, i.e. a
+                                  () -> <ASGI app> callable.
+  --help                          Show this message and exit.
+```
+
+
+See the [settings documentation](../settings.md) for more details on the supported options for running uvicorn.
+
+### Running programmatically
```
adriangb, line 42: I think we should also give an example of using `uvicorn.Config` + `uvicorn.Server` manually. I know that @Tinche and I think @agronholm  (sorry if I am misremembering) use this style a lot . It lets you take control of the event loop as well as startup/shutdown, configuring logging and several other things.

adriangb (commented):
Sorry for the long delay in getting _anything_ done with this, and thank you for putting in the work to write this up.

I think you've done a great job with the structure here! I reworded some stuff that I thought wasn't very clear, but I don't fancy myself a very eloquent writer, maybe Florimond can give it a look once it's closer to being ready.

The main large scale points I'd like to make:

* We really need to be careful with scope creep. I think these docs should go up until other docs / tutorials / material can take over. For example, I don't think we need to go in depth (code examples, etc.) into building a Docker container.
* Let's add a section on using `await uvicorn.Server(uvicorn.Config(app)).run()`
* I think for process managers we should just cover the built in `--workers` argument, Gunicorn and using k8s/similar.
* We may want to give an example here or in the logging docs of setting up structured logs. This is super valuable if you're deploying something.

docs/settings.md
iudeen: Yes, this was supposed to be in another PR. I'll fix it.

iudeen:
Thanks @adriangb for your valuable feedback. I agree, it was hard for me to limit the scope. I wrote what I felt I would need to get started. But it'd be better to link to other docs/materials for further reading and truncate information here. 

It'd be great to see more people add their expertise! @florimondmanca would you mind taking a stab at this?

docs/deployment/getting-started.md
iudeen: I got this referring from either existing documentation or from FastAPI. It was supposed to mean `main:app` or `app.main:app`

docs/deployment/getting-started.md
iudeen: I was thinking we add these here, and link this in introduction.🧐

docs/deployment/getting-started.md
adriangb: Yeah those sound right but `module:instance.path` does not

docs/deployment/getting-started.md
adriangb: Let's start with the smallest change possible, we can always make more sweeping changes to the docs later on if it's warranted

docs/deployment/getting-started.md
iudeen: Agreed. I'll change that!

docs/deployment/getting-started.md
iudeen: I haven’t tried that implementation yet. I’ll add that too!

docs/deployment/run-in-prod.md
iudeen: I think it’s fare to include the recommended Docker image and how to build and run it. I kept it really short.  We can trim further…

florimondmanca:
@iudeen Sure, keeping this one on my list.

docs/deployment/run-in-prod.md
iudeen: It was there in the previous Uvicorn docs. I just copy pasted the same. I don't think we need - but that's my opinion.

docs/deployment/deploy-in-prod.md
```diff
@@ -0,0 +1,9 @@
+# Deploying in Production
```
florimondmanca, line 1: Right now, this page is not used yet, so I initially thought we should drop it.

But actually I realize this PR drops existing information on "running behind NGINX", "running behind a CDN", "running with HTTPS".

I'm also not sure about the relevance of separating "Running in a production environment" vs "Deploying in production". Running in production involves several bits of pieces, the arrangement of which ultimately depends on the user's particular setup and needs (as the very first sentence of the deployment guide reminds us). So, functionally, I think the deployment guide ought to be a series of "How to configure technology X if you're using Uvicorn". That's what the existing guide sticks to and I think that works quite well.

mkdocs.yml
```diff
@@ -24,12 +26,18 @@ edit_uri: ""
 nav:
   - Introduction: "index.md"
   - Settings: "settings.md"
-  - Deployment: "deployment.md"
```
florimondmanca, line 27: The `deployment.md` file was replaced with `deployment/getting-started.md` (mostly), but it still exists, so mkdocs complains. Should we drop it ?

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,182 @@
...
+As a general rule, you probably want to:
+
+* Run `uvicorn --reload` from the command line for local development.
+* Run `gunicorn -k uvicorn.workers.UvicornWorker` for production (non containerized).
+* Run behind Nginx for self-hosted deployments.
+* Finally, run everything behind a CDN for caching support, and serious DDOS protection.
+
+## Quick Start
+
+This section is meant to get you started serving an ASGI application as quick as possible. ASGI applications are applications that adhere to the [ASGI Specification], which is the contract between the ASGI server (Uvicorn) and the ASGI application. Starlette  is just one ASGI framework amongst many, see [awesome-asgi] for a listing of some of the others.
```
florimondmanca: I'm not sure why we mention Starlette here, as it is not referred to in the rest of the page. I thought the reference to `awesome-asgi` (thanks) was meant to point to other servers, but it points to other _frameworks_. Yet the documentation is about Uvicorn as a _server_. So maybe it's a bit confusing?

Besides, I think by the point users reach this page, they'll have some understanding of what ASGI is already, don't think we need to explain it again.

So here's my suggestion:

```suggestion
This section is meant to get you started serving an ASGI application as quick as possible.
```

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,182 @@
...
+The simplest way to run `uvicorn` is from the command line.
+
+```bash
+$ uvicorn example:app --reload --port 5000
+```
+
+The ASGI application should be specified in the form `path.to.module:instance.path`
+
+Use `--reload` to turn on auto-reloading. 
+???+ note
```
florimondmanca: It seems `???` makes the admonition collapsible, and `+` makes it un-collapsed by default. Not sure we need it as the admonition is very small. Thoughts on sticking to plain `!!!`, i.e. show this small admonition always?
```suggestion
!!! note
```

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,182 @@
...
+    `--reload` is recommended only for local use, during the development phase.
+    
+    The `--reload` and `--workers` arguments are **mutually exclusive**.
+
+
+
+
```
florimondmanca:
```suggestion
```

docs/deployment/getting-started.md
florimondmanca: We might want to deal with this separately as it could be added to the existing docs -- not specifically related to this "enhanced" guide. Also, I believe we already mentioned this somewhere? There might be an old PR about it.

docs/deployment/run-in-prod.md
florimondmanca: Agreed, I was about to post a comment asking if we could shrink the general Docker verbiage waaaay down.  Overall, I believe we can assume people who will be read this section are familiar with Docker as a technology.

So, an introduction phrase with a link to docker.com could be enough, then enter in "How-to guide" mode: "If you want to run Uvicorn within a Docker container, here's one way to do it: ..."

Actually, after reading this section, it seems it mixes the different documentation styles (see [Daniele Procida's talk](https://www.youtube.com/watch?v=t4vKPhjcMZg)). The general style of this page is "How-to", yet this section goes in a "tutorial" mode, with lots of beginner explanations and step-by-step instructions to setup an entire environment. It ends up being a bit confusing I think.

We _could_ have a more detailed "Uvicorn + Docker" tutorial, but it would probably have to live in a different place. It's also possible there are excellent blog posts out there that already explain all this better than we ever could in a concise documentation site.

Also, one frustrating bit with Docker is that it's a rather complex technology when used in production. Here we explain how to "run the container", but what if I have Docker Compose? How would I do this from some kind of CI? Etc. If we try to "explain Docker", the potential documentation scope creep is very far reaching -- that's why it should not be our role.

florimondmanca (commented):
Thanks for this! Here's my review.

Overall, there might be another way to go about this. Perhaps we could keep the existing `deployment.md` page, but restructure it a bit.

The main questions people might ask are:

1. What are the possible was to start Uvicorn? (so I can use one of these ways with a production technology)
2. How do I use Uvicorn with production technology XYZ?

Based on the list in https://github.com/encode/uvicorn/issues/1549#issuecomment-1200362537, we could (eventually) end up with a structure like this:

* Running Uvicorn
  * Running from the command line
  * Running programmatically
* How-to guides
  * Configuring HTTPS on Uvicorn
  * Using a process manager
    * Gunicorn
    * Supervisor
    * Circus
  * Running as a systemd service
  * Running behind a reverse proxy
    * NGINX
    * Traefik
  * Running behind a CDN
  * Deploying with Docker and Docker Compose
  * Deploying with Kubernetes
    * General guidelines
    * Recommended auto-scaling options
  * Deploying on serverless
  * Distributing as an executable

In each how-to guide, we'd probably want a short introduction sentence about _why_ somebody would want to use the technology at hand. Readers with a simple blog may see the list and be overwhelmed. On the other hand, advanced users in a multi cloud environment would need guidelines for their setup. So readers should be able to quickly answer "Do I need to read this for my use case?". A consequence of this would be to order the "How-to guides" by increasing "complexity" or "advanced-ness".

Anyways, if this sounds sensible, one very first step would be to restructure the _existing_ documentation in this style. Then we can add sections for the various technologies we think Uvicorn users might need to integrate against.

docs/deployment/run-in-prod.md
iudeen: Agree with you, I think we can trim all the docker explanation and stick to "how to" alone. The is just the sample `Dockerfile`.

mkdocs.yml
iudeen: The idea was to drop `deployment.md` and have all deployment related doc files under a directory and move the introduction part (or quick guide) into `deployment/getting-started.md`. I did not remove any part of existing documentation, as I just started writing this.

docs/deployment/run-in-prod.md
iudeen: I will make the changes and push a version today  :)

iudeen:
@florimondmanca thanks for your amazing review! If we feel like having them separated into different files is not necessary, we can modify the existing `deployment.md` and make the necessary changes according to the structure you mentioned. Can we do that?

docs/deployment/getting-started.md
adriangb: Ah yes I missed it! https://github.com/encode/uvicorn/blob/master/docs/index.md#running-programmatically

mkdocs.yml
iudeen: I need help from you all here, what should we do? 

Option 1:
Keep a section called `Deployments` and have multiple files under it.

Option 2:
Keep one single file i.e., `deployments.md` and add all our enhancement into that.

Which option do we go with?

iudeen:
I got very busy lately, and will be for next few weeks 😢. Any idea to attract more collaborators?

docs/deployment/deploy-in-prod.md
euri10: I'd remove this page entirely, this is out of scope to me and kind of a rabbit hole, I dont think we should recommand or even teach people on how to use those docker-compose or k8s systemd or whatever,
the only thing I feel is ok is that we'd recommand running it behing a proxy server but that's pretty much it,

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,76 @@
+# Deployment
+
+Server deployment is a complex area, that will depend on what kind of service you're deploying Uvicorn onto.
+
+### TL;DR
+As a general rule, you probably want to:
+
+* Run `uvicorn --reload` from the command line for local development.
```
euri10, line 8:
```suggestion
```

as said, reload is a dev feature so I'd remove that if we talk about deployment, on top of that it's already explained in the docs

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,76 @@
...
+* Run `gunicorn -k uvicorn.workers.UvicornWorker` for production (non containerized).
+* Run behind Nginx for self-hosted deployments.
+* Finally, run everything behind a CDN for caching support, and serious DDOS protection.
```
euri10, line 11:
```suggestion
```
what does everything mean ? I'd remove that, it's too generic, not actionnable

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,76 @@
+# Deployment
+
+Server deployment is a complex area, that will depend on what kind of service you're deploying Uvicorn onto.
+
+### TL;DR
+As a general rule, you probably want to:
+
+* Run `uvicorn --reload` from the command line for local development.
+* Run `gunicorn -k uvicorn.workers.UvicornWorker` for production (non containerized).
```
euri10, line 9: why non-containerized ? some people may like to have it thar way and it's a viable option,

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,76 @@
...
+* Run behind Nginx for self-hosted deployments.
```
euri10, line 10: why for self-hosted deplyments, the use of nginx could be totally justified in a good number of other cases, this doesnt seem right to me

docs/deployment/getting-started.md
```diff
@@ -0,0 +1,76 @@
...
+    uvicorn.run(app, host="127.0.0.1", port=5000, log_level="info")
+    ```
+
+    However, this style only works if you are not using multiprocessing (`workers=NUM`)
+    or reloading (`reload=True`), so we recommend using the import string style.
+
+    Also note that in this case, you should put `uvicorn.run` into `if __name__ == '__main__'` clause in the main module.
+
+    Remember, the `reload` and `workers` parameters are **mutually exclusive**.
+
```
euri10, line 76: all this is toatlly redundant with all that part is redundant with https://github.com/encode/uvicorn/blob/master/docs/deployment.md I dont see the interest of having twice the same info disseminated in the docs

docs/deployment/run-in-prod.md
```diff
@@ -0,0 +1,155 @@
...
+    Uvicorn provides a lightweight way to run multiple worker processes, for example `--workers 4`, but does not provide any process monitoring.
+
+There are several tools you can use as process manager. In this section we will see how to use the following tools:
+
+- Gunicorn
+- Supervisor
+- Circus
+
+!!! tldr
+    The recommended process manager is Gunicorn.
```
euri10, line 23: why ?

docs/deployment/run-in-prod.md
```diff
@@ -0,0 +1,155 @@
...
+
+Dockerfile:
+```commandline
+FROM python:3.9
+
+WORKDIR /code
+
+COPY ./requirements.txt /code/requirements.txt
+
+RUN pip install --no-cache-dir --upgrade -r /code/requirements.txt
```
euri10, line 139: well, this is the kind of rabbit hole I was talking about when I begun reviewing this, let's teach poeple how to use docker...

euri10 (commented):
well I begun reviewing this, then stopped, I'm leaving the few comments I made nonetheless.
gloabally speaking I think this is a pretty bad idea, and kind of a rabbit hole where this ends up being a collection of do this do that without real explanations as to why.
My take on it, we should totally drop this PR or at least discuss it throughly in advance, in the current state it doesn't bring anything interesting and will lead to a sheer amout of totally useless questions for us that should be asked in stackoverflow over how to deploy things,

docs/deployment/getting-started.md
iudeen: Most of these lines were from current documentation. The deployments.md file.

My initial thought was to replace that with getting-started.md and add section wise files.

iudeen:
Closing this for now. I think we can discuss and plan this properly and then start! 

I will archive all the suggestions given here for future reference :D and of course, thanks everyone for the help!

PR #1628, opened 2022-09-01
Remove HTTPX from setup.py

aminalaee:
Contents were copied from HTTPX MR. Changing httpx to uvicorn.


Kludex:
Why not just replace `httpx` by `uvicorn`?

Kludex (approved):

PR #1632, opened 2022-09-02
Fix lifespan=on and interface=wsgi cannot be used at the same time

ryomahan:
to resolve #1631


Kludex:
As mentioned on the #1631, this is only an issue when the user sets `--lifespan=on`, and `--interface=wsgi`.

This behavior is acceptable to me.

We may need to validate the input to let the user be aware of issues beforehand. For example, when `reload` and `workers` are set, only one of those works... :eyes:

Issue #111, opened 2018-07-04
RuntimeError in uvicorn for some requests


ChillarAnand:
`uvicorn==0.2.5` is throwing errors for some requests.

```
ERROR: Exception in ASGI application
Traceback (most recent call last):
  File "/home/chillar/.virtualenvs/library/lib/python3.6/site-packages/uvicorn/protocols/http/httptools.py", line 196, in run_asgi
    result = await asgi(self.receive, self.send)
  File "/home/chillar/.virtualenvs/library/lib/python3.6/site-packages/channels/http.py", line 190, in __call__
    await self.handle(body)
  File "/home/chillar/.virtualenvs/library/lib/python3.6/site-packages/asgiref/sync.py", line 110, in __call__
    return await asyncio.wait_for(future, timeout=None)
  File "/usr/lib/python3.6/asyncio/tasks.py", line 339, in wait_for
    return (yield from fut)
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/chillar/.virtualenvs/library/lib/python3.6/site-packages/asgiref/sync.py", line 125, in thread_handler
    return self.func(*args, **kwargs)
  File "/home/chillar/.virtualenvs/library/lib/python3.6/site-packages/channels/http.py", line 229, in handle
    self.send(response_message)
  File "/home/chillar/.virtualenvs/library/lib/python3.6/site-packages/asgiref/sync.py", line 64, in __call__
    return call_result.result()
  File "/usr/lib/python3.6/concurrent/futures/_base.py", line 432, in result
    return self.__get_result()
  File "/usr/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/home/chillar/.virtualenvs/library/lib/python3.6/site-packages/asgiref/sync.py", line 78, in main_wrap
    result = await self.awaitable(*args, **kwargs)
  File "/home/chillar/.virtualenvs/library/lib/python3.6/site-packages/uvicorn/protocols/http/httptools.py", line 308, in send
    protocol.transport.write(body)
  File "uvloop/handles/stream.pyx", line 636, in uvloop.loop.UVStream.write
  File "uvloop/handles/handle.pyx", line 165, in uvloop.loop.UVHandle._ensure_alive
RuntimeError: unable to perform operation on <TCPTransport closed=True reading=False 0x1a48ad8>; the handler is closed
```

[...]

PR #1636, opened 2022-09-07
fix: keep waiting if connection is not ready to close

Yangruipis:
related to https://github.com/encode/uvicorn/issues/111

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -306,8 +306,8 @@ async def test_keepalive_timeout_concurrency(protocol_cls):
 
     # Assume we have two request(A, B), receiving data at the same time. A completes quickly,
     # B takes more than keepalive_timeout(=5s) to await
-    protocol.data_received(SIMPLE_GET_REQUEST) # request A
-    protocol.data_received(SIMPLE_GET_REQUEST) # request B
+    protocol.data_received(SIMPLE_GET_REQUEST)  # request A
+    protocol.data_received(SIMPLE_GET_REQUEST)  # request B
 
     await protocol.loop.run_one()
     assert not protocol.transport.is_closing()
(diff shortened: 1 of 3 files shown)


Kludex:
I need more information here...

Yangruipis:
[Here](https://github.com/encode/uvicorn/issues/1637#issuecomment-1242987329)  is the way to reproduce this problem.

Kludex:
If you can explain the problem, and the solution, it would save me time, and I'd really appreciate it.

Yangruipis:
> If you can explain the problem, and the solution, it would save me time, and I'd really appreciate it.

Sorry for my late response.

## Problem 

The close of the connection after `timeout_keep_alive` may raise error like that:

```
h11._util.LocalProtocolError: can't handle event type ConnectionClosed when role=SERVER and state=SEND_RESPONSE
```

Because the function `timeout_keep_alive_handler` is called when the connection is still sending data. This may happen with a very little probability under high concurrency, like a locust test.

https://github.com/encode/uvicorn/blob/c3c2db3f9ed8e331c6757c9c302ef21cf7338848/uvicorn/protocols/http/h11_impl.py#L352-L360

## Solution

1. if the connection is still sending data, do not close it
2. (optional) put it back to the loop, which will be called after another timeout. Or we do nothing until another timeout_keep_alive_task is launched on response complete.

https://github.com/encode/uvicorn/blob/47475dd30e3801c5ebbdf00087123a54ab46e8e0/uvicorn/protocols/http/h11_impl.py#L352-L366

Kludex:
I cannot reproduce your example. Are you able to create a test case on this PR?

Yangruipis:
> I cannot reproduce your example. Are you able to create a test case on this PR?

I have added a new case and some comments. But it's interesting that the new case passed locally, but failed in CI. I'll check my local package version firstly, and reopen this PR if I can make sure about it. 


However, if you want to reproduce this issue, just run the new test case with the master branch code. It will raise error like that:

image

PR #1639, opened 2022-09-11

### LINKED ISSUE #1618: "Date" header not changing between requests
### Discussed in https://github.com/encode/uvicorn/discussions/1610

<sup>Originally posted by **arjwilliams** August 22, 2022</sup>
I am testing out some cache-control headers in my web app, but the auto-generated "Date" header seems to only be changing when a new connection is established, rather than with each new request, which is preventing the browser from caching new responses, as it sees the response as being older than it actually is.

I would expect that the "Date" header to change between requests, even when using an existing connection.

I am testing on Python 3.10.5, with Uvicorn 0.18.2.

Possible test case, which is currently failing for me:
```python
import asyncio

from email.utils import parsedate_to_datetime

import httpx

async def test_default_date_header_different():
    config = Config(app=app, loop="asyncio", limit_max_requests=2)
            response1 = await client.get("http://127.0.0.1:8000")
            await asyncio.sleep(2)
            response2 = await client.get("http://127.0.0.1:8000")

            response1_date = parsedate_to_datetime(response1.headers["date"])
            response2_date = parsedate_to_datetime(response2.headers["date"])
            assert(response2_date > response1_date)

    await send({"type": "http.response.start", "status": 200, "headers": []})
    await send({"type": "http.response.body", "body": b"", "more_body": False})

@asynccontextmanager
async def run_server(config: Config):
    cancel_handle = asyncio.ensure_future(server.serve())
    await asyncio.sleep(0.1)
    try:
        yield server
    finally:
        await server.shutdown()
        cancel_handle.cancel()

asyncio.run(test_date_header_different())
```</div>

### PR DESCRIPTION
**Pull request by Kludex**
- Closes #1618

The idea here is to use the `ServerState` object instead of passing the `server_state.default_headers` to the `<protocol>.default_headers` attribute.

## How to test it?

- Use the same snippet provided on #1618.

### DIFF
--- a/tests/test_default_headers.py
+++ b/tests/test_default_headers.py
@@ -1,3 +1,6 @@
+import asyncio
+import datetime as dt
+
 import httpx
 import pytest
 
@@ -18,6 +21,24 @@ async def test_default_default_headers():
         async with httpx.AsyncClient() as client:
             response = await client.get("http://127.0.0.1:8000")
             assert response.headers["server"] == "uvicorn" and response.headers["date"]
+
+
+@pytest.mark.anyio
+async def test_date_headers_update():
+    config = Config(app=app, loop="asyncio")
+    async with run_server(config):
+        async with httpx.AsyncClient() as client:
+            response = await client.get("http://127.0.0.1:8000")
+            date = response.headers["date"]
+            first_date = dt.datetime.strptime(date, "%a, %d %b %Y %H:%M:%S %Z")
+
+            await asyncio.sleep(1)
+
+            response = await client.get("http://127.0.0.1:8000")
+            date = response.headers["date"]
+            second_date = dt.datetime.strptime(date, "%a, %d %b %Y %H:%M:%S %Z")
+
+            assert second_date - first_date == dt.timedelta(seconds=1)
 
 
 @pytest.mark.anyio
--- a/tests/test_default_headers.py
+++ b/tests/test_default_headers.py
@@ -42,8 +42,8 @@ async def test_date_headers_update():
             async def sleep_and_cancel(tg: anyio.abc.TaskGroup):
                 nonlocal cancelled
                 await asyncio.sleep(2)
-                cancelled = True
-                await tg.cancel_scope.cancel()
+                cancelled = True  # pragma: no cover
+                await tg.cancel_scope.cancel()  # pragma: no cover
 
             async def ensure_different_date(tg: anyio.abc.TaskGroup):
                 nonlocal second_date
--- a/tests/test_default_headers.py
+++ b/tests/test_default_headers.py
@@ -1,6 +1,3 @@
-import asyncio
-import datetime as dt
-
 import httpx
 import pytest
 
@@ -21,24 +18,6 @@ async def test_default_default_headers():
         async with httpx.AsyncClient() as client:
             response = await client.get("http://127.0.0.1:8000")
             assert response.headers["server"] == "uvicorn" and response.headers["date"]
-
-
-@pytest.mark.anyio
-async def test_date_headers_update():
-    config = Config(app=app, loop="asyncio")
-    async with run_server(config):
-        async with httpx.AsyncClient() as client:
-            response = await client.get("http://127.0.0.1:8000")
-            date = response.headers["date"]
-            first_date = dt.datetime.strptime(date, "%a, %d %b %Y %H:%M:%S %Z")
-
-            await asyncio.sleep(1)
-
-            response = await client.get("http://127.0.0.1:8000")
-            date = response.headers["date"]
-            second_date = dt.datetime.strptime(date, "%a, %d %b %Y %H:%M:%S %Z")
-
-            assert second_date - first_date == dt.timedelta(seconds=1)
 
 
 @pytest.mark.anyio
--- a/tests/test_default_headers.py
+++ b/tests/test_default_headers.py
@@ -1,3 +1,6 @@
+import asyncio
+import datetime as dt
+
 import httpx
 import pytest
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by euri10 — CHANGES_REQUESTED** [review]
You should probably add the test from the discussion so that we're sure it is ok in the future.

This said, I played a little with the test, the PR fixes the failed test, BUT, I find that as soon as the sleep is below 1s, the test fails again, any idea ?

**Comment by Kludex** [comment]
> You should probably add the test from the discussion so that we're sure it is ok in the future.
> 
> This said, I played a little with the test, the PR fixes the failed test, BUT, I find that as soon as the sleep is below 1s, the test fails again, any idea ?

Yeah, the `on_tick` function runs every second.

Ideas for testing this?

**Comment by euri10** [comment]
ok so just the test added will be fine I suppose:  with a 1.1s sleep it should do the job, 
maybea add a note refering to the on_tick so we know that's a limitation and headers cannot be refreshed more that once per second because of this.
Ideally on_tick should be configurable, but not too sure

**Comment by Kludex** [comment]
This doesn't work. I've set 1-second sleep, and I got 2 seconds difference.

I tried to play with mocking the `asyncio.sleep(0.1)`, and using `time-machine` package, but none work properly.

**Comment by euri10** [comment]
works fine on linux, I just added prints, what mac is doing ???
```
============================= test session starts ==============================
collecting ... collected 1 item

test_default_headers.py::test_date_headers_update 

============================== 1 passed in 1.48s ===============================

PASSED                 [100%]INFO:     127.0.0.1:42396 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:42396 - "GET / HTTP/1.1" 200 OK
2022-09-15 03:08:36
2022-09-15 03:08:37
INFO:     Started server process [1871053]
INFO:     Shutting down
```

**Comment by Kludex** [comment]
By "I got" I mean the pipeline got. 😟

The job failing is with 2 seconds diff.

**Comment by euri10** [comment]
can we add microseconds on the CI just to check
is it possible to reproduce locally, I have no mac >_)

**Comment by Kludex** [comment]
Why would the microseconds matter here? Also, the pipeline on mac also fails with `timedelta < 1` :sweat_smile:

**Comment by Kludex** [comment]
> Why would the microseconds matter here?

It's a real question, seeing the `on_tick` and `main_loop` I don't know how that would help (nor where I should be printing it).

**Comment by euri10** [comment]
I was thinking maybe there could have been a rounding effect in the difference leading to having a difference of 2 instead of one,
But apparently there is no microseconds set in the headers so that wont help.
However removing the test is probably not a solution, would be interesting to dig further to see why macos is doing that, it's waiting for one more seconds so there's likely somethig fishy going on

**Comment by Kludex** [comment]
> However removing the test is probably not a solution, would be interesting to dig further to see why macos is doing that, it's waiting for one more seconds so there's likely somethig fishy going on

Not necessarily... As I said:

> Also, the pipeline on mac also fails with timedelta < 1 :sweat_smile:

The pipeline is failing for >1 and <1 for Mac.

**Comment by euri10** [comment]
> The pipeline is failing for >1 and <1 for Mac.

but locally it's ok ?

**Comment by Kludex** [comment]
> > The pipeline is failing for >1 and <1 for Mac.
> 
> but locally it's ok ?

I don't know. I've been using Ubuntu. I'll try on the Mac later this week.

**Comment by Kludex** [comment]
@euri10 Can you check it? It should be less flaky now.

File: tests/test_default_headers.py [hunk 1]
```diff
@@ -20,6 +27,41 @@ async def test_default_default_headers():
...
+            first_date = dt.datetime.strptime(date, DATE_FORMAT)
+            second_date = first_date
+
+            cancelled = False
+
+            async def sleep_and_cancel(tg: anyio.abc.TaskGroup):
+                nonlocal cancelled
+                await asyncio.sleep(2)
+                cancelled = True
+                await tg.cancel_scope.cancel()
```
Kludex:
```suggestion
                cancelled = True  # pragma: no cover
                await tg.cancel_scope.cancel()  # pragma: no cover
```

tests/test_default_headers.py
```diff
@@ -1,9 +1,16 @@
...
+
+import anyio
+import anyio.abc
 import httpx
 import pytest
 
 from tests.utils import run_server
 from uvicorn import Config
 
+DATE_FORMAT = "%a, %d %b %Y %H:%M:%S GMT"
```
euri10, line 12: what about we use the same format mechanism as in the codebase, iirc email.formatdate ?

tests/test_default_headers.py
```diff
@@ -20,6 +27,41 @@ async def test_default_default_headers():
...
+                        second_date = dt.datetime.strptime(date, DATE_FORMAT)
+                tg.cancel_scope.cancel()
+
+            async with anyio.create_task_group() as tg:
+                tg.start_soon(sleep_and_cancel, tg)
+                tg.start_soon(ensure_different_date, tg)
+
+            assert not cancelled
+            assert second_date - first_date == dt.timedelta(seconds=1)
+
```
euri10, line 63: seems convoluted to me but if this works. 
do we know why previous test was failing on mac os ?

euri10 (approved):

tests/test_default_headers.py
Kludex: Going to change it

tests/test_default_headers.py
Kludex: I agree it's convoluted. I can explain the code on the docstring... 

I don't know why MacOS fails. I can try to investigate that before merging it. But the thing is that it failed for timestamp <1 and >1.

PR #1640, opened 2022-09-11
Remove `--debug` flag

### LINKED ISSUE #480: Phase out `debug` in favour of `reload`

Kludex:
- Closes #480

I've created https://github.com/encode/uvicorn/pull/1190 some time ago, but at the time I didn't remember why I was creating that PR.

The goal of this PR is to remove the `--debug` flag. This flag is actually hidden, so to be aware of its existence, you need to check `uvicorn`s source code.

There's also a conditional on the `DebugMiddleware`, that makes the flag only useful if the exception happens before the `http.response.start` is sent from the application.

Consider the application:

```python
async def app(scope, send, receive):
    raise Exception("HAHAH")
```

What the `--debug` flag returns for the client is:
```bash
❯ http GET :8000/         
HTTP/1.1 500 Internal Server Error
Transfer-Encoding: chunked
content-type: text/plain; charset=utf-8
date: Sun, 11 Sep 2022 12:15:46 GMT
server: uvicorn

  File "/home/marcelo/Development/encode/uvicorn/uvicorn/middleware/debug.py", line 104, in __call__
    await self.app(scope, receive, inner_send)
  File "/home/marcelo/Development/test/./main.py", line 2, in app
    raise Exception("HAHAH")
Exception: HAHAH
```

Without the flag, we have the same log on the server side, but the client receives the normal:

```bash
❯ http GET :8000/
HTTP/1.1 500 Internal Server Error
Transfer-Encoding: chunked
connection: close
content-type: text/plain; charset=utf-8
date: Sun, 11 Sep 2022 12:26:05 GMT
server: uvicorn

Internal Server Error
```

If we are using a web framework that handles exceptions, like Starlette, the `--debug` flag is completely useless, as the application already sets the `Internal Server Error` by default, see:

```python


async def endpoint(request):
    raise Exception("haha")


app = Starlette(routes=[Route("/", endpoint)])
```

With the above application, and the `--debug` flag set, we have the following on the client side:

```bash
❯ http GET :8000/
HTTP/1.1 500 Internal Server Error
content-length: 21
content-type: text/plain; charset=utf-8
date: Sun, 11 Sep 2022 12:28:52 GMT
server: uvicorn

Internal Server Error
```

To summarize: this flag is not used, it doesn't bring much benefit, and we have a 3 years old issue to remove it.


euri10 (approved):
agreed, just need to fix the coverage, not sure why removing all that code makes it drop..

Kludex:
This is quite interesting actually... I have a different behavior on line 193 in the `config.py` file on two different python versions... I didn't expect that. There must be something there that I'm not aware... Which is not related to this PR.

I'll check later on.

Kludex:
I've rerun the pipeline and the line that was marked as not covered on Python 3.7 is now covered, and another line became uncovered... What's happening here... :sweat_smile:

Kludex:
There's some random behavior on the `reload_directory_structure` for Python 3.7.

I'm dropping 0.04 coverage for this. It would be cool to increase this number to 100%. :disappointed:

Kludex:
> agreed, just need to fix the coverage, not sure why removing all that code makes it drop..

Answering this... If you have 100 potatoes, and 10 are bad potatoes, 90% are good. If you eat 10 potatoes, you have 80 potatoes, 80/90 = 88.8% potatoes are good. Same idea when removing lines of code that are covered.

edmorley:
@Kludex Hi! Should the middleware file itself be deleted too? It's currently both unused and untested:
https://github.com/encode/uvicorn/blob/master/uvicorn/middleware/debug.py

Kludex:
That's why the coverage dropped...

Yes. I missed that. Would you like to create a PR @edmorley ? 👀

Issue #1501, opened 2022-05-25
Gracefully handle HTTP/2 upgrade request


ChristianCiach:
I know that HTTP/2 is out of scope of this project. But I am not asking to support HTTP/2. 

More and more http clients try to upgrade the connection to HTTP/2. Uvicorn is free to not honor this request. Unfortunately, instead of ignoring the upgrade request, Uvicorn responds with status `400 Bad Request` and the message "Unsupported upgrade request".

According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism the server should just ignore the upgrade request:

> If the server decides to upgrade the connection, it sends back a [101 Switching Protocols](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/101) response status with an Upgrade header that specifies the protocol(s) being switched to. If it does not (or cannot) upgrade the connection, it ignores the Upgrade header and sends back a regular response (for example, a [200 OK](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200)).

We continue to encounter this issue because the `HttpClient` class of modern OpenJDK versions tries to upgrade the connection to HTTP/2 by default. Uvicorn should just ignore these headers and process the requests as if these headers were not present.

euri10:
interesting, low priority but something to consider

ChristianCiach:
I don't agree with the "low priority" assessment. This is a blatant violation of the http specification and more and more (standards compliant) http clients break just by using them with their default settings.

Kludex:
PR welcome.

ChristianCiach:
The existence of a PR has nothing to do with the priority assessment. In fact, giving this issue a higher priority may incentivize people to actually fix this issue. Saying that this violation of the http spec is "low prio" is more a statement about the level of professionalism of this software, or the lack thereof.

For the record: Yes, I would try to fix this myself, but you have to realize that not every user of your product is a software developer that uses the language that your software is written in. I am fluent in Java and Go, but I cannot fix python code. Or at least not at a level that you would ever consider to accept contributions.

Kludex:
Locking the issue as I'm not liking the tone of the reporter.

PR is still welcome. 🙏 

If no one implements it, I will, at some point, on my free time (unpaid free time).

euri10:
omg, happy to come back to see your fluency in being a total asshat @ChristianCiach 
raising the issue as high priority :+1:  !

Kludex:
- Closed by #1661

[...]

PR #1642, opened 2022-09-11
Correctly handle Upgrade-requests for HTTP/2 by ignoring them

hajs:
(Fix for https://github.com/encode/uvicorn/issues/1501)

MDN (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Upgrade) says: 

_The server can choose to ignore the request, for any reason, in which case it should just respond as though the Upgrade header had not been sent (for example, with a 200 OK)._

The previous implementation could only upgrade requests for WebSockets and generated errors for "Upgrade: h2c".

Example for `uvicorn example:app`:

```python

    await send({
        'status': 200,
        'headers': [
        ],
    })
    await send({
        'type': 'http.response.body',

```

With this patch the server will send the correct result:

```bash
$ curl -v --http2 localhost:8000
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET / HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.81.0
> Accept: */*
> Connection: Upgrade, HTTP2-Settings
> Upgrade: h2c
> HTTP2-Settings: AAMAAABkAAQCAAAAAAIAAAAA
>
< HTTP/1.1 200 OK
< date: Sun, 11 Sep 2022 21:14:15 GMT
< server: uvicorn
< content-type: text/plain
< transfer-encoding: chunked
<
* Connection #0 to host localhost left intact
```

Previous behaviour:

```bash
$ curl -v --http2 localhost:8000
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET / HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.81.0
> Accept: */*
> Connection: Upgrade, HTTP2-Settings
> Upgrade: h2c
> HTTP2-Settings: AAMAAABkAAQCAAAAAAIAAAAA
>
< HTTP/1.1 400 Bad Request
< content-type: text/plain; charset=utf-8
< connection: close
< Transfer-Encoding: chunked
<
* Closing connection 0
Unsupported upgrade request.
```

### DIFF
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -168,6 +168,7 @@ class HttpToolsProtocol(asyncio.Protocol):
         for name, value in self.headers:
             if name == b"upgrade":
                 upgrade_value = value.lower()
+
         if upgrade_value == b"h2c":
             return
         elif upgrade_value != b"websocket" or self.ws_protocol_class is None:
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -245,7 +245,16 @@ class HttpToolsProtocol(asyncio.Protocol):
         if http_version != "1.1":
             self.scope["http_version"] = http_version
         if self.parser.should_upgrade():
+<<<<<<< HEAD
             return
+=======
+            http2 = any(
+                name == b"upgrade" and value.lower() == b"h2c"
+                for name, value in self.headers
+            )
+            if not http2:
+                return
+>>>>>>> 7a22ba8 (Alternative work)
         parsed_url = httptools.parse_url(self.url)
         raw_path = parsed_url.path
         path = raw_path.decode("ascii")
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -104,6 +104,7 @@ class HttpToolsProtocol(asyncio.Protocol):
         self.scope: HTTPScope = None  # type: ignore[assignment]
         self.headers: List[Tuple[bytes, bytes]] = None  # type: ignore[assignment]
         self.expect_100_continue = False
+        self.is_http2 = False
         self.cycle: RequestResponseCycle = None  # type: ignore[assignment]
 
     # Protocol interface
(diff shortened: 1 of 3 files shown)


uvicorn/protocols/http/httptools_impl.py
```diff
@@ -245,7 +246,15 @@ def on_headers_complete(self) -> None:
...
-            return
+            upgrade_http2 = False
+            for name, value in self.headers:
+                if name == b"upgrade":
+                    upgrade_value = value.lower()
+                    if upgrade_value == b"h2c":
+                        upgrade_http2 = True
+                        break
+            if not upgrade_http2:
+                return
```
Kludex: We can add a `self.is_http2`, and set as `True` on the `on_header` method.

Kludex (commented):
If we look for `should_upgrade` on the `httptools` implementation, why is the same logic not needed on `on_body` and `on_message_complete`?

uvicorn/protocols/http/httptools_impl.py
hajs: Oh indeed, the code looks like, we should also replace the `self.parser.should_upgrade()` in `on_body` and `on_message_complete` with `(self.parser.should_upgrade() and not self.is_http2)`. 
Sorry, I had overlooked that because no unit test failed for this edge case.

Kludex: We can probably benefit from more tests that consider those conditionals 🙌

Kludex (commented):
The drop on coverage threshold shouldn't be needed if we add the necessary tests here. Would you like to work on those, @hajs ?

hajs:
Yeah this is normally not a good idea. The coverage declined only after re-formatting this commit https://github.com/encode/uvicorn/pull/1642/commits/0cffedcbc1c392fc1edbc5c6e59e8298578fcfbb into multiple lines.

I am not sure how to test for different request-response-cycles. Do you have any idea?

Kludex:
Thanks for the PR @hajs! 🙏 

I'll be closing this, and continue on https://github.com/encode/uvicorn/pull/1661, as it's the changes are a superset of these here.

PR #1650, opened 2022-09-16
Change Server.started to asyncio.Event

davidbrochart:
It doesn't seem to possible to get notified when the server has started, because `Server.started` is currently a `bool`.
Changing it to an `asyncio.Event` would make it possible, by awaiting `Server.started.wait()`, but this is a breaking change. Other options would be to add a new `Server.started_event: asyncio.Event`, or to be able to register a callback.

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -50,7 +50,7 @@ class Server:
         self.config = config
         self.server_state = ServerState()
 
-        self.started = False
+        self.started = asyncio.Event()
         self.should_exit = False
         self.force_exit = False
         self.last_notified = 0.0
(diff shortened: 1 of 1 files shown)


Kludex:
Why do you want to know that?

davidbrochart:
For instance, by default JupyterLab opens a browser window when it is launched, but it should not happen before the server has started.

Kludex:
It's already possible.

```python
import asyncio

import uvicorn
from fastapi import FastAPI

app = FastAPI()

@app.get("/")
def home():
    ...

@app.on_event("startup")
async def startup():
    await asyncio.sleep(1)

async def main():
    config = uvicorn.Config("main:app", port=5000, log_level="info")
    server_task = asyncio.create_task(server.serve())
    while not server.started:
        print("not started")
        await asyncio.sleep(0.1)
    print("started")
    await server_task


if __name__ == "__main__":
    asyncio.run(main())
```

Kludex:
Please create a discussion if the above solution doesn't satisfy your needs. :pray:

davidbrochart:
> It's already possible.

Sure, although polling is not great.

PR #1661, opened 2022-09-21
Upgrade request handling: ignore http/2 and optionally ignore websocket

Argannor:
This PR addresses the discussion #1659 (raised by me) and issue #1501. 

Rationale: As outlined in the discussion and issue above the HTTP/1.x spec allows servers to ignore upgrade requests by responding with a normal response.

Changes made:
- Since uvicorn currently only supports upgrades to websocket protocol, all other upgrade requests are ignored by default.
- Websocket requests can be optionally ignored by setting the new config property `ws_ignore_upgrades` to `True`
- Introduced a new method `_should_upgrade` in both http protocol implementations to be overwritten by users that need a more fine grained logic to decide on upgrades
- Test cases added for both cases
- Documentation updated

After making the change for #1659 I also stumbled across the PR #1642 (adressing #1501). Since both changes affect the same area I took the liberty to refactor my change to include graceful handling of #1501 as well.

Looking forward to your feedback!

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -71,6 +71,7 @@ Using Uvicorn with watchfiles will enable the following options (which are other
 * `--ws-max-size <int>` - Set the WebSockets max message size, in bytes. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-interval <float>` - Set the WebSockets ping interval, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-timeout <float>` - Set the WebSockets ping timeout, in seconds. Please note that this can be used only with the default `websockets` protocol.
+* `--ws-ignore-upgrade` - Will ignore all upgrade requests and instead continue replying with HTTP responses.
 * `--lifespan <str>` - Set the Lifespan protocol implementation. **Options:** *'auto', 'on', 'off'.* **Default:** *'auto'*.
 * `--h11-max-incomplete-event-size <int>` - Set the maximum number of bytes to buffer of an incomplete event. Only available for `h11` HTTP protocol implementation. **Default:** *'16384'* (16 KB).
 
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -71,7 +71,7 @@ Using Uvicorn with watchfiles will enable the following options (which are other
 * `--ws-max-size <int>` - Set the WebSockets max message size, in bytes. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-interval <float>` - Set the WebSockets ping interval, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-timeout <float>` - Set the WebSockets ping timeout, in seconds. Please note that this can be used only with the default `websockets` protocol.
-* `--ws-ignore-upgrade` - Will ignore all upgrade requests and instead continue replying with HTTP responses.
+* `--ws-ignore-upgrade` - Will ignore all upgrade requests, and HTTP/1.1 will continue to be used.
 * `--lifespan <str>` - Set the Lifespan protocol implementation. **Options:** *'auto', 'on', 'off'.* **Default:** *'auto'*.
 * `--h11-max-incomplete-event-size <int>` - Set the maximum number of bytes to buffer of an incomplete event. Only available for `h11` HTTP protocol implementation. **Default:** *'16384'* (16 KB).
 
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -67,11 +67,10 @@ Using Uvicorn with watchfiles will enable the following options (which are other
 
 * `--loop <str>` - Set the event loop implementation. The uvloop implementation provides greater performance, but is not compatible with Windows or PyPy. **Options:** *'auto', 'asyncio', 'uvloop'.* **Default:** *'auto'*.
 * `--http <str>` - Set the HTTP protocol implementation. The httptools implementation provides greater performance, but it not compatible with PyPy. **Options:** *'auto', 'h11', 'httptools'.* **Default:** *'auto'*.
-* `--ws <str>` - Set the WebSockets protocol implementation. Either of the `websockets` and `wsproto` packages are supported. Use `'none'` to deny all websocket requests. **Options:** *'auto', 'none', 'websockets', 'wsproto'.* **Default:** *'auto'*.
+* `--ws <str>` - Set the WebSockets protocol implementation. Either of the `websockets` and `wsproto` packages are supported. Use `'none'` to ignore all websocket requests. **Options:** *'auto', 'none', 'websockets', 'wsproto'.* **Default:** *'auto'*.
 * `--ws-max-size <int>` - Set the WebSockets max message size, in bytes. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-interval <float>` - Set the WebSockets ping interval, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-timeout <float>` - Set the WebSockets ping timeout, in seconds. Please note that this can be used only with the default `websockets` protocol.
-* `--ws-ignore-upgrade` - Will ignore all upgrade requests, and HTTP/1.1 will continue to be used.
 * `--lifespan <str>` - Set the Lifespan protocol implementation. **Options:** *'auto', 'on', 'off'.* **Default:** *'auto'*.
 * `--h11-max-incomplete-event-size <int>` - Set the maximum number of bytes to buffer of an incomplete event. Only available for `h11` HTTP protocol implementation. **Default:** *'16384'* (16 KB).
 
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -744,11 +744,12 @@ async def test_ignored_ws_upgrade_request(protocol_cls):
 
 @pytest.mark.anyio
 @pytest.mark.parametrize("protocol_cls", HTTP_PROTOCOLS)
-async def test_http2_upgrade_request(protocol_cls):
+@pytest.mark.parametrize("ws_ignore_upgrade", [True, False])
+async def test_http2_upgrade_request(protocol_cls, ws_ignore_upgrade):
     app = Response("Hello, world", media_type="text/plain")
 
     protocol = get_connected_protocol(
-        app, protocol_cls, ws="wsproto", ws_ignore_upgrade=True
+        app, protocol_cls, ws="wsproto", ws_ignore_upgrade=ws_ignore_upgrade
     )
     protocol.data_received(UPGRADE_HTTP2_REQUEST)
     await protocol.loop.run_one()
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -2,12 +2,13 @@ import logging
 import socket
 import threading
 import time
+from typing import List
 
 import pytest
 
 from tests.response import Response
 from uvicorn import Server
-from uvicorn.config import Config
+from uvicorn.config import WS_PROTOCOLS, Config
 from uvicorn.main import ServerState
 from uvicorn.protocols.http.h11_impl import H11Protocol
 
@@ -18,6 +19,7 @@ except ImportError:  # pragma: nocover
 
 
 HTTP_PROTOCOLS = [p for p in [H11Protocol, HttpToolsProtocol] if p is not None]
+WEBSOCKET_PROTOCOLS = WS_PROTOCOLS.keys()
 
 SIMPLE_GET_REQUEST = b"\r\n".join([b"GET / HTTP/1.1", b"Host: example.org", b"", b""])
 
@@ -172,6 +174,15 @@ class MockLoop:
             else:
                 later.append((delay, callback, args))
         self._later = later
+
+
+class MockLogger(logging.Logger):
+    def __init__(self, name: str):
+        super().__init__(name)
+        self.warnings: List[str] = []
+
+    def warning(self, msg: str):
+        self.warnings += [msg]
 
 
 class MockTask:
@@ -709,17 +720,6 @@ async def test_100_continue_not_sent_when_body_not_consumed(protocol_cls):
 
 @pytest.mark.anyio
 @pytest.mark.parametrize("protocol_cls", HTTP_PROTOCOLS)
-async def test_unsupported_upgrade_request(protocol_cls):
-    app = Response("Hello, world", media_type="text/plain")
-
-    protocol = get_connected_protocol(app, protocol_cls, ws="none")
-    protocol.data_received(UPGRADE_REQUEST)
-    assert b"HTTP/1.1 400 Bad Request" in protocol.transport.buffer
-    assert b"Unsupported upgrade request." in protocol.transport.buffer
-
-
-@pytest.mark.anyio
-@pytest.mark.parametrize("protocol_cls", HTTP_PROTOCOLS)
 async def test_supported_upgrade_request(protocol_cls):
     app = Response("Hello, world", media_type="text/plain")
 
@@ -730,12 +730,10 @@ async def test_supported_upgrade_request(protocol_cls):
 
 @pytest.mark.anyio
 @pytest.mark.parametrize("protocol_cls", HTTP_PROTOCOLS)
-async def test_ignored_ws_upgrade_request(protocol_cls):
+async def test_unsupported_ws_upgrade_request(protocol_cls):
     app = Response("Hello, world", media_type="text/plain")
 
-    protocol = get_connected_protocol(
-        app, protocol_cls, ws="wsproto", ws_ignore_upgrade=True
-    )
+    protocol = get_connected_protocol(app, protocol_cls, ws="none")
     protocol.data_received(UPGRADE_REQUEST)
     await protocol.loop.run_one()
     assert b"HTTP/1.1 200 OK" in protocol.transport.buffer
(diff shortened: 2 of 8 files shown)


uvicorn/main.py
```diff
@@ -169,6 +169,13 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     help="WebSocket per-message-deflate compression",
     show_default=True,
 )
+@click.option(
+    "--ws-ignore-upgrade",
+    type=bool,
```
Kludex: From:
```bash
uvicorn main:app --ws-ignore-upgrade true
```
To:
```bash
uvicorn main:app --ws-ignore-upgrade
```
```suggestion
    type=bool,
    is_flag=True,
```

uvicorn/protocols/http/h11_impl.py
```diff
@@ -155,6 +155,23 @@ def _unset_keepalive_if_required(self) -> None:
...
+    def _should_upgrade(
+        self, headers: List[Tuple[bytes, bytes]], event: H11Event
+    ) -> bool:
+        connection = []
+        upgrade = None
+        for name, value in headers:
+            if name == b"connection":
+                connection = [token.lower().strip() for token in value.split(b",")]
+            if name == b"upgrade":
+                upgrade = value.lower().strip()
```
Kludex: Is there a reason for stripping?

```suggestion
                upgrade = value.lower()
```

Kludex (commented):
If we reach `handle_upgrade`, we know that `upgrade_value == b"websocket"`, so we can remove the header verification over there.

uvicorn/protocols/http/h11_impl.py
Argannor: No specific reason, kept it for symmetry reasons and because I generally like to sanitize inputs. Removed it

uvicorn/main.py
Argannor: Thanks done

Argannor:
> If we reach `handle_upgrade`, we know that `upgrade_value == b"websocket"`, so we can remove the header verification over there.

You're right, I kept it in if we want to support additional protocols, but that's a bad reason

uvicorn/protocols/http/h11_impl.py
```diff
@@ -155,6 +155,23 @@ def _unset_keepalive_if_required(self) -> None:
             self.timeout_keep_alive_task.cancel()
             self.timeout_keep_alive_task = None
 
+    def _should_upgrade(
+        self, headers: List[Tuple[bytes, bytes]], event: H11Event
+    ) -> bool:
```
Kludex: I didn't notice before, but we already have the `headers` attribute at this point. 
```suggestion
    def _should_upgrade(self, event: H11Event) -> bool:
```

Argannor: Yep, decided to pass them around instead. Changed it to use the attribute.

uvicorn/protocols/http/h11_impl.py
```diff
@@ -155,6 +155,21 @@ def _unset_keepalive_if_required(self) -> None:
             self.timeout_keep_alive_task.cancel()
             self.timeout_keep_alive_task = None
 
+    def _should_upgrade(self, event: H11Event) -> bool:
```
Kludex: Neither the event is used 🤔 

```suggestion
    def _should_upgrade(self) -> bool:
```

uvicorn/main.py
```diff
@@ -169,6 +169,14 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
...
+    default=False,
+    help="Upgrade requests will be ignored and answered with normal HTTP responses",
```
```suggestion
    help="Upgrade requests will be ignored, and HTTP/1.1 will continue to be used.",
```

uvicorn/protocols/http/h11_impl.py
Argannor: Yes that is done in accordance to my proposal to enable easy overwriting for more complex use cases (s. #1659).
Would you prefer simplicity over customizability?

Kludex (commented):
I'm not sure if we should rename the `handle_upgrade` to `handle_websocket_upgrade`, but that's not a blocker.

Those two last comments were my last here. 👍

uvicorn/protocols/http/h11_impl.py
Kludex: It's a private method, I don't think we should think about customization here. 🤔 

But... Then this PR doesn't fit your needs? 🤔

uvicorn/main.py
Argannor: I think renaming would be in order, I had the same thought before. Changing the help as well.

uvicorn/main.py
```diff
@@ -173,7 +173,7 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     "--ws-ignore-upgrade",
     type=bool,
     default=False,
-    help="Upgrade requests will be ignored and answered with normal HTTP responses",
+    help="Upgrade requests will be ignored, and HTTP/1.1 will continue to be used",
```
Kludex: I don't want to be pedantic, but it was in the previous comment... 😅

```suggestion
    help="Upgrade requests will be ignored, and HTTP/1.1 will continue to be used.",
```

uvicorn/protocols/http/h11_impl.py
Argannor: The PR will still fit my needs, however when thinking about the implementation I thought maybe someone will be happy because they only want to ignore on certain paths. During implementation I realized it won't be as simple for the httptools protocol.
YAGNI? :D

We could then think of extracting the method to a common place for both protocols, since they're identical

uvicorn/main.py
Argannor: Hahaha sorry and thank you for being pedantic!

uvicorn/main.py
Argannor: Done

uvicorn/main.py
Argannor: Done

uvicorn/protocols/http/h11_impl.py
Kludex: In case you need to do something custom, uvicorn supports a way to implement your own HTTP protocol class:

```python
uvicorn.run("main:app", http=CustomHTTPProtocol)
```
jfyk

docs/settings.md
```diff
@@ -71,6 +71,7 @@ Using Uvicorn with watchfiles will enable the following options (which are other
 * `--ws-max-size <int>` - Set the WebSockets max message size, in bytes. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-interval <float>` - Set the WebSockets ping interval, in seconds. Please note that this can be used only with the default `websockets` protocol.
 * `--ws-ping-timeout <float>` - Set the WebSockets ping timeout, in seconds. Please note that this can be used only with the default `websockets` protocol.
+* `--ws-ignore-upgrade` - Will ignore all upgrade requests and instead continue replying with HTTP responses.
```
```suggestion
* `--ws-ignore-upgrade` - Will ignore all upgrade requests, and HTTP/1.1 will continue to be used.
```

Kludex (dismissed):
I'm fine with this implementation (just one last thing on the docs).

Thanks @Argannor! 🙌

@euri10 Would you like to check this as well?

uvicorn/protocols/http/h11_impl.py
Kludex: > YAGNI

Yep, please. 👀

docs/settings.md
Argannor: Done, good catch

uvicorn/protocols/http/h11_impl.py
Argannor: Want me to keep the duplication or introduce a utils file (`upgrade.py`?)

Argannor: would make it necessary to pass the headers and config around again though

Kludex: keep it for now, before merge I can rearrange if we change our minds, dw 👍

Just removing the event is fine

Kludex (dismissed):

Argannor:
Sry the previous commit only changed the signature not the call, have been doing multiple things at once..
@Kludex Thank you for your review!

tests/protocols/test_http.py
```diff
@@ -716,6 +728,34 @@ async def test_supported_upgrade_request(protocol_cls):
...
+    assert b"Hello, world" in protocol.transport.buffer
+
+
+@pytest.mark.anyio
+@pytest.mark.parametrize("protocol_cls", HTTP_PROTOCOLS)
+async def test_http2_upgrade_request(protocol_cls):
+    app = Response("Hello, world", media_type="text/plain")
+
+    protocol = get_connected_protocol(
+        app, protocol_cls, ws="wsproto", ws_ignore_upgrade=True
```
Kludex: I've been thinking about this, and maybe it will be better if we remove the ws-ignore-upgrade in this test. 🤔

Kludex: This is a note to myself.

Argannor: Hm, I think it would be best to do this test with ws-ignore-upgrade set to both True and False, as the test case should have no knowledge of the implementation and it should in both cases be ignored.
I can make that change if you agree with me

Kludex: yes

Argannor: done

Kludex: thanks

euri10:
> @euri10 Would you like to check this as well?

wont have the time sorry

Kludex:
> > @euri10 Would you like to check this as well?
> 
> wont have the time sorry

Ok. I'll deal with this. Thanks for letting me know. 🙏

Kludex:
@Argannor I actually have a question that I didn't think before... But if I have one of the `WebSocket`s package installed, why should we ignore the upgrade request to `websocket`?

Argannor:
@Kludex The application might run in a shared environment (globally installed packages) but not implement websocket endpoints.

Kludex:
> @Kludex The application might run in a shared environment (globally installed packages) but not implement websocket endpoints.

But you can set the `ws=None` even if the packages are installed. I'm trying to understand something here: isn't the real issue the fact that the current implementation returns a 400 instead of ignoring the upgrade when the `ws` parameter is `None`? 🤔

uvicorn/protocols/http/h11_impl.py
```diff
@@ -155,6 +155,21 @@ def _unset_keepalive_if_required(self) -> None:
...
+        for name, value in self.headers:
+            if name == b"connection":
+                connection = [token.lower().strip() for token in value.split(b",")]
+            if name == b"upgrade":
+                upgrade = value.lower()
+
+        return (
+            b"upgrade" in connection
+            and upgrade == b"websocket"
+            and not self.config.ws_ignore_upgrade
```
Argannor: Hm I think we could eliminate the new flag by changing it to this
```suggestion
            and self.ws_protocol_class is not None
```
and removing the corresponding block from `handle_websocket_upgrade`. I think that would achieve the same thing.

Kludex: I think the warning should stay tho 🤔

Argannor: So trying to understand what that block tries to achieve is 
- logging a warning
- logging another warning if no protocol was found
- responding with 400

But should these warnings really be logged on every upgrade request again? Especially the second one.
In my use case this will lead to a lot of unnecessary log messages.

Argannor: What do you think about moving the second log statement to the run method in main.py? I feel like this indicates a possible configuration / setup issue which would be easier to catch if printed up front on startup. We should rephrase to account for the fact that there are applications which will use the default setting `auto` without the need for an implementation to be present.

I am still undecided on what to do with the first warning. Since I already have a custom log handler, I could filter this message out to prevent the broken window for me.

Alternatively we could document the behaviour and necessary remediation inside the docs to help confused devs figure out why their request gets ignored and or turn down the log level.

Side note: I am not available today, but will come back to this tomorrow

lovelydinosaur (commented):
Hi folks.

Thanks for looping me into the review @kludex. There's two different things being addressed here, and I think we need to treat them independently...

* Ignore `Upgrade` requests that we don't handle. As raised by #1501.
* Support websockets being configured as "off" without raising warnings on ws upgrades. As discussed in #1659.

I would suggest that we start off by dealing with the first of these. That looks clearly sensible to me, and we can improve our behaviour here without adding any additional configuration.

It looks to me like the second of these cases should be handled by ensuring that `--ws none` has the desired behaviour, rather than adding an additional configuration option? (As it happens that would probably be a much better default than `auto`, but that's a different consideration.)

Argannor:
Hi @tomchristie thanks for having a look at this too!

I pushed a new iteration of this PR and integrated the feedback of both of you. The check is now based on the `ws_protocol_class` but if it isn't set to `None` warnings will be logged.

I also tried to split the concerns of #1501 and #1659, however I think that it makes the code more complicated in the case of httptools. While still trying to use the natively implemented `parser.should_upgrade` in the hope of it being more performant.

lovelydinosaur:
This seems sensible to me now.

tests/protocols/test_http.py
```diff
@@ -697,23 +720,54 @@ async def test_100_continue_not_sent_when_body_not_consumed(protocol_cls):
...
 @pytest.mark.anyio
 @pytest.mark.parametrize("protocol_cls", HTTP_PROTOCOLS)
-async def test_supported_upgrade_request(protocol_cls):
+async def test_unsupported_ws_upgrade_request_warn_on_auto(protocol_cls):
     app = Response("Hello, world", media_type="text/plain")
 
-    protocol = get_connected_protocol(app, protocol_cls, ws="wsproto")
+    protocol = get_connected_protocol(app, protocol_cls, ws="auto")
+    protocol.ws_protocol_class = None
+    protocol.logger = MockLogger(protocol.logger.name)
```
Kludex: Can't we use [`caplog`](https://docs.pytest.org/en/7.1.x/how-to/logging.html#caplog-fixture) instead?

You can see examples on our source code.

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -189,7 +201,7 @@ def handle_upgrade(self) -> None:
         for name, value in self.scope["headers"]:
             output += [name, b": ", value, b"\r\n"]
         output.append(b"\r\n")
-        protocol = self.ws_protocol_class(  # type: ignore[call-arg]
+        protocol = self.ws_protocol_class(  # type: ignore[call-arg, misc]
```
Kludex, line 204: What is this `misc` for?

uvicorn/protocols/http/h11_impl.py
```diff
@@ -155,6 +155,28 @@ def _unset_keepalive_if_required(self) -> None:
...
+        for name, value in self.headers:
+            if name == b"connection":
+                connection = [token.lower().strip() for token in value.split(b",")]
+            if name == b"upgrade":
+                upgrade = value.lower()
+        if b"upgrade" in connection:
+            return upgrade
+        return None
+
+    def _should_upgrade_to_ws(self) -> bool:
```
Kludex, line 170: Why the one in the other protocol is implemented in one way, and this one in another? 🤔

Argannor: Both protocols work a bit different, and in the case of httptools the "check and return" is repeated in several places, but the actual upgrade handling is only done once in the except block. While the h11 implementation only does it in one single place and both the returning and handling is in the same place.

tests/protocols/test_http.py
Argannor: Thank you, wasn't aware will fix

uvicorn/protocols/http/httptools_impl.py
Argannor: removing the `if self.ws_protocol_class is None` block from the method raised the lint warning that `None is not callable` because it could potentially be `None`. This is for us not the case however since we outsourced this check.

tests/protocols/test_http.py
Kludex: Thanks 🙏

Argannor: done

BilalAlpaslan (approved):

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -244,7 +256,7 @@ def on_headers_complete(self) -> None:
         self.scope["method"] = method.decode("ascii")
         if http_version != "1.1":
             self.scope["http_version"] = http_version
-        if self.parser.should_upgrade():
+        if self.parser.should_upgrade() and self._should_upgrade():
```
Kludex, line 259: Why is it an "and"? :thinking:

Kludex:
We are checking all the headers 3 times on the `httptools` implementation now. Is it something we should be concerned?

uvicorn/protocols/http/httptools_impl.py
Argannor: The C implementation of parser.should_upgrade looks like this:
```
uint8_t llhttp_get_upgrade(llhttp_t* parser) {
  return parser->upgrade;
}
```
And the flag is set during parsing anyways, so this is faster than our iterate over headers implementation. If the parser didn't detect an upgrade request we don't need to do our checks. If the parser detected an upgrade we still need to decide if we want/can upgrade.
So the first part is functionally not needed but should perform better

Argannor:
Hi @Kludex, thank you for looking into this. Are there any updates, or changes that I should make?

Kludex (approved):
Thanks @Argannor :pray:

PR #1664, opened 2022-09-25
Revert "Use updated headers from server state"

Kludex:
Reverts encode/uvicorn#1639

The test is flaky, we need to revert.

See: https://github.com/encode/uvicorn/actions/runs/3122428066/jobs/5064360211

### DIFF
--- a/tests/test_default_headers.py
+++ b/tests/test_default_headers.py
@@ -1,15 +1,8 @@
-import asyncio
-import datetime as dt
-
-import anyio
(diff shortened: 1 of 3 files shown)


Kludex:
I have no idea how this is possible.

euri10 (approved):

euri10:
will try to take a look, inho the simpler the test the better and maybe we skip CI for mac os until someone gets to the root of it

PR #1665, opened 2022-09-25

Kludex:
- Second chance to https://github.com/encode/uvicorn/pull/1639/
- Related to https://github.com/encode/uvicorn/pull/1664

### DIFF
--- a/tests/test_default_headers.py
+++ b/tests/test_default_headers.py
@@ -26,7 +26,7 @@ async def test_default_default_headers():
 
 @pytest.mark.skipif(sys.platform == "darwin", reason="Unexpected behavior on MacOS.")
 @pytest.mark.anyio
-async def test_date_headers_update():
+async def test_date_headers_update():  # pragma: py-darwin
     config = Config(app=app, loop="asyncio")
     async with run_server(config):
         async with httpx.AsyncClient() as client:
--- a/tests/test_default_headers.py
+++ b/tests/test_default_headers.py
@@ -24,9 +24,12 @@ async def test_default_default_headers():
             assert response.headers["server"] == "uvicorn" and response.headers["date"]
 
 
-@pytest.mark.skipif(sys.platform == "darwin", reason="Unexpected behavior on MacOS.")
+@pytest.mark.skipif(
+    sys.platform != "linux",
+    reason="This fails on other OSs. If you know why, please help us.",
+)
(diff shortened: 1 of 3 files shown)


tests/test_default_headers.py
```diff
@@ -20,6 +24,25 @@ async def test_default_default_headers():
...
 
+@pytest.mark.skipif(sys.platform == "darwin", reason="Unexpected behavior on MacOS.")
+@pytest.mark.anyio
+async def test_date_headers_update():
+    config = Config(app=app, loop="asyncio")
+    async with run_server(config):
+        async with httpx.AsyncClient() as client:
+            response = await client.get("http://127.0.0.1:8000")
+            date = response.headers["date"]
+            first_date = dt.datetime.strptime(date, "%a, %d %b %Y %H:%M:%S %Z")
```
Kludex, line 38: The `email.utils` function only works in one way :thinking:

euri10: mm ok  sorry

euri10:
this time test doesn't pass on windows lol
ok to move forward, let's skip this on windows too, I dont think anyways someone considering that feature important would run in prod on either macos or windows, but one can always be surprised...

this said this leaves a bad feeling having no comprehension as to why this happens

euri10 (approved):
approved once the CI pass with the skip @Kludex

Kludex:
I'm not confident in merging this test. It looks like it's just more likely for it to pass on Ubuntu. 🤔

Are you fine with this anyway? We also have the option to not add any test, even if it looks bad saying it. 😔

euri10:
well I'd prefer to the test pass on at least linux, and we skip on dev only plateforms.
but there's definitely something fishy going on

tests/test_default_headers.py
```diff
@@ -20,6 +24,28 @@ async def test_default_default_headers():
             assert response.headers["server"] == "uvicorn" and response.headers["date"]
 
 
+@pytest.mark.skipif(
+    sys.platform != "linux",
+    reason="This fails on other OSs. If you know why, please help us.",
```
euri10, line 29: :rofl:

euri10 (approved):

tests/test_default_headers.py
Kludex: 😎 👍

PR #1666, opened 2022-09-27
Revert "Use updated headers from server state"

Kludex:
Reverts encode/uvicorn#1665

It fails on Ubuntu as well. It was just luck. cc @euri10 

I'll create a PR without any test, unless you have a better idea.

See: https://github.com/encode/uvicorn/pull/1606#issuecomment-1259219877


Kludex:
Actually, let me create a PR without the test.

PR #1667, opened 2022-09-27
Remove date headers update test

Kludex:
- Replaces #1666

### DIFF
--- a/tests/test_default_headers.py
+++ b/tests/test_default_headers.py
@@ -1,6 +1,4 @@
 import asyncio
-import datetime as dt
-import sys
 
 import httpx
 import pytest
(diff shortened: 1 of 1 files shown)


euri10:
I'd prefer we revert entirely, if test fails then is it a proper fix ?
One needs to understand why it is behaving this way

iudeen:
`default_headers` test failed saying the time delta was 0 seconds. 

To investigate, I added a print statement to the `on_tick` function.

```python
 async def on_tick(self, counter: int) -> bool:
        # Update the default headers, once per second.
        if counter % 10 == 0:
            print(datetime.datetime.now())
            current_time = time.time()
            current_date = formatdate(current_time, usegmt=True).encode()
```

I noticed this..
```
test_default_headers.py::test_date_headers_update FAILED                 [100%]2022-09-27 16:17:51.230887
INFO:     127.0.0.1:53643 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:53643 - "GET / HTTP/1.1" 200 OK
2022-09-27 16:17:52.340582
```
~~I think, the `on_tick` does not run on the first second after starting the server.~~  Or the way windows schedules task is causing it go out of sequence/sync.

euri10:
interesting, I created https://github.com/encode/uvicorn/pull/1668 so that we can maybe understand where the delay might come from

Kludex:
> I'd prefer we revert entirely, if test fails then is it a proper fix ? One needs to understand why it is behaving this way

Well... I'm not sure if I agree that is the case here, the issue is on the test. But what you said makes sense. I'll create it.

Issue #157, opened 2018-07-25
Further timeouts and resource limiting.


lovelydinosaur:
- [ ] --timeout-request-start (Timeout unless request headers complete within this time. ~10s)
- [ ] --timeout-request-complete (Timeout if request body is still being read after this time. ~60s)
- [x] ~--limit-request-line-size  (4096, [See gunicorn](http://docs.gunicorn.org/en/stable/settings.html#limit-request-line))~
- [x] ~--limit-request-header-size  (8190, [See gunicorn](http://docs.gunicorn.org/en/stable/settings.html#limit-request-field-size))~
- [x] ~--limit-request-header-count  (100, [See gunicorn](http://docs.gunicorn.org/en/stable/settings.html#limit-request-fields))~
- [x] --limit-ws-message-size

## @Kludex notes

- We are not going to implement `--limit-request-header-count`. The header size limitation is enough.
- Should we consider timeout on `startup` and `shutdown` lifespans?

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/uvicorn/issues/157">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/157/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/uvicorn/issues/157/pledge.svg">
</picture>
</a>

hoopes:
Hello, 

Apologies for chiming in on this, but my use case could _really_ use this feature. I have a very large sql query that uses sqlalchemy (so it is not async, from the start). This is an internal "big data" app, so there is no pressure for super low latency per request. I'm streaming lines of json with a starlette `StreamingResponse` from a `fastapi` application.

However, the query takes about 90-120 seconds to get started actually producing results. `uvicorn` appears to time out, whether i run it directly or under gunicorn. `daphne` has a timeout arg, so i'm using that now. I'd _really_ like to use uvicorn as a work in gunicorn, to get easy multi-process service.

Anyway, I just wanted to bump this. If you pointed me to the code where the timeout actually occurs, I might be interested in trying a PR, if you'd be interested (at least for the `request start`, and the others if they fall out easily).

Thanks!

Kludex:
[...]

PR #1682, opened 2022-10-01
Implement `--limit-request-header-count` flag

Kludex:
## Description

This PR implements the flag `--limit-request-header-count` proposed on #157.

## Notes

- I thought [431](https://www.webfx.com/web-development/glossary/http-status-codes/what-is-a-431-status-code/) would make sense, but no. That's only on the size, not the number of headers. Also, [`gunicorn` also sends 400](https://github.com/benoitc/gunicorn/blob/cf55d2cec277f220ebd605989ce78ad1bb553c46/gunicorn/workers/base.py#L208-L233).

## Checklist

- [ ] Make sure the flag is properly passed from `gunicorn` to `uvicorn`.
	`limit_request_fields` :arrow_right: `limit_request_header_count`.

## Questions

Q.: Should we actually match the name to `gunicorn`?
A.: Nah, I think the name here makes more sense.

Q.: Is this implementation correct? We should not check the length of the headers when we already have the size, right? We should check it when we already know that we have more headers than what is allowed!
A.: ???

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -256,6 +256,12 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     " HTTP 503 responses.",
 )
 @click.option(
+    "--limit-request-header-count",
+    type=int,
+    default=100,
+    help="Maximum number of HTTP headers to accept.",
+)
+@click.option(
     "--backlog",
     type=int,
     default=2048,
@@ -385,6 +391,7 @@ def main(
     forwarded_allow_ips: str,
     root_path: str,
     limit_concurrency: int,
+    limit_request_header_count: int,
     backlog: int,
     limit_max_requests: int,
     timeout_keep_alive: int,
@@ -432,6 +439,7 @@ def main(
         forwarded_allow_ips=forwarded_allow_ips,
         root_path=root_path,
         limit_concurrency=limit_concurrency,
+        limit_request_header_count=limit_request_header_count,
         backlog=backlog,
         limit_max_requests=limit_max_requests,
         timeout_keep_alive=timeout_keep_alive,
@@ -484,6 +492,7 @@ def run(
     forwarded_allow_ips: typing.Optional[typing.Union[typing.List[str], str]] = None,
     root_path: str = "",
     limit_concurrency: typing.Optional[int] = None,
+    limit_request_header_count: int = 100,
     backlog: int = 2048,
     limit_max_requests: typing.Optional[int] = None,
     timeout_keep_alive: int = 5,
@@ -534,6 +543,7 @@ def run(
         forwarded_allow_ips=forwarded_allow_ips,
         root_path=root_path,
         limit_concurrency=limit_concurrency,
+        limit_request_header_count=limit_request_header_count,
         backlog=backlog,
         limit_max_requests=limit_max_requests,
         timeout_keep_alive=timeout_keep_alive,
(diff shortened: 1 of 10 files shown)


uvicorn/protocols/http/app_factory.py
```diff
@@ -0,0 +1,43 @@
+from typing import TYPE_CHECKING
```
Kludex, line 1: Reviewer, this is a new file. Are you fine with this?

uvicorn/protocols/http/app_factory.py
```diff
@@ -0,0 +1,43 @@
...
+
+    async def app(
+        scope: "Scope", receive: "ASGIReceiveCallable", send: "ASGISendCallable"
+    ) -> None:
+        response_start: "HTTPResponseStartEvent" = {
+            "type": "http.response.start",
+            "status": status_code,
+            "headers": [
+                (b"content-type", b"text/plain; charset=utf-8"),
+                (b"content-length", str(len(message)).encode("ascii")),
```
Kludex, line 30: The `content-length` header is new! I noticed that the `content-length` was missing from the previous `service_unavailable` application.

uvicorn/main.py
```diff
@@ -255,6 +255,12 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     help="Maximum number of concurrent connections or tasks to allow, before issuing"
     " HTTP 503 responses.",
 )
+@click.option(
+    "--limit-request-header-count",
+    type=int,
+    default=100,
+    help="Maximum number of HTTP headers to accept.",
```
Kludex, line 262: We should mention here that the limit is `MAX_HEADERS`.

Kludex:
Yeah, I think this implementation is actually wrong. This is quite funny, as I knew this when I started, and then I forgot. :sweat_smile:

Is there a way to send 400 to the client on the `httptools` implementation on `on_header` callback? :thinking:
If I raise an exception on `on_header`, the exception becomes `HttpParserError`:
```python
  File "httptools/parser/parser.pyx", line 284, in httptools.parser.parser.cb_on_header_field
  File "httptools/parser/parser.pyx", line 112, in httptools.parser.parser.HttpParser._on_header_field
  File "httptools/parser/parser.pyx", line 108, in httptools.parser.parser.HttpParser._maybe_call_on_header
  File "/home/marcelo/Development/encode/uvicorn/uvicorn/protocols/http/httptools_impl.py", line 242, in on_header
    raise Exception("Potato")
Exception: Potato

  File "/home/marcelo/Development/encode/uvicorn/uvicorn/protocols/http/httptools_impl.py", line 156, in data_received
    self.parser.feed_data(data)
httptools.parser.errors.HttpParserCallbackError: `on_header_field` callback error
```

Kludex:
Got it! I need to verify the length on `on_message_complete` as well!

Kludex:
But wait... I can't stop `httptools` from calling `on_header`! I can let the protocol know that we have many headers, but I can't stop the execution... Unless... I raise an exception there, and check it on the `data_received`!

Writing stuff to myself is very useful. :eyes:

Kludex:
Closing it again, as the implementation is wrong.

Kludex:
I don't think it's possible to implement the logic that I want on the `h11` implementation, because when I have the `Request` event available, all the headers were already included.

I've created this issue on `h11`: https://github.com/python-hyper/h11/issues/155

PR #1683, opened 2022-10-01
Implement `--limit-request-header-count` flag

Kludex:
- Replaces #1682

Please check #1682 before reviewing this.

## Checklist

- [ ] h11 implementation. See https://github.com/python-hyper/h11/issues/155 for more information.
- [ ] Check if the flag `limit_request_fields` on `gunicorn` is correctly passed to `uvicorn`.
- [ ] Add an entry to "Resource Limits" section in the documentation.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -97,6 +97,10 @@ Options:
   --limit-concurrency INTEGER     Maximum number of concurrent connections or
                                   tasks to allow, before issuing HTTP 503
                                   responses.
+  --limit-request-header-count INTEGER
+                                  Maximum number of HTTP headers to accept.
+                                  Defaults to 100, and it can't be larger than
+                                  32768.
   --backlog INTEGER               Maximum number of connections to hold in
                                   backlog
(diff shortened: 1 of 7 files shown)


njsmith:
I'm not sure this approach actually does anything useful -- presumably the *reason* people want to limit header count is to avoid people sending huge numbers of headers and overwhelming the server. But this PR checks the number of headers *after* the server has already accepted and parsed the entire header block, so if the header block was going to overwhelm the server, it's already happened before the check runs.

In practice though h11 already puts a limit on the total size of the header block, so this option probably isn't doing anything useful anyway. https://github.com/python-hyper/h11/issues/155#issuecomment-1264480829 has some more details.

Kludex:
I see... Thanks @njsmith :pray: 

I'll look for just limiting the size then. For `h11`, we already support `h11-max-incomplete-event-size`, but for `httptools` can I limit this on `uvicorn` itself? If I do that check on `data_received`, it would slow down `uvicorn`... :thinking:

PR #1684, opened 2022-10-01
feat: add support for limit header size

iudeen:
Implemented a feature discussed in #157

Note: 
I am still figuring out how to add this in `h11` implementation. But as per https://github.com/encode/uvicorn/issues/157#issuecomment-1264326300 we do not need there.


iudeen:
Some tests are failing because of huge headers. Should we not default to 8190, and use only if the argument is passed?

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -237,6 +241,12 @@ def on_header(self, name: bytes, value: bytes) -> None:
         if name == b"expect" and value.lower() == b"100-continue":
             self.expect_100_continue = True
         self.headers.append((name, value))
+        if self.config.limit_request_header_size and (
```
iudeen, line 242: I wanted to check lengths of name and values individually. Is this the right approach? Or should we combine? `len(name+value)`?

uvicorn/main.py
```diff
@@ -258,6 +258,15 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     help="Maximum number of concurrent connections or tasks to allow, before issuing"
     " HTTP 503 responses.",
 )
+@click.option(
+    "--limit-request-header-size",
+    type=int,
+    default=0,
+    is_flag=False,
+    flag_value=8190,
```
iudeen, line 263: 8190 is standard value used by Gunicorn, IBM server, etc.

iudeen:
Thanks @Kludex for pointing out an example test :)

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -154,7 +155,10 @@ def data_received(self, data: bytes) -> None:
         try:
             self.parser.feed_data(data)
         except httptools.HttpParserError:
-            msg = "Invalid HTTP request received."
+            if self.header_too_big:
+                msg = "Header size exceeds limit"
```
BilalAlpaslan, line 159: it sould be return http 431 Request Header Fields Too Large
https://www.rfc-editor.org/rfc/rfc6585#section-5

BilalAlpaslan (approved):

uvicorn/protocols/http/httptools_impl.py
iudeen: Makes sense, thanks!

But I'm not sure if there's a straightforward way to raise or respond with status code other than 400. 

May be we need a helper function similar to send400response for other status codes.

Kludex:
We should be able to stop when we are about to reach the limit, not from the chunk itself. When we have the chunk of data, it's already too late.

I'm not sure if this is possible without changing httptools itself. Let's follow https://github.com/MagicStack/httptools/issues/87

iudeen:
So shall we close this PR now?

PR #1704, opened 2022-10-12
change reload warning to info

gertcuykens:
https://github.com/encode/uvicorn/discussions/1701


Kludex:
By now, I agree with you that it shouldn't be a warning, but given that it's used only in development, it actually makes my eyes happy when I see the yellow message.

But the real argument for not accepting this is that it's not an obvious change that should get in, and users may be unhappy.

In any case, thanks for bringing this up.

@euri10 if you disagree with my judgement here, feel free to reopen, and merge it. :pray:

PR #1706, opened 2022-10-12

### LINKED ISSUE #1618: "Date" header not changing between requests

Kludex:
- Closes #1618 

Does anyone know a way to test this feature? ;)

Testing with `asyncio.sleep` is not reliable.

## Related PRs

This was already worked, and reverted twice. I'm sorry for that.

- https://github.com/encode/uvicorn/pull/1639
- https://github.com/encode/uvicorn/pull/1664
- https://github.com/encode/uvicorn/pull/1665
- https://github.com/encode/uvicorn/pull/1671
- https://github.com/encode/uvicorn/pull/1668

This PR is a blocker for the next release.

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -92,7 +92,6 @@ class H11Protocol(asyncio.Protocol):
         self.server_state = server_state
         self.connections = server_state.connections
         self.tasks = server_state.tasks
-        self.default_headers = server_state.default_headers
 
         # Per-connection state
         self.transport: asyncio.Transport = None  # type: ignore[assignment]
@@ -230,7 +229,7 @@ class H11Protocol(asyncio.Protocol):
                     logger=self.logger,
                     access_logger=self.access_logger,
                     access_log=self.access_log,
-                    default_headers=self.default_headers,
+                    default_headers=self.server_state.default_headers,
                     message_event=asyncio.Event(),
                     on_response=self.on_response_complete,
                 )
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -90,7 +90,6 @@ class HttpToolsProtocol(asyncio.Protocol):
         self.server_state = server_state
         self.connections = server_state.connections
         self.tasks = server_state.tasks
-        self.default_headers = server_state.default_headers
 
         # Per-connection state
         self.transport: asyncio.Transport = None  # type: ignore[assignment]
@@ -199,7 +198,7 @@ class HttpToolsProtocol(asyncio.Protocol):
     def send_400_response(self, msg: str) -> None:
 
         content = [STATUS_LINE[400]]
-        for name, value in self.default_headers:
+        for name, value in self.server_state.default_headers:
             content.extend([name, b": ", value, b"\r\n"])
         content.extend(
             [
@@ -274,7 +273,7 @@ class HttpToolsProtocol(asyncio.Protocol):
             logger=self.logger,
             access_logger=self.access_logger,
             access_log=self.access_log,
-            default_headers=self.default_headers,
+            default_headers=self.server_state.default_headers,
             message_event=asyncio.Event(),
             expect_100_continue=self.expect_100_continue,
             keep_alive=http_version != "1.0",


lovelydinosaur:
> Does anyone know a way to test this feature?

For a first pass can we just describe the change in behaviour against this issue, with a manual demonstration.
I assume this change resolves #1618, right?

Kludex:
Application:
```python
    await send(
        {
            "status": 200,
            "headers": [
                [b"content-type", b"text/plain"],
            ],
        }
    )
    await send(
        {
            "type": "http.response.body",
            "body": b"Hello, world!",
        }
    )
```

Server command (set `timeout-keep-alive` to make sure the connection stays alive):
```bash
uvicorn main:app --reload --timeout-keep-alive 100
```

Client:
```python
from time import sleep

from httpx import Client

with Client() as client:
    response = client.get("http://localhost:8000")
    first_date_header = response.headers["date"]

    sleep(3)

    response = client.get("http://localhost:8000")
    second_date_header = response.headers["date"]
    assert first_date_header != second_date_header, "Date headers should be different"
```

You can test the code with and without this PR. 🙏

Kludex:
> > Does anyone know a way to test this feature?
> 
> For a first pass can we just describe the change in behaviour against this issue, with a manual demonstration. I assume this change resolves #1618, right?

Correct. It solves #1618.

lovelydinosaur:
Pragmatically I'd be inclined to just accept this pull request.

euri10:
as far as I'm ocncerned I'm not inclined in accepting it without a test. This is a reather simple test, just sending 2 queries waiting 1s in between, there's no magic at all, so the test fails smells really bad, the fact the sleep is not respected in the CI looks bad as well, if the logs displayed are correct of course, the fact noone here can understand why it fails is probably the worse.

anyways, the fix looks good, you can merge it without a test, but this will resurface one way or the other because it's highly illogical it's failing at the moment.

lovelydinosaur:
> noone here can understand why it fails is probably the worse

We know why it fails, and why this is the right fix...

The `ServerState` holds the default headers that should be used, and is updated ~once per second, which includes setting the `Date` field...

https://github.com/encode/uvicorn/blob/a94781dcd122b9727b85f86fcc60318eed72d223/uvicorn/server.py#L224-L250

This PR changes where we include the `ServerState.default_headers`.
Instead of being read once at the start of the connection, they're instead read for every request.

It's a simple fix. It's just awkward to write a test for is all.

euri10:
> > noone here can understand why it fails is probably the worse
> 
> We know why it fails, and why this is the right fix...

I'm not debating the fix, which is good, I was talking about the test that fails.
Anyways do as you want.

Kludex:
The problem on the test is that comparing time with time.time can lead to this lack of precision.

I can mock patch the datetime format on the date header, and also mock patch time.time to use loop.time, and then making sure that is bigger than one second and less than two...

But I guess things become too complex?

Kludex:
@tomchristie @euri10 Can I have an approval here to proceed with the release? :pray: 

In any case, I'll try to increase the test coverage to 100% until the end of the year (I'm not sure if I can cover this scenario anyway).

lovelydinosaur (approved):
Absolutely, yup. 👍

PR #1709, opened 2022-10-16
Adding SIGQUIT as a handled signal

adnaanbheda:
Gunicorn sends a SIGQUIT instead of a SIGINT even when it receives a SIGINT, in any case, uvicorn should be able to handle SIGQUIT as well, so this is a small fix that does that.

Fixes

Ref: https://github.com/benoitc/gunicorn/issues/2604


Kludex:
Let's proceed with #1710. 👍

PR #566, opened 2020-02-05
Check that socket exists on shutdown

wagoodman:
- Reverses the order during `Server.shutdown()` such that a server (which may have a reference to sockets) is closed first followed by a safe close of any remaining sockets.
- Adds a test ensuring when binding to a socket (usually done in `main,py run()`) that `server.serve(sockets=[sock])` does not raise any exceptions (as seen in #565).
- black formatting for `setup.py`

Closes #565

PR #1710, opened 2022-10-17
Add `SIGQUIT` handler to `UvicornWorker`

adnaanbheda:
Gunicorn sends a SIGQUIT instead of a SIGINT even when it receives a SIGINT, in any case, uvicorn should be able to handle SIGQUIT as well, so this is a small fix that does that.

Fixes

Ref: https://github.com/benoitc/gunicorn/issues/2604

### DIFF
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -3,6 +3,7 @@ import logging
 import signal
 import sys
 from typing import Any
+import threading
 
 from gunicorn.arbiter import Arbiter
 from gunicorn.workers.base import Worker
@@ -76,9 +77,22 @@ class UvicornWorker(Worker):
         # Don't let SIGUSR1 disturb active requests by interrupting system calls
         signal.siginterrupt(signal.SIGUSR1, False)
 
+    def _install_sigquit_handler(self) -> None:
+        """Workaround to install a SIGQUIT handler on workers.
+        - https://github.com/encode/uvicorn/issues/1116
+        - https://github.com/benoitc/gunicorn/issues/2604
+        """
+        if threading.current_thread() is not threading.main_thread():
+            # Signals can only be listened to from the main thread.
+            return
+
+        loop = asyncio.get_running_loop()
+        loop.add_signal_handler(signal.SIGQUIT, self.handle_exit, signal.SIGQUIT, None)
+
     async def _serve(self) -> None:
         self.config.app = self.wsgi
         server = Server(config=self.config)
+        self._install_sigquit_handler()
         await server.serve(sockets=self.sockets)
         if not server.started:
             sys.exit(Arbiter.WORKER_BOOT_ERROR)
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -77,7 +77,8 @@ class UvicornWorker(Worker):
         signal.siginterrupt(signal.SIGUSR1, False)
 
     def _install_sigquit_handler(self) -> None:
-        """Workaround to install a SIGQUIT handler on workers.
+        """Install a SIGQUIT handler on workers.
+
         - https://github.com/encode/uvicorn/issues/1116
         - https://github.com/benoitc/gunicorn/issues/2604
         """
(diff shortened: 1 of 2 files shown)


uvicorn/workers.py
```diff
@@ -76,9 +77,22 @@ def init_signals(self) -> None:
...
+        """Workaround to install a SIGQUIT handler on workers.
+        - https://github.com/encode/uvicorn/issues/1116
+        - https://github.com/benoitc/gunicorn/issues/2604
+        """
+        if threading.current_thread() is not threading.main_thread():
+            # Signals can only be listened to from the main thread.
+            return
+
+        loop = asyncio.get_running_loop()
+        loop.add_signal_handler(signal.SIGQUIT, self.handle_exit, signal.SIGQUIT, None)
```
Kludex: Is all this logic necessary? Can't we add part of this in the `init_signals` method above?

Kludex: Maybe the question is misplaced because the signals are added to different processes? 🤔

adnaanbheda: Seems like it, the signal handler needs to be attached to the main process's event loop, which will be available after `Worker.run()` calls `_serve()` inside an event loop (`asyncio.run()`).

The question is whether the thread check is necessary 🤔 because when a `SIGINT` signal is sent, the main thread will handle it anyways.

Kludex:
I have this `Error while closing socket [Errno 9] Bad file descriptor` when shutting down with this PR. We need to solve that before this gets in. :eyes:

uvicorn/workers.py
```diff
@@ -76,9 +77,22 @@ def init_signals(self) -> None:
...
         signal.siginterrupt(signal.SIGUSR1, False)
 
+    def _install_sigquit_handler(self) -> None:
+        """Workaround to install a SIGQUIT handler on workers.
+        - https://github.com/encode/uvicorn/issues/1116
+        - https://github.com/benoitc/gunicorn/issues/2604
+        """
+        if threading.current_thread() is not threading.main_thread():
+            # Signals can only be listened to from the main thread.
+            return
```
Kludex: This shouldn't be needed. It's on the `Server` because `uvicorn` can be started in a thread, but the `UvicornWorker` cannot.

Kludex:
Hi @adnaanbheda , are you still interested on working on this PR?

adnaanbheda:
@Kludex yep, lemme check on this tomorrow.

adnaanbheda:
@Kludex Managed to fix this by modifying the server, closing down / releasing the sockets first instead of the servers.

I understand that's not the way we want to go ahead, we would like to steer this change towards `UvicornWorker` ! 👀 
But, I think the change is appropriate in the scope of `uvicorn` as well, let me know what you think.
Anyways, I'll look at this again and try to keep it worker-only!

Kludex:
> But, I think the change is appropriate in the scope of uvicorn as well, let me know what you think.

It's not. The order was changed on https://github.com/encode/uvicorn/pull/566/.

Kludex:
So... Maybe the `sock.close()` logic shouldn't even be here, as the `server.close()` closes the socket as well, see [here](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_server).

That said, the issue we are facing can be found on https://github.com/benoitc/gunicorn/issues/1877.

Kludex (changes requested):
Please revert the last commit, and we are good. :+1: 

The Bad descriptor issue is on `gunicorn`, as mentioned in the last comment.

uvicorn/workers.py
```diff
@@ -76,9 +76,19 @@ def init_signals(self) -> None:
         # Don't let SIGUSR1 disturb active requests by interrupting system calls
         signal.siginterrupt(signal.SIGUSR1, False)
 
+    def _install_sigquit_handler(self) -> None:
+        """Workaround to install a SIGQUIT handler on workers.
```
Kludex:
```suggestion
        """Install a SIGQUIT handler on workers.

```

Kludex (approved):
Thanks @adnaanbheda ! :+1:

PR #1716, opened 2022-10-19
Allowing log_config to accept ConfigParser instance or a file object

antonymayi:
The existing logic for the `log_config` option accepts:
* `dict` or `str` representing a JSON or YAML file name to be parsed using `logging.config.dictConfig`
* `str` representing an INI file name to be parsed using `logging.config.fileConfig`

The `logging.config.fileConfig`, however, accepts not only the file name, but possibly also an instance of `configparser.RawConfigParser` and even an open file object. Passing the instance of `configparser.RawConfigParser` (or its subclass) would be especially useful as this can accumulate configs from multiple files (`cp.read()` can be called repeatedly on different files updating its config values) instead of just a single one.

I am proposing this trivial change to allow the `log_config` option to also accept the `*ConfigParser` instance or a general file object.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,3 +1,5 @@
+import configparser
+import io
 import json
 import logging
 import os
@@ -414,15 +416,21 @@ def test_log_config_yaml(
     mocked_logging_config_module.dictConfig.assert_called_once_with(logging_config)
 
 
-def test_log_config_file(mocked_logging_config_module: MagicMock) -> None:
+@pytest.mark.parametrize(
+    "config_file", ["log_config.ini", configparser.ConfigParser(), io.StringIO()]
+)
+def test_log_config_file(
+    mocked_logging_config_module: MagicMock,
+    config_file: typing.Union[str, configparser.RawConfigParser, typing.IO],
+) -> None:
     """
     Test that one can load a configparser config from disk.
     """
-    config = Config(app=asgi_app, log_config="log_config")
+    config = Config(app=asgi_app, log_config=config_file)
     config.load()
 
     mocked_logging_config_module.fileConfig.assert_called_once_with(
-        "log_config", disable_existing_loggers=False
+        config_file, disable_existing_loggers=False
     )
 
 
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -7,8 +7,10 @@ import os
 import socket
 import ssl
 import sys
+from configparser import RawConfigParser
 from pathlib import Path
 from typing import (
+    IO,
     TYPE_CHECKING,
     Any,
     Awaitable,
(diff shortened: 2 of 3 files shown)


Kludex:
I didn't even know about `ConfigParser` :rofl: 

Anyway... Can you show me an example on how you use this? A self-contained repository that uses this branch is enough.

Also, we need to add a test for it. :eyes:

antonymayi:
re the usage example:

I use `uvicorn` as a specific plugin within the [ForML](https://github.com/formlio/forml) framework which itself is using [its own logging config](https://github.com/formlio/forml/blob/1b0f578feb307255612bab44a5bc4edf2c789855/forml/setup/_logging.py#L32) potentially combining multiple user-provided INI files loaded (via the [LOGGING.read()](https://github.com/formlio/forml/blob/1b0f578feb307255612bab44a5bc4edf2c789855/forml/setup/_logging.py#L38)) into a single `ConfigParser` instance.

It would be convenient to be able to pass that instance to the uvicorn setup [as shown here](https://github.com/antonymayi/forml/blob/7d9903e7806a8ee774dfaed9bea2ee64ebaff7fe/forml/provider/gateway/rest.py#L139) (working example based on my PR).

I've updated the PR with the `test_log_config_file` now covering not just a string file name but also an instance of a `ConfigParser` and a file-like object.

antonymayi:
Hi @Kludex, happy to update the PR based on your further guidelines in case you have any?

iudeen (commented):
I like this! 👍

Kludex:
Is there any big project that uses `ConfigParser`?

antonymayi:
> Is there any big project that uses `ConfigParser`?

Depends on the definition of a _big project_ but I believe [Apache Airflow](https://github.com/apache/airflow/blob/main/airflow/configuration.py) for instance should be a fit beyond any doubt.

Kludex (commented):
What is the amount of work needed for you if this PR is not accepted?

antonymayi:
> What is the amount of work needed for you if this PR is not accepted?

Making my life better or worse shouldn't be the concern here (even though I appreciate the consideration).

Can you please briefly clarify your reservations about this feature? Do you see any issues or do you just think it is a marginal use case and hence not worth it?

The way I see it, this really makes it way easier to integrate `uvicorn` logging within any embedding project. Without it, `uvicorn` is forcing the embedding application to use a single physical file (why single? why physical?) or essentially a hand-written (since there is no standard API to produce it programmatically) dictionary. This seems unnecessarily too demanding from a component within an embedding project (especially if it is not even a central component like in my case).

Had `uvicorn` supported just `ConfigParser` and not any of the other current options (files/dict), it would still have supported all these usecases (it is possible to create `ConfigParser` instance from files/dict) but not the other way round. So `ConfigParser` should have imho always been the primary option and only then considering the other ways (explicit files/dict) as convenient cosmetical shortcuts.

antonymayi:
I accidentally deleted the forked repository which automatically closed this PR. I am opening a new one as #1976, apologies for the inconvenience.

PR #1718, opened 2022-10-19
Ideas to test gunicorn

Kludex:


setup.cfg
```diff
@@ -75,8 +75,7 @@ filterwarnings=
     ignore: 'cgi' is deprecated and slated for removal in Python 3\.13:DeprecationWarning
 
 [coverage:run]
-omit = venv/*
-include = uvicorn/*, tests/*
+source_pkgs = uvicorn, tests
```
Kludex, line 78: This should change anyway because we are missing files in the coverage report :man_facepalming:

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -23,9 +23,9 @@
 )
 from uvicorn.server import ServerState
 
-if sys.version_info < (3, 8):
+if sys.version_info < (3, 8):  # pragma: py-gte-38
     from typing_extensions import Literal
-else:
+else:  # pragma: py-lt-38
     from typing import Literal
```
Kludex, line 29: Those should also get in anyway.

PR #1722, opened 2022-10-21
Fixed arguments mismatch in docs

kamil-cy:
Hi, I've found this mismatch when I strictly followed documentation from this link https://www.uvicorn.org/settings/

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -3,7 +3,7 @@
 Use the following options to configure Uvicorn, when running from the command line.
 
 If you're running programmatically, using `uvicorn.run(...)`, then use
-equivalent keyword arguments, eg. `uvicorn.run("example:app", port=5000, reload=True, access_log=False)`.
+equivalent keyword arguments, eg. `uvicorn.run("example:app", port=5000, reload=True, access_log=False)`. It's worth to mention that for three of them  `--reload-dir`, `--reload-include` and `--reload-exclude` you have to use plural forms `reload_dirs`, `reload_includes` and `reload_excludes` accordingly.
 Please note that in this case, if you use `reload=True` or `workers=NUM`,
 you should put `uvicorn.run` into `if __name__ == '__main__'` clause in the main module.
 


Kludex:
It's not a mismatch... It's intended. :thinking: 

On a CLI, you can pass many arguments, but you do something like: `uvicorn --reload-dir potato --reload-dir banana`. On Python, you have a list of directories.

kamil-cy:
Ok, so in that case maybe it would be great to add: 
```If you are running programmatically use `reload_includes=[]` and pass a list of strings.```
and
```If you are running programmatically use `reload_excludes=[]` and pass a list of strings.```
at the end of options description like it is now only for `--reload-dir`. What do you think about that?

Kludex:
I'm trying to understand why exactly this was an issue in the first place. 🤔 

Assuming you are using a modern IDE (PyCharm or VSCode), when you do `uvicorn.run(reload_dirs=` you are able to see the type of it... Isn't it the case?

kamil-cy:
When I'm learning something I use a very simple text editor. I see a `--reload-include` option in the documentation, so I'm trying to use `uvicorn.run(reload_include=` (without `s`, at the end, just simple conversion" removing `--` and replacing `-` with `_`) and got an error as a result. So I switched to the modern IDE and saw that the valid parameter name is the `reload_excludes` (with `s` at the end). Problem solved, but then I decided to let you know about it, because maybe someone else might bump onto it in the future. Eventually I think that we can close this PR without any improvements to the documentation, because probably almost 100% of people use a modern IDE. Thank you for the reply and for your time.

Kludex:
I see. I use `nvim` for small things, and I don't have any extension that let me know about the parameters as well. 👀

Anyway, thanks for the PR. I'll check this later, maybe we can rearrange the words here. 🙏

kamil-cy:
Sure. No problem, take your time. This is not urgent. Documentation is written very good. I just wanted to make these parameters look more explicit, so that any newbie to uvicorn who is doing this a hard way without a modern IDE (like you or me) to not hesitate at all when follows documentation.

Kludex:
@kamil-cy I've thought about this.

I think we need a bit of refactoring on our docs to make it more visual appealing, and more logical in the sense of reading. I also think that we should have a separate section only for running `uvicorn` via Python code.

I'll be closing this PR for now, and think on how we can do what I mentioned above.

Thanks for the PR tho, it's appreciated.

PR #1727, opened 2022-10-26
improve coverage

iudeen:


iudeen:
There are many instances where coverage is dropped because of a implicit `return None`. I understand that mypy does not like functions without a return statement, but I am not sure if we can write tests for those. 

Note to self: Also I had to skip coverage for many. When I get time, I will revisit to check if a test can be written for those.

Kludex:
> I understand that mypy does not like functions without a return statement, but I am not sure if we can write tests for those.

If you can't, the function is wrong.

Kludex:
Removing the `return None` just hides the fact that we are not testing when the function returns `None`... As every function that doesn't return anything returns `None`... :eyes:

iudeen:
> 

Agreed. No return implies `None`. Do we check for private functions as well? (eg: `_get_headers()`)?

iudeen:
I will have to spend more time on this before marking this as final. I hope its okay if this remains in draft for a few days!

Kludex:
@iudeen To be more efficient on the _100% test coverage_ milestone, I recommend adding a single test per PR.

I'm also inclined to not accept PRs with `# pragma: no cover`, unless they are justified by previous `pragma: no cover`s that we have in code. Also, even in the scenario that they are justified by previous ones, maybe we did it wrong on the previous ones and we should correct them.

iudeen:
Okay, I shall do that. One test PR makes much more sense. Yes, we would definitely would want to avoid pragma no cover. 

Closing this for now. Will open individual test PR.

Thanks!

Kludex:
Thanks :+1:

PR #1731, opened 2022-10-27
Warn user when reload and workers flag are used together

Kludex:
Either this, or `RuntimeError` will do.


iudeen:
Since we are ignoring, warning is better IMHO.

lovelydinosaur (approved):
Thanks folks. 🙏🏼

PR #1734, opened 2022-10-27
Make pyproject.toml PEP 621 compliant

simontegelid:
The issue is (had issues with building this with buildroot), which requires this to be correct:

```text
$ validate-pyproject pyproject.toml                          
Invalid file: pyproject.toml
[ERROR] `project.license` must be valid exactly by one definition (2 matches found):

    - keys:
        'file': {type: string}
      required: ['file']
    - keys:
        'text': {type: string}
      required: ['text']
```

For more info: https://peps.python.org/pep-0621/#license

After this change:

```text
$ validate-pyproject pyproject.toml
Valid file: pyproject.toml
```

For the record:

```text
$ validate-pyproject --version     
validate_pyproject 0.10.1
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "uvicorn"
 dynamic = ["version"]
 description = "The lightning-fast ASGI server."
 readme = "README.md"
-license = "BSD-3-Clause"
+license = {text = "BSD-3-Clause"}
 requires-python = ">=3.7"
 authors = [
     { name = "Tom Christie", email = "tom@tomchristie.com" },


Kludex:
@ofek do you have instructions here?

iudeen:
Based on the [Hatch Documentation](https://hatch.pypa.io/latest/config/metadata/#license) the existing one seems fine. 
Also check https://peps.python.org/pep-0639/#add-string-value-to-license-key

PEP 639 supersedes PEP 621?

simontegelid:
PEP 621 is a bit vague in that "The table may have one of two keys" (just "may") and at the same time "A practical string value for the license key has been purposefully left out to allow for a future PEP to specify support for [SPDX](https://spdx.dev/) expressions". Seems like PEP 639 deprecate license sub keys https://peps.python.org/pep-0639/#deprecate-license-key-table-subkeys (had missed this PEP until now). What a mess. So, maybe it's an error to merge this after all.

Kludex:
So... Do we close this?

ofek:
related: https://github.com/psf/black/pull/3326#issuecomment-1274126507

iudeen:
Is it only the moon launch that's delaying the approval of PEP639?😂

ofek:
yup lollll

Kludex:
> So, maybe it's an error to merge this after all.

Ok. Thanks. :pray:

PR #1737, opened 2022-10-28
Check if handshake is completed before sending frame on wsproto shutdown

### LINKED ISSUE #596: Handshaking may not be completed yet at `shutdown` in wsproto impl

Kludex:
- Closes #596

I'm extremely happy about this PR. I understand WebSocket things now. :sweat_smile:

## How to test it?

Application:
```python
import asyncio


    event = await receive()
    assert event["type"] == "websocket.connect"

    print("Go ahead, stop the server...")
    await asyncio.sleep(10)
    print("Too late!")

```
Client (I've used curl):
```bash
curl --include \
     --no-buffer \
     --header "Connection: Upgrade" \
     --header "Upgrade: websocket" \
     --header "Host: example.com:80" \
     --header "Origin: http://example.com:80" \
     --header "Sec-WebSocket-Key: SGVsbG8sIHdvcmxkIQ==" \
     --header "Sec-WebSocket-Version: 13" \
     http://localhost:8000/
```

## Checklist

- [x] Include every behavior introduced here on `websockets` implementation.
- [x] Tests

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -82,7 +82,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 97.82
+fail_under = 97.87
 show_missing = true
 skip_covered = true
 exclude_lines =
--- a/setup.cfg
+++ b/setup.cfg
@@ -82,7 +82,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 97.87
+fail_under = 97.92
 show_missing = true
 skip_covered = true
 exclude_lines =
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -528,7 +528,6 @@ async def test_client_connection_lost(ws_protocol_cls, http_protocol_cls):
         while True:
             message = await receive()
             if message["type"] == "websocket.connect":
-                print("accepted")
                 await send({"type": "websocket.accept"})
             elif message["type"] == "websocket.disconnect":
                 break
@@ -549,6 +548,34 @@ async def test_client_connection_lost(ws_protocol_cls, http_protocol_cls):
             got_disconnect_event_before_shutdown = got_disconnect_event
 
     assert got_disconnect_event_before_shutdown is True
+
+
+@pytest.mark.anyio
+@pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
+@pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
+async def test_not_accept_on_connection_lost(ws_protocol_cls, http_protocol_cls):
+    send_accept_task = asyncio.Event()
+
+    async def app(scope, receive, send):
+        while True:
+            message = await receive()
+            if message["type"] == "websocket.connect":
+                await send_accept_task.wait()
+                await send({"type": "websocket.accept"})
+            elif message["type"] == "websocket.disconnect":
+                break
+
+    async def websocket_session(uri):
+        async with websockets.client.connect(uri):
+            while True:
+                await asyncio.sleep(0.1)
+
+    config = Config(app=app, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off")
+    async with run_server(config):
+        task = asyncio.create_task(websocket_session("ws://127.0.0.1:8000"))
+        await asyncio.sleep(0.1)
+        task.cancel()
+        send_accept_task.set()
 
 
 @pytest.mark.anyio
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -581,6 +581,38 @@ async def test_not_accept_on_connection_lost(ws_protocol_cls, http_protocol_cls)
 @pytest.mark.anyio
 @pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
 @pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
+async def test_send_close_on_server_shutdown(ws_protocol_cls, http_protocol_cls):
+    disconnect_message = {}
+
+    async def app(scope, receive, send):
+        nonlocal disconnect_message
+        while True:
+            message = await receive()
+            if message["type"] == "websocket.connect":
+                await send({"type": "websocket.accept"})
+            elif message["type"] == "websocket.disconnect":
+                disconnect_message = message
+                break
+
+    async def websocket_session(uri):
+        async with websockets.client.connect(uri):
+            while True:
+                await asyncio.sleep(0.1)
+
+    config = Config(app=app, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off")
+    async with run_server(config):
+        task = asyncio.create_task(websocket_session("ws://127.0.0.1:8000"))
+        await asyncio.sleep(0.1)
+        disconnect_message_before_shutdown = disconnect_message
+
+    assert disconnect_message_before_shutdown == {}
+    assert disconnect_message == {"type": "websocket.disconnect", "code": 1012}
+    task.cancel()
+
+
+@pytest.mark.anyio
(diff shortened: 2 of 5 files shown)


uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -125,9 +125,12 @@ def resume_writing(self):
         self.writable.set()
 
     def shutdown(self):
-        self.queue.put_nowait({"type": "websocket.disconnect", "code": 1012})
-        output = self.conn.send(wsproto.events.CloseConnection(code=1012))
-        self.transport.write(output)
+        if self.handshake_complete:
+            self.queue.put_nowait({"type": "websocket.disconnect", "code": 1012})
+            output = self.conn.send(wsproto.events.CloseConnection(code=1012))
+            self.transport.write(output)
```
Kludex, line 131: This is the core here.

We want to check if the handshake was completed, because `CloseConnection` is not a valid event for `wsproto.WSConnection.send()` when the handshake is not completed.

This is the error these lines solve:
```bash
wsproto.utilities.LocalProtocolError: Event CloseConnection(code=1012, reason=None) cannot be sent during the handshake
```

More about it on #596.

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -125,9 +125,12 @@ def resume_writing(self):
...
+        else:
+            self.send_500_response()
```
Kludex, line 133: The thing is... If the above is `False`, we still need to send something to the client, otherwise we are being rude. :disappointed: 

Without these lines, the client will have a:
```bash
```

With these lines:
```
HTTP/1.1 500 
content-length: 0
```

The curl command used was:
```bash
curl --include \
     --no-buffer \
     --header "Connection: Upgrade" \
     --header "Upgrade: websocket" \
     --header "Host: example.com:80" \
     --header "Origin: http://example.com:80" \
     --header "Sec-WebSocket-Key: SGVsbG8sIHdvcmxkIQ==" \
     --header "Sec-WebSocket-Version: 13" \
     http://localhost:8000/
```

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -222,9 +225,8 @@ def send_500_response(self):
     async def run_asgi(self):
         try:
             result = await self.app(self.scope, self.receive, self.send)
-        except BaseException as exc:
-            msg = "Exception in ASGI application\n"
-            self.logger.error(msg, exc_info=exc)
+        except BaseException:
+            self.logger.exception("Exception in ASGI application\n")
```
Kludex, line 229: These are analogous. Just making it more readable.

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -257,14 +259,15 @@ async def send(self, message):
...
+                    output = self.conn.send(
+                        wsproto.events.AcceptConnection(
+                            subprotocol=subprotocol,
+                            extensions=extensions,
+                            extra_headers=extra_headers,
+                        )
                     )
-                )
-                self.transport.write(output)
+                    self.transport.write(output)
```
Kludex, line 270: This is necessary on this scenario, because the transport was closed on the `shutdown()`, it can also be that the client disconnected before accepting the connection...

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -345,6 +345,8 @@ async def asgi_receive(
             data = await self.recv()
         except ConnectionClosed as exc:
             self.closed_event.set()
+            if self.ws_server.closing:
+                return {"type": "websocket.disconnect", "code": 1012}
```
Kludex, line 349: This is to match the behavior on both implementations.

If we close the connection on `shutdown`, we should be sending 1012.

tests/utils.py
```diff
@@ -13,6 +14,8 @@ async def run_server(config: Config, sockets=None):
     await asyncio.sleep(0.1)
     try:
         yield server
+    except BaseException:  # pragma: no cover
+        traceback.print_exc()
```
Kludex, line 18: This is actually to have more visibility.

Let me explain a bit... Well... The `run_server` is a context manager that prevents the test to fail. We need to refactor this logic, or only have assertions outside the `run_server`. In any case, the idea here is just to see the potential exceptions when running `pytest` with `-s`.

setup.cfg
```diff
@@ -82,7 +82,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 97.82
+fail_under = 97.87
```
Kludex: Yey! :tada:

Kludex: Let's try to increase it! :smiling_imp: 
```suggestion
fail_under = 97.92
```

setup.cfg
```diff
@@ -82,7 +82,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 97.82
+fail_under = 97.92
```
Kludex, line 85: Yey! :sunglasses: :+1:

iudeen (approved):
Handles everything that was attempted in https://github.com/encode/uvicorn/pull/1738 😄

tests/utils.py
Kludex:
```suggestion
```

tests/utils.py
```diff
@@ -1,5 +1,6 @@
 import asyncio
 import os
+import traceback
```
Kludex:
```suggestion
```

uvicorn/protocols/websockets/wsproto_impl.py
lovelydinosaur: Much neater. 😌

I did spend some time trying to figure out if we should we also include a textual description here, but I think it's probably okay as it currently stands.

uvicorn/protocols/websockets/wsproto_impl.py
lovelydinosaur: Makes sense to me. 👍

lovelydinosaur (approved):
> I'm extremely happy about this PR.

😊

Kludex:
Thanks @tomchristie 🙏

tests/protocols/test_websocket.py
```diff
@@ -551,6 +550,66 @@ async def app(scope, receive, send):
...
+        async with websockets.client.connect(uri):
+            while True:
+                await asyncio.sleep(0.1)
+
+    config = Config(app=app, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off")
+    async with run_server(config):
+        task = asyncio.create_task(websocket_session("ws://127.0.0.1:8000"))
+        await asyncio.sleep(0.1)
+        task.cancel()
+        send_accept_task.set()
```
Kludex, line 578: This test actually fails on the `websockets` implementation.

This fails before and after my changes, so it is actually not this PR who broke it.

There's a single small change on the `websockets` implementation on this PR, which actually handles the next test.

PR #1738, opened 2022-10-28
Fix shutdown event in Websocket Implementations

iudeen:


iudeen:
Current `websockets` implementation is giving status code 1006, which is not right. 

>1006 is a reserved value and MUST NOT be set as a status code in a  Close control frame by an endpoint.  It is designated for use in applications expecting a status code to indicate that the connection was closed abnormally, e.g., without sending or receiving a Close control frame.

Ref: https://www.rfc-editor.org/rfc/rfc6455#section-7.4.1

This implies in websockets implementation, connections are not gracefully closed when the server is shutdown.

iudeen:
As per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code) even 1012 is wrong. The right response code should be 1001.

Kludex:
See https://github.com/django/asgiref/issues/340 for `WebSocket` codes on disconnect/close.

iudeen:
So 1005 then? There is already a PR, but that is for disconnect. Is it the same for shutdown?

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -141,7 +141,9 @@ def connection_lost(self, exc: Optional[Exception]) -> None:
 
     def shutdown(self) -> None:
         self.ws_server.closing = True
-        self.transport.close()
+        task = asyncio.create_task(self.close(code=1012))
+        task.add_done_callback(self.on_task_complete)
+        self.tasks.add(task)
```
iudeen, line 147: Help needed here. 

`self.transport.close()` causes unexpected closure of connection at this place. However the transport is getting closed at a later stage.  Is this okay?

setup.cfg
```diff
@@ -82,7 +82,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 97.82
```
iudeen, line 85: 🍪

Kludex:
I've added the tests for `shutdown` on https://github.com/encode/uvicorn/pull/1737. :pray:

iudeen:
Ok, cool! I'll close this PR then.

PR #1739, opened 2022-10-28
Use `asyncio.create_task` instead of `asyncio.ensure_future`

iudeen:


Kludex (approved):
Thank you.

For reference, this is the recommended way, see: https://docs.python.org/3/library/asyncio-future.html#asyncio.ensure_future.

PR #1725, opened 2022-10-26
Fix crash on exist with "--uds" if socket not exists

JGoutin:
Currently uvicorn may crash when exiting if run with the "--uds" option, and if Uvicorn exits before the socket creation (For instance if the application to run crash on statup).

This PR fix the issue by ignoring the exception in that case.

PR #1741, opened 2022-10-28
Add test scenario for `--uds` flag when file exists

iudeen:
~~This PR is to test a hypothesis.~~

Hypothesis holds true. A change in #1725 caused a gap in coverage.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -82,7 +82,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 97.69
+fail_under = 97.81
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -82,7 +82,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 97.69
+fail_under = 97.81
```
Kludex: Adjusting...
```suggestion
fail_under = 97.79
```

Kludex (approved):

PR #1746, opened 2022-10-30
Annotate CONFIG_KWARGS in UvicornWorker class

ukalwa:
Avoid mypy error when adding non-str values to the arg in subclass

Discussion ref: https://github.com/encode/uvicorn/discussions/1745


Kludex (approved):
Thanks @ukalwa :pray:

PR #1753, opened 2022-11-01
Use correct WebSocket error codes

Kludex:
This is a very confusing subject.

There are 4 ways for us to send `websocket.disconnect` to the application:

| WSProtocol | WebSocketProtocol | Event | Code |
| :---: | :---: | --- | --- |
| :white_check_mark: | :white_check_mark: | Server shutdown | [`1012`](https://mailarchive.ietf.org/arch/msg/hybi/P_1vbD9uyHl63nbIIbFxKMfSwcM/) (Service Restart) |
| :white_check_mark: | :white_check_mark: | Client sent close frame | Send the code passed to the server i.e. `wsproto` sends `event.code`, and `websockets` sends `exc.code` |
| :white_check_mark: | :white_check_mark: | Handshake failed | [`1006`](https://www.rfc-editor.org/rfc/rfc6455.html) (Abnormal Closure) |
| :white_check_mark: | :white_check_mark: | Connection lost without close frame | [`1005`](https://www.rfc-editor.org/rfc/rfc6455.html) (No Status Rcvd) |

Note: If the connection was lost before the handshake was completed, then it should be a `1006`.

This PR aims to match the behavior on both websockets implementations, and fixes https://github.com/encode/uvicorn/pull/1737/files#r1009855193.

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -553,7 +553,9 @@ async def test_client_connection_lost(ws_protocol_cls, http_protocol_cls):
 @pytest.mark.anyio
 @pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
 @pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
-async def test_not_accept_on_connection_lost(ws_protocol_cls, http_protocol_cls):
+async def test_connection_lost_before_handshake_complete(
+    ws_protocol_cls, http_protocol_cls
+):
     send_accept_task = asyncio.Event()
     disconnect_message = {}
 
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -568,9 +568,7 @@ async def test_connection_lost_before_handshake_complete(
         disconnect_message = await receive()
 
     async def websocket_session(uri):
-        async with websockets.client.connect(uri):
-            while True:
-                await asyncio.sleep(0.1)
+        await websockets.client.connect(uri)
 
     config = Config(app=app, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off")
     async with run_server(config):
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -731,7 +731,7 @@ async def test_server_can_read_messages_in_buffer_after_close(
     ws_protocol_cls, http_protocol_cls
 ):
     frames = []
-    client_close_connection = asyncio.Event()
+    disconnect_message = {}
 
     class App(WebSocketResponse):
         async def websocket_connect(self, message):
@@ -739,7 +739,11 @@ async def test_server_can_read_messages_in_buffer_after_close(
             # Ensure server doesn't start reading frames from read buffer until
             # after client has sent close frame, but server is still able to
             # read these frames
-            await client_close_connection.wait()
+            await asyncio.sleep(0.2)
+
+        async def websocket_disconnect(self, message):
+            nonlocal disconnect_message
+            disconnect_message = message
 
         async def websocket_receive(self, message):
             frames.append(message.get("bytes"))
@@ -753,9 +757,9 @@ async def test_server_can_read_messages_in_buffer_after_close(
     config = Config(app=App, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off")
     async with run_server(config):
         await send_text("ws://127.0.0.1:8000")
-        client_close_connection.set()
 
     assert frames == [b"abc", b"abc", b"abc"]
+    assert disconnect_message == {"type": "websocket.disconnect", "code": 1000}
 
 
 @pytest.mark.anyio
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -555,15 +555,15 @@ async def test_client_connection_lost(ws_protocol_cls, http_protocol_cls):
 @pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
 async def test_not_accept_on_connection_lost(ws_protocol_cls, http_protocol_cls):
     send_accept_task = asyncio.Event()
+    disconnect_message = {}
 
     async def app(scope, receive, send):
-        while True:
-            message = await receive()
-            if message["type"] == "websocket.connect":
-                await send_accept_task.wait()
-                await send({"type": "websocket.accept"})
-            elif message["type"] == "websocket.disconnect":
-                break
+        nonlocal disconnect_message
+        message = await receive()
+        if message["type"] == "websocket.connect":
+            await send_accept_task.wait()
+            await send({"type": "websocket.accept"})
+        disconnect_message = await receive()
 
     async def websocket_session(uri):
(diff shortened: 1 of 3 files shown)


iudeen:
1012 means server restarted and indicates client that it can retry connection right? What if the server is gone away completely? Shouldn't it be 1001?🤔

tests/protocols/test_websocket.py
```diff
@@ -553,22 +553,22 @@ async def app(scope, receive, send):
 @pytest.mark.anyio
 @pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
 @pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
-async def test_not_accept_on_connection_lost(ws_protocol_cls, http_protocol_cls):
+async def test_connection_lost_before_handshake_complete(
```
Kludex, line 556: A more meaningful name on what scenario we are testing.

tests/protocols/test_websocket.py
```diff
@@ -577,6 +577,8 @@ async def websocket_session(uri):
         task.cancel()
         send_accept_task.set()
 
+    assert disconnect_message == {"type": "websocket.disconnect", "code": 1006}
```
Kludex, line 580: The connection was closed without a close frame, so we send a `1006`.

tests/protocols/test_websocket.py
```diff
@@ -750,8 +758,11 @@ async def send_text(url):
     config = Config(app=App, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off")
     async with run_server(config):
         await send_text("ws://127.0.0.1:8000")
+        await asyncio.sleep(0.1)
+        client_close_connection.set()
 
     assert frames == [b"abc", b"abc", b"abc"]
+    assert disconnect_message == {"type": "websocket.disconnect", "code": 1000}
```
Kludex, line 762: The client closed the connection with a close frame, and the code interpreted by the `WebSocket` packages is `1000` i.e. normal closure.

tests/protocols/test_websocket.py
```diff
@@ -553,22 +553,22 @@ async def app(scope, receive, send):
...
+        if message["type"] == "websocket.connect":
+            await send_accept_task.wait()
+            await send({"type": "websocket.accept"})
+        disconnect_message = await receive()
 
     async def websocket_session(uri):
-        async with websockets.client.connect(uri):
-            while True:
-                await asyncio.sleep(0.1)
+        await websockets.client.connect(uri)
```
Kludex, line 571: This was changed because `connect` hangs until we are able to connect with the server. There was a previous misconception that we'd actually hit the `while True` logic.

We are never able to complete the handshake, as we don't send the `websocket.accept`, so we cancel the task below and let the application continue.

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -134,6 +135,9 @@ def connection_lost(self, exc: Optional[Exception]) -> None:
             prefix = "%s:%d - " % self.client if self.client else ""
             self.logger.log(TRACE_LOG_LEVEL, "%sWebSocket connection lost", prefix)
 
+        self.lost_connection_before_handshake = (
+            not self.handshake_completed_event.is_set()
+        )
```
Kludex, line 140: We need to know if the connection was lost before the handshake was completed to determine what code we'll be sending to the application.

- If we lost connection before the handshake is completed, then we should send a `1006`.
- If we lost connection after the handshake is completed, then we should send a `1005`.

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -335,11 +339,13 @@ async def asgi_receive(
 
         await self.handshake_completed_event.wait()
 
-        if self.closed_event.is_set():
-            # If client disconnected, use WebSocketServerProtocol.close_code property.
+        if self.lost_connection_before_handshake:
```
Kludex, line 342: Instead of `lost_connection_before_handshake`, I can create an attribute called `lost_connection`, and do the conditional before the `await self.handshake_completed_event.wait()` above. What do you prefer? :thinking: 

(I think this alternative will work, I'm not sure if I'm missing something...)

Kludex:
> 1012 means server restarted and indicates client that it can retry connection right? What if the server is gone away completely? Shouldn't it be 1001?thinking

Conceptually, if we think about a web server, I think it doesn't make sense to send a code that says that the server has stopped for good. We think about it as a long-living software.

In other words, when we call "shutdown", it's because it's going to be up again.

If the application wants, it can send a `websocket.close` with `1001`, and shutdown the server. :man_shrugging:

iudeen:
> In other words, when we call "shutdown", it's because it's going to be up again.

Thanks! This justifies!

uvicorn/protocols/websockets/websockets_impl.py
iudeen: Meaning we wouldn’t hit `await self.handshake_completed_event.wait()` at all in connection is lost?

If that’s the case, it seems better 🤔

Kludex:
Maybe this is not trivial, but I really want to include this in the release. Mainly because this is the "WebSockets release". :eyes:

iudeen (approved):
This looks good!😄

It'd be great if someone can confirm the comment on `lost_connection_before_handshake` change where Kludex asked about an alternative approach 😊

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -70,13 +70,15 @@ def connection_made(self, transport):
...
-        self.queue.put_nowait({"type": "websocket.disconnect"})
+        code = 1005 if self.handshake_complete else 1006
+        self.queue.put_nowait({"type": "websocket.disconnect", "code": code})
         self.connections.remove(self)
 
         if self.logger.level <= TRACE_LOG_LEVEL:
             prefix = "%s:%d - " % tuple(self.client) if self.client else ""
             self.logger.log(TRACE_LOG_LEVEL, "%sWebSocket connection lost", prefix)
 
+        self.handshake_complete = True
```
Kludex, line 81: This is matching the behavior with the `websockets` implementation.

uvicorn/protocols/websockets/websockets_impl.py
Kludex: What I said doesn't work.

tests/protocols/test_websocket.py
abey369: > A more meaningful name on what scenario we are testing.

ETS TRADEX SERVICES LLP
BeNADABURAMbVADAKARA
NADAPURAM , VADAKARA, Kerala Waterm98562820968
Server:4
10:10AM
366/1
02-21-2023
1/10379
SALE
210000083
403226
CardXXXXXXXXXXXXXXXXmber
CardPresent to Remove Card EntrySwipe/Chomark
APPROVAL:3BC02
1 for 192.168.89.191Rs 210000083.00
GSTs 16926006.69
GSRs 12705005.02
AMRs 239631094.71Mem to Remove Watermark
†TIP
=TOTAL
TIP SUGGESTION:
Tip 15%
Tip 18%
Tip20%
R$31500012.45 Rs37800014.94 Rs4200C

PR #1756, opened 2022-11-02
Add activate virtual environment before running testing scripts

ys-wu:
You will get an error if you don't run testing scripts in a virtual environment, which is not mentioned in the documentation. I wrote a discussion [#1755 ](https://github.com/encode/uvicorn/discussions/1755) about it.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -66,6 +66,11 @@ $ scripts/install
 We use custom shell scripts to automate testing, linting,
 and documentation building workflow.
 
+Activate virtual environment before testing:
+```shell
+$ source venv/bin/activate
+```
+
 To run the tests, use:
 
 ```shell


Kludex:
But you shouldn't need this... :thinking:

ys-wu:
> But you shouldn't need this... 🤔

This line needs a `uvicorn` which is installed in the virtual environment:
```python
res = subprocess.run(["uvicorn", "--help"], stdout=subprocess.PIPE)
```

This is my understanding.
Please correct me if I misunderstood something. Thank you!

iudeen:
What if I don't have virtual environment?🤔

ys-wu:
> What if I don't have virtual environment?🤔

Thank you for your attention!

You mean to skip the `scripts/install` and run the test directly?
```
git clone https://github.com/MY-USERNAME/uvicorn
cd uvicorn
scripts/test
```
No way. The test needs dependencies. Then the error will be:
```
+ '[' -z ']'
+ scripts/check
+ ./scripts/sync-version
+ black --check --diff --target-version=py37 uvicorn tests
scripts/check: line 12: black: command not found
```

iudeen:
I believe `scripts/install` doesn't create an environment, it just installs the dependencies. 

VENV is created only for GitHub actions I suppose.

I was talking about cases where I have all my dependencies created in my base python (or even Conda). The command would not work in those cases. 

IMHO we should let this upto the user to decide how he wants his development environment to be.

Kludex:
No, that's not correct. Hold on.

ys-wu:
> I believe `scripts/install` doesn't create an environment, it just installs the dependencies.

I suppose it's easy to verify.
If you run `scripts/install` in your local machine, it does create an environment:
```sh
if [ -z "$GITHUB_ACTIONS" ]; then
    "$PYTHON" -m venv "$VENV"
    PIP="$VENV/bin/pip"
```

I'm just curious.
Can the error be reproduced in your local machine if you start from scratch as follows?
```sh
git clone https://github.com/MY-USERNAME/uvicorn
cd uvicorn
scripts/install
scripts/test
```

iudeen:
I was wrong. Apologies, it does create venv.

Kludex:
> Can the error be reproduced in your local machine if you start from scratch as follows?

No.

You aren't supposed to activate the venv. Just running `./scripts/test` should be enough. Maybe there's something wrong with your venv. Let's try to figure this out on #1755. :pray:

PR #1757, opened 2022-11-06
Add tests to check for invalid hosts and server start with socket

iudeen:


tests/test_main.py
```diff
@@ -122,3 +124,24 @@ def test_run_match_config_params() -> None:
...
+
+
+@pytest.mark.anyio
+async def test_run_invalid_host() -> None:
+    with pytest.raises(SystemExit) as e:
+        config = Config(app=app, host="illegal_host")
+        server = Server(config=config)
+        await server.serve()
+    assert e.type == SystemExit
+    assert e.value.code == 1
```
Kludex, line 132: I think we can assert something more meaningful with the `caplog` fixture... Can't we?
```suggestion
async def test_run_invalid_host(caplog: pytest.LogCaptureFixture) -> None:
    with pytest.raises(SystemExit):
        config = Config(app=app, host="illegal_host")
        server = Server(config=config)
        await server.serve()
```

tests/test_main.py
```diff
@@ -122,3 +124,24 @@ def test_run_match_config_params() -> None:
...
+            await asyncio.sleep(0.1)
+
+
+@pytest.mark.anyio
+async def test_run_invalid_host() -> None:
+    with pytest.raises(SystemExit) as e:
+        config = Config(app=app, host="illegal_host")
+        server = Server(config=config)
+        await server.serve()
+    assert e.type == SystemExit
```
Kludex, line 131: This is redundant. You are already using the `pytest.raises`.

tests/test_main.py
```diff
@@ -122,3 +124,24 @@ def test_run_match_config_params() -> None:
...
+
+@pytest.mark.anyio
+async def test_run_multiprocess_with_sockets():
+    config = Config(app=app, workers=2, limit_max_requests=1)
+    with socket.socket() as sock:
+        sock.bind(("localhost", 0))
+        async with run_server(config, sockets=[sock]) as server:
+            while not server.started:
+                await asyncio.sleep(0.1)
+            await asyncio.sleep(0.1)
```
Kludex, line 122: What's being covered by this?

iudeen: This covers https://github.com/encode/uvicorn/blob/62f19c1c39929c84968712c371c9b7b96a041dec/uvicorn/server.py#L118

tests/test_main.py
iudeen: I will check this tonight

Kludex:
@iudeen I hope you don't mind, but I'll close this, since I think we should have a PR per test case. :pray:

PR #1758, opened 2022-11-06
Add test case for `port` being zero

iudeen:


Kludex:
I've tweaked the test a bit to make it a bit simpler to understand, as the goal is already achieved with that assertion.

Thanks :pray:

Kludex (approved):

PR #1759, opened 2022-11-10
Remove unused `handle_no_connect` on `wsproto`

Kludex:
Neither [`RejectConnection`](https://github.com/python-hyper/wsproto/blob/a7d9ff0c34567b042d025db555595bb6234f89e8/src/wsproto/events.py#L91-L127) nor [`RejectData`](https://github.com/python-hyper/wsproto/blob/a7d9ff0c34567b042d025db555595bb6234f89e8/src/wsproto/events.py#L134-L151) have the `reason` field that `handle_no_connect` assumes both have, for that, I'm pretty confident this is never reached.

What those lines try to solve, it's already handled on `data_receive`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -82,7 +82,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 98.28
+fail_under = 98.50
 show_missing = true
 skip_covered = true
 exclude_lines =
--- a/uvicorn/protocols/websockets/wsproto_impl.py
+++ b/uvicorn/protocols/websockets/wsproto_impl.py
@@ -100,10 +100,6 @@ class WSProtocol(asyncio.Protocol):
                 self.handle_text(event)
             elif isinstance(event, events.BytesMessage):
                 self.handle_bytes(event)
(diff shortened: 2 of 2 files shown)


iudeen (commented):
Agreed. This looks good IMO. 

Do we have a logic for `RejectData` event in `data_received`?

Kludex:
I don't understand the question.

The RemoteProtocolError is raised on data_received.

iudeen:
What I meant to ask was if we need to handle [`RejectData`](https://github.com/python-hyper/wsproto/blob/a7d9ff0c34567b042d025db555595bb6234f89e8/src/wsproto/events.py#L134-L151) explicitly. If `RemoteProtocolError` is good enough, then okay.

lovelydinosaur:
> If RemoteProtocolError is good enough, then okay.

@iudeen - If you're happy with this, then you're welcome to review & accept it.

iudeen (approved):
Looks good! 😄

Kludex:
Thanks! :raised_hands:

PR #1764, opened 2022-11-13
Experimental: integrate httparse

florimondmanca:
Hi folks,

I'm opening this PR as the result of playing with @adriangb's [httparse](https://github.com/adriangb/httparse) port this morning.

Overall, integration went fine! As this (hopefully passing) PR will attest.

This is mostly an experiment to play with both Uvicorn and httparse. I'm not sure we want to integrate, though that might be a nice [experimental feature](https://github.com/encode/uvicorn/discussions/1762) possibility? Edit: Ha! https://github.com/Kludex/uvicorn-extensions/tree/main/src/python/uvicorn-httparse @Kludex ;-)

On the Uvicorn side, there are a few things that came out of this experiment:

* I had to add `protocol.eof_received()` calls in tests. Out of consistency, I added it even when it's not _required_ (e.g. in HTTP GET requests, which don't have a body). Maybe I'm missing something here. Both `h11` and `httptools` are able to detect "no more request content will be sent" on their own (`h11` triggers the `EndOfMessage` event, and `httptools` calls the `.on_message_complete()` callback). But `httparse` works on an external `buffer`, and when asyncio gets request EOF, `data_received()` isn't called with `b""`, instead `eof_received()` is called. If I find the way h11 and httptools do this, perhaps I can do the same and not use `eof_received()`. Until then, calling this in tests has become required for `httparse`.
  * Edit: further investigation. As far as I understand, `httptools` is based on the C++ library [llhttp](https://llhttp.org/). It seems like it uses `Content-Length` to tell "enough data has been sent, request is done". Rust's `httparse` only deals with request headers (not the body), so that's the kind of checks we'd have to do ourselves. But what about `Transfer-Encoding: chunked` requests, which don't have `Content-Length`? It seems we really do want to leverage `eof_received()`.

On the `httparse` side, some interesting bits cc @adriangb:

* Rust's `httparse` doesn't do HTTP method checking. If that's correct, I think it's fair that the port doesn't do checks either. So I had to implement some basic checks myself.
* `httparse` may raise errors in case of invalid HTTP content, such as `InvalidHttpVersion`, but those exceptions aren't exposed on the library. E.g. I had to do `except httparse._httparse.ParsingError` to handle the "invalid HTTP request" case.
* Not sure which types Rust's `httparse` uses, but `parsed.method` and `parsed.path` might be better as `bytes` rather than pre-decoded `str`. h11 uses bytes, and so does httptools.

I guess the next step could be to contribute this to [Kludex/uvicorn-extensions](https://github.com/Kludex/uvicorn-extensions/tree/main/src/python/uvicorn-httparse)? This brings questions about making sure the test suite of `uvicorn-extensions` matches up to the extensive checks we do in "main" uvicorn. (Here we add `httparser` as an extra pytest parameter to `test_http.py`, but in extensions `test_http.py` would have to be copied over.)

### DIFF
--- a/uvicorn/protocols/http/httparse_impl.py
+++ b/uvicorn/protocols/http/httparse_impl.py
@@ -1,0 +1,616 @@
+import asyncio
+import http
+import logging
+import re
+import sys
+import urllib
+from asyncio.events import TimerHandle
+from collections import deque
+from typing import TYPE_CHECKING, Callable, Deque, List, Optional, Tuple, Union, cast
+
+import httparse
+
+from uvicorn.config import Config
+from uvicorn.logging import TRACE_LOG_LEVEL
+from uvicorn.protocols.http.flow_control import (
+    CLOSE_HEADER,
+    HIGH_WATER_LIMIT,
+    FlowControl,
+    service_unavailable,
+)
+from uvicorn.protocols.utils import (
+    get_client_addr,
+    get_local_addr,
+    get_path_with_query_string,
+    get_remote_addr,
+    is_ssl,
+)
+from uvicorn.server import ServerState
+
+if sys.version_info < (3, 8):  # pragma: py-gte-38
+    from typing_extensions import Literal
+else:  # pragma: py-lt-38
+    from typing import Literal
+
+if TYPE_CHECKING:
+    from asgiref.typing import (
+        ASGI3Application,
+        ASGIReceiveEvent,
+        ASGISendEvent,
+        HTTPDisconnectEvent,
+        HTTPRequestEvent,
+        HTTPResponseBodyEvent,
+        HTTPResponseStartEvent,
+        HTTPScope,
+    )
+
+HEADER_RE = re.compile(b'[\x00-\x1F\x7F()<>@,;:[]={} \t\\"]')
+HEADER_VALUE_RE = re.compile(b"[\x00-\x1F\x7F]")
+
+
+def _get_status_line(status_code: int) -> bytes:
+    try:
+        phrase = http.HTTPStatus(status_code).phrase.encode()
+    except ValueError:
+        phrase = b""
+    return b"".join([b"HTTP/1.1 ", str(status_code).encode(), b" ", phrase, b"\r\n"])
+
+
+STATUS_LINE = {
+    status_code: _get_status_line(status_code) for status_code in range(100, 600)
+}
+
+
+class HttparseProtocol(asyncio.Protocol):
+    def __init__(
+        self,
+        config: Config,
+        server_state: ServerState,
+        _loop: Optional[asyncio.AbstractEventLoop] = None,
+    ) -> None:
+        if not config.loaded:
+            config.load()
+
+        self.config = config
+        self.app = config.loaded_app
+        self.loop = _loop or asyncio.get_event_loop()
+        self.logger = logging.getLogger("uvicorn.error")
+        self.access_logger = logging.getLogger("uvicorn.access")
+        self.access_log = self.access_logger.hasHandlers()
+        self.parser = httparse.RequestParser()
+        self.ws_protocol_class = config.ws_protocol_class
+        self.root_path = config.root_path
+        self.limit_concurrency = config.limit_concurrency
+
+        # Timeouts
+        self.timeout_keep_alive_task: Optional[TimerHandle] = None
+        self.timeout_keep_alive = config.timeout_keep_alive
+
+        # Global state
+        self.server_state = server_state
+        self.connections = server_state.connections
+        self.tasks = server_state.tasks
+
+        # Per-connection state
+        self.transport: asyncio.Transport = None  # type: ignore[assignment]
+        self.flow: FlowControl = None  # type: ignore[assignment]
+        self.server: Optional[Tuple[str, int]] = None
+        self.client: Optional[Tuple[str, int]] = None
+        self.scheme: Optional[Literal["http", "https"]] = None
+        self.pipeline: Deque[Tuple[RequestResponseCycle, ASGI3Application]] = deque()
+
+        # Per-request state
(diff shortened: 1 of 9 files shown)


Kludex:
> I guess the next step could be to contribute this to [Kludex/uvicorn-extensions](https://github.com/Kludex/uvicorn-extensions/tree/main/src/python/uvicorn-httparse?rgh-link-date=2022-11-13T11%3A39%3A50Z)? This brings questions about making sure the test suite of uvicorn-extensions matches up to the extensive checks we do in "main" uvicorn. (Here we add httparser as an extra pytest parameter to test_http.py, but in extensions test_http.py would have to be copied over.)

Not sure if I get 100% the message, but I've created a fixture to include all the extensions:
https://github.com/Kludex/uvicorn-extensions/blob/201aec4c3b38b896f16a712cc9c0ebf0775b2de2/tests/conftest.py#L12-L17

Which I think we can back port that idea to `uvicorn` itself.

uvicorn/protocols/http/httparse_impl.py
```diff
@@ -0,0 +1,616 @@
...
+            self.cycle.more_body = False
+            self.cycle.message_event.set()
+
+        self._parsed = None
+
+    def _attempt_to_parse_request(self) -> None:
+        try:
+            parsed = self.parser.parse(self._buffer)
+        except httparse._httparse.ParsingError:
+            return
```
Kludex: Aren't we suppose to do something if there's an error when parsing the headers? I guess this is not partial, as partial results is parsed being `None`, right?

florimondmanca: Yes, a partial (non-invalid) request will lead to `None`.

Maybe I should add a comment here. We don't do anything special here, because that leaves `_parsed` as `None`. Then, when `eof_received()` is called, it sees that `_parsed is None`, we surely means we didn't get any valid request — and that's where we send the 400 Bad Request response. This also covers the case when no data was sent at all, or the client sent a partial request and never finished it. So I think it's the most complete way to do things. But I agree this "just `return`" line here prompts questions... I'll defo add a comment.

florimondmanca:
> Which I think we can back port that idea to uvicorn itself.

Actually I was not sure what the status of `uvicorn-extensions` was. Is it a temporary repo to figure out some mechanism we will eventually bring into uvicorn? https://github.com/encode/uvicorn/discussions/1762 mentions `uvicorn.experimental` but then we created `uvicorn-extensions`, so I got confused. 😄 

If what we're targeting is indeed an `uvicorn.experimental` namespace, then that means all "extension" code will live in `uvicorn`, allowing us to leverage the existing test suite, which sounds perfect and addresses my questions in that last paragraph.

Kludex:
> Actually I was not sure what the status of uvicorn-extensions was. Is it a temporary repo to figure out some mechanism we will eventually bring into uvicorn? https://github.com/encode/uvicorn/discussions/1762 mentions uvicorn.experimental but then we created uvicorn-extensions, so I got confused. smile

I can change the name hahaha

My idea was to release different "extensions" that implement a single "experimental" feature on each package, and also have a package that uses all the ideas together. The core of this is to understand what people are using, so we can track it via PyPI downloads.

Also, as the HTTP test suite is shared between the implementations, it avoids a lot of duplication work if it's in the same repository.

I'm going to implement an alert on `uvicorn-extensions` (or `uvicorn-experimental` :eyes:) to watch for `test_http.py` and `uvicorn/protocols/http` changes. But... I'm actively maintaining here, so the alert is really just to make sure.

> Actually I was not sure what the status of uvicorn-extensions was.

I've created it some days ago. It's WIP. I've implemented `asgi-trailers` (maybe out of scope... Still figuring it out what makes sense to be there), and `uvicorn-trailers` yesterday.

adriangb:
> * Rust's `httparse` doesn't do HTTP method checking. If that's correct, I think it's fair that the port doesn't do checks either. So I had to implement some basic checks myself.

That’s right. I kept it a 1:1 map of the original library.

> * `httparse` may raise errors in case of invalid HTTP content, such as `InvalidHttpVersion`, but those exceptions aren't exposed on the library. E.g. I had to do `except httparse._httparse.ParsingError` to handle the "invalid HTTP request" case.

Oh ya this is a TODO. It definitely should expose rich errors, I was just too lazy at first.

> * Not sure which types Rust's `httparse` uses, but `parsed.method` and `parsed.path` might be better as `bytes` rather than pre-decoded `str`. h11 uses bytes, and so does httptools.

This also comes from it being a 1:1 port. I’m not sure why the Rust httparse decided to return strings but it does.

Overall there’s a lot of things I’d do different if this library “knew” it was going to be used within an ASGI server: https://github.com/adriangb/httparse/pull/1

Kludex:
@florimondmanca I have some questions.

- Do you want to push this code to `uvicorn-extensions`? (I can also push it, if you allow me to)
- If not, do you want me to transfer the `uvicorn-httparse` PyPI package name?
- Is `uvicorn-extensions` a good name? :thinking:

I've organized a bit more the `uvicorn-extensions` repository, and I'll keep organizing it...

adriangb:
> I had to do except httparse._httparse.ParsingError to handle the "invalid HTTP request" case

Fixed in `0.1.2` (building)

florimondmanca:
@Kludex I'm happy to let anyone reuse the code in this PR. To be honest, I had some time to do some OSS this weekend, but that may not be the case starting next week. 😅 So feel free to port this code to `uvicorn-extensions`. Also yes, I think that name is good, actually.

Kludex:
I've released `uvicorn-httparse`, but I didn't benchmark it yet: https://pypi.org/project/uvicorn-httparse/

I'm closing this just to clean PRs. Further discussions can be swapped to https://github.com/Kludex/uvicorn-extensions, or even here. :eyes:

PR #1768, opened 2022-11-18
feat: compose already existing signal handling

JoanFM:
This is a draft PR to spark some discussion.

I would like UviServer to be able to not override `signal handlers` already installed, so I was thinking that perhaps overriding an already existing signal handler with a composition of the previous and the one needed by `UviCorn` could be a good idea.

I know that we are accessing a private element that not every event loop implements and so on, but I would like to have your opinion about this problem and potential solutions before moving forward.

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -292,9 +292,19 @@ class Server:
 
         loop = asyncio.get_event_loop()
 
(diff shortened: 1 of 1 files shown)


JoanFM:
Duplicated concern as https://github.com/encode/uvicorn/pull/1600

PR #1774, opened 2022-11-25
Remove `try..except`s from `Config.bind_socket()`

Kludex:
The idea here is the `try..except` is being redundant.

If an exception happens, the same will be raised, and the same `1` code will happen.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,7 +51,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 98.50
+fail_under = 98.60
 show_missing = true
 skip_covered = true
 exclude_lines =
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,7 +51,7 @@ plugins =
 
 [coverage:report]
 precision = 2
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -51,7 +51,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 98.50
+fail_under = 98.60
```
Kludex:
```suggestion
fail_under = 98.55
```

lovelydinosaur:
Okay. I'd double check the behaviour in the console if you start uvicorn in one terminal, and then attempt to start another instance with the same port. It's a pretty common thing to bump into, so it'd be worth making sure what the UX is for this particular case.

Eg... do we end up with a stack trace after this change, but a plain error message before this change? Or is the behaviour the same?

Kludex:
Ok. Actually, I prefer the way it was before this PR. It was my bad, it's not redundant.

I'm going to close this. Thanks @tomchristie :+1:

PR #1786, opened 2022-11-30
add pyyaml to reqs

r0fls:
When using the `--log-config` flag, uvicorn fails if pyyaml is not installed:

```
File "/usr/local/lib/python3.9/site-packages/uvicorn/config.py", line 411, in configure_logging
    loaded_config = yaml.safe_load(file)
NameError: name 'yaml' is not defined
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,6 +10,9 @@ wsproto==1.2.0
 build==0.8.0
 twine==4.0.1
 
+# Application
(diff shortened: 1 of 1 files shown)


Kludex:
This is mentioned on the first page of our documentation.

If you need you can either install uvicorn with standard extras i.e. `pip install uvicorn [standard]`, or just install pyyaml separately.

PR #1794, opened 2022-12-03
Fix one example in REAME

Et7f3:
error: List item 0 has incompatible type "List[bytes]"; expected "Tuple[bytes, bytes]"  [list-item]

Linked #998

I just spotted this example because I used it. Maybe other example contains typing mistake. This one should be fine at runtime but the type is more wide so I just fixed it. I didn't typyfied the exemple completely but can if you want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ async def app(scope, receive, send):
         'type': 'http.response.start',
         'status': 200,
         'headers': [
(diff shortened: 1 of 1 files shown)


Kludex:
Thanks @Et7f3 :pray:

Kludex (approved):

PR #1804, opened 2022-12-11
Use unused TCP ports on the test suite

Kludex:
There are two goals here:
1. The test suite doesn't care about other applications running on the same machine.
2. Be able to run the tests concurrently.

For 2, there's another step that needs to be done, which I don't intend to suggest yet, but people can already install `pytest-xdist`, and run `pytest -n auto`. On my machine, I'm able to run the test suite in 6 seconds, instead of the normal 53 seconds.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,6 @@
+import contextlib
 import os
+import socket
 import ssl
 from copy import deepcopy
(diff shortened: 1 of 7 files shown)


lovelydinosaur (approved):
Okey dokes.

PR #1807, opened 2022-12-14
Remove MacOS pipeline

Kludex:
The MacOS pipeline is 10x more expensive than Linux. See on https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers, and on the image below.

image

This affects the organization, and the people forking `uvicorn` - I run a lot of pipelines on a fork, and exhausted my free plan. :cry:

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        os: [windows-latest, ubuntu-latest]
     steps:
       - uses: "actions/checkout@v3"
       - uses: "actions/setup-python@v4"


iudeen:
What works on Linux should technically (most of the time) work on mac as well right? 

Further adding to discussion:

GitHub says there are no limitations on public repositories right? Is it not so?🤔

Also, is macOS computationally more power demanding? If so, by removing we also contribute towards sustainable software development 😁♻️

zanieb:
@iudeen I develop on macOS (an unfortunate constraint of my job) and run CI on Linux and there are a lot of differences :/ I'm not sure how often Uvicorn encounters them, but they definitely exist.

iudeen:
@madkinsz oh, I was not aware of it. I have not developed on macOS. Thanks for pointing out 🙂

Kludex:
Hmm... So, the problem is that I'm running the pipeline on a private fork...

I did solve all issues on MacOS when I had one, and I propose to add it to the pipeline. The issues were mainly related to the size of the file path, which has a smaller limitation than the other OSs. Now I don't have a Mac anymore (🙏), and since the issues on the test suite were related mainly to the issue I described, I think it's fine to remove it.

Also, if someone creates a PR to fix issues on the test suite for MacOS, I don't see any problem on accepting it. 👀

zanieb:
I agree it seems fine to remove.

If you want an alternative, you could cancel the macOS run unless a pull request has the label "macos" (or similar) attached to reduce the frequency of runs while keeping it easy to toggle on where needed.

euri10:
most of the differences between os we have in the codebase iirc are related to

1. sockets creation, defaults sometimes are different 
2. process (with the reloader / workers) forks
3. loops, they're different a least in windows 

I think just for that we should keep it, I cant remember the specifics, but macos didnt deal with forked processes the same way , etc..

Kludex:
Fair enough.

PR #1813, opened 2022-12-20
tests: test to start server with invalid host

iudeen:
Covers `server.py` lines 160-163

Before PR (Ubuntu):
```
Name                                           Stmts   Miss     Cover   Missing
-------------------------------------------------------------------------------
uvicorn/server.py                                171     22    87.13%   115-125, 141, 160-163, 245-247, 263, 274-277, 307-310
-------------------------------------------------------------------------------
TOTAL                                           5102     67    98.69%
```

After PR (Ubuntu):
```
Name                                           Stmts   Miss     Cover   Missing
-------------------------------------------------------------------------------
uvicorn/server.py                                171     18    89.47%   115-125, 141, 245-247, 263, 274-277, 307-310
-------------------------------------------------------------------------------
TOTAL                                           5110     63    98.77%
```

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -117,9 +117,9 @@ def test_run_match_config_params() -> None:
 
 
 @pytest.mark.anyio
-async def test_run_invalid_host(caplog: pytest.LogCaptureFixture) -> None:
-    with pytest.raises(SystemExit):
+async def test_exit_on_create_server_with_invalid_host() -> None:
+    with pytest.raises(SystemExit) as exc_info:
         config = Config(app=app, host="illegal_host")
         server = Server(config=config)
         await server.serve()
-    assert "ERROR" == caplog.records[-1].levelname
+    assert exc_info.value.code == 1
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -122,4 +122,4 @@ async def test_run_invalid_host(caplog: pytest.LogCaptureFixture) -> None:
         config = Config(app=app, host="illegal_host")
         server = Server(config=config)
         await server.serve()
-    assert "[Errno 11001] getaddrinfo failed" in caplog.records[-1].message
+    assert "ERROR" == caplog.records[-1].levelname
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -6,6 +6,7 @@ import httpx
 import pytest
 
 from tests.utils import run_server
+from uvicorn import Server
 from uvicorn.config import Config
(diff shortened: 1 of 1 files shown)


tests/test_main.py
```diff
@@ -113,3 +114,12 @@ def test_run_match_config_params() -> None:
...
     assert config_params == run_params
+
+
+@pytest.mark.anyio
+async def test_run_invalid_host(caplog: pytest.LogCaptureFixture) -> None:
+    with pytest.raises(SystemExit):
+        config = Config(app=app, host="illegal_host")
+        server = Server(config=config)
+        await server.serve()
+    assert "ERROR" == caplog.records[-1].levelname
```
iudeen: Reviewer,

The error message is different on different OSes. I asserted the log level instead, is this okay?

Kludex: What's the goal of this test?

Test that `uvicorn` exits on invalid host?

iudeen: Yes. That is the purpose.

Kludex:
```suggestion
```

tests/test_main.py
```diff
@@ -113,3 +114,12 @@ def test_run_match_config_params() -> None:
         if key not in ("app_dir",)
     }
     assert config_params == run_params
+
+
+@pytest.mark.anyio
+async def test_run_invalid_host(caplog: pytest.LogCaptureFixture) -> None:
```
Kludex:
```suggestion
async def test_exit_on_invalid_host() -> None:
```

tests/test_main.py
Kludex: Can you also add an assertion on the `code` from the `SystemExit`? :pray:

e.g.

```python
assert exc.code == 1
```

Kludex (changes requested):

tests/test_main.py
iudeen: I added this based on your suggestion here, want me to remove still? https://github.com/encode/uvicorn/pull/1757#discussion_r1033017140

Kludex: Hold on... Then is the goal of the test really to test that uvicorn exits on an invalid host, or the goal was to "test that we send an error message on an invalid host"?

iudeen: I would be happy to keep it to test if Uvicorn exits properly (with intended error code). Uvicorn doesn't send a standard error messages at this moment. Different OS gives different error message.

Kludex (approved):

PR #1816, opened 2022-12-26
Send code 1012 on shutdown for websockets

### LINKED ISSUE #1344: Incorrect websocket close code on server shutdown
### Checklist

### Describe the bug

This seems related to #1140 but may be a simpler issue.

When the server is shutdown with `SIGTERM`, websocket connections are terminated with a close code of `1006`. This is seen on both the server and the client.

According to [RFC 6455](https://www.rfc-editor.org/rfc/rfc6455.html), `1006` indicates an abnormal connection termination (no `Close` control frame) and `1001` indicates `Going Away`.

1. Create `websocket_test.py` with:
    ```
    class App:
        def __init__(self, scope, receive, send):
            self.scope = scope
            self.receive = receive
            self.send = send
    
        def __await__(self):
            return self.asgi().__await__()
    
        async def asgi(self):
            while True:
                message = await self.receive()
                message_type = message["type"].replace(".", "_")
                print(message)
                handler = getattr(self, message_type, None)
                if handler is not None:
                    await handler(message)
                if message_type == "websocket_disconnect":
                    break
    
        async def websocket_connect(self, message):
            await self.send({"type": "websocket.accept"})
    ```
2. Run `uvicorn --ws-ping-interval 5 --ws-ping-timeout 5 --log-level debug --lifespan off websocket_test:App`
3. Connect using a websocket client
4. Kill the server with `Ctrl+C` or sending a `SIGTERM` signal
5. Observe that both the server and client connections are closed with code `1006`.

### Expected behavior

uvicorn should send a `Close` frame and the client connection should be closed with code `1001`.

### Actual behavior

Client connection is closed with code `1006`.

### Debugging material

_No response_

### Environment

- uvicorn 0.17.1 with CPython 3.10.1 on Darwin
- `uvicorn --ws-ping-interval 5 --ws-ping-timeout 5 --log-level debug --lifespan off websocket_test:App`

### Additional context

When testing with daphne, the server close code is 1001 and the client close code is 1006.

Looking at the `WebSocketCommonProtocol`, there is a `close` method that properly implements the close handshake. However, it is not called from the `WebSocketProtocol` class on shutdown in uvicorn (although it is called as part of unwinding `handler` in `WebSocketServerProtocol`). Not calling `transport.close` from `shutdown` may solve the problem, but need to test it.

Kludex:
- Closes #1344

## Changes
- Send 500 or 1012 to the **client** on server shutdown, instead of the previous 1006. 
- Improve test suite to check what the client receives, instead of only the ASGI application.
- Increase coverage from 98.5% to 98.8%.

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -719,7 +719,6 @@ async def test_connection_lost_before_handshake_complete(
     response: typing.Optional[httpx.Response] = None
 
     async def websocket_session(uri):
-        # await websockets.client.connect(uri)
         nonlocal response
         async with httpx.AsyncClient() as client:
             response = await client.get(
@@ -747,7 +746,6 @@ async def test_connection_lost_before_handshake_complete(
         send_accept_task.set()
 
     task.cancel()
-
     assert response is not None
     assert response.status_code == 500, response.text
     assert response.text == "Internal Server Error"
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -1,4 +1,5 @@
 import asyncio
+import typing
 
 import httpx
 import pytest
@@ -713,11 +714,23 @@ async def test_connection_lost_before_handshake_complete(
         message = await receive()
         if message["type"] == "websocket.connect":
             await send_accept_task.wait()
-            await send({"type": "websocket.accept"})
         disconnect_message = await receive()
 
+    response: typing.Optional[httpx.Response] = None
+
     async def websocket_session(uri):
-        await websockets.client.connect(uri)
+        # await websockets.client.connect(uri)
+        nonlocal response
+        async with httpx.AsyncClient() as client:
+            response = await client.get(
+                f"http://127.0.0.1:{unused_tcp_port}",
+                headers={
+                    "upgrade": "websocket",
+                    "connection": "upgrade",
+                    "sec-websocket-version": "13",
+                    "sec-websocket-key": "dGhlIHNhbXBsZSBub25jZQ==",
+                },
+            )
 
     config = Config(
         app=app,
@@ -731,9 +744,13 @@ async def test_connection_lost_before_handshake_complete(
             websocket_session(f"ws://127.0.0.1:{unused_tcp_port}")
         )
         await asyncio.sleep(0.1)
-        task.cancel()
         send_accept_task.set()
 
+    task.cancel()
+
+    assert response is not None
+    assert response.status_code == 500, response.text
+    assert response.text == "Internal Server Error"
     assert disconnect_message == {"type": "websocket.disconnect", "code": 1006}
 
 
@@ -744,6 +761,7 @@ async def test_send_close_on_server_shutdown(
     ws_protocol_cls, http_protocol_cls, unused_tcp_port: int
 ):
     disconnect_message = {}
+    server_shutdown_event = asyncio.Event()
 
     async def app(scope, receive, send):
         nonlocal disconnect_message
@@ -755,10 +773,13 @@ async def test_send_close_on_server_shutdown(
                 disconnect_message = message
                 break
 
+    websocket: typing.Optional[websockets.client.WebSocketClientProtocol] = None
+
     async def websocket_session(uri):
-        async with websockets.client.connect(uri):
-            while True:
-                await asyncio.sleep(0.1)
+        nonlocal websocket
+        async with websockets.client.connect(uri) as ws_connection:
+            websocket = ws_connection
+            await server_shutdown_event.wait()
 
     config = Config(
         app=app,
@@ -773,7 +794,10 @@ async def test_send_close_on_server_shutdown(
         )
         await asyncio.sleep(0.1)
         disconnect_message_before_shutdown = disconnect_message
+    server_shutdown_event.set()
 
+    assert websocket is not None
+    assert websocket.close_code == 1012
     assert disconnect_message_before_shutdown == {}
     assert disconnect_message == {"type": "websocket.disconnect", "code": 1012}
     task.cancel()
--- a/uvicorn/protocols/websockets/websockets_impl.py
+++ b/uvicorn/protocols/websockets/websockets_impl.py
@@ -144,8 +144,11 @@ class WebSocketProtocol(WebSocketServerProtocol):
 
     def shutdown(self) -> None:
         self.ws_server.closing = True
-        if not self.transport.is_closing():
+        if self.handshake_completed_event.is_set():
             self.fail_connection(1012)
+        else:
+            self.send_500_response()
+        self.transport.close()
 
     def on_task_complete(self, task: asyncio.Task) -> None:
         self.tasks.discard(task)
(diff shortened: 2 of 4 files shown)


uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -145,7 +144,8 @@ def connection_lost(self, exc: Optional[Exception]) -> None:
 
     def shutdown(self) -> None:
         self.ws_server.closing = True
-        self.transport.close()
+        if not self.transport.is_closing():
```
Kludex: Do I need to check if the transport is closing? I don't think so...

Do I need to send 500 if handshake is not completed?

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -90,7 +90,6 @@ def __init__(
         self.connect_sent = False
         self.lost_connection_before_handshake = False
         self.accepted_subprotocol: Optional[Subprotocol] = None
-        self.transfer_data_task: asyncio.Task = None  # type: ignore[assignment]
```
Kludex, line 93: This is necessary to call `fail_connection()`, because internally there's a check like `if hasattr(self, "transfer_data_task")`, which should succeed.

uvicorn/protocols/websockets/websockets_impl.py
Kludex: No, I don't need to check if the transport is closing.

Yes, I do need to send 500 if handshake is not completed.

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -232,17 +232,14 @@ def send_500_response(self) -> None:
...
             )
-            output = self.conn.send(msg)
-            msg = h11.Data(data=b"Internal Server Error")
-            output += self.conn.send(msg)
-            msg = h11.EndOfMessage()
-            output += self.conn.send(msg)
+        )
+        output += self.conn.send(
+            wsproto.events.RejectData(data=b"Internal Server Error")
+        )
```
Kludex, line 241: Don't be afraid here. I'll explain what's happening.

1. We were sending `RejectConnection(status_code=500)` without a body, but on `websockets` implementation we were sending the "Internal Server Error" body on the analogous behavior. The `RejectData` matches the behavior.
2. We are removing the conditional because it's never reached, and the reason for it is that we only call `send_500_response` is we didn't complete the handshake - which makes a lot of sense to remove it.

Hope it's clear.

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -145,6 +144,10 @@ def connection_lost(self, exc: Optional[Exception]) -> None:
 
     def shutdown(self) -> None:
         self.ws_server.closing = True
+        if self.handshake_completed_event.is_set():
+            self.fail_connection(1012)
+        else:
+            self.send_500_response()
```
Kludex, line 150: This is the change that motivated this PR.

We were only sending 1006 to the client, even when the handshake was not completed.

tests/protocols/test_websocket.py
```diff
@@ -731,9 +743,12 @@ async def websocket_session(uri):
             websocket_session(f"ws://127.0.0.1:{unused_tcp_port}")
         )
         await asyncio.sleep(0.1)
-        task.cancel()
         send_accept_task.set()
 
+    task.cancel()
+    assert response is not None
+    assert response.status_code == 500, response.text
+    assert response.text == "Internal Server Error"
```
Kludex, line 751: We are now testing what the client receives, instead of only the ASGI application.

tests/protocols/test_websocket.py
```diff
@@ -731,9 +743,12 @@ async def websocket_session(uri):
...
     assert disconnect_message == {"type": "websocket.disconnect", "code": 1006}
```
Kludex, line 752: Hmm... On a second thought... This doesn't make sense, does it? Why are we even sending a `websocket.disconnect` when the handshake was not even completed? :thinking: 

I think this was on purpose because then the application could receive a `websocket` event, but thinking about it again, does it make sense?

Kludex: I need to check the issues/PRs about this decision. It shouldn't be a blocker for this PR tho.

uvicorn/protocols/websockets/websockets_impl.py
jalaziz: Why `self.fail_connection(1012)` instead of just calling `self.close(1012)`?

I've been look at `websockets` [server.close()](https://github.com/aaugustin/websockets/blob/e4fcab16a70344e356268e50dc0c0cf541920c5d/src/websockets/legacy/server.py#L780-L783) and [procotol.close()](https://github.com/aaugustin/websockets/blob/e4fcab16a70344e356268e50dc0c0cf541920c5d/src/websockets/legacy/protocol.py#L741-L796) to see why uvicorn isn't calling `close(), but haven't gotten terribly far yet.

Kludex: Because `close()` is async.

jalaziz: Ahh, missed the lack of async here.

Looking deeper at the difference, it _seems_ like the primary difference is `fail_connection` will proactively cancel the data transfer task but `close` will simply send the close frame and then wait.

Is there any consequence to that distinction here? My very, very basic testing seems to suggest with uvicorn tasks are given a chance to shutdown cleanly, but I'm still testing.

jalaziz:
Can confirm that `wscat`, Chrome, and Firefox all show the correct close codes. I also checked that the client receives the correct close code if an app sends `{"type": "websocket.close", "code": code}` and that works too.

~Just trying to figure out what's going on with the `websockets` client.~
~Figured it out, `websockets` is fine, it's just the `__main__.py` script doesn't show the correct code when the server initiates the disconnect.~
I'm an idiot and was accidentally using an testing with an older version of `websockets`. The latest version (10.4) shows the correct codes.

uvicorn/protocols/websockets/wsproto_impl.py
Kludex: Am I missing `EndOfMessage`? I need to check this.

tests/protocols/test_websocket.py
Kludex: According to https://github.com/django/asgiref/issues/364 it's fine.

uvicorn/protocols/websockets/wsproto_impl.py
Kludex: No, I'm not missing it. `wsproto` internally adds the `h11.EndOfMessage`. :pray:

uvicorn/protocols/websockets/websockets_impl.py
Kludex: Yeah, this code still waits for it to finish. :+1:

lovelydinosaur (approved):
So PRs like this do highlight to me how awkward it is to try to follow the flow when using protocols rather than streams.

Anyways.

Kludex:
> So PRs like this do highlight to me how awkward it is to try to follow the flow when using protocols rather than streams.
> 
> Anyways.

I agree. :shrug:

PR #1818, opened 2022-12-27
Introduce lifespan state

adriangb:


uvicorn/config.py
```diff
@@ -471,7 +473,9 @@ def load(self) -> None:
         else:
             self.ws_protocol_class = self.ws
 
-        self.lifespan_class = import_from_string(LIFESPAN[self.lifespan])
+        self.lifespan_class: Type[Union[LifespanOn, LifespanOff]] = import_from_string(
+            LIFESPAN[self.lifespan]
+        )
```
adriangb: This is just nice to have typing that avoids adding typing downstream in `server.py`

adriangb:
Needs more tests but we don't currently have any tests that test both lifespans and requests. I suppose this should be some sort of end to end test in `test_main.py`?

Kludex:
> Needs more tests but we don't currently have any tests that test both lifespans and requests. I suppose this should be some sort of end to end test in `test_main.py`?

Hmm... I think we can test it on `test_http.py` with an application that stores something on the state.

uvicorn/config.py
Kludex: We can't use the `Type[Lifespan]` here?

btw, the `LifespanOn` and `LifespanOff` are in the `TYPE_CHECKING`, but here they are not string annotated.

Kludex (commented):
As I told you, I think this is a great idea. :+1:

I'm just not sure about the new `lifespan` parameter on the HTTP protocol implementations. Maybe your previous idea about adding it to the `ServerState` may be better (just thinking here, do not change it yet).

Also, what about the WebSockets? The `state` is not included on the WebSocket scope?

adriangb:
I like having Lifespan be the thing creating it since it is “owned” by the lifespan. We could pass in `lifespan_state: Dict[str, Any]` instead of the Lifespan object itself. I don’t think it’s all that bad for the request/response lifecycle to have a reference to it’s corresponding lifespan.

uvicorn/config.py
adriangb: Changed/fixed

adriangb:
Added support for websockets and tests for http, websockets and the lifespan itself.

adriangb:
I think the main questions are:
- Who creates the lifespan state dict? Is it the "server" or the "lifespan"? There's (currently at least) a 1:1 relationship between server and lifespan, so I guess it could be in the server, but I think it fits better in the lifespan. Putting it in the lifespan would also allow a server to be started multiple times (even concurrently) which is I think a sign of the correct design (at least in theory, there's other stateful stuff in the server instances I think).
- What do we pass to the protocols? It could be the lifespan, the server state or just a dict. I implemented this with the lifespan, but I think just the dict would be reasonable as well.
- Do we need to include default values in the protos? They're technically public classes and adding a new required param to the constructor is a breaking change. But I don't think anyone uses Uvicorn by importing just a proto class, so I expect the close to 0% of users impacted aren't worth the extra line of code / logic branch to handle default values.

Kludex:
The creation of the dict should be on the lifespan class, but I don't agree with the lifespan parameter on the protocol classes. I think it can be passed through the ServerState. The way I see it, the protocol classes should avoid as much as possible ASGI related concepts.

After the startup runs the lifespan, we can pass the state to the server state object. Can't we?

---

On the test suite, you made assertions inside the app. That doesn't work, as every exception inside the `run_server` is not caught by pytest. You can store the values you want to assert from a nonlocal, and check at the end. If you search for "nonlocal" in the test suite, you'll find what I mean.

---

This is probably going to be an extension, so we need to provide the "lifespan.state" on the extensions on each scope on the protocol classes.

And if is not, then we'll probably need to bump the spec - which I doubt this will happen. 👀

adriangb:
I moved to passing in the already copied dict to the protocols. This actually fixed a bug (or at least non-intentional behavior) where the state would be double copied for websockets. So now protocols have no references to lifespans. I used the name `app_state` which I think it's appropriate: it's the state that the ASGI server is managing for the ASGI app.

Working on fixing the assertions, thank you for pointing that out.

adriangb:
We need to move this out of the extensions namespace into the top level namespace

tests/protocols/test_http.py
```diff
@@ -980,3 +994,34 @@ async def app(scope, receive, send):
     protocol.data_received(SIMPLE_GET_REQUEST)
     await protocol.loop.run_one()
     assert b"x-test-header: test value" in protocol.transport.buffer
+
+
+@pytest.mark.anyio
+@pytest.mark.parametrize("protocol_cls", HTTP_PROTOCOLS)
+async def test_lifespan_state(protocol_cls):
+    app = Response("Hello, world", media_type="text/plain")
+
```
```suggestion
```

tests/protocols/test_websocket.py
```diff
@@ -1065,3 +1065,56 @@ async def open_connection(url):
...
+        scope["state"]["b"] = [1]
+        await send({"type": "lifespan.startup.complete"})
+        message = await receive()
+        assert message["type"] == "lifespan.shutdown"
+        await send({"type": "lifespan.shutdown.complete"})
+
+    class App(WebSocketResponse):
+        async def websocket_connect(self, message):
+            expected_state = expected_states.pop(0)
+            assert self.scope["state"] == expected_state
```
Kludex: Assertions here shouldn't work...

Kludex (changes requested):

tests/protocols/test_websocket.py
adriangb: Why not?

uvicorn/protocols/http/h11_impl.py
```diff
@@ -68,6 +79,7 @@ def __init__(
         self,
         config: Config,
         server_state: ServerState,
+        app_state: Dict[str, Any],
```
adriangb, line 83: Since we already copied (via `.copy()`) should we rename this to `request_state` or something like that? To make it clear via the name that it is a copy of the app state.

tests/protocols/test_websocket.py
Kludex: > On the test suite, you made assertions inside the app. That doesn't work, as every exception inside the run_server is not caught by pytest. You can store the values you want to assert from a nonlocal, and check at the end. If you search for "nonlocal" in the test suite, you'll find what I mean.

uvicorn/protocols/http/h11_impl.py
Kludex: I prefer `app_state`. :+1:

adriangb: Maybe `copied_app_state`? No biggie if you prefer to just keep `app_state`

Kludex: Just `app_state` is fine. :pray:

tests/protocols/test_websocket.py
adriangb: done

tests/protocols/test_websocket.py
```diff
@@ -1087,3 +1088,60 @@ async def open_connection(url):
...
+@pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
+async def test_lifespan_state(ws_protocol_cls, http_protocol_cls, unused_tcp_port: int):
+    expected_states = [
+        {"a": 123, "b": [1]},
+        {"a": 123, "b": [1, 2]},
+    ]
+
+    actual_states = []
+
+    state_equals_expected_state = True
```
```suggestion
```

tests/protocols/test_websocket.py
```diff
@@ -1087,3 +1088,60 @@ async def open_connection(url):
...
+        scope["state"]["a"] = 123
+        scope["state"]["b"] = [1]
+        await send({"type": "lifespan.startup.complete"})
+        message = await receive()
+        assert message["type"] == "lifespan.shutdown"
+        await send({"type": "lifespan.shutdown.complete"})
+
+    class App(WebSocketResponse):
+        async def websocket_connect(self, message):
+            nonlocal state_equals_expected_state
```
```suggestion
```

Kludex: Is this being used? :thinking:

tests/protocols/test_websocket.py
Kludex: Since you removed the one below, I'm fixing the lint here.

Kludex (approved):

NargiT:
hello, I am here because of renovate bot. I try to understand what this lifespan state is for and didn't find anything relevant in uvicorn doc. maybe you could link the discussion/issue this refers to ?

Kludex:
> hello, I am here because of renovate bot. I try to understand what this lifespan state is for and didn't find anything relevant in uvicorn doc. maybe you could link the discussion/issue this refers to ?

You can read more about it here:
- https://www.starlette.io/lifespan/#lifespan-state
- https://github.com/django/asgiref/pull/354

PR #1822, opened 2022-12-28
Delete setup.py

Kludex:
It looks like this is not needed anymore. See FastAPI: https://github.com/tiangolo/fastapi.


iudeen (approved):
Yes, we don't need!

PR #1824, opened 2022-12-30
Add `--timeout-graceful-shutdown` parameter

### LINKED ISSUE #451: How do I cancel a streaming response on shutdown?

### LINKED ISSUE #675: Make force_exit configure-able from command line
Hello!

In development mode, with `--debug`, uvicorn waits for background tasks before reloading the server.

This is annoying if the server has background tasks in an infinite loop. You have to manually press Ctrl+C to reload code, which (I think) defeats the whole purpose of having an "automatic" reloader.

```
 $ uvicorn <app> --debug
INFO:     Started reloader process [84365] using statreload
INFO:     Started server process [84367]
.
.
.
.
WARNING:  Detected file change in '<file>'. Reloading...
INFO:     Shutting down
INFO:     Waiting for background tasks to complete. (CTRL+C to force quit) <--- ???
^CINFO:     Shutting down <--- Ctrl+C
```

---

As a workaround, I tried setting `force_exit = True`  manually and it seems to work great.

```python
config = uvicorn.Config("<app>", debug=True)
server = uvicorn.Server(config=config)

server.force_exit = True  # <---

sock = config.bind_socket()
supervisor = ChangeReload(config, target=server.run, sockets=[sock])
supervisor.run()
``` 

---

Happy to submit a PR adding this as a click option if you think this is a valid concern.

Thanks for this great project and others :)

Kludex:
- Closes https://github.com/encode/uvicorn/issues/451
- Closes https://github.com/encode/uvicorn/issues/675

## Description

This PR adds the `--timeout-graceful-shutdown` parameter, which sets a timeout on tasks that the server is waiting to perform the `shutdown` lifespan event.

This timeout doesn't act on the `shutdown` lifespan event i.e. the `shutdown` event is supposed to run even in the scenario the tasks are timed out.

I need to think of a nice test for this. Ideas?

## Notes

1. The `--graceful-timeout` from `gunicorn` is conceptually different from `--timeout-graceful-shutdown`. The first forcefully kills the worker, and the latter cancels the tasks that are running after `uvicorn` said "I'm not receiving more requests, I'm just going to finish some tasks". 

2. An alternative name for this parameter could be `--timeout-tasks-shutdown`, or similar.

3. The logs are horrible when a task is cancelled. Should it be fine tho? :shrug: 

## Missing

- [ ] Add an entry on the documentation, and mention the difference with `--graceful-timeout` from `gunicorn`.
- [ ] Add a test case.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -104,6 +104,9 @@ Options:
   --timeout-keep-alive INTEGER    Close Keep-Alive connections if no new data
                                   is received within this timeout.  [default:
                                   5]
+  --timeout-graceful-shutdown INTEGER
+                                  Maximum number of seconds to wait for
+                                  graceful shutdown.
   --ssl-keyfile TEXT              SSL key file
   --ssl-certfile TEXT             SSL certificate file
   --ssl-keyfile-password TEXT     SSL keyfile password
--- a/docs/index.md
+++ b/docs/index.md
@@ -171,6 +171,9 @@ Options:
   --timeout-keep-alive INTEGER    Close Keep-Alive connections if no new data
                                   is received within this timeout.  [default:
                                   5]
+  --timeout-graceful-shutdown INTEGER
+                                  Maximum number of seconds to wait for
+                                  graceful shutdown.
   --ssl-keyfile TEXT              SSL key file
   --ssl-certfile TEXT             SSL certificate file
   --ssl-keyfile-password TEXT     SSL keyfile password
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -240,6 +240,7 @@ class Config:
         backlog: int = 2048,
         timeout_keep_alive: int = 5,
         timeout_notify: int = 30,
+        timeout_graceful_shutdown: Optional[int] = None,
         callback_notify: Optional[Callable[..., Awaitable[None]]] = None,
         ssl_keyfile: Optional[str] = None,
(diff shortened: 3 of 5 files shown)


JonasKs (approved):
Reminds me of https://github.com/samuelcolvin/arq/pull/345 😁 I'd love to see this merged.   
As for a test, I agree, not an obvious pretty comes to mind.
I think I would've went for caplog test on this one personally.

Kludex:
The problem on this PR is that `run_asgi` task gets cancelled, and there's a horrible `asyncio.CancelledError` message.

Also, I guess we need to cancel the tasks (?) in case `TimeoutError` happens. :thinking:

Kludex:
- Close in favor of https://github.com/encode/uvicorn/pull/1950

Issue #177, opened 2018-08-15
Document Uvicorn as a WSGI & ASGI webserver.


lovelydinosaur:
Given that Uvicorn now has `--wsgi` support, we ought to change the documentation around a bit, and promote it as being suitable for either WSGI or ASGI.

This'll be a more obvious sell once we have HTTP/2 support, but it's probably already a decent alternative to gunicorn/uwsgi/waitress/bjeorn.

Kludex:
@tomchristie Do you still think we should promote Uvicorn as WSGI & ASGI webserver?

lovelydinosaur:
Not sure. I'd be open to switching up the documentation if anyone particularly wants to push for it, but I think it's probably just fine as it is.

PR #1825, opened 2022-12-31
Replace current WSGIMiddleware implementation by a2wsgi one

### LINKED ISSUE #371: WSGI middleware should stream request body, rather than loading it all at once.

humrochagf:
Hello!

This should fix issue #371
And this is also a continuation of PR #1303 

Edit by @Kludex:
- Closes #371
- Relates to #177
- Relates to #1303


humrochagf:
I kept most of wsgi tests as a backwards compatibility validation, and that's also why I kept the wsgi module, to avoid breaking someone importing it directly.

If that's not needed I can drop the module and the tests for it and move it inside the config as the previous PR was.

I also added a test for the missing module case since in #177 there was no adhesion to push towards wsgi feature so I understood it as an optional feature, that is part of the [standard] package and not the minimal one.

Kludex:
I know I said [this](https://github.com/encode/uvicorn/issues/371#issuecomment-1293105755), but did you check what is needed for us to just solve #371 without adding `a2wsgi` as optional dependency?

I'd like to explore that option if it's not that complicated.

---

btw, I invited you to encode. Thanks for all the help here. :+1:

humrochagf:
Hey, sure what is required for the fix is a rework of how the body is handled to stream the payload instead of loading it into memory.

@euri10 did this change in #1049 PR.

And thanks for inviting me to encode looking forward to help more with uvicorn :smiley:

Kludex:
> @euri10 did this change in https://github.com/encode/uvicorn/pull/1049 PR.

Hmm... That has a footprint that I believe it would be worth it just replacing to `a2wsgi`. Is all that needed?

> And thanks for inviting me to encode looking forward to help more with uvicorn :smiley:

:sunglasses: :+1:

humrochagf:
Yup, all there, also tested running a django/takahe instance replacing fully gunicorn by uvicorn and doing some image uploads and everything is running fine

Kludex:
I'm not sure if I was clear. I mean: is it possible to achieve a fix with less footprint?

humrochagf:
Oh okay, I can try some things, but for sure we need a body that streams content on demand, I'll write a small test scenario with a bigfile that is more reproducible than trying things manually and try to reduce the scope of this change.

humrochagf:
Yeah, there is not much that can be done to reduce footprint.

WSGI expects a file like structure in the `wsgi.input`, so we need to have that bridge between the async reader and the sync file like stream to deliver the body without preload the content in memory and explode the pod with big file uploads.

Another viable path to follow is a deprecation warning on wsgi interface + adding a pluggable interface option + docs on how to plug a2wsgi if maintaining wsgi isn't a focus of uvicorn.

Kludex:
I'll see this tomorrow. Thanks!

Feliz ano novo Humberto! 😁

humrochagf:
Alright!

Feliz ano novo pra vc tb :grin:

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
 
 [project.optional-dependencies]
 standard = [
+    "a2wsgi>=1.6.0,<2.0.0",
```
Kludex: Let's add `a2wsgi` as `uvicorn` dependency, and change the documentation as mentioned on https://github.com/encode/uvicorn/pull/1303#discussion_r803623891.

Notes about this decision:
- `a2wsgi` is [less than 15kB](https://pypi.org/project/a2wsgi/#files).
- `a2wsgi` is already the [recommended `WSGIMiddleware` on `Starlette`](https://github.com/encode/starlette/blob/master/starlette/middleware/wsgi.py).

Kludex (changes requested):
@abersheeran If you have time, can you check if this is fine?

pyproject.toml
abersheeran: Yeah, is fine. 😊

humrochagf: So the change will be the move from optional dependencies to the base dependecies, is that it?

Kludex: Yes, and to add on the documentation that Uvicorn is also a WSGI server.

humrochagf: Et voilà :slightly_smiling_face: 

Since `WSGIMiddleware` converts WSGI into ASGI I described it as an `ASGI first web server` what do you think?

Kludex:
I gave a bit more thought about this.

This is what I suggest us to do:
1. Don't add `a2wsgi` as a dependency. Instead, ask the user to download `a2wsgi` if needed.
2. Don't remove uvicorn's `WSGIMiddleware`, but deprecate it, mentioning that people should install `a2wsgi`.
3. Add a logic to check if `a2wsgi` is installed (just import), and if it's use it. If not, use the one from uvicorn.
4. Don't change documentation in regard to Uvicorn being a WSGI server.

The idea here is to not add another minimal dependency - as we always want to avoid this as much as possible - and to not break our users.

What do you think @humrochagf ?

humrochagf:
Sounds good 🙂

uvicorn/middleware/wsgi.py
```diff
@@ -188,3 +194,9 @@ def wsgi(self, environ: Environ, start_response: StartResponse) -> None:
         }
         self.send_queue.append(empty_body)
         self.loop.call_soon_threadsafe(self.send_event.set)
+
+
+try:
+    from a2wsgi import WSGIMiddleware  # type: ignore # noqa
+except ImportError:
+    pass
```
humrochagf: Not the most attractive solution, but it's the best option to deprecate it without touching the config module since for testing purposes we can't reload the config module import table without breaking unrelated tests that share the config module state.

It also keeps backward compatibility with someone importing it directly in code.

humrochagf:
I think it's all there, the only weird thing is that when running uvicorn command without a2wsgi I don't see the deprecation warning, but pytest detects it.

Is there any kind of filter to not log it in terminal?

**Update:** If I change the filter in my client code to show it, then I can see the deprecation warning:

```python
import warnings
warnings.simplefilter("always")
```

uvicorn/middleware/wsgi.py
```diff
@@ -75,6 +76,11 @@ def build_environ(
 
 class WSGIMiddleware:
     def __init__(self, app: WSGIApp, workers: int = 10):
+        warnings.warn(
+            '"WSGIMiddleware" is deprecated, you should switch '
+            "to a2wsgi (`pip install a2wsgi`).",
```
Kludex:
```suggestion
            'Uvicorn's native "WSGIMiddleware" is deprecated, you '
            "should switch to a2wsgi (`pip install a2wsgi`).",
```

uvicorn/middleware/wsgi.py
```diff
@@ -75,6 +76,11 @@ def build_environ(
 
 class WSGIMiddleware:
```
Kludex:
```suggestion
class _WSGIMiddleware:
```
Doing this, we can improve how our tests looks like, and also make the import below more meaningful.

uvicorn/middleware/wsgi.py
Kludex:
```suggestion
try:
    from a2wsgi import WSGIMiddleware  # type: ignore # noqa
except ModuleNotFound:
    WSGIMiddleware = _WSGIMiddleware
```

tests/middleware/test_wsgi.py
```diff
@@ -1,17 +1,38 @@
...
+            with mock.patch.dict(sys.modules, {"a2wsgi": None}):
+                reload(wsgi)
+
+                await async_test()
+
+            reload(wsgi)
+
+    return test_wrapper
+
+
```
Kludex: Now we do this, and avoid the `@environ_switcher` magic, and the individual parametrize on each test.

```suggestion
@pytest.fixtures(param=wsgi._WSGIMiddleware, a2wsgi.WSGIMiddleware)
def wsgi_middleware_class(request):
    return request.param
```

tests/middleware/test_wsgi.py
```diff
@@ -34,7 +55,7 @@ def echo_body(environ: Environ, start_response: StartResponse) -> List[bytes]:
     return [output]
 
 
-def raise_exception(environ: Environ, start_response: StartResponse) -> RuntimeError:
+def raise_exception(environ: Environ, start_response: StartResponse) -> List[bytes]:
```
Kludex, line 38: I understand why you did this, but I don't think it's the right type. It should be `NoReturn` :thinking:

docs/index.md
```diff
@@ -460,6 +460,35 @@ async def app(scope, receive, send):
     })
 ```
 
+## The WSGI Support
+
+Uvicorn offer support to WSGI applications throught `WSGIMiddleware` which converts
+your WSGI into ASGI application seamlessly just by running it with `--interface wsgi`
```
Kludex: Is this the first mention to the `WSGIMiddleware` on the documentation? I think this is just an implementation detail.

docs/index.md
```diff
@@ -460,6 +460,35 @@ async def app(scope, receive, send):
...
+
+!!! warning
+    The present WSGIMiddleware will be deprecated in favour of [a2wsgi](https://github.com/abersheeran/a2wsgi).
+    If you are using Uvicorn to run WSGI applications, please add it as part of your project requirements.
```
Kludex: It will be **removed**, but this PR already deprecates it.
```suggestion
    The present WSGIMiddleware is deprecated in favour of [a2wsgi](https://github.com/abersheeran/a2wsgi).
    If you are using Uvicorn to run WSGI applications, please add it as part of your project requirements.
```

docs/index.md
```diff
@@ -460,6 +460,35 @@ async def app(scope, receive, send):
...
+```python
+def app(environ, start_response):
+    start_response('200 OK', [('Content-Type', 'text/plain')])
+    return [b'Hello World!']
+```
+
+Run the server:
+
+```shell
+$ uvicorn --interface wsgi example:app
```
Kludex: Is the `--interface` needed, or does uvicorn infers it? I never used WSGI with uvicorn. :thinking:

Kludex (changes requested):
I'm not so sure about the documentation.

Can we just add a small note on https://www.uvicorn.org/settings/#application-interface mentioning that the user should install `a2wsgi` for WSGI support, and that the Uvicorn's native WSGI integration is deprecated?

tests/middleware/test_wsgi.py
humrochagf: From the typechecker perspective WSGIMiddleware expects a function that has `List[bytes]` as return type, if it aways fails into exception is not that important.

This type error started to be more clear because the a2wsgi WSGIMiddleware is less permissive in terms of functions it accepts.

docs/index.md
humrochagf: It fails to infer it, maybe a bug but it falls into asgi2 case that also uses a function instead of a coroutine in the config code:
```python
...
            elif inspect.isfunction(self.loaded_app):
                use_asgi_3 = asyncio.iscoroutinefunction(self.loaded_app)
...
```

docs/index.md
humrochagf: Yup, updated the docs

docs/index.md
humrochagf: Also cut it and put just the mention in the settings like you said

humrochagf:
ok, I'll need to stop for today, tomorrow I'll fix what is missing

tests/middleware/test_wsgi.py
humrochagf: Done, much better now :smile:

uvicorn/middleware/wsgi.py
humrochagf: Updated as well :slightly_smiling_face:

humrochagf:
Alright!
I think I checked everything :slightly_smiling_face:

humrochagf:
Just saw your test simplification,  didn't knew fixtures could behave like parametrize

humrochagf:
Done, updated the tests with the code simplification

Kludex:
I'm on the phone, but I can comment some stuff.

There are two things that I think we can improve before merging:
1. Improve the warning message. I think the user shouldn't be aware that uvicorn has an entity called "WSGIMiddleware". Maybe something like "uvicorn's native WSGI implementation is deprecated in favor of a2wsgi package. Please install it with pip install a2wsgi." - or something like that.
2. We can specify which "# type: ignore[here]". There's an option on mypy to make this mandatory, that we have on starlette already. And the noqa I believe it can be removed? 🤔

Besides this, everything is cool here. Sorry for the round-trips here.

humrochagf:
Oh, no worries and all done :slightly_smiling_face: 

Had to ignore 2 kinds of type error codes:

```shell
uvicorn/middleware/wsgi.py:202: error: Cannot assign to a type  [misc]
uvicorn/middleware/wsgi.py:202: error: Incompatible types in assignment (expression has type "Type[_WSGIMiddleware]", variable has type "Type[WSGIMiddleware]")  [assignment]
```

Kludex (approved):
Thanks @humrochagf :pray: 

Feel free to merge it :)

humrochagf:
Yay, can I delete the branch after the merge as well?

Kludex:
It's in your fork anyway :eyes:

humrochagf:
Oh right,  the way the ui presented it I thought it had a branch copy in encode/uvicorn side 😜

PR #1835, opened 2023-01-05
Change default `--app-dir` from from "." (dot) to "" (empty string).

mila:
Both the dot and empty string are interpreted as the current working directory but Python, but the dot causes ugly `__file__` of imported modules. The empty string is consistent with Python behavior, inserting an empy string to sys.path when running interactively.

Discussion: https://github.com/encode/uvicorn/discussions/1814

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -119,7 +119,7 @@ Options:
   --version                       Display the uvicorn version and exit.
   --app-dir TEXT                  Look for APP in the specified directory, by
                                   adding this to the PYTHONPATH. Defaults to
-                                  the current working directory.  [default: .]
+                                  the current working directory.
   --h11-max-incomplete-event-size INTEGER
                                   For h11, the maximum number of bytes to
                                   buffer of an incomplete event.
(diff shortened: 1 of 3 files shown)


Kludex (changes requested):
Does `os.getcwd()` instead empty string has the same behavior here?

Also, can we add a test for it? Just to validate exactly what you said in the discussion. Ideally, the test would fail on master right now, and pass here.

mila:
> Does os.getcwd() instead empty string has the same behavior here?

The behavior would be different if the application changed cwd at runtime. Both `"."` and `""` mean the cwd at the time of import, `os.getcwd()` would freeze the value.

> Also, can we add a test for it? Just to validate exactly what you said in the discussion. Ideally, the test would fail on master right now, and pass here.

I was considering that and but I am not sure what I should test and whether it is worth it. 

Today, a [test](https://github.com/encode/uvicorn/blob/eec7d22ecc0c8926feb365ac7708918e32725f4f/tests/test_cli.py#L174) for `--app-dir` asserts that `sys.path` is modified. I can copy this test and assert that the empty string is inserted there. But that would just ensure that default value is honored by Click.

It would be safer to test that it is possible to import an app from `--app-dir` (empty or custom). But all CLI tests now use mocks, so I did not want to introduce a test that would run a real server.

Kludex (approved):
It's fine. :+1:

Thanks! :pray: 

Ref.: https://docs.python.org/3/library/sys.html#sys.path

Issue #1389, opened 2022-02-23
CHANGELOG.md versions have wrong markdown


euri10:
### Discussed in https://github.com/encode/uvicorn/discussions/1388

<div type='discussions-op-text'>

<sup>Originally posted by **HarrySky** February 23, 2022</sup>
Hey, just noticed, that Renovate can't parse your `CHANGELOG.md` since `0.17.4` version :sweat_smile:

You seem to have wrong title style applied

**Actual**:
```
### 0.17.5 - 2022-02-16
...
### 0.17.4 - 2022-02-04
```

**Expected**:
```
## 0.17.5 - 2022-02-16
...
## 0.17.4 - 2022-02-04
```</div>

PR #1836, opened 2023-01-06
Replace CHANGELOG.md by docs/release-notes.md

Kludex:
This PR adds the changelog to the documentation. It makes easier for users to check the PRs.

## Changes

- Replace `CHANGELOG.md` by `docs/release-notes.md`.
- Add links to PRs.


lovelydinosaur:
⚡️ *deflection* ⚡️

image

aminalaee:
What's the reasoning behind it? I think having a CHANGEOIG.md is still a Github standard.

Kludex:
> What's the reasoning behind it? I think having a CHANGEOIG.md is still a Github standard.

Complying with Starlette: https://www.starlette.io/

Can we have both worlds? In the docs, and the CHANGELOG.md? 🤔

br3ndonland:
> Can we have both worlds? In the docs, and the CHANGELOG.md? 🤔

Yes - if they are auto-generated. But with the manual copy/paste we do right now, I think it would be too error-prone.

In the future, maybe we should think about some ways to automate changelog maintenance. Maybe [scriv](https://scriv.readthedocs.io/en/latest/)?

br3ndonland:
One other thing to consider is how bots like Dependabot pick up release notes.

- Dependabot doesn't really have docs for this, but I _think_ as long as we have a file with "release" in the title, we are good (https://github.com/dependabot/dependabot-core/issues/2773#issuecomment-1227243377)?
- [Renovate looks for a "changelog" file, but it can pick up our GitHub Releases instead](https://docs.renovatebot.com/python/#how-it-works) (referencing re: encode/uvicorn#1389)

br3ndonland (approved):
Thanks @Kludex. Release notes look good in general - looks like you:

- Added full URLs to the PRs (works for me - we could also use [`pymdownx.magiclink`](https://facelessuser.github.io/pymdown-extensions/extensions/magiclink/) if we wanted to automate it)
- Removed the PR dates (they were a mix of mm/dd/yy and dd/mm/yy, so I agree, probably best to just remove those dates)
- Cleaned up some formatting

It would be helpful to mention this in the next release notes. I know contributing.md says "Things that should not go in the changelog: changes to documentation..." but maybe we should mention this, just so people don't think we stopped keeping a changelog.

It wouldn't hurt to put a link in the README so people can find the new release notes.

Kludex:
Given the comments, I'm not convinced about this anymore hahaha

Can I don't move the CHANGELOG.md, but add it to mkdocs? Or do I need to change the folder structure?

Kludex:
I'll close this for now. If someone knows the above, I'd like to hear more about it :eyes:

PR #1837, opened 2023-01-06
Specified unit of time for timeout keep alives

ghost:
We encountered an issue with how we assumed timeouts were handled and had to adjust this setting. Throughout the call stack, the unit of measure (minutes) was never specified so we hope this detail helps save others from some trial and error.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -110,4 +110,4 @@ connecting IPs in the `forwarded-allow-ips` configuration.
 
 ## Timeouts
 
-* `--timeout-keep-alive <int>` - Close Keep-Alive connections if no new data is received within this timeout. **Default:** *5*.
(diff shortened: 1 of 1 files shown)


euri10 (commented):
I'm pretty sure it is seconds, been a while though

ghost:
In testing the default setting of 5 by using a process that simply sleeps indefinitely, we get a timeout at 5 minutes. If I change to --timeout-keep-alive 2 , then we get a timeout at 2 minutes. We have found the setting to be in minutes. Either way, just trying to help clarify for others.

euri10:
I think the minutes unit is incorrect,
it is used here https://github.com/encode/uvicorn/blob/23b9f05a389a476aee151698cede171d915f6c53/uvicorn/protocols/http/httptools_impl.py#L332-L334
you can check docs of call_later, the delay is in seconds https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.call_later
if you experince effectively a 2minutes timeout setting it at 2 then we have a bug, one should try set a MRE to clarify this

PR #1845, opened 2023-01-12
Fix missing self.servers exception

Archmonger:
If another webserver is currently parked on uvicorn's port, then this exception will occur:

```
  File "C:\Users\Markg\Documents\Repositories\idom-docs\.venv\lib\site-packages\idom\backend\default.py", line 38, in serve_development_app
    return await _default_implementation().serve_development_app(
  File "C:\Users\Markg\Documents\Repositories\idom-docs\.venv\lib\site-packages\idom\backend\starlette.py", line 76, in serve_development_app
    await serve_development_asgi(app, host, port, started)
  File "C:\Users\Markg\Documents\Repositories\idom-docs\.venv\lib\site-packages\idom\backend\_common.py", line 53, in serve_development_asgi
    await asyncio.wait_for(server.shutdown(), timeout=3)
  File "C:\Users\Markg\AppData\Local\Programs\Python\Python310\lib\asyncio\tasks.py", line 445, in wait_for
    return fut.result()
  File "C:\Users\Markg\Documents\Repositories\idom-docs\.venv\lib\site-packages\uvicorn\server.py", line 258, in shutdown
    for server in self.servers:
AttributeError: 'Server' object has no attribute 'servers'
```

This PR ensures `self.servers` is always initialized so the error can be handled gracefully.

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -115,7 +115,7 @@ class Server:
 
             # Ensure list does not have stale items if the server is run multiple times.
             self.servers.clear()
-            
+
             for sock in sockets:
                 if config.workers > 1 and platform.system() == "Windows":
                     sock = _share_socket(  # type: ignore[assignment]
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -113,6 +113,9 @@ class Server:
                 sock_data = sock.share(os.getpid())  # type: ignore[attr-defined]
                 return fromshare(sock_data)
 
+            # Ensure list does not have stale items if the server is run multiple times.
+            self.servers.clear()
+            
             for sock in sockets:
                 if config.workers > 1 and platform.system() == "Windows":
                     sock = _share_socket(  # type: ignore[assignment]
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -10,7 +10,7 @@ import threading
 import time
 from email.utils import formatdate
 from types import FrameType
-from typing import TYPE_CHECKING, List, Optional, Sequence, Set, Tuple, Union
+from typing import TYPE_CHECKING, Any, List, Optional, Sequence, Set, Tuple, Union
 
 import click
 
@@ -54,7 +54,7 @@ class Server:
         self.should_exit = False
         self.force_exit = False
         self.last_notified = 0.0
-        self.servers = []
+        self.servers: List[Any] = []
 
     def run(self, sockets: Optional[List[socket.socket]] = None) -> None:
         self.config.setup_event_loop()
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -54,6 +54,7 @@ class Server:
         self.should_exit = False
         self.force_exit = False
         self.last_notified = 0.0
+        self.servers = []
 
     def run(self, sockets: Optional[List[socket.socket]] = None) -> None:
         self.config.setup_event_loop()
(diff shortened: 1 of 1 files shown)


Kludex:
What's your goal?

Archmonger:
To prevent an unexpected exception from occuring if something is taking the requested Uvicorn port during `startup`.

Now, uvicorn will now properly log that the port is busy during the self-activated `shutdown`.

Kludex:
Do you have a snippet that demonstrates what you mean?

uvicorn/server.py
```diff
@@ -112,7 +113,6 @@ def _share_socket(
                 sock_data = sock.share(os.getpid())  # type: ignore[attr-defined]
                 return fromshare(sock_data)
 
-            self.servers = []
```
zanieb, line 115: Might want to retain a `self.servers.clear()` so that the list does not have stale items if the server is run multiple times.

Archmonger:
Not exactly a snippet, but the process I've done to create this exception is:

1) Run another webserver on a specific port (ex. running `mkdocs serve`  on port 8000)
2) Run `uvicorn` on an arbitrary ASGI application using that specific port. In my case, I was utilizing [IDOM to run uvicorn](https://idom-docs.herokuapp.com/docs/guides/getting-started/running-idom.html#running-idom-in-debug-mode)

uvicorn/server.py
Archmonger: Should this occur at the end of the `shutdown` method?

Kludex:
I'd appreciate an MRE. :eyes:

Archmonger:
Sure I'll draft one in a bit

Kludex:
> Sure I'll draft one in a bit

Ok :eyes:

Archmonger:
Oops, thanks for the reminder.

MRE generated. See the readme for steps to reproduce.

https://github.com/Archmonger/uvicorn-server-exception

---

This issue can only be reproduced when `uvicorn` is run within asyncio.run([serve_development_asgi](https://github.com/idom-team/idom/blob/bd874072f0899dea72d9d1647d0c24c5e48dc724/src/idom/backend/_common.py#L28)) in `idom`.

With the changes in this PR, everything seems to be handled gracefully.

Kludex:
So it cannot be reproduced with uvicorn standalone?

Archmonger:
Not from my one quick test. 

[`serve_development_asgi`](https://github.com/idom-team/idom/blob/bd874072f0899dea72d9d1647d0c24c5e48dc724/src/idom/backend/_common.py#L28) seems to be a basic asyncio runner though.

Not entirely sure how that's causing this problem, but it does look like uvicorn's `Server.shutdown()` is making some unsafe assumptions that `self.servers` always exists. But `self.servers` will only exist if at least one successful `Server.startup()` has taken place.

Kludex:
This is the first time I'm seeing this issue, and it doesn't look like is reproducible against uvicorn standalone... Even if it's one line, I need a proof that the issue is here, and I can only have that if it's possible to reproduce with only uvicorn.

Archmonger:
No problem. I'm actually seeing that this issue is ultimately caused by [this LOC](https://github.com/idom-team/idom/blob/bd874072f0899dea72d9d1647d0c24c5e48dc724/src/idom/backend/_common.py#L53) in IDOM.

Although the variable safety issue still exists within `Server.shutdown`, it at least doesn't seem to be breaking things yet.

I will refactor the server running logic in IDOM.

PR #1848, opened 2023-01-16
Add no-lifespan worker class

niccolomineo:
I'd like to suggest a courtesy worker class with the lifespan disabled, proving useful for Django users in particular.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -269,6 +269,8 @@ gunicorn example:app -w 4 -k uvicorn.workers.UvicornWorker
 
 For a [PyPy][pypy] compatible configuration use `uvicorn.workers.UvicornH11Worker`.
 
+For a [Django][django] compatible configuration use `uvicorn.workers.UvicornNoLifespanWorker`.
+
 For more information, see the [deployment documentation](deployment.md).
 
 ### Application factories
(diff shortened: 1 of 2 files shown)


pauloxnet (approved):

Kludex:
Hi @niccolomineo,

Thanks for the PR!

I'd prefer to not have another custom worker, it's too simple to be created from the user's point of view.

And I think the documentation explains how to add a custom configuration for the worker? If not, PR welcome for that, if we can find a good spot in the docs.

Please don't hate me when you see me at PyCon IT. 🙏

niccolomineo:
Hi Marcelo, no hurt feelings! Do you believe it could still be useful to mention the no-lifespan config in a specific section of the docs / README to help Django users in dark?

Kludex:
I just read the docs, this part specifically: https://www.uvicorn.org/deployment/#gunicorn

Isn't it enough? Or what do you propose?

niccolomineo:
What about mentioning that turning off the lifespan is the way to have the worker class function properly with Django?

Kludex:
I think the docs are pretty agnostic in regard to web frameworks, wouldn't that make more sense to be on https://docs.djangoproject.com/en/4.1/howto/deployment/asgi/uvicorn/? :thinking:

pauloxnet:
Thanks for your useful feedback on this PR @Kludex and you will be always welcome at PyCon Italia 😄

niccolomineo:
> I think the docs are pretty agnostic in regard to web frameworks, wouldn't that make more sense to be on https://docs.djangoproject.com/en/4.1/howto/deployment/asgi/uvicorn/? :thinking: 

Sounds reasonable to me. Thank you.

PR #1842, opened 2023-01-10
Draft: asgiref tls extension

brussee:
This is fully based on: #1119 and rebased on master (23b9f05)
Added similar implementation for websockets (untested)

PR #1861, opened 2023-02-02
Update h11_Impl.py

siddharthakkapatruni:
This update enables client cert authentication in Uvicorn  Giving access to TCP layer enables the access to cert info


Kludex:
Thanks for the PR, but we would need to implement https://asgi.readthedocs.io/en/latest/specs/tls.html.

- Related to https://github.com/encode/uvicorn/pull/1842

PR #1863, opened 2023-02-05
Fix pyright issue on `uvicorn.run` strict mode

### LINKED ISSUE #1855: Type of "run" is partially unknown
When using VSCode with `pylance` in "strict" mode:

```python
from uvicorn import run
```
Warning is shown: Type of "run" is partially unknown

Kludex:
- Closes #1855

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -451,7 +451,7 @@ def main(
 
 
 def run(
-    app: typing.Union["ASGIApplication", typing.Callable, str],
+    app: typing.Union["ASGIApplication", typing.Callable[..., None], str],
     *,
     host: str = "127.0.0.1",
     port: int = 8000,
@@ -472,7 +472,7 @@ def run(
     reload_excludes: typing.Optional[typing.Union[typing.List[str], str]] = None,
     reload_delay: float = 0.25,
     workers: typing.Optional[int] = None,
-    env_file: typing.Optional[typing.Union[str, os.PathLike]] = None,
+    env_file: typing.Optional[typing.Union[str, os.PathLike[typing.Any]]] = None,
     log_config: typing.Optional[
         typing.Union[typing.Dict[str, typing.Any], str]
     ] = LOGGING_CONFIG,
@@ -488,7 +488,7 @@ def run(
     limit_max_requests: typing.Optional[int] = None,
     timeout_keep_alive: int = 5,
     ssl_keyfile: typing.Optional[str] = None,
-    ssl_certfile: typing.Optional[typing.Union[str, os.PathLike]] = None,
+    ssl_certfile: typing.Optional[typing.Union[str, os.PathLike[typing.Any]]] = None,
     ssl_keyfile_password: typing.Optional[str] = None,
     ssl_version: int = SSL_PROTOCOL_VERSION,
     ssl_cert_reqs: int = ssl.CERT_NONE,
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -472,7 +472,7 @@ def run(
     reload_excludes: typing.Optional[typing.Union[typing.List[str], str]] = None,
     reload_delay: float = 0.25,
     workers: typing.Optional[int] = None,
-    env_file: typing.Optional[typing.Union[str, os.PathLike[typing.Any]]] = None,
+    env_file: typing.Optional[typing.Union[str, "os.PathLike[typing.Any]"]] = None,
     log_config: typing.Optional[
         typing.Union[typing.Dict[str, typing.Any], str]
     ] = LOGGING_CONFIG,
@@ -488,7 +488,7 @@ def run(
     limit_max_requests: typing.Optional[int] = None,
     timeout_keep_alive: int = 5,
     ssl_keyfile: typing.Optional[str] = None,
-    ssl_certfile: typing.Optional[typing.Union[str, os.PathLike[typing.Any]]] = None,
+    ssl_certfile: typing.Optional[typing.Union[str, "os.PathLike[typing.Any]"]] = None,
     ssl_keyfile_password: typing.Optional[str] = None,
     ssl_version: int = SSL_PROTOCOL_VERSION,
     ssl_cert_reqs: int = ssl.CERT_NONE,


br3ndonland (commented):
If we want to support pylance, shouldn't we run [pyright in CI](https://github.com/microsoft/pyright/blob/main/docs/ci-integration.md) (the underlying type checker that powers pylance) with GitHub Actions?

I'm also not sure if we should expect to pass pyright strict mode, because we don't pass mypy strict mode either (we have tons of un-parametrized types, for example). If we want to avoid pyright errors, maybe we should add a [pyright config](https://github.com/microsoft/pyright/blob/main/docs/configuration.md) like we have for mypy, and match it as closely as possible with our mypy config.

Kludex:
My intention was not to comply with pyright. My goal was to specifically fix the problem mentioned on the issue for VSCode users.

michaeloliverx:
On a side note not related to this PR.

> because we don't pass mypy strict mode either

`httpx` now has mypy's [`--strict` mode enabled](https://github.com/encode/httpx/blob/562b4b40aa88dbe88e0eb5b814c51a80126bcd9e/setup.cfg#L7) so maybe this should also be a goal for `uvicorn` and or other projects in the org?

Kludex:
> On a side note not related to this PR.
> 
> > because we don't pass mypy strict mode either
> 
> `httpx` now has mypy's [`--strict` mode enabled](https://github.com/encode/httpx/blob/562b4b40aa88dbe88e0eb5b814c51a80126bcd9e/setup.cfg#L7) so maybe this should also be a goal for `uvicorn` and or other projects in the org?

Usually, Uvicorn and Starlette follow `httpx` decisions. But... I'm unsure if the effort would be worth it in Uvicorn. Maybe we can start removing the `type: ignore`s around... :shrug:

Kludex:
`os.PathLike` for old versions is not a generic... :eyes:

br3ndonland:
> My intention was not to comply with pyright. My goal was to specifically fix the problem mentioned on the issue for VSCode users.

That's fine, but then how do we know that this PR fixes the issue?

I'm not a pylance user, so I'll recuse myself and let someone else assess.

abersheeran:
> > My intention was not to comply with pyright. My goal was to specifically fix the problem mentioned on the issue for VSCode users.
> 
> That's fine, but then how do we know that this PR fixes the issue?
> 
> I'm not a pylance user, so I'll recuse myself and let someone else assess.

Maybe we can add pyright checks in CI. pyright tends to be smarter than mypy.

rafrafek:
> Maybe we can add pyright checks in CI. pyright tends to be smarter than mypy.

Is there a way to run pyright in strict mode with single flag e.g. `--strict`? In VS Code and Pylance it is easy to change type checking to strict with single option in settings. I don't know how to do it in CI.

abersheeran:
> > Maybe we can add pyright checks in CI. pyright tends to be smarter than mypy.
> 
> Is there a way to run pyright in strict mode with single flag e.g. ? In VS Code and Pylance it is easy to change type checking to strict with single option in settings. I don't know how to do it in CI.`--strict`

https://github.com/microsoft/pyright/blob/main/docs/configuration.md#main-pyright-config-options

strict [array of paths, optional]: Paths of directories or files that should use “strict” analysis if they are included. This is the same as manually adding a “# pyright: strict” comment. In strict mode, most type-checking rules are enabled. Refer to [this table](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#diagnostic-rule-defaults) for details about which rules are enabled in strict mode. Paths may contain wildcard characters ** (a directory or multiple levels of directories), * (a sequence of zero or more characters), or ? (a single character).

Kludex:
If we run mypy strict only on `main.py` it should probably also solve this issue (?)... :thinking:

rafrafek:
> > > Maybe we can add pyright checks in CI. pyright tends to be smarter than mypy.
> > 
> > 
> > Is there a way to run pyright in strict mode with single flag e.g. ? In VS Code and Pylance it is easy to change type checking to strict with single option in settings. I don't know how to do it in CI.`--strict`
> 
> https://github.com/microsoft/pyright/blob/main/docs/configuration.md#main-pyright-config-options
> 
> strict [array of paths, optional]: Paths of directories or files that should use “strict” analysis if they are included. This is the same as manually adding a “# pyright: strict” comment. In strict mode, most type-checking rules are enabled. Refer to [this table](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#diagnostic-rule-defaults) for details about which rules are enabled in strict mode. Paths may contain wildcard characters ** (a directory or multiple levels of directories), * (a sequence of zero or more characters), or ? (a single character).

I am aware of the description in the official docs, but I don't understand it. Does "optional" mean that the strict parameter is optional, or that the array of paths is optional? I want standard behaviour for paths, I don't want to set them manually. I just want the strict behaviour to be turned on for all the paths.

abersheeran:
> > > > Maybe we can add pyright checks in CI. pyright tends to be smarter than mypy.
> > > 
> > > 
> > > Is there a way to run pyright in strict mode with single flag e.g. ? In VS Code and Pylance it is easy to change type checking to strict with single option in settings. I don't know how to do it in CI.`--strict`
> > 
> > 
> > https://github.com/microsoft/pyright/blob/main/docs/configuration.md#main-pyright-config-options
> > strict [array of paths, optional]: Paths of directories or files that should use “strict” analysis if they are included. This is the same as manually adding a “# pyright: strict” comment. In strict mode, most type-checking rules are enabled. Refer to [this table](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#diagnostic-rule-defaults) for details about which rules are enabled in strict mode. Paths may contain wildcard characters ** (a directory or multiple levels of directories), * (a sequence of zero or more characters), or ? (a single character).
> 
> I am aware of the description in the official docs, but I don't understand it. Does "optional" mean that the strict parameter is optional, or that the array of paths is optional? I want standard behaviour for paths, I don't want to set them manually. I just want the strict behaviour to be turned on for all the paths.

In uvicorn, this option should be set to `["uvicorn/"]`

rafrafek:
> In uvicorn, this option should be set to `["uvicorn/"]`

Can you specify it as a CLI argument, without modifying any files?

Kludex:
We can follow `httpx`, and try to get a more strict configuration for mypy, but I think that adding pyright is something that at organization level. I don't think it will be my focus, but if someone is interested, I'm happy to discuss the plan, and review PRs. :+1:

If someone wants to fix this pipeline, and open this PR to not have issues on VSCode, I think we can merge it, it's not a big deal.

Issue #1862, opened 2023-02-03
DeflatePerMessage VS memory consumption


Kludex:
### Discussed in https://github.com/encode/uvicorn/discussions/1850

<div type='discussions-op-text'>

<sup>Originally posted by **M1ha-Shvn** January 27, 2023</sup>
Hi.
I'm devloping a server serving websockets connections using FastAPI.
I've noticed, that creating several thousands of simultinious websocket connections leads to high memory usage (4 Gb per 5-8k websocket connections in my case). I've started debugging it with [tracemalloc](https://docs.python.org/3/library/tracemalloc.html) and found out that the largest amount of this memory is consumed by websockets deflate extension in [this line](https://github.com/aaugustin/websockets/blob/main/src/websockets/extensions/permessage_deflate.py#L64).
After that I've digged into [websockets deflate mechanics](https://websockets.readthedocs.io/en/stable/topics/compression.html) and found out that it can be tuned wisely in order to achive lower memory consumption using custom `ServerPerMessageDeflateFactory`. I've tried searching for it in FastApi => Starlette => Uvicorn code and it lead me [here](https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/websockets/websockets_impl.py#L98).
What is the source of memory leak:
1. Uvicorn creates separate `PerMessageDeflate` instance for each websocket connection (using `ServerPerMessageDeflateFactory`). From my point of view, it is a disadvantageous behaviour: it would be much better if it is created not for each websocket connection, but for each combination of connection parameters (like [Singleton](https://betterprogramming.pub/what-is-a-singleton-2dc38ca08e92?gi=1632bfde6b86) pattern, but created for each combination of parameters. Something like [lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache)).
2. The only setting uvicorn gives to tune deflate is just disabling it with `--ws-per-message-deflate`. It is not flexible for different cases.</div>

Kludex:
@M1ha-Shvn I still don't have a clear view on how to solve the issue with the `1` that you mention above. Would you like to show me in a PR?

Kludex:
Also, do you have a MRE for me to confirm the issue?

M1ha-Shvn:
I'll try making MRE and PR a little bit later, no time today. What I can give you fast is the following log of tracemalloc. It has bin made in conditions:
1. --workers = 16
[...]

PR #1864, opened 2023-02-06
Optimize creating PerMessageDeflate

M1ha-Shvn:
[Issue](https://github.com/encode/uvicorn/issues/1862)


Kludex:
Based on what was discussed on #1862, it doesn't look like this is the solution we want.

Let's continue the discussion on the GH issue to see if it's really a problem.

PR #1881, opened 2023-02-28
Add type annotation on `test_logging.py`

Kludex:


michaeloliverx (approved):

tests/middleware/test_logging.py
```diff
@@ -104,7 +104,7 @@ async def websocket_app(scope, receive, send):
                 break
 
     async def open_connection(url):
-        async with websockets.connect(url) as websocket:
+        async with websockets.client.connect(url) as websocket:
```
euri10, line 107: why this change ?

Kludex: Because there's a lazy import logic on the `websockets.__init__.py`, and that kind prevents the type checker to infer that `connect` is actually part of it... So I need to go to the submodule that actually creates it - or ignore with `# type: ignore`.

PR #1890, opened 2023-03-07
Improve discoverability when `--port=0` is used

robbtraister:
It is possible to request the server to run on any available port. This already works by using `--port 0`. However, the resulting logs show `Uvicorn running on http://127.0.0.1:0 (Press CTRL+C to quit)`, making it very difficult to determine what port was selected.

This change will read the port number directly from the socket to ensure the bound port is displayed in all cases.


Kludex:
Thanks! :pray:

Kludex (approved):

PR #1895, opened 2023-03-12
config-refactoring

Voldemat:
Refactoring uvicorn/config.py into separate files in uvicorn.config module. **NO BREAKING CHANGES**


Kludex:
Thanks, but I'd prefer to not do this.

PR #1906, opened 2023-03-17
hoist timeout notify config option into main/cli args

timkpaine:


Kludex:
I think the reason is not there it's because is only for Gunicorn 🤔

timkpaine:
🤷‍♂️

PR #1907, opened 2023-03-19
Support websocket.http.response ASGI extension

kristjanvalur:
uvicorn now can return a response from websocket apps that choose to reject a connection with a custom response.
See https://github.com/encode/starlette/pull/2041/files for related work.

Note: The `websockets` module currently does not read the response body from a websockets upgrade requests, so we cannot effectively test that functionality, in the unittests.  The exception thrown at the client contains no body, only status code.

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -51,6 +51,21 @@ class WebSocketResponse:
                 break
 
 
+async def wsresponse(url):
+    """
+    A simple websocket connection request and response helper
+    """
+    url = url.replace("ws:", "http:")
+    headers = {
+        "connection": "upgrade",
+        "upgrade": "websocket",
+        "Sec-WebSocket-Key": "x3JJHMbDL1EzLkh9GBhXDw==",
+        "Sec-WebSocket-Version": "13",
+    }
+    async with httpx.AsyncClient() as client:
+        return await client.get(url, headers=headers)
+
+
 @pytest.mark.anyio
 @pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
 @pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
@@ -967,11 +982,9 @@ async def test_server_reject_connection_with_response(
         disconnected_message = await receive()
 
     async def websocket_session(url):
-        with pytest.raises(websockets.exceptions.InvalidStatusCode) as exc_info:
-            async with websockets.client.connect(url):
-                pass  # pragma: no cover
-        assert exc_info.value.status_code == 400
-        # Websockets module currently does not read the response body from the socket.
+        response = await wsresponse(url)
+        assert response.status_code == 400
+        assert response.content == b"goodbye"
 
     config = Config(
         app=app,
@@ -1022,11 +1035,9 @@ async def test_server_reject_connection_with_multibody_response(
         disconnected_message = await receive()
 
     async def websocket_session(url):
-        with pytest.raises(websockets.exceptions.InvalidStatusCode) as exc_info:
-            async with websockets.client.connect(url):
-                pass  # pragma: no cover
-        assert exc_info.value.status_code == 400
-        # Websockets module currently does not read the response body from the socket.
+        response = await wsresponse(url)
+        assert response.status_code == 400
+        assert response.content == (b"x" * 10) + (b"y" * 10)
 
     config = Config(
         app=app,
@@ -1068,10 +1079,9 @@ async def test_server_reject_connection_with_invalid_status(
         await send(message)
 
     async def websocket_session(url):
-        with pytest.raises(websockets.exceptions.InvalidStatusCode) as exc_info:
-            async with websockets.client.connect(url):
-                pass  # pragma: no cover
-        assert exc_info.value.status_code == 500
+        response = await wsresponse(url)
+        assert response.status_code == 500
+        assert response.content == b"Internal Server Error"
 
     config = Config(
         app=app,
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -1090,9 +1090,6 @@ async def test_server_reject_connection_with_invalid_status(
 async def test_server_reject_connection_with_invalid_msg(
     ws_protocol_cls, http_protocol_cls, unused_tcp_port: int
 ):
-    if ws_protocol_cls is WSProtocol:
-        pytest.skip("Cannot supporess asynchronously raised errors")
-
     async def app(scope, receive, send):
         assert scope["type"] == "websocket"
         assert "websocket.http.response" in scope["extensions"]
@@ -1107,12 +1104,8 @@ async def test_server_reject_connection_with_invalid_msg(
             "headers": [(b"Content-Length", b"0"), (b"Content-Type", b"text/plain")],
         }
         await send(message)
-        # send invalid message
-        try:
-            await send(message)
-        except Exception:
-            # swallow the invalid message error
-            pass
+        # send invalid message.  This will raise an exception here
+        await send(message)
 
     async def websocket_session(url):
         with pytest.raises(websockets.exceptions.InvalidStatusCode) as exc_info:
@@ -1124,6 +1117,8 @@ async def test_server_reject_connection_with_invalid_msg(
             # undo that, we will get the initial 404 response
             assert exc_info.value.status_code == 404
         else:
+            # websockets protocol sends its response in one chunk
+            # and can override the already started response with a 500
             assert exc_info.value.status_code == 500
 
     config = Config(
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -923,11 +923,48 @@ async def test_server_reject_connection(
         assert message["type"] == "websocket.disconnect"
 
     async def websocket_session(url):
-        try:
+        with pytest.raises(websockets.exceptions.InvalidStatusCode) as exc_info:
             async with websockets.client.connect(url):
                 pass  # pragma: no cover
-        except Exception:
-            pass
+        assert exc_info.value.status_code == 403
+
+    config = Config(
+        app=app,
+        ws=ws_protocol_cls,
+        http=http_protocol_cls,
+        lifespan="off",
+        port=unused_tcp_port,
+    )
+    async with run_server(config):
+        await websocket_session(f"ws://127.0.0.1:{unused_tcp_port}")
+
+from tests.response import Response
+
+@pytest.mark.anyio
+@pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
+@pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
+async def test_server_reject_connection_with_response(
+    ws_protocol_cls, http_protocol_cls, unused_tcp_port: int
+):
+    async def app(scope, receive, send):
+        assert scope["type"] == "websocket"
+        assert "websocket.http.response" in scope["extensions"]
+
+        # Pull up first recv message.
+        message = await receive()
+        assert message["type"] == "websocket.connect"
+
+        # Reject the connection with a response
+        response = Response(b"goodbye", status_code=400)
+        await response(scope, receive, send)
+        message = await receive()
+        assert message["type"] == "websocket.disconnect"
+
+    async def websocket_session(url):
+        with pytest.raises(websockets.exceptions.InvalidStatusCode) as exc_info:
+            async with websockets.client.connect(url):
+                pass  # pragma: no cover
+        assert exc_info.value.status_code == 400
 
     config = Config(
         app=app,
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -1087,6 +1087,59 @@ async def test_server_reject_connection_with_invalid_status(
 @pytest.mark.anyio
 @pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
 @pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
+async def test_server_reject_connection_with_invalid_msg(
+    ws_protocol_cls, http_protocol_cls, unused_tcp_port: int
+):
+    if ws_protocol_cls is WSProtocol:
+        pytest.skip("Cannot supporess asynchronously raised errors")
+
+    async def app(scope, receive, send):
+        assert scope["type"] == "websocket"
+        assert "websocket.http.response" in scope["extensions"]
+
+        # Pull up first recv message.
+        message = await receive()
+        assert message["type"] == "websocket.connect"
+
+        message = {
+            "type": "websocket.http.response.start",
+            "status": 404,
+            "headers": [(b"Content-Length", b"0"), (b"Content-Type", b"text/plain")],
+        }
+        await send(message)
+        # send invalid message
+        try:
+            await send(message)
+        except Exception:
+            # swallow the invalid message error
+            pass
+
+    async def websocket_session(url):
+        with pytest.raises(websockets.exceptions.InvalidStatusCode) as exc_info:
+            async with websockets.client.connect(url):
+                pass  # pragma: no cover
+        if ws_protocol_cls == WSProtocol:
+            # ws protocol has started to send the response when it
+            # fails with the subsequent invalid message so it cannot
+            # undo that, we will get the initial 404 response
+            assert exc_info.value.status_code == 404
+        else:
+            assert exc_info.value.status_code == 500
+
+    config = Config(
+        app=app,
+        ws=ws_protocol_cls,
+        http=http_protocol_cls,
+        lifespan="off",
+        port=unused_tcp_port,
+    )
+    async with run_server(config):
+        await websocket_session(f"ws://127.0.0.1:{unused_tcp_port}")
+
+
+@pytest.mark.anyio
+@pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
+@pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
 async def test_server_can_read_messages_in_buffer_after_close(
     ws_protocol_cls, http_protocol_cls, unused_tcp_port: int
 ):
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -938,7 +938,9 @@ async def test_server_reject_connection(
     async with run_server(config):
         await websocket_session(f"ws://127.0.0.1:{unused_tcp_port}")
 
+
 from tests.response import Response
+
 
 @pytest.mark.anyio
 @pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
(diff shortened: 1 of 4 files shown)


uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -269,49 +273,94 @@ async def send(self, message: "ASGISendEvent") -> None:
...
+                    body_finished = not message.get("more_body", False)
+                    event = events.RejectData(
+                        data=message["body"], body_finished=body_finished
+                    )
+                    output = self.conn.send(event)
+                    self.transport.write(output)
+                    if body_finished:
+                        self.queue.put_nowait(
+                            {"type": "websocket.disconnect", "code": 1006}
+                        )
```
Kludex, line 354: Why would you send this message?

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -269,49 +273,94 @@ async def send(self, message: "ASGISendEvent") -> None:
...
+                        self.handshake_complete = True
+                        self.close_sent = True
```
Kludex, line 356: I think you need to close the transport here, don't you?

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -278,43 +282,83 @@ async def asgi_send(self, message: "ASGISendEvent") -> None:
...
-                )
-                self.initial_response = None
-                self.accepted_subprotocol = cast(
-                    Optional[Subprotocol], message.get("subprotocol")
-                )
-                if "headers" in message:
-                    self.extra_headers.extend(
-                        # ASGI spec requires bytes
-                        # But for compatibility we need to convert it to strings
+            if self.response_body is None:
```
Kludex, line 285: Can't we remove this conditional and add `self.handshake_started_event.set()` on the `websocket.http.response.start`?

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -278,43 +282,83 @@ async def asgi_send(self, message: "ASGISendEvent") -> None:
...
+                    self.initial_response = (http.HTTPStatus.FORBIDDEN, [], b"")
+                    self.handshake_started_event.set()
+                    self.closed_event.set()
+
+                elif message_type == "websocket.http.response.start":
+                    message = cast("WebSocketResponseStartEvent", message)
+                    try:
+                        status = http.HTTPStatus(message["status"])
+                    except AttributeError:
+                        status = http.HTTPStatus.FORBIDDEN
```
Kludex: By the spec, `status` should always be provided.
```suggestion
```

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -278,43 +282,83 @@ async def asgi_send(self, message: "ASGISendEvent") -> None:
...
+                    headers = [
                         (name.decode("latin-1"), value.decode("latin-1"))
-                        for name, value in message["headers"]
+                        for name, value in message.get("headers", [])
+                    ]
+                    self.initial_response = (status, headers, b"")
```
Kludex, line 331:
```suggestion
                    self.initial_response = (message["status"], headers, b"")
```

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -278,43 +282,83 @@ async def asgi_send(self, message: "ASGISendEvent") -> None:
...
+                    self.response_body.append(message["body"])
+
+                    if not message.get("more_body", False):
+                        self.logger.info(
+                            '%s - "WebSocket %s" %i',
+                            self.scope["client"],
+                            get_path_with_query_string(self.scope),
+                            self.initial_response[0],
+                        )
+
```
Kludex: Access logs should be sent on `start` messages.
```suggestion
```

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -278,43 +282,83 @@ async def asgi_send(self, message: "ASGISendEvent") -> None:
...
+                        '%s - "WebSocket %s" 403',
+                        self.scope["client"],
+                        get_path_with_query_string(self.scope),
+                    )
+                    self.initial_response = (http.HTTPStatus.FORBIDDEN, [], b"")
+                    self.handshake_started_event.set()
+                    self.closed_event.set()
+
+                elif message_type == "websocket.http.response.start":
+                    message = cast("WebSocketResponseStartEvent", message)
```
Kludex, line 318:
```suggestion
                    message = cast("WebSocketResponseStartEvent", message)
                    self.logger.info(
                        '%s - "WebSocket %s" %d',
                        self.scope["client"],
                        get_path_with_query_string(self.scope),
                        message["status"],
                    )
```

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -269,49 +273,94 @@ async def send(self, message: "ASGISendEvent") -> None:
...
-                if self.config.ws_per_message_deflate:
-                    extensions.append(PerMessageDeflate())
-                if not self.transport.is_closing():
-                    self.handshake_complete = True
-                    output = self.conn.send(
-                        wsproto.events.AcceptConnection(
-                            subprotocol=subprotocol,
-                            extensions=extensions,
-                            extra_headers=extra_headers,
+            if not self.response_started:
```
Kludex, line 276: Can't we do the same as suggested for `websockets` i.e. remove this line from accept and close?

Kludex (dismissed):

uvicorn/protocols/websockets/wsproto_impl.py
kristjanvalur: yes, fixed.

uvicorn/protocols/websockets/wsproto_impl.py
kristjanvalur: This is the same logic as for the `websocket.close`.  The application, when it closes from its end (with `websocket.close` or `websocket.http.response.*` also gets a close message on its input queue.

uvicorn/protocols/websockets/websockets_impl.py
kristjanvalur: Ah, it is an `AttributeError` which happens if the lookup from `http.HTTPStatus` fails.  That is, if the use has provided
a code which isn't part of the enum.  This needs to be an enum, otherwise sending the response later will quietly fail.
I guess we could just allow the lookup here to fail, I was trying to be defensive.  Do you agree, or should we fallback to this code if an invalid code is provided?

uvicorn/protocols/websockets/websockets_impl.py
kristjanvalur: No, the status needs to be the enum value from http.HTTPStatus.  I spent an hour debugging why sending it this way would result in an error.  Turns out that the `websockets` library is finicky:

https://github.com/aaugustin/websockets/blob/ba1ed7a65cc876ff4e0fcd4dd4711402836475e2/src/websockets/legacy/server.py#L227-L230

The attribute error raised on `status.value` and `status.phrase` is later caught here and disappears:

https://github.com/aaugustin/websockets/blob/ba1ed7a65cc876ff4e0fcd4dd4711402836475e2/src/websockets/legacy/server.py#L250-L255

uvicorn/protocols/websockets/websockets_impl.py
Kludex: > This needs to be an enum

Does `websockets` package "doesn't work" if you provide the int `message["status"]`?

kristjanvalur: yes, see below.

tests/protocols/test_websocket.py
```diff
@@ -923,11 +924,99 @@ async def app(scope, receive, send):
...
+    config = Config(
+        app=app,
+        ws=ws_protocol_cls,
+        http=http_protocol_cls,
+        lifespan="off",
+        port=unused_tcp_port,
+    )
+    async with run_server(config):
+        await websocket_session(f"ws://127.0.0.1:{unused_tcp_port}")
+
```
Kludex: Unfortunately, exceptions in the `app` are ignored by the test suite, so even if you have an AssertionError, the test will never fail.

```suggestion
):
    disconnected_message = {}

    async def app(scope, receive, send):
        nonlocal disconnected_message
        assert scope["type"] == "websocket"
        assert "websocket.http.response" in scope["extensions"]

        # Pull up first recv message.
        message = await receive()
        assert message["type"] == "websocket.connect"

        # Reject the connection with a response
        response = Response(b"goodbye", status_code=400)
        await response(scope, receive, send)
        disconnected_message = await receive()

    async def websocket_session(url):
        with pytest.raises(websockets.exceptions.InvalidStatusCode) as exc_info:
            async with websockets.client.connect(url):
                pass  # pragma: no cover
        assert exc_info.value.status_code == 400
        # Websockets module currently does not read the response body from the socket.

    config = Config(
        app=app,
        ws=ws_protocol_cls,
        http=http_protocol_cls,
        lifespan="off",
        port=unused_tcp_port,
    )
    async with run_server(config):
        await websocket_session(f"ws://127.0.0.1:{unused_tcp_port}")

    assert disconnected_message == {"type": "websocket.disconnect", "code": 1006"}
```

tests/protocols/test_websocket.py
```diff
@@ -923,11 +924,99 @@ async def app(scope, receive, send):
...
+            "more_body": True,
+        }
+        await send(message)
+        message = {
+            "type": "websocket.http.response.body",
+            "body": b"y" * 10,
+        }
+        await send(message)
+        message = await receive()
+        assert message["type"] == "websocket.disconnect"
```
Kludex: Same as last comment.

uvicorn/protocols/websockets/websockets_impl.py
kristjanvalur: We need to somehow distinguish between the case when handshake is started with `websocket.http.response.start` as opposed to the others, because it sets us in a state where the only allowable messages are `websocket.http.response.body`. Otherwise, we cannot perform state validation.

We also cannot set the event until the body is ready, because we `await` this event (line 214) and expect any response to be ready when it is set.  This event really should be renamed, I could call it `handshake_done_event` if you agree.

I could however avoid the extra `if` clause (and indentation) and keep handling of all these handshake messages on the same level, but it would complicate the message validation logic.

uvicorn/protocols/websockets/wsproto_impl.py
Kludex: https://github.com/django/asgiref/issues/381

Kludex: What does `hypercorn` do? Let's follow their decision.

uvicorn/protocols/websockets/websockets_impl.py
kristjanvalur: this is done.

uvicorn/protocols/websockets/websockets_impl.py
kristjanvalur: done

uvicorn/protocols/websockets/websockets_impl.py
Kludex: Let's not silently replace it, tho... Let's add a log message here, please :+1:

uvicorn/protocols/websockets/wsproto_impl.py
kristjanvalur: Again, we need to somehow know that we got an initial `websocket.http.response.start` message and that the only allowed messages now are `websocket.http.response.body`.  Is there some other way we can signal this?  Or do you think we shouldn't validate the messages?

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -278,43 +282,82 @@ async def asgi_send(self, message: "ASGISendEvent") -> None:
...
+                    self.logger.info(
+                        '%s - "WebSocket %s" %s',
```
Kludex: Isn't it %d?
```suggestion
                        '%s - "WebSocket %s" %d',
```

kristjanvalur: That was your suggested change, but yes :)

Kludex: image

:sweat_smile:

uvicorn/protocols/websockets/websockets_impl.py
Kludex: Do you think we could benefit from another event to be used on line 214?

Kludex: if too much work, don't worry

uvicorn/protocols/websockets/websockets_impl.py
kristjanvalur: I must be getting confused. sorry.

uvicorn/protocols/websockets/websockets_impl.py
kristjanvalur: I think the only reason there is an `event` there is because a task is awaiting the handshake to be performed, i.e. it needs to wait until the app has either accepted the connection or delivered a rejection reponse.  An event makes sense for that.  for other internal state logic, regular `bool` members are sufficient.  But I do think that the name of the event, at least, is not clear, so if you approve, I will rename it to `handshake_done_event`, because it indicates that a connect message has been delivered to the app, and the app has responded.

kristjanvalur:
I don't see hypercorn sending any messages to the app:
https://github.com/pgjones/hypercorn/blob/8ae17ca68204d9718389fb3649ca0ed6ba851906/src/hypercorn/protocol/ws_stream.py#L322

uvicorn/protocols/websockets/websockets_impl.py
kristjanvalur: Adding a test for this case (trying to send a message with invalid status) showed that the wsproto will cause a 500 error to be raised.  So I think it is fine if we also get that error for websockets.  I'm removing the check and adding a test to verify that the server produces a 500 error.

Kludex: I'm fine with this compromise. Thanks :+1:

uvicorn/protocols/websockets/wsproto_impl.py
kristjanvalur: Hypercorn does not seem to send anything to the app when it closes the connection.
https://github.com/pgjones/hypercorn/blob/8ae17ca68204d9718389fb3649ca0ed6ba851906/src/hypercorn/protocol/ws_stream.py#L322

Kludex: Let's wait for Andrew's reply on the issue above, but to me, you shouldn't expect a message when sending a response. :thinking:

Kludex:
I don't have more comments here. Beautiful PR, beautiful tests. :+1:

Thanks @kristjanvalur :pray: 

For me, before approving:
- [ ] Check if all lines are covered.
- [ ] Check Andrew's reply to the disconnect message.

kristjanvalur:
Please see test `test_server_reject_connection_with_invalid_msg()`

I am unable to silence the wsproto internal error which happens when it is trying to send a 500 error.

1. wsproto will start a response as soon as it gets the `.start` message
2. It then gets an invalid message and attempts to send a 500 response
3. but it cannot because it has already started a response.
4. An exception is not raised via the async `send()` method but somewhere else which I cannot fully understand, so the test fails.

I've added a skip for the WSProto case, but maybe there is a better solution?
Skipping is not ideal, because this test was designed to increase coverage.

uvicorn/protocols/websockets/wsproto_impl.py
kristjanvalur: Yes, I kind of agree.  In case that turns out to be the case, should we also remove the message fore the `websocket.close` case?

Kludex: I thought I had asked that here: https://github.com/django/asgiref/issues/364

But it's actually a different case... :thinking: There was also a discussion here: https://github.com/django/asgiref/issues/340.

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -97,7 +97,7 @@ def __init__(
 
         # Connection events
         self.scope: WebSocketScope = None  # type: ignore[assignment]
-        self.handshake_started_event = asyncio.Event()
+        self.handshake_completed_event = asyncio.Event()
         self.handshake_completed_event = asyncio.Event()
```
Kludex: Eh... :thinking:

kristjanvalur: I see!  I didn't notice this one.  Let me take another look at this.  Silly me.

kristjanvalur: Undone.  sorry for the noise.

kristjanvalur:
A separate question:  The `websockets` library does not read the response body from a websocket upgrade request.  Do you think it is worth making a PR against that repo to add that functionality?

uvicorn/protocols/websockets/wsproto_impl.py
Kludex: Andrew replied. It's good to send the disconnect message. :+1:

uvicorn/protocols/websockets/websockets_impl.py
```diff
@@ -278,43 +282,81 @@ async def asgi_send(self, message: "ASGISendEvent") -> None:
...
+                if message_type == "websocket.http.response.body":
+                    message = cast("WebSocketResponseBodyEvent", message)
+                    self.response_body.append(message["body"])
+
+                    if not message.get("more_body", False):
+                        assert self.initial_response is not None
+                        self.initial_response = self.initial_response[:2] + (
+                            b"".join(self.response_body),
+                        )
+                        self.response_body = None
```
Kludex: This doesn't make any difference, does it? After the response is sent, the connection is closed.
```suggestion
```

Kludex:
> A separate question: The `websockets` library does not read the response body from a websocket upgrade request. Do you think it is worth making a PR against that repo to add that functionality?

That would be great :+1: 

First ask the maintainer if he would accept such a PR, please :eyes:

Kludex:
> Please see test `test_server_reject_connection_with_invalid_msg()`
> 
> I am unable to silence the wsproto internal error which happens when it is trying to send a 500 error.
> 
> 1. wsproto will start a response as soon as it gets the `.start` message
> 2. It then gets an invalid message and attempts to send a 500 response
> 3. but it cannot because it has already started a response.
> 4. An exception is not raised via the async `send()` method but somewhere else which I cannot fully understand, so the test fails.
> 
> I've added a skip for the WSProto case, but maybe there is a better solution? Skipping is not ideal, because this test was designed to increase coverage.

We should have the analogous behavior that we have when using `http.response.start` and `http.response.body`. What's the behavior with those events (I think they should be the same on httptools and h11)?

uvicorn/protocols/websockets/websockets_impl.py
kristjanvalur: No, it's just me not wanting to leave already processed data lying around.  I'll remove it.

kristjanvalur:
> We should have the analogous behavior that we have when using `http.response.start` and `http.response.body`. What's the behavior with those events (I think they should be the same on httptools and h11)?

Okay, I looked at the similar test for http (`test_duplicate_start_message`) and it, in fact, asserts that it **does not** get a 500 response, precisely because a 200 response has been previously written into the buffer.

for websockets, and the `websockets` protocol, we can't send the response incrementally.  Instead, we prepare it and return it all at once.  That means that we can in fact issue a 500 error in this case.

It is not a problem to modify the test for both cases.
The problem remains for me, how to run the test with wsproto, because it fails in some weird way.  I'll try some things.

kristjanvalur:
> It is not a problem to modify the test for both cases.
> The problem remains for me, how to run the test with wsproto, because it fails in some weird way. I'll try some things.

Turns out that we needed extra tests to check if response had been started before attempting to send a 500 response, similar to what is done in h11.

kristjanvalur:
seems like workflows are disabled 🤔

Kludex:
I cancelled a pipeline yesterday that it was not able to stop, maybe it's related to that. I don't know how to enable it again here.

kristjanvalur:
Weird, yes, the last time a workflow ran in this PR it timed out.

Kludex:
You can try to reopen a new PR. 🤔

PR #1584, opened 2022-07-29
Fix shutdown event on Windows in reloader

StarHeartHunt:
On Windows, using `Process.terminate()` will immediately kill the child process so that the asgi event shutdown cannot be triggered in reloader. This PR is going to implement a graceful exit of child process.

PR #1909, opened 2023-03-22
Handle `SIGBREAK` for Windows

yanyongyu:
On Windows, `subprocess.CTRL_BREAK_EVENT` may be sent to subprocess using `os.kill` or [`process.send_signal`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.send_signal) as `SIGBREAK` signal. This signal can be catched by subprocess to do a graceful shutdown instead of terminating the subprocess directly.  
And, this signal is also catched in `hypercorn`.

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -28,6 +28,8 @@ HANDLED_SIGNALS = (
     signal.SIGINT,  # Unix signal 2. Sent by Ctrl+C.
     signal.SIGTERM,  # Unix signal 15. Sent by `kill <pid>`.
 )
+if sys.platform == "win32":
+    HANDLED_SIGNALS += (signal.SIGBREAK,)  # Windows signal 21. Sent by Ctrl+Break.
 
 logger = logging.getLogger("uvicorn.error")
 
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -28,7 +28,7 @@ HANDLED_SIGNALS = (
     signal.SIGINT,  # Unix signal 2. Sent by Ctrl+C.
     signal.SIGTERM,  # Unix signal 15. Sent by `kill <pid>`.
 )
-if sys.platform == "win32":
+if sys.platform == "win32":  # pragma py-linux pragma: py-darwin
     HANDLED_SIGNALS += (signal.SIGBREAK,)  # Windows signal 21. Sent by Ctrl+Break.
 
 logger = logging.getLogger("uvicorn.error")
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -28,7 +28,7 @@ HANDLED_SIGNALS = (
     signal.SIGINT,  # Unix signal 2. Sent by Ctrl+C.
     signal.SIGTERM,  # Unix signal 15. Sent by `kill <pid>`.
 )
-if sys.platform == "win32":  # pragma py-linux pragma: py-darwin
+if sys.platform == "win32":  # pragma: py-not-win32
     HANDLED_SIGNALS += (signal.SIGBREAK,)  # Windows signal 21. Sent by Ctrl+Break.
 
 logger = logging.getLogger("uvicorn.error")


Kludex:
Is this an alternative to https://github.com/encode/uvicorn/pull/1584 ?

yanyongyu:
> Is this an alternative to #1584 ?

This pr may be related to #1584 but not an alternative. The reloader should do something else on Windows including send `CTRL_BREAK_EVENT`(or maybe `CTRL_C_EVENT`) to worker process and preventing `SIGBREAK`(or `CTRL_C_EVENT`) being handled in main process when reloading.

In our application, we use `subprocess` to run uvicorn now and `SIGBREAK` is sent to subprocess on Windows when shutting down (as `process.terminate` will call winapi to force exit). uvicorn uses `multiprocessing` to spawn workers, i'm not sure if the signal handling is the same. maybe @StarHeartHunt we need a further testing on uvicorn reloader.

Another problem is that `CTRL_C_EVENT` will be sent to all processes in the process group on Windows. We cannot use it as we have multiple subprocess workers running. We finally change to using [`subprocess.CREATE_NEW_PROCESS_GROUP`](https://docs.python.org/3/library/subprocess.html#subprocess.CREATE_NEW_PROCESS_GROUP) and [`signal.CTRL_BREAK_EVENT`](https://docs.python.org/3/library/signal.html#signal.CTRL_BREAK_EVENT).😓 

Reference: 
1. note in python docs: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.send_signal
2. our code: https://github.com/nonebot/nb-cli/blob/0221588fdc132891fbe7ccd8a8b15487380e722c/nb_cli/handlers/process.py#L56-L104
3. windows docs: https://learn.microsoft.com/en-us/windows/console/generateconsolectrlevent, https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags

StarHeartHunt:
As mentioned above, The Ctrl + Break signal is supposed to be handled by `uvicorn` as a kind of exit signal too, for external subprocess calls with `subprocess.CREATE_NEW_PROCESS_GROUP` to gracefully shutdown the server process, This is what this PR meant for.

And #1584 tries fixing the graceful exit in reloader when uvicorn it self is the root process, it seems in this case we can use Ctrl+C signal to achieve this goal (Maybe the internal code is different in multiprocessing?). The solution has been proved to be valid on my Windows 10 laptop. 

TL;DR: Based on current investigation and discussions in our team @nonebot, To implement a graceful exit:
1. `subprocess` with a `subprocess.CREATE_NEW_PROCESS_GROUP` flag or multiprocessing
2. The main process can only use `CTRL_BREAK_EVENT` as exit signal in subprocess situations, but multiprocessing could use `CTRL_C_EVENT`, and in both situations, the main process should implement a shield context to avoid being terminated by itself.
3. A signal (`CTRL_BREAK_EVENT` / `CTRL_C_EVENT`) is sent to the subprocess.

uvicorn/server.py
```diff
@@ -28,6 +28,8 @@
     signal.SIGINT,  # Unix signal 2. Sent by Ctrl+C.
     signal.SIGTERM,  # Unix signal 15. Sent by `kill <pid>`.
 )
+if sys.platform == "win32":
+    HANDLED_SIGNALS += (signal.SIGBREAK,)  # Windows signal 21. Sent by Ctrl+Break.
```
Kludex: Can you add the pragma ignore here for other OSs, please?

Kludex (changes requested):

Kludex:
Can you also show me the reference on hypercorn that you mentioned in the description, please?

yanyongyu:
> Can you also show me the reference on hypercorn that you mentioned in the description, please?

That's here: https://github.com/pgjones/hypercorn/blob/8ae17ca68204d9718389fb3649ca0ed6ba851906/src/hypercorn/run.py#L61-L63

By the way, `hypercorn` uses a multiprocessing spawn Event instead of signals to notify the worker to exit.

uvicorn/server.py
```diff
@@ -28,6 +28,8 @@
     signal.SIGINT,  # Unix signal 2. Sent by Ctrl+C.
     signal.SIGTERM,  # Unix signal 15. Sent by `kill <pid>`.
 )
+if sys.platform == "win32":  # pragma py-linux pragma: py-darwin
```
Kludex: don't we have the negative of win32 or something? :thinking:

yanyongyu: The negative pragma rule is added in https://github.com/encode/uvicorn/pull/1584/files#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52, but not yet merged :joy:. I just copy the comment from bellow. Should i add the rule in this pull request?

Kludex: Would you mind reviewing that PR as well?

yanyongyu: I don't have a Windows machine 😢 . I may ask someone else to test it.

yanyongyu: Since #1584 is merged, i just sync the branch and update the pragma comment.

Kludex (approved):

Kludex:
Did this caused https://github.com/encode/uvicorn/issues/1972 ?

yanyongyu:
Our team member test the 0.22.0 version but can not repreduce the issue on windows 10. The detailed info is shown in the https://github.com/encode/uvicorn/pull/1584#issuecomment-1546865551

PR #1915, opened 2023-03-28
Fix instructions so they work in zsh as well as Bash

simonw:
The installation instructions in a few places currently tell the user to do this:

    pip install uvicorn[standard]

This works fine with Bash, but on zsh (the default these days on macOS) results in this confusing error:

    zsh: no matches found: uvicorn[standard]

The following works fine on both Bash and zsh:

    pip install 'uvicorn[standard]'


Kludex (approved):
Thanks @simonw :)

PR #1916, opened 2023-03-29
Support the WebSocket Denial Response ASGI extension

kristjanvalur:
uvicorn now can return a response from websocket apps that choose to reject a connection with a custom response.
See https://github.com/encode/starlette/pull/2041/files for related work.

This is a re-submission of pr #1907

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -1134,6 +1134,55 @@ async def test_server_reject_connection_with_invalid_status(
 @pytest.mark.anyio
 @pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
 @pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
+async def test_server_reject_connection_with_body_nolength(
+    ws_protocol_cls, http_protocol_cls, unused_tcp_port: int
+):
+    # test that the server can send a response with a body but no content-length
+    async def app(scope, receive, send):
+        assert scope["type"] == "websocket"
+        assert "websocket.http.response" in scope["extensions"]
+
+        # Pull up first recv message.
+        message = await receive()
+        assert message["type"] == "websocket.connect"
+
+        message = {
+            "type": "websocket.http.response.start",
+            "status": 403,
+            "headers": [],
+        }
+        await send(message)
+        message = {
+            "type": "websocket.http.response.body",
+            "body": b"hardbody",
+        }
+        await send(message)
+
+    async def websocket_session(url):
+        response = await wsresponse(url)
+        assert response.status_code == 403
+        assert response.content == b"hardbody"
+        if ws_protocol_cls == WSProtocol:
+            # wsproto automatically makes the message chunked
+            assert response.headers["transfer-encoding"] == "chunked"
+        else:
+            # websockets automatically adds a content-length
+            assert response.headers["content-length"] == "8"
+
+    config = Config(
+        app=app,
+        ws=ws_protocol_cls,
+        http=http_protocol_cls,
+        lifespan="off",
+        port=unused_tcp_port,
+    )
+    async with run_server(config):
+        await websocket_session(f"ws://127.0.0.1:{unused_tcp_port}")
+
+
+@pytest.mark.anyio
+@pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
+@pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
 async def test_server_reject_connection_with_invalid_msg(
     ws_protocol_cls, http_protocol_cls, unused_tcp_port: int
 ):
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -1124,6 +1124,59 @@ async def test_server_reject_connection_with_invalid_status(
 @pytest.mark.anyio
 @pytest.mark.parametrize("ws_protocol_cls", WS_PROTOCOLS)
 @pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
+async def test_server_reject_connection_with_invalid_msg(
+    ws_protocol_cls, http_protocol_cls, unused_tcp_port: int
+):
+    if ws_protocol_cls is WSProtocol:
+        pytest.skip("Cannot supporess asynchronously raised errors")
+
+    async def app(scope, receive, send):
+        assert scope["type"] == "websocket"
+        assert "websocket.http.response" in scope["extensions"]
+
+        # Pull up first recv message.
+        message = await receive()
+        assert message["type"] == "websocket.connect"
+
+        message = {
+            "type": "websocket.http.response.start",
+            "status": 404,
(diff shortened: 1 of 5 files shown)


uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -249,14 +252,15 @@ async def run_asgi(self) -> None:
...
+            if not self.response_started:
                 self.send_500_response()
             self.transport.close()
         else:
             if not self.handshake_complete:
                 msg = "ASGI callable returned without completing handshake."
                 self.logger.error(msg)
-                self.send_500_response()
+                if not self.response_started:
+                    self.send_500_response()
```
Kludex: Why are those necessary?

What's the behavior if they are not present?

kristjanvalur: This is similar to the code in http/h11_impl.py, e.g.
```python
        except BaseException as exc:
            msg = "Exception in ASGI application\n"
            self.logger.error(msg, exc_info=exc)
            if not self.response_started:
                await self.send_500_response()
            else:
                self.transport.close()
```
If an attempt is made to send a 500_response() after the sending of a response has already begun, then the application fails in a nasty way.  send_500_response() fails in a way which isn't caught, no error is output or anything.  This is what was causing me to comment this:

```
I am unable to silence the wsproto internal error which happens when it is trying to send a 500 error.

wsproto will start a response as soon as it gets the .start message
It then gets an invalid message and attempts to send a 500 response
but it cannot because it has already started a response.
An exception is not raised via the async send() method but somewhere else which I cannot fully understand, so the test fails.
I've added a skip for the WSProto case, but maybe there is a better solution? Skipping is not ideal, because this test was designed to increase coverage.
```

We could of course try to fix that error, it is bad if an internal exception is hard to debug, but that's not within the scope of this PR.  
I guess the check could be moved into `send_500_response()` but I was following the style of `h11_impl.py`

kristjanvalur: If you approve, I´ll move this check into `send_500_response()` where it properly belongs, along with a comment, and do the same in h11_impl.py for consistency.  We can also log the situation.

kristjanvalur: still something broken with the GHA.  I'll leave this until we sort out what's going on.

Kludex:
> still something broken with the GHA. I'll leave this until we sort out what's going on.

It looks like it's related to this PR. I just cancelled the two previous jobs.

tests/protocols/test_websocket.py
```diff
@@ -919,15 +938,289 @@ async def app(scope, receive, send):
...
+    async def websocket_session(url):
+        response = await wsresponse(url)
+        assert response.status_code == 403
+        assert response.content == b"hardbody"
+        if ws_protocol_cls == WSProtocol:
+            # wsproto automatically makes the message chunked
+            assert response.headers["transfer-encoding"] == "chunked"
+        else:
+            # websockets automatically adds a content-length
+            assert response.headers["content-length"] == "8"
```
Kludex: It makes sense for WebSockets to add the `content-length`, since we can't have a chunked one there.

Kludex:
Interesting... I've just found this:

> The protocol server must not start sending the response to the client until it has received at least one Response Body event.

Ref.: https://asgi.readthedocs.io/en/latest/specs/www.html#response-start-send-event

Meaning that we shouldn't be sending the 404 if nobody is provided... But that also means that we have an issue with the HTTP implementations :thinking:

tests/protocols/test_websocket.py
kristjanvalur: well, this is only about what happens if the app doesn't provide a content-length and only provides the raw body data.  the defaults for these two protocols are different.  Technically, you _can_ provide chunked data over the websockets:  The app, provides the "chunked" header, and then provides the chunks, with the appropriate chunk headers (as an application should do if it provides that header (I think).  I think that nowadays it is rare for an application to specify content length or transfer encoding...

kristjanvalur:
> Meaning that we shouldn't be sending the 404 if nobody is provided... But that also means that we have an issue with the HTTP implementations 🤔

Yes, h11 definitely starts sending the header before receiving the first body.

This can be fixed, I can fix both h11 and websockets to wait for the first body...    I can do it for websockets in this PR and then a followup one for h11_impl.py

kristjanvalur:
> It looks like it's related to this PR. I just cancelled the two previous jobs.

Well, I haven't touched any of the .github/ code, and this is what the actions log says:
`GitHub Actions has encountered an internal error when running your job.`
The actions don't even start.  I have never seen this before.

kristjanvalur:
seems things are running smoothly again.  there was hanging test due to an introduced bug, but that should not have caused these weird symptoms.  Now fixed.

paulo-raca:
Awesome, thanks!

kristjanvalur:
I guess there are some changes upstream.  Do you want me to update this PR?

Kludex:
Yes. I'll get to this tomorrow.

uvicorn/protocols/websockets/wsproto_impl.py
```diff
@@ -254,49 +262,110 @@ async def send(self, message: "ASGISendEvent") -> None:
...
+                    )
+                    event = events.RejectConnection(
+                        status_code=message["status"],
+                        headers=list(message["headers"]),
+                        has_body=True,
+                    )
+                    # Create the event here but do not send it, the ASGI spec
+                    # suggest that we wait for the body event before sending.
+                    # https://asgi.readthedocs.io/en/latest/specs/www.html#response-start-send-event
+                    self.reject_event = event
```
Kludex: We already discussed this: https://github.com/django/asgiref/pull/387

uvicorn/protocols/websockets/wsproto_impl.py
Kludex: This is fine 👍

uvicorn/_types.py
```diff
@@ -199,7 +199,7 @@ class WebSocketResponseStartEvent(TypedDict):
 class WebSocketResponseBodyEvent(TypedDict):
     type: Literal["websocket.http.response.body"]
     body: bytes
-    more_body: bool
+    more_body: NotRequired[bool]
```
Kludex, line 202: This is more correct.

Kludex (approved):
@kristjanvalur Please check my changes, and if we are fine, let's merge this.

uvicorn/_types.py
kristjanvalur: If you say so :), We didn't have those fancy type decorations in the old days.

uvicorn/protocols/websockets/wsproto_impl.py
kristjanvalur: I know.  I disagree, I think the _recommendation_  is a good one because it enlarges the window during which the server can still send a 500 error in case the application messes up.  It is not a requirement but (IMHO) good practice in a library, and that is the reason I wrote the code that way.
But it is your call of course, happy to get this merged either way.

Kludex: I understand, and it makes sense... But I'll rather not have a mismatch in how the HTTP implementations behave, and the WebSockets.

PR #1917, opened 2023-03-31
Do not send response until first ASGI body is received for http protocols

kristjanvalur:
ASGI spec states that a server must not start sending a response until the first _body_ message has been sent to it:

https://asgi.readthedocs.io/en/latest/specs/www.html#response-start-send-event

This PR fixes both the `h11` and `httptools` based http implementations to adhere to this.

- Internal errors, such as sending two `http.response.start` messages can now result in a 500 response being sent.
- the test suite is updated accordingly

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -486,11 +486,7 @@ async def test_duplicate_start_message(protocol_cls):
     protocol = get_connected_protocol(app, protocol_cls)
     protocol.data_received(SIMPLE_GET_REQUEST)
     await protocol.loop.run_one()
-    if protocol_cls == H11Protocol:
-        assert b"HTTP/1.1 500 Internal Server Error" in protocol.transport.buffer
-    else:
-        # TODO: This should be a 500 error, but we're not currently handling it.
-        assert b"HTTP/1.1 500 Internal Server Error" not in protocol.transport.buffer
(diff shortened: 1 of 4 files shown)


Kludex:
Sorry, I didn't mean that I think we should change this behavior...

kristjanvalur:
Fixes a few problems though.

kristjanvalur:
Interesting error, and completely unrelated.  This does not repro on my machine.
My investigation indicates that there is a race condition in the websockets_ipml protocol.
When the protocol detects the ASGI error of a return value, it does a self.transport.close() at websockets_impl.py:275.
But this just logs a close, and then returns normally.

Later in websocket.legacy.server, the handler has determined that the application had a normal close, and will send a close frame.
Whether this close frame gets through or not is a race, depending on how the scheduled close from the websockets_impl.py was handled, and when it got executed.

The websockets_impl.py must do things differently and cause an error return in case of error, or perform a more forceful close, to avoid the server.py to send a normal close after the error.

kristjanvalur:
I incorporated the elements of pr #1923 to enable the unit tests to run successfully.

kristjanvalur:
Closing this, since this is no longer a requirement of ASGI:
https://github.com/django/asgiref/pull/387

PR #1923, opened 2023-04-02
Set the `closed_event` at the end, when the application is fully done.

kristjanvalur:
This PR fixes a race condition, made apparent by pr #1917 affecting the `websockets` implementation for running websockets apps.

If the `closed_event` is set too early, the `websockets.legacy.server` module will believe the application ended successfully and send a closed-handshake.  This can occur, even if the application explicitly calls `self.transport.close()` in some cases.

This pr makes sure that the `closed_event` is set at the end, giving the application the chance to perform an abortive close before returning.

Unfortunately, it is not simple to create a unit-test demonstrating this race condition.


kristjanvalur:
@Kludex , what do you think?

kristjanvalur:
I still think this should go in.

Kludex:
@kristjanvalur Sorry for the delay here.

> If the closed_event is set too early, the websockets.legacy.server module will believe the application ended successfully and send a closed-handshake. This can occur, even if the application explicitly calls self.transport.close() in some cases.

What do you mean here? The `closed_event` is only used by the uvicorn's implementation, and there's no logic around that, is there?. The `WebSocketServerProtocol` from `websockets` doesn't do anything with it as well.

kristjanvalur:
Sorry, it's been 11 months since I wrote this.  Not a great turnaround time.
It was a real problem made evident in a PR that I was working on at the time, and this PR fixed it.  I´ll try to refresh my memory now.

kristjanvalur:
ISo, my mind is a bit rusty in how all of this fits together.
But there is a  `ws_handler` method.  This is used by a `websockets` implementation, such as `websockets.legacy`.  It is part of the WebSocketServerProtocol.
It will return once the closed_event is set, and the server will then finish the connection.
But the closed event is set early, Notice how during the error handling, uvicorn does a `self.transport.close()` after having set the event and done another await.  Because `closed_event` is set at this point, the server may have already closed the connection.

the `closed_event` must therefore be set as the end, when the hander task has completely finished dealing with the connection.

I can't recall the exact problem this caused, but IIRC, it meant that an error case, which should result in an error termination, instead was closed like a normal termination had taken place.  effectively the `self.transport.close()` line didn't do anything because the server had already performed a graceful close (having returned from `await ws_handler()`

Kludex:
But the `closed_event` is not taking in consideration in the `ws_handler`. 🤔

Kludex:
> ISo, my mind is a bit rusty in how all of this fits together. But there is a `ws_handler` method. This is used by a `websockets` implementation, such as `websockets.legacy`. It is part of the WebSocketServerProtocol. It will return once the closed_event is set, and the server will then finish the connection. But the closed event is set early, Notice how during the error handling, uvicorn does a `self.transport.close()` after having set the event and done another await. Because `closed_event` is set at this point, the server may have already closed the connection.
> 
> the `closed_event` must therefore be set as the end, when the hander task has completely finished dealing with the connection.
> 
> I can't recall the exact problem this caused, but IIRC, it meant that an error case, which should result in an error termination, instead was closed like a normal termination had taken place. effectively the `self.transport.close()` line didn't do anything because the server had already performed a graceful close (having returned from `await ws_handler()`

But the `closed_event` is not taken in consideration on the `ws_handler`...

Kludex:
@kristjanvalur I'll close this given my last comment, but I'm happy to reopen and merge if you get the time to reply.

Sorry the delay here.

kristjanvalur:
I see.
I was pretty sure it used to be the case.  Maybe things have changed.  I'm pretty sure that there was an await for this event at the time.  The issue I saw at the time was real (oopped up as a fix for a unit test of a different PR) and my pr fixed it then.

Never mind then, and sorry for the noise.

PR #1925, opened 2023-04-04
Use ruff instead of flake8, autoflake and isort

Kludex:
- Analogous to https://github.com/encode/starlette/pull/2110

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,3 +61,11 @@ path = "uvicorn/__init__.py"
 include = [
     "/uvicorn",
 ]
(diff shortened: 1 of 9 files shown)


michaeloliverx (approved):
Great stuff!

Another rule that is useful is `RUF100` to find unused `noqa` comments.

euri10:
> Another rule that is useful is `RUF100` to find unused `noqa` comments.

Oo nice, TIL I learnt how to get rid of those old E999 on match, thanks :)

PR #1929, opened 2023-04-06
Fix WebSocket tests

### LINKED ISSUE #1927: WebSockets 11.0 broke Uvicorn test suite
As the title says... Changelog: https://websockets.readthedocs.io/en/stable/project/changelog.html

aminalaee:
Closes https://github.com/encode/uvicorn/issues/1927

From WebSockets changelog:

> Closing a connection without an empty close frame is OK.
> Receiving an empty close frame now results in [ConnectionClosedOK](https://websockets.readthedocs.io/en/stable/reference/exceptions.html#websockets.exceptions.ConnectionClosedOK) instead of [ConnectionClosedError](https://websockets.readthedocs.io/en/stable/reference/exceptions.html#websockets.exceptions.ConnectionClosedError).
>As a consequence, calling WebSocket.close() without arguments in a browser isn’t reported as an error anymore.


tests/protocols/test_websocket.py
```diff
@@ -584,7 +584,7 @@ async def connect(url):
     async with run_server(config):
         with pytest.raises(websockets.exceptions.ConnectionClosed) as exc_info:
             await connect(f"ws://127.0.0.1:{unused_tcp_port}")
-        assert exc_info.value.code == 1006
+        assert exc_info.value.code == 1000 or 1006
```
aminalaee: This is silly, will look for  a better way :)

Kludex: What would be needed for the websockets implementation to continue to send 1006?

aminalaee: This is the definition of their `ConnectionClosedError`:

```A close frame with a code other than 1000 (OK) or 1001 (going away) was received or sent, or the closing handshake didn't complete properly.```

Comment (aminalaee) [reply tests/protocols/test_websocket.py]: I'll leave this up for review but I think it makes sense to keep this way.
I tried changing the test to something like:

```py
        await send({"type": "websocket.send", "code": 1002})
        return 123
```
To cause websockets fail with 1006 again, but that will lead to these lines not being covered:

https://github.com/encode/uvicorn/blob/b54e24874d3765c21f193e2da4c2ac5e6d3f51fe/uvicorn/protocols/websockets/websockets_impl.py#L271-L275

Kludex:
The issue is on the client side, and not the server? :thinking: Or is the server sending 1000 now?

Kludex:
I'd prefer consistency between the two implementations. Can we achieve that?

aminalaee:
I might be wrong, but I think it's the client side.
I don't think we can keep it consistent between the two implementations because this test is specifically trying to return not None in the ASGI but the clients have different behaviors now. Open to suggestions of course.

Kludex:
It's on the server side. I run it with the previous version and with the new one. The previous had an error.

I don't think we should have an "or" statement in the assertion. We should try to match the behavior on the websocket implementations.

Kludex:
Can you add the code to the parametrization of the websockets implementations?

aminalaee:
I will take a look into this.

aminalaee:
I think this is caused in here https://github.com/encode/uvicorn/blob/ffa5b1ac96b10976ed0e092a0bc1dd5526101356/uvicorn/protocols/websockets/websockets_impl.py#L265-L267

With websockets 10.4, checking `self.state` before and after ` await self.handshake_completed_event.wait()` shows that the connection is `CONNECTING` and then `OPEN`.
But with websockets 11.0, the same check shows that state is `CONNECTING` and then `CLOSING`, that's why trying to manually close the connection with a different code doesn't work.

I think one way is to change the test to make the connection open, and then make it fail with a 1006 code. 
If that doesn't seem right, we can go back to the option you mentioned to parametrize it.

I think @aaugustin 's input would be very helpful here.

aaugustin:
I looked into this. At this ponit, it seems to me that we need to update the code, not the tests, because it was relying on subtle timing effects of asyncio tasks that changed in 11.0. See my comments in #1927.

aaugustin:
This PR move the test to "after the connection handshake is complete" while it was previously happening "during the connection handshake".

#2061 is an alternative approach that keeps the test "during the connection handshake". I'm removing a line that was added in #174, making it difficult to know why this particular line is present.

aminalaee:
Thanks for taking a look, I think your approach and explanation makes much more sense.

PR #1937, opened 2023-04-11
Add type hints to test_auto_detection.py

aminalaee:
Add typing to test_auto_detection.py

### DIFF
--- a/tests/test_auto_detection.py
+++ b/tests/test_auto_detection.py
@@ -8,21 +8,9 @@ from uvicorn.main import ServerState
 from uvicorn.protocols.http.auto import AutoHTTPProtocol
 from uvicorn.protocols.websockets.auto import AutoWebSocketsProtocol
 
-try:
-    import uvloop
-except ImportError:  # pragma: no cover
-    uvloop = None
-
-try:
-    import httptools
-except ImportError:  # pragma: no cover
-    httptools = None
-
-try:
-    import websockets
-except ImportError:  # pragma: no cover
-    # Note that we skip the websocket tests completely in this case.
-    websockets = None
+pytest.importorskip("uvloop")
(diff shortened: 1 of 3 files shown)


tests/test_auto_detection.py
```diff
@@ -9,20 +10,22 @@
 from uvicorn.protocols.websockets.auto import AutoWebSocketsProtocol
 
 try:
-    import uvloop
+    importlib.import_module("uvloop")
+    expected_loop = "uvloop"
 except ImportError:  # pragma: no cover
-    uvloop = None
+    expected_loop = "asyncio"
```
aminalaee, line 16: Open to discuss unused imports vs using importlib just to make sure package is installed.

Kludex: What was wrong with the import statement?

aminalaee: The import statement does `import uvloop` just to see if uvloop is installed, so causes two issues:
- One is that we have unused import
- Second is that if package is not installed, we set `uvloop = None` which needs another `type: ignore` because module cannot be None.

Since this is just for purpose of checking package installation, I thought it makes sense to use importlib.

uvicorn/protocols/websockets/auto.py
```diff
@@ -1,12 +1,13 @@
 import asyncio
```
Kludex: On the tests is fine, but can we revert it here? I guess the cost is higher to import importlib + call import_module than import the packages itself?

aminalaee: Yeah, I guess it's not worth it here.

Kludex (approved):
Please merge after last comment addressed. Thanks @aminalaee :pray:

PR #1938, opened 2023-04-11
Fix typo in deprecation warning

aminalaee:
I think it was supposed to be `deprecated` instead of `depreciated` in the warning.


Kludex (approved):
I think it was not a typo tbh hahaha but fine :+1:

PR #1940, opened 2023-04-12
Handle missing trustme/cryptography gracefully in the test suite

mgorny:
Skip tests requiring trustme/cryptography if either of the packages are not installed.  This makes it possible to run most of the test suite on architectures that do not feature a working Rust toolchain at the time.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,9 +11,15 @@ from time import sleep
 from uuid import uuid4
 
 import pytest
-import trustme
-from cryptography.hazmat.backends import default_backend
-from cryptography.hazmat.primitives import serialization
+
+try:
+    import trustme
+    from cryptography.hazmat.backends import default_backend
+    from cryptography.hazmat.primitives import serialization
+
+    HAVE_TRUSTME = True
+except ImportError:  # pragma: nocover
+    HAVE_TRUSTME = False
 
 from uvicorn.config import LOGGING_CONFIG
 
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -11,9 +11,13 @@
...
-import trustme
-from cryptography.hazmat.backends import default_backend
-from cryptography.hazmat.primitives import serialization
+
+try:
+    import trustme
+    from cryptography.hazmat.backends import default_backend
+    from cryptography.hazmat.primitives import serialization
+except ImportError:  # pragma: nocover
+    trustme = None  # type: ignore
```
mgorny: I've done this to avoid mypy complaining about incompatible type of `trustme`. If you prefer another approach, just let me know and I'll update it.

Kludex: Can we follow this way: https://adamj.eu/tech/2021/12/29/python-type-hints-optional-imports, please?

mgorny: Sure, done that now.

tests/conftest.py
```diff
@@ -30,12 +36,14 @@
 
 
 @pytest.fixture
-def tls_certificate_authority() -> trustme.CA:
+def tls_certificate_authority() -> "trustme.CA":
+    if not HAVE_TRUSTME:
+        pytest.skip("trustme not installed")
```
Kludex:
```suggestion
        pytest.skip("trustme not installed")  # pragma: no cover
```

tests/conftest.py
```diff
@@ -11,9 +11,15 @@
...
-from cryptography.hazmat.backends import default_backend
-from cryptography.hazmat.primitives import serialization
+
+try:
+    import trustme
+    from cryptography.hazmat.backends import default_backend
+    from cryptography.hazmat.primitives import serialization
+
+    HAVE_TRUSTME = True
+except ImportError:  # pragma: nocover
```
Kludex:
```suggestion
except ImportError:  # pragma: no cover
```

Kludex (changes requested):

mgorny:
Thanks. Please lemme know if I should rebase.

Kludex (approved):

Kludex:
Thanks Michał :pray: 

With similar subjects, you can create a PR directly, no need for discussions. :+1:

Kludex:
Do you need me to make a release, or can you already make use of those changes?

mgorny:
Thanks. We already have a hack in place, so no hurry for a new release.

PR #1942, opened 2023-04-13
Simplified worker monitoring and restart, without gunicorn.

### LINKED ISSUE #517: Master process should restart expired workers.

gnat:
Drastically simplifies app deployment for Starlette and FastAPI for many users.

Survive worker crashes directly in Uvicorn- go ahead and run your ffmpeg and machine learning tasks. Uvicorn will auto restart your workers up to the desired `--workers` count, reliably, even under heavy load.

* For many users, this removes gunicorn as a dependency. @tomchristie @tiangolo
  * https://github.com/encode/uvicorn/issues/517#issuecomment-564090865 
  * Simplifies production documentation for many scenarios.
* Tested, benchmarked on both Windows 11 :window: and Linux! @euri10 @kludex
* **Under 20 lines**, backwards compatible, no effect on uvicorn performance.
* You can now "reload" your app by killing all of the child processes. :stuck_out_tongue: 

### Related issues

* Closes: https://github.com/encode/uvicorn/issues/517
* Good resolution for: https://github.com/encode/starlette/issues/671
* Many tangentially related FastAPI issues which would have never been reported if this was included in Uvicorn.

### Deployments right now...

caddy/nginx  ➡️  gunicorn  ➡️  uvicorn  ➡️  starlette/fastapi

### For many users can be simplified to...

caddy/nginx  ➡️  uvicorn  ➡️  starlette/fastapi

### Stress tested to 100,000's of r/s using using [hey](https://github.com/rakyll/hey).

Thoughts, feedback appreciated. 

I realise this isn't as "feature rich" as some may want (handling various signals, etc), but we're eliminating an entire dependency with only a few lines. It works well and is a relatively tiny change for big benefits, and can be easily re-factored out if something better is implemented in the future.

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -6,6 +6,7 @@ on:
     branches: ["master"]
   pull_request:
     branches: ["master"]
+  workflow_dispatch:
 
 jobs:
   tests:
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -75,7 +75,10 @@ class Multiprocess:
         logger.info(message, extra={"color_message": color_message})
 
     def monitor(self) -> None:
-        while not self.should_exit.wait(0.5):
+        while not self.should_exit.is_set():
+            time.sleep(0.5)
+            if self.should_exit.is_set():
+                break
             # Restart expired workers.
             for process in self.processes:
                 if not process.is_alive():
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -75,10 +75,7 @@ class Multiprocess:
         logger.info(message, extra={"color_message": color_message})
 
     def monitor(self) -> None:
-        while not self.should_exit.is_set():
-            time.sleep(0.5)
-            if self.should_exit.is_set():
-                break
+        while not self.should_exit.wait(0.5):
             # Restart expired workers.
             for process in self.processes:
                 if not process.is_alive():
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import signal
+import time
 import threading
 from multiprocessing.context import SpawnProcess
 from socket import socket
@@ -42,7 +43,7 @@ class Multiprocess:
 
     def run(self) -> None:
         self.startup()
-        self.should_exit.wait()
+        self.monitor()
         self.shutdown()
 
     def startup(self) -> None:
@@ -72,3 +73,18 @@ class Multiprocess:
             click.style(str(self.pid), fg="cyan", bold=True)
         )
         logger.info(message, extra={"color_message": color_message})
+
+    def monitor(self) -> None:
+        while not self.should_exit.is_set():
+            time.sleep(0.5)
+            if self.should_exit.is_set():
+                break
+            # Restart expired workers.
+            for process in self.processes:
+                if not process.is_alive():
+                    self.processes.remove(process)
+                    process_new = get_subprocess(
+                        config=self.config, target=self.target, sockets=self.sockets
+                    )
+                    process_new.start()
+                    self.processes.append(process_new)
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -1,8 +1,8 @@
 import logging
 import os
 import signal
-import time
 import threading
+import time
 from multiprocessing.context import SpawnProcess
 from socket import socket
 from types import FrameType
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 98.68
+fail_under = 98.65
 show_missing = true
 skip_covered = true
 exclude_lines =
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 98.80
+fail_under = 98.68
 show_missing = true
 skip_covered = true
 exclude_lines =
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,7 @@ plugins =
 
 [coverage:report]
 precision = 2
-fail_under = 98.65
+fail_under = 98.63
 show_missing = true
 skip_covered = true
 exclude_lines =


humrochagf:
There's a behavior change that may be unexpected for those running on k8s and expect the pod to be restarted during a crash. To cover those, I recommend having an unmanaged option

gnat:
I would agree, but I'm not sure it matters because this new functionality is only present if `--workers=2` or greater. (`multiprocess.py` only)

Unless I'm mistaken, k8s pods are running single worker mode (1 worker = 1 pod), because k8s is being used as the "worker manager".

In the current `multiprocess.py` behavior: the parent uvicorn process awkwardly stays alive (and unusable) even when all workers are dead: #517 This mode has never been suitable for use with an external worker manager, AFAIK.

uvicorn/supervisors/multiprocess.py
```diff
@@ -72,3 +73,18 @@ def shutdown(self) -> None:
...
         logger.info(message, extra={"color_message": color_message})
+
+    def monitor(self) -> None:
+        while not self.should_exit.is_set():
+            time.sleep(0.5)
+            if self.should_exit.is_set():
+                break
+            # Restart expired workers.
+            for process in self.processes:
+                if not process.is_alive():
```
Kludex, line 84: This is not reliable. We can't know if the process is stuck. It will only give the false impression of a good process manager.

gnat: I've considered this but I don't think it's reasonable to expect Uvicorn to be so involved, because determining "stuck" and the handling of "stuck" could be quite different for every project: It's up to the developer to decide if/when/how to kill a "stuck" worker.

Nor is it the point of this PR.

We want Uvicorn to just mimic k8s here- and only become involved when the worker is dead/crashed/exited, which is what `is_alive()` does perfectly.

ahopkins: This is a good change. I agree with @gnat. I think the feature that @Kludex is after is a slightly different scope.

humrochagf:
> I would agree, but I'm not sure it matters because this new functionality is only present if `--workers=2` or greater. (`multiprocess.py` only)
> 
> Unless I'm mistaken, k8s pods are running single worker mode (1 worker = 1 pod), because k8s is being used as the "worker manager".
> 
> In the current `multiprocess.py` behavior: the parent uvicorn process awkwardly stays alive (and unusable) even when all workers are dead: #517 This mode has never been suitable for use with an external worker manager, AFAIK.

Sorry my bad, you are right, the multiprocess isn't used in the single worker case

uvicorn/supervisors/multiprocess.py
```diff
@@ -72,3 +73,18 @@ def shutdown(self) -> None:
             click.style(str(self.pid), fg="cyan", bold=True)
         )
         logger.info(message, extra={"color_message": color_message})
+
+    def monitor(self) -> None:
+        while not self.should_exit.is_set():
+            time.sleep(0.5)
+            if self.should_exit.is_set():
+                break
```
zanieb, line 81: I believe this can be replaced with the simpler:
```suggestion
        while not self.should_exit.wait(0.5):
```

gnat: Agreed! This is a good simplification. Tested, works identically.

gnat: @madkinsz Actually there are implications of this call--- `wait()` here uses `Condition` which introduces a lock allocation, and can raise an exception: https://github.com/python/cpython/blob/2b5dbd1f237a013defdaf0799e0a1a3cbd0b13cc/Lib/threading.py#L288

On the other hand, `is_set()` just returns the value:
https://github.com/python/cpython/blob/2b5dbd1f237a013defdaf0799e0a1a3cbd0b13cc/Lib/threading.py#L568

We're only calling it once per half a second, so the performance difference is not noticeable- but I don't think it's worth introducing a lock, I'm going to revert this for now.

.github/workflows/test-suite.yml
```diff
@@ -6,6 +6,7 @@ on:
     branches: ["master"]
   pull_request:
     branches: ["master"]
+  workflow_dispatch:
```
Kludex, line 9: Why did you add this?

uvicorn/supervisors/multiprocess.py
Kludex: On the description it says: "we're eliminating an entire dependency with only a few lines.", but gunicorn does check if the process is stuck, so it's the same scope, and this is not a replacement for it.

Kludex (commented):
I don't think we should include this change given the "simplicity" and how it "lies" to users about being reliable.

But... If we want this to get in, at least those should happen:
1. Add tests.
2. Answer the question: why this shouldn't be an opt-in feature?

uvicorn/supervisors/multiprocess.py
zanieb: Hm interesting. I guess another benefit is that it'd wake up early with the `wait` implementation. Is it worth adding 0.5s of slowdown to shutdown? It seems like it's probably fine in the context of uvicorn multiprocessing.

.github/workflows/test-suite.yml
gnat: This enables dev branches to manually run the test suite. 

image

Kludex: Without creating a PR you mean?

gnat: Yes, that's correct. It allows for running the test suite manually on your own fork or branch prior to creating a pull request.

zanieb:
> 2. Answer the question: why this shouldn't be an opt-in feature?

It seems like advanced monitoring should be opt-in and simple "is alive" monitoring should be the default since Python doesn't let you "opt-out" of dependencies.

gnat:
> 2. Answer the question: why this shouldn't be an opt-in feature?

The current behavior is broken: `multiprocess.py` becomes a zombie process when all the workers are dead.

`multiprocess.py` is not used when running under k8s or gunicorn, so it's not obvious that this is broken.

Kludex:
We're not doing this. I prefer to have reliable restart, as I said on https://github.com/encode/uvicorn/pull/1942#discussion_r1167559989.

I'll make @abersheeran 's PR happen instead: https://github.com/encode/uvicorn/pull/2183 👍

PR #1943, opened 2023-04-14
Use a ChainMap for the lifespan state

adriangb:
This lets applications differentiate between the lifespan and request states and even update the lifespan state from a request (previously not possible). For now let's keep this an implementation detail, but if it looks good we can advertise it and try to get it in the ASGI spec.

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -2,7 +2,8 @@ import logging
 import socket
 import threading
 import time
-from typing import Optional, Union
+from copy import deepcopy
+from typing import Any, List, Optional, Union
 
 import pytest
 
@@ -998,15 +999,23 @@ async def test_iterator_headers(protocol_cls):
 @pytest.mark.anyio
 @pytest.mark.parametrize("protocol_cls", HTTP_PROTOCOLS)
 async def test_lifespan_state(protocol_cls):
-    expected_states = [{"a": 123, "b": [1]}, {"a": 123, "b": [1, 2]}]
+    expected_states = [
+        [{}, {"a": 123, "b": [1]}],
+        [{"a": 456, "c": 42}, {"a": 123, "b": [1, 2]}],
+        [{}, {"a": 123, "b": [1, 2]}],
+        [{"a": 456, "c": 42}, {"a": 123, "b": [1, 2, 2]}],
+    ]
+    actual_states: List[Any] = []
 
     async def app(scope, receive, send):
-        expected_state = expected_states.pop(0)
-        assert scope["state"] == expected_state
-        # modifications to keys are not preserved
+        actual_states.append(deepcopy(scope["state"].maps))
+        # modifications to keys or new keys are not preserved
         scope["state"]["a"] = 456
+        scope["state"]["c"] = 42
         # unless of course the value itself is mutated
         scope["state"]["b"].append(2)
+        # the ChainMap is accessible
+        actual_states.append(deepcopy(scope["state"].maps))
         return await Response("Hi!")(scope, receive, send)
 
     lifespan = LifespanOn(config=Config(app=app))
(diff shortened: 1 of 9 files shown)


Kludex:
Given the reason, I don't think we should include this. The request state is a concept that exists in Starlette (in opposed to being specified in ASGI), and it doesn't need to use the same namespace. If we want to differentiate both we could use another namespace.

adriangb:
> The request state is a concept that exists in Starlette (in opposed to being specified in ASGI)

But it is a thing in the ASGI spec now as of the lifespan state PR we merged.

> If we want to differentiate both we could use another namespace.

Well we could create a new namespace in Starlette but then it would (1) be a breaking change in Starlette and (2) be outside of the ASGI spec.

The reasons I like this approach are:
- It stays compliant with and uses the current asgi spec
- It is backwards compatible for both Uvicorn and Starlette
- If we wanted to include it in the asgi spec that would be backwards compatible as well
- There's no extra cognitive load for users that treat this like a normal dict and unless they "opt into" treating it as a ChainMap they are protected from doing anything potentially dangerous like persisting information from one request to another.
- ChainMap is in the standard library, we don't have to write any logic ourselves or add any dependencies.
- The idea of a "stack of contexts" is similar to things that [Flask already has](https://flask.palletsprojects.com/en/1.1.x/appcontext/#manually-push-a-context) and I think other web frameworks have similar things.

Kludex:
> > The request state is a concept that exists in Starlette (in opposed to being specified in ASGI)
> 
> But it is a thing in the ASGI spec now as of the lifespan state PR we merged.

That's not what I meant. The namespace ["scope"]["state"] being used for passing data on the request/response cycle is a concept that exists in Starlette. It's not mention in the spec that this should be in this namespace...

humrochagf:
Not sure I got the advantage of ChainMap in this case, you remain unable to persist values in the state because the chainmap child creation / dict shallow copy is made at request parsing and dies at the end of the request (and this is an asgi specification).

The expected flow of state usage is to set all persistent states at `'lifespan.startup'` to be available to all requests.

The ephemeral values can be set to the state per requests on both (chainmap / dict) implementations and this use case is the one used by middleware where you set things to the state to be used in the next layer, and at the end of the request all of this data is destroyed.

The chainmap on the other hand has a downside that is adding an o(n) complexity where n = number of layers you add to your chain, so to add it to the core of every request we need to be sure the benefit of adding it will be really paying off the complexity increase.

adriangb:
> Not sure I got the advantage of ChainMap in this case, you remain unable to persist values in the state because the chainmap child creation / dict shallow copy is made at request parsing and dies at the end of the request

That's not true. The whole point is that you can persist data to other requests, but you have to be explicit about it by accessing `scope["state"].maps[-2]` or something like that.

> and this is an asgi specification

True, but I think since this is API-compatible with the intention of the ASGI specification it should be fine. Besides, if this does work out and we end up wanting it for FastAPI or something we can always amend the specification. The nice thing about this approach is that you can totally ignore it and not ever acknowledge that there is a ChainMap involved and everything will work just fine.

> The expected flow of state usage is to set all persistent states at `'lifespan.startup'` to be available to all requests.

Yes but there may be use cases where you want to put stuff in there later.
For example if FastAPI lets you mark dependencies as being cached at the lifespan level, it may not want to eagerly run them all on startup an instead lazily store them in this cache as endpoints need them. That's just one use case I'm thinking of.

> The chainmap on the other hand has a downside that is adding an o(n) complexity where n = number of layers you add to your chain, so to add it to the core of every request we need to be sure the benefit of adding it will be really paying off the complexity increase.

The number of layers is always going to be 2, and O(2) == O(1) so there is no difference really.

Kludex:
I'm going to close this, since it's not needed for now. In case it's proven to be needed, we can reopen it.

Thanks @adriangb 🙏

PR #1945, opened 2023-04-15
Log exception traceback in case of invalid HTTP request when using h11

ZhymabekRoman:
Author: florimondmanca (https://github.com/florimondmanca)
Source: https://github.com/encode/uvicorn/pull/889
Revert PR: https://github.com/encode/uvicorn/pull/1518

Motivation: I absolutely don't understand why this commit was reverted, if there are errors during execution you shouldn't keep quiet about them. Today I encountered the error `Invalid HTTP request received.` and I was stumped for hours because of it. After adding a traceback of the error it immediately became clear: `h11._util.RemoteProtocolError: Missing mandatory Host: header`. The funny thing about this error is that it appeared to me only if I used reverse proxy NGINX,

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -205,9 +205,9 @@ class H11Protocol(asyncio.Protocol):
         while True:
             try:
                 event = self.conn.next_event()
-            except h11.RemoteProtocolError:
+            except h11.RemoteProtocolError as exc:
                 msg = "Invalid HTTP request received."
-                self.logger.warning(msg)
+                self.logger.warning(msg, exc_info=exc)
                 self.send_400_response(msg)
                 return
             event_type = type(event)


euri10:
> Author: florimondmanca (https://github.com/florimondmanca) Source: #889 Revert PR: #1518
> 
> Motivation: I absolutely don't understand why this commit was reverted, if there are errors during execution you shouldn't keep quiet about them. Today I encountered the error `Invalid HTTP request received.` and I was stumped for hours because of it. After adding a traceback of the error it immediately became clear: `h11._util.RemoteProtocolError: Missing mandatory Host: header`. The funny thing about this error is that it appeared to me only if I used reverse proxy NGINX,

it is reverted because it serves no purpose anymore, it was added as a way for us to get the info needed to solve #344 which is now solved

I think a switch to opt-in for logging the warning could be considered, in which case we shoul dhave it for both h11 and httptools and not just h11 as it is now.

Just having the warning without opt-in is imho highly annoying to say the least, the default should be not to have it

PR #1954, opened 2023-04-26
remove a few mypy excludes

graingert:
I accidentally read the "If you are seeing this, feel free to create a PR to cover the below files." note and became obligated

# Summary

# Checklist


Kludex (approved):
Thanks @graingert ! Glad to have you back! 😃

Issue #1868, opened 2023-02-15
Reload options doesn't work when you use uvicorn.Server(config).serve()


vgulaev:
Looks like forgotten add ChangeReload initialization in serve()
My app:
**main.py**
```
import asyncio
import uvicorn


async def main():
    config = uvicorn.Config(
        "example:app", port=5000, log_level="info", host="0.0.0.0", reload=True
    )
    server = uvicorn.Server(config)
    await server.serve()


if __name__ == "__main__":
    asyncio.run(main())

````
**example.py**
```
async def app(scope, receive, send):
    assert scope["type"] == "http"

    body = f'Received {scope["method"]} request ++++* to {scope["path"]}'
    await send(
        {
            "type": "http.response.start",
            "status": 200,
            "headers": [
                [b"content-type", b"text/plain"],
            ],
        }
    )
    await send(
        {
            "type": "http.response.body",
            "body": body.encode("utf-8"),
        }
    )

```

when use everything works correct:
**main.py**
```
import uvicorn

if __name__ == "__main__":
    uvicorn.run("example:app", port=5000, log_level="info", reload=True, host="0.0.0.0")
```

the only place where ChangeReload exists is in **site-packages/uvicorn/main.py**, sorry for my debug code here
![image](https://user-images.githubusercontent.com/2358096/219000913-266a273c-7153-4179-9cf0-b8cd54305b28.png)

**Comment by Kludex** [comment]
As you pointed yourself, it's intentional.

**Comment by slyapustin** [comment]
@Kludex 
That is worth to be mentioned in the `Running programmatically` section of the documentation:
https://github.com/encode/uvicorn/blob/master/docs/index.md#config-and-server-instances

**Comment by Kludex** [comment]
PR to improve documentation is welcome.

**Comment by slyapustin** [comment]
@Kludex I just created PR #1955

**Comment by declension** [comment]
Came here to say it's not just docs, it's uvicorn logging too, so that might want updating

```
INFO:     Will watch for changes in these directories: ['/home/foo/bar...]
```

PR #1955, opened 2023-04-26
[GH-1868] Updated documentation to note what auto-reload is disabled when serve() used directly.

slyapustin:
# Summary

Updated documentation as suggested by @Kludex on the issue: https://github.com/encode/uvicorn/issues/1868#issuecomment-1521690175

# Checklist

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -253,7 +253,7 @@ if __name__ == "__main__":
     asyncio.run(main())
 ```
     
-When the `uvicorn.Server.serve()` is used directly, the `auto-reload` feature is not available.
+When the `uvicorn.Server.serve()` is used directly, the hot reload feature is not available.
 
 This is because the `uvicorn.Server.serve()` function is usually used in production environments where automatic reloading is not desired. In production environments, the application is expected to be stable and changes are typically made through a deployment process rather than by modifying the code directly.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -255,7 +255,7 @@ if __name__ == "__main__":
     
 When the `uvicorn.Server()`'s `serve()` and `run()` methods are used directly, the hot reload feature is not available.
 
-This is because the `uvicorn.Server.serve()` function is usually used in production environments where automatic reloading is not desired. In production environments, the application is expected to be stable and changes are typically made through a deployment process rather than by modifying the code directly.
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -252,6 +252,10 @@ async def main():
 if __name__ == "__main__":
     asyncio.run(main())
 ```
+    
+When the `uvicorn.Server.serve()` is used directly, the `auto-reload` feature is not available.
```
Kludex:
```suggestion
When the `uvicorn.Server.serve()` is used directly, the hot reload feature is not available.
```

Kludex (commented):
It may be useful to add on the documentation the way to programmatically use the reload, instead of saying just that is not possible.

martinky24:
We ran into this issue as well (hot reload not working when using `server.run()`)-- I was about to open a PR, glad to see it already exists.

@slyapustin Would it be reasonable to also note that `server.run` ([here](https://github.com/encode/uvicorn/blob/f02b80b3d37c847044ccb02bb9e7a5e48b9e7ace/docs/index.md?plain=1#L235)) also does not enable hot reloading? I am happy to throw in a commit to note this explicit addition.

slyapustin:
> I am happy to throw in a commit to note this explicit addition.
@martinky24 That would be great

martinky24:
Ah, I realize I don't have permissions to update the branch directly I think...

docs/index.md
```diff
@@ -252,6 +252,10 @@ async def main():
 if __name__ == "__main__":
     asyncio.run(main())
 ```
+    
+When the `uvicorn.Server.serve()` is used directly, the hot reload feature is not available.
```
martinky24:
```suggestion
When the `uvicorn.Server()`'s `serve()` and `run()` methods are used directly, the hot reload feature is not available.
```

docs/index.md
```diff
@@ -252,6 +252,10 @@ async def main():
...
+
+This is because the `uvicorn.Server.serve()` function is usually used in production environments where automatic reloading is not desired. In production environments, the application is expected to be stable and changes are typically made through a deployment process rather than by modifying the code directly.
```
martinky24:
```suggestion
This is because the `serve()` and `run()` methods functions are usually used in production environments where automatic reloading is not desired. In production environments, the application is expected to be stable and changes are typically made through a deployment process rather than by modifying the code directly.
```

slyapustin:
Thanks @martinky24, I updated my PR with your suggestions.

PR #1966, opened 2023-05-02
Load MessageLoggerMiddleware based on logger's effective level

### LINKED ISSUE #1958: logger.level vs logger.getEffectiveLevel() on config load()
Hello,

I was wondering why `logger.level` is checked [here](https://github.com/encode/uvicorn/blob/2c0ea0a9b80facd2902634b74987b9165c6018ab/uvicorn/config.py#L507) instead of `logger.getEffectiveLevel()` because at the final stage, the level returned by the `logger.getEffectiveLevel()` is the one that is going to be respected.

So for example, if you set a level INFO for _uvicorn_ logger with a dictConfig (and you don't declare any more loggers) uvicorn will load `MessageLoggerMiddleware`, but in the reality we are not going to see any logs from there because our logger is not going to capture at that level.

So what I think is that on that line, `logger.level` should be changed to `logger.getEffectiveLevel()`. 

I can send a MR if you agree with me.

Sorry if I'm missing something and thanks for taking care of this project.

josumoreno-BP:
# Summary

Fixes https://github.com/encode/uvicorn/issues/1958

I did an isolated test for this one, but if you think that I should merge this one with `test_config_log_level` and check there both (level and effective level) let me know.

I've not updated the documentation because I feel that it is not needed for this one.

# Checklist

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -5,6 +5,7 @@ import socket
 import sys
 import typing
 from pathlib import Path
+from typing import Optional
 from unittest.mock import MagicMock
 
 import pytest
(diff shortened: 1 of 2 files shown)


Kludex (approved):
Thanks 👍

josumoreno-BP:
Ups, I see that this failed. I only tested it isolated and worked.

I've been checking what is going on to suffer side effects from other tests and the reality is that once a test has loaded [the default config](https://github.com/encode/uvicorn/blob/e712ee8012d1aa9c033dc4acdd7420626245d918/uvicorn/config.py#L82) I cannot use `None` as a parameter for `uvicorn_logger_level` because it will not "override" the existing level (see [here](https://github.com/python/cpython/blob/main/Lib/logging/config.py#L871)). Even trying to set it to None is an error, but in this case passed inadvertent because python ignores it.

So I'm sending an amend removing that `None` from `uvicorn_logger_level` parameters.

Kludex:
Thanks @josumoreno-BP 🙏

PR #1976, opened 2023-05-13
Allow `ConfigParser` or a `io.IO[Any]` on `log_config`

antonymayi:
This is a new instance of PR #1716 closed accidentally by deleting the forked repository.

Copy of the previous description:

> The existing logic for the `log_config` option accepts:
> * `dict` or `str` representing a JSON or YAML file name to be parsed using `logging.config.dictConfig`
> * `str` representing an INI file name to be parsed using `logging.config.fileConfig`
>
> The `logging.config.fileConfig`, however, accepts not only the file name, but possibly also an instance of `configparser.RawConfigParser` and even an open file object. Passing the instance of `configparser.RawConfigParser` (or its subclass) would be especially useful as this can accumulate configs from multiple files (`cp.read()` can be called repeatedly on different files updating its config values) instead of just a single one.
> 
> I am proposing this trivial change to allow the `log_config` option to also accept the `*ConfigParser` instance or a general file object.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,3 +1,7 @@
+from __future__ import annotations
+
+import configparser
+import io
 import json
 import logging
 import os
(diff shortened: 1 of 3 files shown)


karpetrosyan (commented):
These two options, which are listed in the `logging` documentation, can help with project configuration management 
> Changed in version 3.4: An instance of a subclass of [RawConfigParser](https://docs.python.org/3/library/configparser.html#configparser.RawConfigParser) is now accepted as a value for fname. This facilitates:

> Use of a configuration file where logging configuration is just part of the overall application configuration.

> Use of a configuration read from a file, and then modified by the using application (e.g. based on command-line parameters or other aspects of the runtime environment) before being passed to fileConfig.

I like how `fileConfig` is used in this way.

Kludex (approved):

PR #1996, opened 2023-06-07
Drop support for Python 3.7

Kludex:
# Summary

Python 3.7 EOL is approaching: https://endoflife.date/python.

This PR drops support for Python 3.7.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: "actions/checkout@v3"
       - uses: "actions/setup-python@v4"
         with:
-          python-version: 3.7
+          python-version: "3.8"
       - name: "Install dependencies"
         run: "scripts/install"
       - name: "Build package & docs"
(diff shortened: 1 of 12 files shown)


zanieb (approved):
Thoughts on including a warning about this first? e.g. https://github.com/PrefectHQ/prefect/blob/382ebdcc54b1a45ce6e4cc0dfcfaf1fd09939ead/src/prefect/__init__.py#L23-L34

Kludex:
> Thoughts on including a warning about this first? e.g. [PrefectHQ/prefect@`382ebdc`/src/prefect/__init__.py#L23-L34](https://github.com/PrefectHQ/prefect/blob/382ebdcc54b1a45ce6e4cc0dfcfaf1fd09939ead/src/prefect/__init__.py#L23-L34)

Since we change the "requires-python" value, I don't think that's needed.

Also, is there someone that does the same besides Prefect?

zanieb:
@Kludex I think that just means pip won't pick up new versions if the user is on 3.7 which could be confusing without any additional context 🤷‍♀️ I don't know of other projects that are displaying a warning though. I don't think it's a big deal — Prefect has a lot more beginner-level users.

PR #1998, opened 2023-06-08
Revert #1804

Kludex:
This PR reverts #1804.

The idea there was to make sure you could run the test suite on another port besides 8000, but it's a flaky setup - I'm unsure why, but it fails too much, it's better to revert and then if someone wants, they can try to find out the reason.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,4 @@
-import contextlib
 import os
-import socket
 import ssl
 from copy import deepcopy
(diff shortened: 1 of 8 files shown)


Kludex:
Oh wait! It looks like that test fails anyway 🤣 

So it's not because of this! Amazing 😎 

cc @br3ndonland

Kludex:
Sorry the noise, folks.

PR #1999, opened 2023-06-08
Drop asgiref dependency at all

### LINKED ISSUE #1991: Create own types instead of using `asgiref`
We are using `asgiref` only for typing, but it has been super bothersome over the years.

PR welcome to create a new `_typing.py`/`_types.py` module on Uvicorn with the needed ASGI types.

Kludex:
- Closes #1991

### DIFF
--- a/uvicorn/_types.py
+++ b/uvicorn/_types.py
@@ -61,45 +61,6 @@ WSGIApp = Callable[[Environ, StartResponse], Union[Iterable[bytes], BaseExceptio
 
 # ASGI
 
-
-__all__ = (
-    "ASGIVersions",
-    "HTTPScope",
-    "WebSocketScope",
-    "LifespanScope",
-    "WWWScope",
-    "Scope",
-    "HTTPRequestEvent",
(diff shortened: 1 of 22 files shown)


uvicorn/_types.py
```diff
@@ -1,14 +1,320 @@
...
-    typing.Type[BaseException], BaseException, typing.Optional[types.TracebackType]
+Environ = MutableMapping[str, Any]
+ExcInfo = Tuple[Type[BaseException], BaseException, Optional[types.TracebackType]]
+StartResponse = Callable[[str, Iterable[Tuple[str, str]], Optional[ExcInfo]], None]
+WSGIApp = Callable[[Environ, StartResponse], Union[Iterable[bytes], BaseException]]
+
+# ASGI
+
+
+__all__ = (
```
zanieb: Should this be at the top and include the WSGI types?

Kludex: That should be removed. Good catch

zanieb (approved):

PR #2001, opened 2023-06-08
Wait for client on `test_connection_lost_before_handshake_complete`

Kludex:
This PR solves the _flakyness_ on `test_connection_lost_before_handshake_complete`.

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -11,7 +11,7 @@ if [ -z $GITHUB_ACTIONS ]; then
     scripts/check
 fi
 
-${PREFIX}coverage run --debug config -m pytest "$@" -s
+${PREFIX}coverage run --debug config -m pytest "$@"
 
 if [ -z $GITHUB_ACTIONS ]; then
     scripts/coverage
--- a/scripts/test
+++ b/scripts/test
@@ -11,7 +11,7 @@ if [ -z $GITHUB_ACTIONS ]; then
     scripts/check
 fi
 
-${PREFIX}coverage run --debug config -m pytest "$@"
+${PREFIX}coverage run --debug config -m pytest "$@" -s
 
 if [ -z $GITHUB_ACTIONS ]; then
     scripts/coverage
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -750,7 +750,7 @@ async def test_connection_lost_before_handshake_complete(
         await asyncio.sleep(0.1)
         send_accept_task.set()
 
-    task.cancel()
+    await task
     assert response is not None
     assert response.status_code == 500, response.text
     assert response.text == "Internal Server Error"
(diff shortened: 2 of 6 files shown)


Kludex:
Where is my pipeline? 🤔 
Screenshot 2023-06-08 at 21 05 39

tests/protocols/test_websocket.py
```diff
@@ -744,13 +746,11 @@ async def websocket_session(uri):
         port=unused_tcp_port,
     )
     async with run_server(config):
-        task = asyncio.create_task(
-            websocket_session(f"ws://127.0.0.1:{unused_tcp_port}")
-        )
+        asyncio.create_task(websocket_session(f"ws://127.0.0.1:{unused_tcp_port}"))
```
euri10: the `task =` is important please keep it

euri10 (changes requested):

tests/protocols/test_websocket.py
Kludex: It's important when you don't want the GC to clean it, since it's a weak ref, but in this case since all the logic happens in the same function scope, I don't think I need it?

euri10: it's been a while so I dont remember exactly but when I reworked the whole test suite and added strict settings to pytest this helped a lot.

euri10 (approved):

Kludex:
The test is hanging. I'm investigating.

Kludex:
The test is hanging at  `await response_received.wait()`, but I'm not sure what's the race condition here... 🤔

Kludex:
It looks like the success rate increases with a print 🤣

scripts/test
```diff
@@ -11,7 +11,7 @@ if [ -z $GITHUB_ACTIONS ]; then
     scripts/check
 fi
 
-${PREFIX}coverage run --debug config -m pytest "$@"
+${PREFIX}coverage run --debug config -m pytest "$@" -s
```
Kludex:
```suggestion
${PREFIX}coverage run --debug config -m pytest "$@"
```

tests/protocols/test_websocket.py
```diff
@@ -726,6 +726,7 @@ async def app(scope, receive, send):
     async def websocket_session(uri):
         nonlocal response
         async with httpx.AsyncClient() as client:
+            print("2.5")
```
```suggestion
```

tests/protocols/test_websocket.py
```diff
@@ -749,12 +749,13 @@ async def websocket_session(uri):
         )
         await asyncio.sleep(0.1)
         send_accept_task.set()
+        await asyncio.sleep(0.1)
```
Kludex, line 752: Only this tick was needed 😄

Kludex:
I've run the same commit 8x, and it passed on all of them. Should be fine now (...or I'm lucky). 👍

PR #2002, opened 2023-06-09
Handle malformed HTTP request with multiple Connection headers

zioproto:
This PR fixes a problem with Websockets when running Uvicorn on GitHub Codespaces

Detailed discussion:
https://github.com/orgs/community/discussions/57596

# Summary

Please do not merge quickly, I am not sure if this is a Uvicorn bug or a problem with the GitHub codespaces reverse proxy sending a malformed HTTP requests.

According to [RFC7230](https://datatracker.ietf.org/doc/html/rfc7230) there should not be duplicated Connection headers.

> A sender MUST NOT generate multiple header fields with the same field
name in a message unless either the entire field value for that
header field is defined as a comma-separated list [i.e., #(values)]
or the header field is a well-known exception (as noted below).

I am running Uvicorn in GitHub Codespaces. I am forwarding a port with visibility public and port protocol HTTP.
My first Websocket connection works, but after a few seconds the Websocket endpoint starts returning HTTP 404.

I captured traffic at the Uvicorn server with tcpdump and I figured out that when I receive a HTTP 404 the HTTP request has a duplicated `Connection` header.

Screenshot 2023-06-09 at 08 46 22

This is a problem and looking at the Uvicorn implementation it will make the Websockets connection fail to upgrade randomly depending on which one of the duplicated `Connection` header is read first.

https://github.com/encode/uvicorn/blob/1cb58c707ae7203f3b24e714947e7ab025b20895/uvicorn/protocols/http/h11_impl.py#L173-L183

I confirm that with the patch proposed in this PR fixes the problem completely because `connection` will be a list with the values `['upgrade', 'keep-alive']`. However it could be that the request is just malformed and Uvicorn should not implement anything to cope with malformed HTTP requests.

Please review.

# Checklist

- [X] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -175,7 +175,7 @@ class H11Protocol(asyncio.Protocol):
         upgrade = None
         for name, value in self.headers:
             if name == b"connection":
-                connection = [token.lower().strip() for token in value.split(b",")]
+                connection += [token.lower().strip() for token in value.split(b",")]
             if name == b"upgrade":
                 upgrade = value.lower()
         if b"upgrade" in connection:


Kludex:
image

zioproto:
@Kludex do you think I need to update the current tests in place at `tests/protocols/test_websocket.py` ?

Like implementing a test:
`test_websocket_upgrade_with_multiple_connection_headers` ?

Do you think that is a valid HTTP Request ?

Kludex:
> According to [RFC7230](https://datatracker.ietf.org/doc/html/rfc7230) there should not be duplicated Connection headers.

Then we shouldn't change uvicorn. 😅

Kludex:
Does the same happens with hypercorn?

zioproto:
> Does the same happens with hypercorn?

I dont know, I never used it.

This is confirmed being GitHub Codespaces sending duplicated `Connection:` header when port forwarding.

I will change the PR to return a proper HTTP return code in case of multiple `Connection` headers.

Should I just return a HTTP 400 error? 
I read the [4XX client errors](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) but I can't find a more specific one.

Kludex:
> I dont know, I never used it.

It would be interesting to check the behavior of another server implementation.

> I will change the PR to return a proper HTTP return code in case of multiple Connection headers.

No need. I think the current behavior is fine.

Kludex:
Since GitHub is in fault, and you got a solution for the problem, I'll be closing the PR.

Thanks for the investigation. 🙏

PR #2010, opened 2023-06-17
Improve shutdown process in Multiprocess mode

rbtz-openai:
# Summary

This PR explicitly closes listening sockets in parent processe early in the shutdown process.  That way, no new connections can hit workers after shutdown is initiated by the parent process.

While here, speedup shutdown by splitting process shutdown into terminate and join loop, so all child processes are shutdown in parallel.

# Current behaviour

Here is a sample application:
```
import asyncio

from fastapi import FastAPI

app = FastAPI()


@app.get("/")
async def root():
    return {"Hello": "World"}


@app.on_event("shutdown")
async def shutdown_event():
    print("Shutdown event")
    await asyncio.sleep(10)
    print("Shutdown event done")


@app.get("/sleep")
async def sleep():
    """can be used to test graceful shutdown"""
    await asyncio.sleep(60)

```

It is ran with:
```
uvicorn app:app --workers 2
```

In a second window we can run simulated traffic generator, a curtesy of curl:
```
while :; do curl --connect-timeout 0.1 --max-time 1 localhost:8000; sleep 1; done
```

In third window we `pkill(1)` the parent:
```
pkill -f 'uvicorn app:app'
```

What is seen in uvicorn logs:
```
(fast) ➜  app uvicorn app:app --workers 2
INFO:     Started parent process [87167]
INFO:     Started server process [87194]
INFO:     Started server process [87195]
```
Normal traffic:
```
INFO:     127.0.0.1:63935 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:63936 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:63938 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:63939 - "GET / HTTP/1.1" 200 OK
```

`SIGTERM` received:
```
INFO:     Shutting down
Shutdown event
```

At this point only a single worker receives traffic:
```
INFO:     127.0.0.1:63940 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:63942 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:63943 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:63944 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:63945 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:63946 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:63947 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:63948 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:63949 - "GET / HTTP/1.1" 200 OK
INFO:     127.0.0.1:63950 - "GET / HTTP/1.1" 200 OK
Shutdown event done
INFO:     Finished server process [87194]
INFO:     Shutting down
```

Now we closed the last workers' sockets but parent still has opened socket, so we are dropping requests on the floor:
```
Shutdown event
Shutdown event done
INFO:     Finished server process [87195]
INFO:     Stopping parent process [87167]
```
After that point we are throwing connection-refused.

From the client it looks like:
```
$ while :; do curl --connect-timeout 0.1 --max-time 1 localhost:8000; sleep 1; done
{"Hello":"World"}{"Hello":"World"}{"Hello":"World"}
# At this point we sent SIGTERM.  All the following responses come from a single alive worker:
{"Hello":"World"}{"Hello":"World"}{"Hello":"World"}{"Hello":"World"}{"Hello":"World"}{"Hello":"World"}{"Hello":"World"}{"Hello":"World"}{"Hello":"World"}{"Hello":"World"}{"Hello":"World"}
# Here this worker dies and we start dropping requests on the floor (non-retrieable).
curl: (28) Operation timed out after 1001 milliseconds with 0 bytes received
curl: (28) Operation timed out after 1003 milliseconds with 0 bytes received
curl: (28) Operation timed out after 1004 milliseconds with 0 bytes received
curl: (28) Operation timed out after 1002 milliseconds with 0 bytes received
curl: (28) Operation timed out after 1005 milliseconds with 0 bytes received
# Here master quits and we get clean (retriable connection refused)
curl: (7) Failed to connect to localhost port 8000 after 4 ms: Couldn't connect to server
curl: (7) Failed to connect to localhost port 8000 after 4 ms: Couldn't connect to server
curl: (7) Failed to connect to localhost port 8000 after 4 ms: Couldn't connect to server
```

# New behaviour
New version:
* Closes listening socket in parent process immediately.
* Shuts down all worker processes in parallel, which has two side effects:
  * No load imbalance between workers during shutdown.
  * Faster shutdown porcess.

PS. The same situation can be reproduced w/o shutdown lifecycle event with `--timeout-graceful-shutdown 30` and an endless stream of slow requests

# Checklist

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -267,8 +267,6 @@ class Server:
         # Stop accepting new connections.
         for server in self.servers:
             server.close()
-        for sock in sockets or []:
-            sock.close()
         for server in self.servers:
             await server.wait_closed()
 
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -63,8 +63,13 @@ class Multiprocess:
             self.processes.append(process)
 
     def shutdown(self) -> None:
+        for sock in self.sockets:
+            sock.close()
+
         for process in self.processes:
             process.terminate()
+
+        for process in self.processes:
             process.join()
 
         message = "Stopping parent process [{}]".format(str(self.pid))


uvicorn/supervisors/multiprocess.py
```diff
@@ -63,8 +63,13 @@ def startup(self) -> None:
             self.processes.append(process)
 
     def shutdown(self) -> None:
+        for sock in self.sockets:
+            sock.close()
+
```
Kludex, line 68: Why do you need to close the sockets here again?

rbtz-openai: In UNIX-based OSes one needs to close all instances of a listening socket for it to unbind (incl. ones inherited by forks).  In this case this includes:
* master process.
* all childen.

If you're hinting that in some OS'es (e.g. Windows, where my knowledge is quite limited) this would lead to an error I can wrap this in: `try/except socket.error`.

rbtz-openai: You can observe the behaviour of stalled http requests during shutdown with a server that has a health check http handler and a a shutdown lifespan handler:
```
    @app.on_event("shutdown")
    async def shutdown():
        await asyncio.sleep(60)
```

rbtz-openai:
@Kludex A friendly ping on this ^

uvicorn/server.py
```diff
@@ -267,8 +267,6 @@ async def shutdown(self, sockets: Optional[List[socket.socket]] = None) -> None:
         # Stop accepting new connections.
         for server in self.servers:
             server.close()
-        for sock in sockets or []:
-            sock.close()
```
rbtz-openai, line 271: This is not needed.

[asyncio.loop.create_server](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_server) specifies:

> Note The sock argument transfers ownership of the socket to the server created. To close the socket, call the server’s [close()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.Server.close) method.

[asyncio.Server](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.Server) confirms that with:

> close()(https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.Server.close)
Stop serving: close listening sockets and set the [sockets](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.Server.sockets) attribute to None.

rbtz-openai:
ref:
* https://github.com/encode/uvicorn/pull/2183

Kludex:
- Closing this, since we are refactoring the multiprocess module on https://github.com/encode/uvicorn/pull/2183.

PR #1934, opened 2023-04-09
Fix `x_forwarded_proto` for websockets in `proxy_headers.py`

1997cui:
As discussed in #1933, the upstream proxies (i.e., Caddy) will always pass `http` or `https` in `x_forwarded_proto`, even if it is a WebSocket Upgrade request. However, storing the `http` or `https` in a WebSocket scope is not correct. In this pull request, I added some if statements to solve this issue.

I know there are no unit tests, because I don't know how to write, and help wanted.

PR #2011, opened 2023-06-19
Fix `x_forwarded_proto` for websockets in `proxy_headers.py`

1997cui:
# Summary

As discussed in https://github.com/encode/uvicorn/discussions/1933, the upstream proxies (i.e., Caddy) will always pass http or https in x_forwarded_proto, even if it is a WebSocket Upgrade request. However, storing the http or https in a WebSocket scope is not correct. In this pull request, I added some if statements to solve this issue.

I know there are no unit tests, because I don't know how to write, and help wanted.

This PR is based on #1934 based on suggestions from @Kludex and @aminalaee 

# Checklist

- [ x ] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [  ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ x ] I've updated the documentation accordingly.

### DIFF
--- a/uvicorn/middleware/proxy_headers.py
+++ b/uvicorn/middleware/proxy_headers.py
@@ -59,8 +59,15 @@ class ProxyHeadersMiddleware:
                 if b"x-forwarded-proto" in headers:
                     # Determine if the incoming request was http or https based on
                     # the X-Forwarded-Proto header.
-                    x_forwarded_proto = headers[b"x-forwarded-proto"].decode("latin1")
-                    scope["scheme"] = x_forwarded_proto.strip()
+                    x_forwarded_proto = (
+                        headers[b"x-forwarded-proto"].decode("latin1").strip()
+                    )
+                    if scope["type"] == "websocket":
+                        scope["scheme"] = (
+                            "wss" if x_forwarded_proto == "https" else "ws"
+                        )
+                    else:
+                        scope["scheme"] = x_forwarded_proto
 
                 if b"x-forwarded-for" in headers:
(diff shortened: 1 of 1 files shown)


uvicorn/middleware/proxy_headers.py
```diff
@@ -59,8 +59,15 @@ async def __call__(
                 if b"x-forwarded-proto" in headers:
                     # Determine if the incoming request was http or https based on
                     # the X-Forwarded-Proto header.
-                    x_forwarded_proto = headers[b"x-forwarded-proto"].decode("latin1")
-                    scope["scheme"] = x_forwarded_proto.strip()
+                    x_forwarded_proto = (
+                        headers[b"x-forwarded-proto"].decode("latin1").strip()
+                    )
+                    if scope["type"] == "websocket":
+                        scope["scheme"] = (
```
aminalaee, line 66: This line is missing a test coverage. Can you please add that?

aminalaee (changes requested):

uvicorn/middleware/proxy_headers.py
1997cui: > I know there are no unit tests, because I don't know how to write, and help wanted.

Can you give me some examples of how to do a websocket test client? It seems that `httpx` does not support it. #2012

Kludex: That issue is not related

Kludex:
- Closed by #2043

Thanks @1997cui for the interest! :)

PR #2019, opened 2023-06-28
draft: support limitation for header and message body

magiskboy:
# Summary

As a webserver, uvicorn should be implemented limitation of header and message body.

In this PR, uvicorn check length of header and message body when parsing, if those is larger than settings, responds as soon as to client

# Checklist


zanieb:
Hi @magiskboy thanks for contributing!

Is there a discussion or GitHub issue that this is resolving?

Kludex:
> Hi @magiskboy thanks for contributing!
> 
> Is there a discussion or GitHub issue that this is resolving?

I think there's an issue about limiting resources, but I don't think this PR addresses what it's trying to address.

magiskboy:
> 

I see that https://github.com/encode/uvicorn/discussions/1726 but, limit header size only support with h11 protocol. In this pr, limitation will be applied to httptools

magiskboy:
@Kludex could you tell me the reason? Do you have any idea?

Kludex:
> @Kludex could you tell me the reason? Do you have any idea?

You need to read this: https://github.com/encode/uvicorn/pull/1684, and the links there.

magiskboy:
I thought about that so parser should be support limitation because protocol is received only fully headers

PR #2020, opened 2023-06-28
Add worker process guard.

CrazyWolf2014:
# Summary

In some cases, the worker thread cannot be automatically restarted after the exception ends, resulting in the inability to provide services normally. You need to end the main process each time and restart it. To solve this problem, add worker process state detection to the main thread. If the worker process is dead, a new worker process is recreated.

# Checklist

- [*] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -57,6 +57,9 @@ Options:
   --workers INTEGER               Number of worker processes. Defaults to the
                                   $WEB_CONCURRENCY environment variable if
                                   available, or 1. Not valid with --reload.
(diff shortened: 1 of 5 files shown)


zanieb:
This looks like a duplicate of https://github.com/encode/uvicorn/pull/1942; you may want to read the discussion there.

CrazyWolf2014:
okay

Kludex:
> This looks like a duplicate of #1942; you may want to read the discussion there.

Yeah, let's continue there.

Thanks for the PR @CrazyWolf2014 :)

PR #2038, opened 2023-07-09
Improve error message on import errors

mikeedjones:
# Summary

Addresses #2034 by adding additional detail in the error message when an import error is raised. 

# Checklist


iudeen:
Checked this locally. It works. LGTM. 👍

uvicorn/importer.py
```diff
@@ -22,17 +22,19 @@ def import_from_string(import_str: Any) -> Any:
...
+        raise ImportFromStringError(
+            message.format(module_str=module_str, exc=exc)
+        ) from exc
 
     instance = module
     try:
         for attr_str in attrs_str.split("."):
             instance = getattr(instance, attr_str)
-    except AttributeError:
+    except AttributeError as exc:
```
Kludex: Please remove this. We are not using it.

uvicorn/importer.py
```diff
@@ -22,17 +22,19 @@ def import_from_string(import_str: Any) -> Any:
     except ImportError as exc:
         if exc.name != module_str:
             raise exc from None
```
Kludex: Do you know, or would you like to investigate why this conditional exists? There is no test that covers this... 🤔

uvicorn/importer.py
```diff
@@ -22,17 +22,19 @@ def import_from_string(import_str: Any) -> Any:
...
-        message = 'Could not import module "{module_str}".'
-        raise ImportFromStringError(message.format(module_str=module_str))
+        message = 'Could not import module "{module_str}" due to ImportError: {exc}'
+        raise ImportFromStringError(
+            message.format(module_str=module_str, exc=exc)
+        ) from exc
```
Kludex: The error message with this change is horrible. What if we add a `except ModuleNotFoundError` above the current except? Like:
```py
except ModuleNotFoundError:
   ...
except ImportError:
   ...
```

I'm interested to know the conditional above as well...

Kludex (changes requested):

tests/importer/raise_import_error_circular_import_bar.py
Kludex, line 1: We need to remove these 3 files here. We need something like:
```py
# tmp_path is a fixture from pytest
def test(tmp_path: Path):
    d = tmp_path / "dir"
    d.mkdir()
    file = d / "main.py"
    file.touch()
    file.write(<code>)
```

I can do this if you don't want/can't.

tests/importer/test_importer.py
```diff
@@ -1,8 +1,21 @@
+from contextlib import contextmanager
+import os
+from pathlib import Path
+
 import pytest
 
 from uvicorn.importer import ImportFromStringError, import_from_string
 
 
+@contextmanager
```
Kludex: Please use tmp_path from pytest.

mikeedjones: If we write the code to a tmp_path then we also need to do something like monkeypatch-ing the PYTHONPATH env var so that the importer knows the tmp_path is somewhere to look for python - That's an option but I'm not sure how [stable messing with pythonpath is](https://github.com/pytest-dev/pytest/issues/4913). Could also adjust the base temp dir for this test. Both seemed more involved than writing the context manager.

Kludex: Then the previous code was fine. (Sorry)

mikeedjones: could add a mark to adjust the base tmp dir for this test? I think that would work.

uvicorn/importer.py
mikeedjones: If the string-imported module imports a module which does not exist (has a syntax error etc) the exc.name and the module_str will have different values - since the exception is being raised on the import of the sub-module. This is now caught by `ModuleImportError` and reported correctly so i have taken this check out.

The ImportFromString error is handled by an explicit `except: sys.exit(1)` in `config.py`. Slightly concerned that the behaviour now is slightly different.

uvicorn/importer.py
mikeedjones: removed

uvicorn/importer.py
mikeedjones: Added

Kludex:
I'm going to modify the code, please don't push any changes in the meantime.

Kludex:
- I've added you as a co-author of https://github.com/encode/uvicorn/pull/2040. Thanks for the PR.

I prefer to have minimal changes.

PR #2043, opened 2023-07-12
Fix `x_forwarded_proto` for websockets

### LINKED ISSUE #1935: `middleware/proxy_headers.py`: Inconsistent `x-forwarded-proto` for websockets
### Discussed in https://github.com/encode/uvicorn/discussions/1933

<sup>Originally posted by **1997cui** April  9, 2023</sup>
Hi,
I recently noticed a potential bug in the `proxy_headers.py` middleware:
 ## The problem 
Upstream reverse proxy (i.e. Caddy or Nginx [Check scheme variable](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_scheme) ) always puts `http` or `https` in `x-forwarded-proto`, even if the request has an `Upgrade header` and is a WebSocket. [This behavior complies with the protocol](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto)

The `proxy_headers.py` passes the value along into the `scope["scheme"]`, even if it is a Websocket. This is not correct. According to the [spec](https://asgi.readthedocs.io/en/latest/specs/www.html#websocket-connection-scope), the scheme should be either `ws` or `wss`. `https` or `http` in `websocket` type cause routing errors.

### app.py
``` python
from quart import Quart, websocket

app = Quart(__name__)

@app.websocket('/test')
async def test():
    await websocket.send('Hello World!')
```

### Caddyfile
```
example.com {
    route /test {
        reverse_proxy http://localhost:5000 {
            header_up X-Forwarded-For {remote_host}
            header_up X-Forwarded-Proto {scheme}
        }
    }
}
```

### Run it

 - We accept the upstream proxy header
``` bash
gunicorn app:app --worker-class uvicorn.workers.UvicornWorker --bind 127.0.0.1:5000 --forwarded-allow-ips="*"
```
Result
``` bash
wscat -c "wss://slu.t.cuitian1.com/test"
error: Unexpected server response: 403
```
 - We ignore the upstream header
``` bash
gunicorn app:app --worker-class uvicorn.workers.UvicornWorker --bind 127.0.0.1:5000 --forwarded-allow-ips=""
```
result
``` bash
wscat -c "wss://slu.t.cuitian1.com/test"
Connected (press CTRL+C to quit)
< Hello World!
Disconnected (code: 1000, reason: "")
```
</div>

aminalaee:
# Summary

Replaces https://github.com/encode/uvicorn/pull/2011
Closes https://github.com/encode/uvicorn/issues/1935

# Checklist

### DIFF
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -3,9 +3,22 @@ from typing import List, Union
 import httpx
 import pytest
 
+from tests.protocols.test_http import HTTP_PROTOCOLS
 from tests.response import Response
+from tests.utils import run_server
 from uvicorn._types import ASGIReceiveCallable, ASGISendCallable, Scope
+from uvicorn.config import Config
 from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
+from uvicorn.protocols.websockets.wsproto_impl import WSProtocol
+
+try:
+    import websockets.client
+
+    from uvicorn.protocols.websockets.websockets_impl import WebSocketProtocol
+
+    WS_PROTOCOLS = [WSProtocol, WebSocketProtocol]
+except ImportError:  # pragma: nocover
+    WS_PROTOCOLS = []
 
 
 async def app(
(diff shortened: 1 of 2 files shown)


Kludex (approved):

karpetrosyan:
You forgot to include it in the changelog.

Kludex:
> You forgot to include it in the changelog.

We don't do that on per PR yet on `uvicorn`.

PR #2048, opened 2023-07-15
Bump `httpx` and `h11`

Kludex:
`httpx` is limiting `h11` - and `h11` is a mandatory dependency of `uvicorn`.

I'm using the main branch of `h11` because we need the type improvements I've suggested there: https://github.com/python-hyper/h11/pull/161.

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -9,7 +9,6 @@ from typing import (
     Literal,
     Optional,
     Tuple,
-    Union,
     cast,
 )
 from urllib.parse import unquote
@@ -21,7 +20,6 @@ from uvicorn._types import (
     ASGI3Application,
     ASGIReceiveEvent,
     ASGISendEvent,
-    HTTPDisconnectEvent,
     HTTPRequestEvent,
     HTTPResponseBodyEvent,
     HTTPResponseStartEvent,
@@ -192,12 +190,11 @@ class H11Protocol(asyncio.Protocol):
                 self.logger.warning(msg)
                 self.send_400_response(msg)
                 return
-            event_type = type(event)
 
-            if event_type is h11.NEED_DATA:
+            if event is h11.NEED_DATA:
                 break
 
-            elif event_type is h11.PAUSED:
+            elif event is h11.PAUSED:
                 # This case can occur in HTTP pipelining, so we need to
                 # stop reading any more data, and ensure that at the end
                 # of the active request/response cycle we handle any
@@ -267,7 +264,7 @@ class H11Protocol(asyncio.Protocol):
                     self.flow.pause_reading()
                 self.cycle.message_event.set()
 
-            elif event_type is h11.EndOfMessage:
+            elif isinstance(event, h11.EndOfMessage):
                 if self.conn.our_state is h11.DONE:
                     self.transport.resume_reading()
                     self.conn.start_next_cycle()
@@ -296,7 +293,7 @@ class H11Protocol(asyncio.Protocol):
 
     def send_400_response(self, msg: str) -> None:
         reason = STATUS_PHRASES[400]
-        headers = [
+        headers: List[Tuple[bytes, bytes]] = [
             (b"content-type", b"text/plain; charset=utf-8"),
             (b"connection", b"close"),
         ]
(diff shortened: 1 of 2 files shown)


uvicorn/protocols/http/h11_impl.py
```diff
@@ -214,7 +205,7 @@ def handle_events(self) -> None:
                 self.flow.pause_reading()
                 break
 
-            elif event_type is h11.Request:
+            elif isinstance(event, h11.Request):
```
Kludex, line 205: How more expensive are those compared to `is`?

uvicorn/protocols/http/h11_impl.py
```diff
@@ -497,10 +489,10 @@ async def send(self, message: "ASGISendEvent") -> None:
 
             # Write response status line and headers
             reason = STATUS_PHRASES[status_code]
-            event = h11.Response(
-                status_code=status_code, headers=headers, reason=reason
-            )
-            output = self.conn.send(event)
+            # fmt: off
+            response = h11.Response(status_code=status_code, headers=headers, reason=reason) # noqa: E501
```
Kludex: I'm starting to have the feeling that 88 is far worse than 120 for `uvicorn`.

uvicorn/protocols/http/h11_impl.py
Kludex: Well... I think we care far more about correctness than performance on `h11`.

uvicorn/protocols/http/h11_impl.py
```diff
@@ -557,13 +547,9 @@ async def receive(self) -> "ASGIReceiveEvent":
...
-                "body": self.body,
-                "more_body": self.more_body,
-            }
-            self.body = b""
-
+            return {"type": "http.disconnect"}
+        # fmt: off
+        message = {"type": "http.request", "body": self.body, "more_body": self.more_body}  # noqa: E501
+        # fmt: on
+        self.body = b""
```
zanieb: While you're here, a comment about why we're clearing `self.body` might be nice

uvicorn/protocols/http/h11_impl.py
```diff
@@ -543,10 +532,11 @@ async def send(self, message: "ASGISendEvent") -> None:
...
         if self.waiting_for_100_continue and not self.transport.is_closing():
-            event = h11.InformationalResponse(
-                status_code=100, headers=[], reason="Continue"
-            )
-            output = self.conn.send(event)
+            # fmt: off
+            headers: List[Tuple[str, str]] = []
+            informational_response = h11.InformationalResponse(status_code=100, headers=headers, reason="Continue") # noqa: E501
+            output = self.conn.send(event=informational_response)
+            # fmt: on
```
zanieb: I don't really mind the `event` variable (i.e. instead of `informational_response`). It feels a little verbose to specify the type of the response in the variable name (since there is type inference here). You're of course welcome to use whatever style makes you happy here though :)

uvicorn/protocols/http/h11_impl.py
zanieb: These fmt on/off E501 ignores are definitely sad to see. Why not split the response construction over multiple lines? Maybe you _should_ bump the line length for the project, although 120 chars is pretty long.

uvicorn/protocols/http/h11_impl.py
zanieb: I recall them being concerned about the cost of doing that check when I made a similar change in a hot loop in h11. I think it's okay though. Will this break Uvicorn for users that don't have h11 from git or did this change land in an earlier release?

Kludex: Let me check

Kludex: It does work. 👍

zanieb (approved):

uvicorn/protocols/http/h11_impl.py
Kludex: Yeah, I'm going to remove them.

uvicorn/protocols/http/h11_impl.py
Kludex: Ah, I thought we were already using the `event` variable name in this scope. I've changed for `event` once again.

Kludex:
Before merging this, I need to test once again with previous h11 versions.

Kludex:
Works fine. 👍

PR #2049, opened 2023-07-16
Create `ws_protocol_cls` fixture

Kludex:
This PR aims to remove the `WS_PROTOCOLS` "constant" that we have on the `test_websockets.py`.

### DIFF
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, List, Union
+from typing import TYPE_CHECKING, List, Type, Union
 
 import httpx
 import pytest
@@ -116,7 +116,7 @@ async def test_proxy_headers_invalid_x_forwarded_for() -> None:
 @pytest.mark.anyio
 @pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
 async def test_proxy_headers_websocket_x_forwarded_proto(
-    ws_protocol_cls: "type[WSProtocol] | type[WebSocketProtocol]",
+    ws_protocol_cls: "Type[WSProtocol | WebSocketProtocol]",
     http_protocol_cls,
     unused_tcp_port: int,
 ) -> None:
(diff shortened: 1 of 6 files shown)


tests/middleware/test_proxy_headers.py
```diff
@@ -1,24 +1,19 @@
...
-
-try:
-    import websockets.client
 
+if TYPE_CHECKING:
     from uvicorn.protocols.websockets.websockets_impl import WebSocketProtocol
-
-    WS_PROTOCOLS = [WSProtocol, WebSocketProtocol]
-except ImportError:  # pragma: nocover
-    WS_PROTOCOLS = []
```
Kludex, line 21: This was incorrect. The `WebSocketProtocol` will always be available.

tests/supervisors/test_reload.py
Kludex, line 1: I need to create an issue to investigate those. They are failing locally quite often.

EDIT: I'm usually using an M2.

Kludex (commented):
I need to replace `type` by `Type`. 😢

iudeen (commented):
This looks good!

tests/supervisors/test_reload.py
```diff
@@ -232,7 +253,18 @@ def test_should_not_reload_when_only_subdirectory_is_watched(
...
         reloader.shutdown()
 
-    @pytest.mark.parametrize("reloader_class", [WatchFilesReload, WatchGodReload])
+    @pytest.mark.parametrize(
+        "reloader_class",
+        [
+            pytest.param(
+                WatchFilesReload,
+                marks=pytest.mark.skipif(
+                    sys.platform == "darwin", reason="Flaky on MacOS"
```
aminalaee: Don't you want to narrow it down to only Apple Silicon architecture?
Maybe use platform module for that?

aminalaee (approved):

tests/supervisors/test_reload.py
Kludex: How do I do that?

aminalaee: I think something along this to check both OS and CPU architecture:

```py
import platform

system = platform.system() # "Darwin"
processor = platform.processor() # "arm"

@pytest.mark.skipif(system == "Darwin" and processor == "arm", reason="Flakey on MacOS")
```

Kludex: done

PR #2055, opened 2023-07-18
Add pragma on unreachable code from different OSs (encode#2036)

Kludex:
Ignore.


Kludex:
oh sorry, it was supposed to be on my fork.

PR #2059, opened 2023-07-25
Unblock periodically the main process to check for signal

### LINKED ISSUE #1972: Ctrl+C in terminal broken on windows
With last release of Uvicorn 0.22.0 doing CTRL+C in a windows terminal doesn't work anymore. 
Everything is frozen and it is impossible to stop it. In addition the command --reload is broken too, because it doesn't restart anymore, probably because of the same issue.

Tried on Windows 11 and Powershell 7.3.4.

Downgrading to version 0.21.1 makes it works again.

humrochagf:
# Summary

This adds a periodic unblock of the main process to check for signals sent like SIGINT or SIGTERM.
In windows the main process does not listen to signals when blocked by the wait call.

Closes: #1972

# Checklist

### DIFF
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import signal
+import sys
 import threading
 from multiprocessing.context import SpawnProcess
 from socket import socket
@@ -42,7 +43,9 @@ class Multiprocess:
 
     def run(self) -> None:
         self.startup()
-        self.should_exit.wait()
+        while True:
+            if self.should_exit.wait(self.config.reload_delay):
+                break
         self.shutdown()
 
     def startup(self) -> None:
@@ -64,7 +67,10 @@ class Multiprocess:
 
     def shutdown(self) -> None:
         for process in self.processes:
-            process.terminate()
+            if sys.platform == "win32":
+                self.should_exit.set()  # pragma: py-not-win32
+            else:
+                process.terminate()  # pragma: py-win32
             process.join()
 
         message = "Stopping parent process [{}]".format(str(self.pid))


uvicorn/supervisors/multiprocess.py
```diff
@@ -42,7 +43,9 @@ def signal_handler(self, sig: int, frame: Optional[FrameType]) -> None:
 
     def run(self) -> None:
         self.startup()
-        self.should_exit.wait()
+        while True:
+            if self.should_exit.wait(self.config.reload_delay):
+                break
```
Kludex, line 48: This is a setting that should be used only for `--reload`.

humrochagf: Oh right, do you think it makes sense to add a config just for this? Or is it better to use a fixed timeout/constant value?

Kludex: Isn't there an alternative solution? This problem started happening after a feature we added, didn't it?

humrochagf: I tried to remove the signal added in #1909 to check but the issue was still there on windows. I'll try to diff from 0.21.1 and isolate the features to find the exact change that caused the issue.

Side note: I also noticed also a flaky startup issue happening on windows while debugging this that I'll dig more on it later.

Kludex: Yeah, tests on reload are flaky... I don't know why. You can add the `skipif` on them as well for the time being. I'll accept a PR for it.

humrochagf: Just tested with `0.21.1` and the error of `ctrl+c` is still there.

I'm running it on Windows 10 with powershell 7.3.6 but I also tested with `cmd` and got the same freeze.
Python version used for the test is `3.11.1` and the packages are those:
```
➜ pip freeze
annotated-types==0.5.0
anyio==3.7.1
click==8.1.6
colorama==0.4.6
fastapi==0.100.0
h11==0.14.0
idna==3.4
pydantic==2.1.1
pydantic_core==2.4.0
sniffio==1.3.0
starlette==0.27.0
typing_extensions==4.7.1
uvicorn==0.21.1
```

Also, the app used for the test was the fastapi base example:

```python
from typing import Union

from fastapi import FastAPI

app = FastAPI()


@app.get("/")
def read_root():
    return {"Hello": "World"}


@app.get("/items/{item_id}")
def read_item(item_id: int, q: Union[str, None] = None):
    return {"item_id": item_id, "q": q}
```

And the command I'm running is:

```
uvicorn main:app --workers 2
```

humrochagf:
Closing this one since #2183 seems to solve the issue on windows

Issue #1908, opened 2023-03-22
Switch WebSockets legacy implementation to Sans-I/O


Kludex:
Given https://github.com/aaugustin/websockets/issues/975 and https://github.com/aaugustin/websockets/issues/1310#issuecomment-1479780072 we should try to follow @aaugustin's recommendation.

We can discuss if we should deprecate the current `WebSocketProtocol`, or if we should replace it, and that's it.

aaugustin:
The documentation for this is here: https://websockets.readthedocs.io/en/stable/howto/sansio.html

I don't remember how uvicorn integrates websockets so I'm not sure how much work it would take to upgrade to the Sans-I/O implementation. It would probably result in a more straightforward and testable design, though.

As a reference, here's how websockets would use its own Sans-I/O layer: https://github.com/aaugustin/websockets/pull/1245

You might find it helpful to look at Sanic's integration too: https://github.com/sanic-org/sanic/pull/2158

I'm happy to answer questions and assist with reviews.

Kludex:
Is it possible to send data as soon as I have available on the server with [send_response](https://websockets.readthedocs.io/en/stable/reference/server.html#websockets.server.ServerConnection.send_response)?

I want to send the headers, and then the body, and then more body, and maybe more body after...

Kludex:
I'm a bit confused as to why the `send_*` methods don't return the data that `data_to_send` returns.

It looks like I'm in need to run `data_to_send` all the time... :thinking: Am I doing something wrong?

aaugustin:
That can't be a WebSocket handshake response; it's unsupported in websockets.

You aren't going to involve websockets anymore after sending such a response. You can ignore websockets and write directly to the network connection.

If it makes the implementation cleaner or easier, you could call websockets's `send_response` method with only the headers; then you would write the body as you generate it, without involving websockets.

Kludex:
> If it makes the implementation cleaner or easier, you could call websockets's send_response method with only the headers; then you would write the body as you generate it, without involving websockets.

So I `reject` with `websockets`, and then write the body? Just to confirm.

aaugustin:
Well, `reject` isn't going to be entirely helpful, because it wants to set Content-Length:

https://github.com/aaugustin/websockets/blob/ba1ed7a65cc876ff4e0fcd4dd4711402836475e2/src/websockets/server.py#L505

If you want to rely as much as possible on public API to avoid breaking websockets' invariants, here's my recommendation:

[...]

PR #2060, opened 2023-07-27
Websocket sansio implementataion

gourav-kandoria:
This PR provides the implementation of websocket protocol using websockets sansI/O layer.

task details - https://github.com/encode/uvicorn/issues/1908

## Checklist for @Kludex
- [ ] `data_receive` shouldn't be using `Exception`. What is the right exception there?
- [X] Check if `max_size` is being tested, and if not, test it. EDIT: I've checked it - it's not being tested.
- [ ] Add `WebSocketsSansIOProtocol` test for `max_size`.
- [x] Should we remove `CONT` and `PONG` related code?
- [ ] Add documentation.
- [x] Increase the coverage number.
- [x] Add `--ws websockets-sansio`.
- [ ] Check if https://github.com/encode/uvicorn/pull/2060#discussion_r1277501171 is true.
- [ ] Answer the following questions:
  - [ ] Should we deprecate `websockets` legacy version?
  - [ ] Should we add a `UserWarning` saying that it's experimental, and asking users to comment in a discussion X?


Kludex:
If you make the pipeline green, I can review it.

gourav-kandoria:
sure

gourav-kandoria:
@Kludex Could you please review it. all checks are passing now

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,392 @@
...
+)
+from uvicorn.server import ServerState
+
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal
+else:
+    from typing import Literal
+
+if typing.TYPE_CHECKING:
+    from asgiref.typing import (
```
Kludex: We are using uvicorn._types instead.

Kludex (commented):
I'll review on the weekend.

Kludex (commented):
I'm going to explain +- how is going to be my review process, so you can try to speed it up:

1. I'll run the test suite, and check the coverage, and then I'll try to understand why some lines are not covered.
2. I'll check the type hints, and make sure that they make sense. 
3. There are some tests that only work for `wsproto` or for `websockets`. I'm going to make sure that we are also running this protocol implementation where it makes sense.

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,392 @@
...
+        self.response: Response
+        self.curr_msg_data_type: str
+
+        self.read_paused = False
+        self.writable = asyncio.Event()
+        self.writable.set()
+
+        # Buffers
+        self.bytes: "bytes" = b""
+        print(len(self.tasks))
```
valentin994: I guess you wouldn't leave this in for master

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,392 @@
...
+        self.client = get_remote_addr(transport)
+        self.scheme = "wss" if is_ssl(transport) else "ws"
+
+        if self.logger.level <= TRACE_LOG_LEVEL:
+            prefix = "%s:%d - " % self.client if self.client else ""
+            self.logger.log(TRACE_LOG_LEVEL, "%sWebSocket connection made", prefix)
+
+    def connection_lost(self, exc: typing.Optional[Exception]) -> None:
+        self.connections.remove(self)
+        print("came in connection lost : ", exc)
```
valentin994: leftover print?

valentin994:
> I'm going to explain +- how is going to be my review process, so you can try to speed it up:
> 
>     1. I'll run the test suite, and check the coverage, and then I'll try to understand why some lines are not covered.
> 
>     2. I'll check the type hints, and make sure that they make sense.
> 
>     3. There are some tests that only work for `wsproto` or for `websockets`. I'm going to make sure that we are also running this protocol implementation where it makes sense.

I was looking at this task as well, but as [gourav-kandoria](https://github.com/gourav-kandoria) already did it I can help with pointing out few semantics or if I see something else. If you don't mind 😄

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,392 @@
...
+        if self.logger.level <= TRACE_LOG_LEVEL:
+            prefix = "%s:%d - " % self.client if self.client else ""
+            self.logger.log(TRACE_LOG_LEVEL, "%sWebSocket connection lost", prefix)
+        if self.handshake_initiated and not self.close_sent:
+            self.queue.put_nowait({"type": "websocket.disconnect", "code": 1006})
+
+    def data_received(self, data: bytes) -> None:
```
valentin994: https://websockets.readthedocs.io/en/stable/howto/sansio.html
if I'm following this right after receiving data you should call receive_eof()

uvicorn/protocols/websockets/websockets_sansio_impl.py
gourav-kandoria: @valentin994 yes, would remove these types of statement. For now opened the pr for just review and suggestions

uvicorn/protocols/websockets/websockets_sansio_impl.py
gourav-kandoria: @Kludex sure will change this

gourav-kandoria:
> 



> 

okay sure, got it. till then I will also go through the code at my end.

Kludex:
FYI, if you want to speed up the review process here, I need the lines on the report for the added file to be covered, or to understand why they aren't.

Screenshot 2023-08-08 at 07 42 48

I'll eventually come to this PR even if you don't do it, but it may take some time.

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,389 @@
...
+class WebSocketsSansIOProtocol(asyncio.Protocol):
+    def __init__(
+        self,
+        config: Config,
+        server_state: ServerState,
+        app_state: typing.Dict[str, typing.Any],
+        _loop: typing.Optional[asyncio.AbstractEventLoop] = None,
+    ) -> None:
+        if not config.loaded:
+            config.load()
```
gourav-kandoria: in the test_websockets . this line is not being covered for any implementation of websockets.
as an http upgrade is happening there and httpProtocol classes are passing already loaded config in that cases.
upon inspection, I found that this line was covered for websockets_impl.py due test_websocket_auto.py

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,389 @@
...
+            self.logger.log(TRACE_LOG_LEVEL, "%sWebSocket connection lost", prefix)
+        if self.handshake_initiated and not self.close_sent:
+            self.queue.put_nowait({"type": "websocket.disconnect", "code": 1006})
+
+    def data_received(self, data: bytes) -> None:
+        try:
+            self.conn.receive_data(data)
+        except Exception:
+            self.logger.exception("Exception in ASGI server")
+            self.transport.close()
```
gourav-kandoria: This lines(117-119) are not being covered. Because, I guess in the test suite there not any test which after connection establishment sends data which causes receive_data to throw exception such as data which don't follow websocket spec.

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,389 @@
...
+                self.queue.put_nowait({"type": "websocket.disconnect", "code": 1006})
+            self.transport.close()
+
+    def handle_events(self) -> None:
+        for event in self.conn.events_received():
+            if isinstance(event, Request):
+                self.handle_connect(event)
+            if isinstance(event, Frame):
+                if event.opcode == websockets.frames.Opcode.CONT:
+                    self.handle_cont(event)
```
gourav-kandoria, line 138: This line is not covered. as in test suite there is not any test which is sending data in multiple frames. So the whole data is coming in the first frame itself. with fin also set to true. 
Although I had tested it locally on my system.
where I sent multiple frames. in that case, the sequence of frames would be
websockets.frames.Opcode.TEXT or websockets.frames.Opcode.BINARY
followed by ... websockets.frames.Opcode.CONT
then may be more websockets.frames.Opcode.CONT frames
and finally in last websockets.frames.Opcode.CONT frame, fin would be set to true

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,389 @@
...
+        }
+        self.queue.put_nowait({"type": "websocket.connect"})
+        task = self.loop.create_task(self.run_asgi())
+        task.add_done_callback(self.on_task_complete)
+        self.tasks.add(task)
+
+    def handle_cont(self, event: Frame) -> None:
+        self.bytes += event.data
+        if event.fin:
+            self.send_receive_event_to_app()
```
gourav-kandoria, line 193: These lines(195-197) are not covered due to the same reason as I mentioned for line no. 141

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,389 @@
...
+                if event.opcode == websockets.frames.Opcode.CONT:
+                    self.handle_cont(event)
+                elif event.opcode == websockets.frames.Opcode.TEXT:
+                    self.handle_text(event)
+                elif event.opcode == websockets.frames.Opcode.BINARY:
+                    self.handle_bytes(event)
+                elif event.opcode == websockets.frames.Opcode.PING:
+                    self.handle_ping(event)
+                elif event.opcode == websockets.frames.Opcode.PONG:
+                    self.handle_pong(event)
```
gourav-kandoria: as per this document - https://datatracker.ietf.org/doc/html/rfc6455#autoid-29
[5.5.3] Pong
 The Pong frame contains an opcode of 0xA.
   A Pong frame sent in response to a Ping frame must have identical
   "Application data" as found in the message body of the Ping frame
   being replied to.
   If an endpoint receives a Ping frame and has not yet sent Pong
   frame(s) in response to previous Ping frame(s), the endpoint MAY
   elect to send a Pong frame for only the most recently processed Ping
   frame.
   A Pong frame MAY be sent unsolicited.  This serves as a
   unidirectional heartbeat.  A response to an unsolicited Pong frame is
   not expected.
   
   So, the websockets sansI/O layer chooses not to propagate the received PONG frame event.
   That's why It is not being covered.

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,389 @@
...
+        if not self.read_paused:
+            self.read_paused = True
+            self.transport.pause_reading()
+
+    def handle_ping(self, event: Frame) -> None:
+        output = self.conn.data_to_send()
+        self.transport.writelines(output)
+
+    def handle_pong(self, event: Frame) -> None:
+        pass
```
gourav-kandoria: same reason as mentioned for line number 149

gourav-kandoria:
> 

@Kludex as per your suggestion. found out and have mentioned the reason for those missing lines in the comments
open for discussion about any issue you might see

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,389 @@
...
+from uvicorn.protocols.utils import (
+    get_local_addr,
+    get_path_with_query_string,
+    get_remote_addr,
+    is_ssl,
+)
+from uvicorn.server import ServerState
+
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal
```
gourav-kandoria: @Kludex I guess, this condition can be removed as uvicorn requirement for python is atleast 3.8. right?

Kludex: yes

Kludex:
I'll modify the description to set what is missing for this to be merged.

Kludex:
@aaugustin Do you also want to review it?

I've already reviewed. I need to check the points on the checklist, and help there is also wanted - some are just questions. 🙏

uvicorn/protocols/websockets/websockets_sansio_impl.py
Kludex: We still need to test it. It's unlikely that we want `Exception` there as well.

uvicorn/protocols/websockets/websockets_sansio_impl.py
Kludex: Is this related to HTTP/2?

aaugustin:
I'll try to find time this week. If you haven't heard from me by the end of the week, please consider that I timed out :-)

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,378 @@
...
+                elif event.opcode == Opcode.TEXT:
+                    self.handle_text(event)
+                elif event.opcode == Opcode.BINARY:
+                    self.handle_bytes(event)
+                elif event.opcode == Opcode.PING:
+                    self.handle_ping(event)
+                elif event.opcode == Opcode.PONG:
+                    self.handle_pong(event)
+                elif event.opcode == Opcode.CLOSE:
+                    self.handle_close(event)
```
gourav-kandoria, line 146: @Kludex @aaugustin Could you please help with this?
one scenario, where we can expect close frame is when the client sends it.
What other scenarios are there where we can expect it?
Can we expect it in scenarios where the we received unexpected data
and how can we have access to the close code, so that same code can be passed in disconnect event to asgi app

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,378 @@
...
+                self.transport.writelines(output)
+            elif self.handshake_initiated:
+                self.send_500_response()
+                self.queue.put_nowait({"type": "websocket.disconnect", "code": 1006})
+            self.transport.close()
+
+    def data_received(self, data: bytes) -> None:
+        try:
+            self.conn.receive_data(data)
+        except Exception:
```
gourav-kandoria: @Kludex I guess, the "except Exception" block can be removed from here.  Upon reading code of Protocol, It seems that exception thrown by code are handled there internally and the connection is being made to fail. @aaugustin could you please confirm , if this is the case

uvicorn/protocols/websockets/websockets_sansio_impl.py
gourav-kandoria: @Kludex No, In case HTTP/1, according to websocket spec, this could be the behaviour

uvicorn/protocols/websockets/websockets_sansio_impl.py
Kludex: No, that's not the case... Since you are using `asyncio.Protocol`.

uvicorn/protocols/websockets/websockets_sansio_impl.py
gourav-kandoria: @valentin994 @Kludex we don't have to call receive_eof() after receiving data. Because,  protocol ensures that (https://docs.python.org/3/library/asyncio-protocol.html#asyncio.Protocol.data_received) data_received will only be called only when the data is there. In case when eof is received from client side, the transport will close and connection_lost will be called.

uvicorn/protocols/websockets/websockets_sansio_impl.py
gourav-kandoria: @Kludex oh sorry for miscommunication, here I meant websockets Protocol (the library which we are using)

uvicorn/protocols/websockets/websockets_sansio_impl.py
Kludex: got it

Kludex (changes requested):
`max_size` is not being tested on `WebSocketSansIOProtocol` - There's a test that uses `max_size` on the `test_websockets.py`.

The `Exception` part, we need to add the right exception, and test it.

Kludex:
I think we are almost there. 👍

gourav-kandoria:
> `max_size` is not being tested on `WebSocketSansIOProtocol` - There's a test that uses `max_size` on the `test_websockets.py`.
> 
> The `Exception` part, we need to add the right exception, and test it.


> `max_size` is not being tested on `WebSocketSansIOProtocol` - There's a test that uses `max_size` on the `test_websockets.py`.
> 
> The `Exception` part, we need to add the right exception, and test it.

@Kludex raised an issue regarding this
https://github.com/python-websockets/websockets/issues/1397#issue-1877077219

tests/protocols/test_websocket.py
```diff
@@ -968,7 +968,7 @@ async def get_subprotocol(url: str):
         (MAX_WS_BYTES, MAX_WS_BYTES, 0),
         (MAX_WS_BYTES_PLUS1, MAX_WS_BYTES, 1009),
         (10, 10, 0),
-        (11, 10, 1009),
+        (32, 10, 1009),
```
gourav-kandoria: @Kludex I am facing strange situation, in this test case on keeping the message size greater than 31 . Test is passing, but on values from 11 to 31 , it is failing. not sure, where there is some issue with my code, or ServerProtocol or with the client. Could you please assist in this

Kludex: I can, but it will take some weeks. I'm sorry. I'm a bit busy.

gourav-kandoria: sure, whenever the time permits you :)

gourav-kandoria: Hi, actually I was able to find the reason for it. 
In the test just like, There is max_size limit on the server. There were also max_size limit on the client side also
like this

```
async with websockets.client.connect(
            url, max_size=client_size_sent
   ) as websocket:
```
   
   So, in case when client_size_sent is 11, The close frame is being received with code 1009, along with a reason as "over size limit ({length} > {max_size} bytes)". Now the total size of this close which include the close code and the reason is summing more than 11. So, before the client processes this close frame, it is detecting the payload is bigger than 11 bytes. which is causing the client to close the connection on its end without processing the close frame from the server.
   
 So, for this test to work correctly, I have changed the client initiation logic from 
```
        async with websockets.client.connect(
            url, max_size=client_size_sent
        ) as websocket:
```

to this on latest change

`        async with websockets.client.connect(url, max_size=MAX_WS_BYTES) as websocket:`

Because, anyway, we want to test that atleast server is able send the close frame correctly and don't care what happens at client end

gourav-kandoria:
Hi @Kludex can we proceed on it.

Kludex:
Why?

gourav-kandoria:
oh my bad. branch got deleted by mistake. Due to bandwidth issue, I won't be able to contribute on this. That's why closed the pr. 
Although If you want to continue on this. You may refer this branch https://github.com/gourav-kandoria/uvicorn/tree/websocket-sansio-implementation-2. adding additional typing checks would be sufficient as the code in this branch is following complete websocket asgi specs and passing all test cases

Kludex:
cc @aaugustin

Kludex:
- Let's continue on https://github.com/encode/uvicorn/pull/2540.

PR #2061, opened 2023-07-29
Bump websockets to 11.0.3.

### LINKED ISSUE #1927: WebSockets 11.0 broke Uvicorn test suite

aaugustin:
Fix #1927.

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -577,7 +577,7 @@ async def test_asgi_return_value(
     unused_tcp_port: int,
 ):
     """
-    The ASGI callable should return 'None'. If it doesn't make sure that
+    The ASGI callable should return 'None'. If it doesn't, make sure that
     the connection is closed with an error condition.
     """
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ h11 @ git+https://github.com/python-hyper/h11.git@master
 # Explicit optionals
 a2wsgi==1.7.0
 wsproto==1.2.0
-websockets==10.4
+websockets==11.0.3
 
 # Packaging
 build==0.10.0
--- a/uvicorn/protocols/websockets/websockets_impl.py
+++ b/uvicorn/protocols/websockets/websockets_impl.py
@@ -236,7 +236,7 @@ class WebSocketProtocol(WebSocketServerProtocol):
         'send' and 'receive' events to drive the flow.
         """
         self.handshake_completed_event.set()
-        await self.closed_event.wait()
+        await self.wait_closed()
 
     async def run_asgi(self) -> None:
         """


tests/protocols/test_websocket.py
```diff
@@ -597,9 +597,8 @@ async def connect(url):
         port=unused_tcp_port,
     )
     async with run_server(config):
-        with pytest.raises(websockets.exceptions.ConnectionClosed) as exc_info:
+        with pytest.raises(websockets.exceptions.WebSocketException):
             await connect(f"ws://127.0.0.1:{unused_tcp_port}")
-        assert exc_info.value.code == 1006
```
Kludex: Is the behavior different between wsproto and websockets implementation different here?

aaugustin: I believe that they do. I will double check how the 4 parametrized variants of this test behave and report back.

Kludex: I'd like to avoid having a different behavior, but it's not a blocker. If we do, it would be cool that the test would show how they differ (it can be without the parametrize and create a new test).

aaugustin: Understood. I'll see what I can do.

aaugustin: We get the following errors in the test client:

- wsproto: `ConnectionClosedError(None, None, None)`
- websockets 10.4: `ConnectionClosedError(None, None, None)`
- websockets 11:
    - before this PR: `ConnectionClosedOK(Close(code=1000, reason=''), Close(code=1000, reason=''), True)`
    - after this PR: `InvalidMessage('did not receive a valid HTTP response')`

This is happening because:

- `WebSocketProtocol.run_asgi` sets `self.closed_event` here: https://github.com/encode/uvicorn/blob/8239373ec649b1a41179070b6ba2c8dc549df63e/uvicorn/protocols/websockets/websockets_impl.py#L258
- `WebSocketProtocol.ws_handler` exits here: https://github.com/encode/uvicorn/blob/8239373ec649b1a41179070b6ba2c8dc549df63e/uvicorn/protocols/websockets/websockets_impl.py#L239
- As soon as the connection handler returns, websockets performs the closing handshake automatically.

The last step used to be delayed by one cycle of the event loop in websockets 10.4. This used to allow `WebSocketProtocol.run_asgi` to close the transport first here: https://github.com/encode/uvicorn/blob/8239373ec649b1a41179070b6ba2c8dc549df63e/uvicorn/protocols/websockets/websockets_impl.py#L268

aaugustin: I am updating the PR with a small change that restores the previous behavior.

aaugustin:
FYI: I tried to remove `self.closed_event` entirely:

- `await self.closed_event.wait()` -> `await self.wait_closed()`
- `self.closed_event.is_set()` -> `self.closed`

but it caused another test failure. I didn't investigate further.

aaugustin:
This change guarantees that the connection handler never exits before the TCP connection is closed. Effectively, this disables websockets' automatic closing handshake, which is clearly the intent of the current implementation (but not always true).

Kludex:
For the record (for myself), I've tested locally with the application in the test suite:

```py
    return 123
```

Using a different WebSockets client (wscat):
```bash
❯ wscat -c "ws://localhost:8000/"
```
And I got the same 1006 as before.

Kludex (approved):
Thanks @aaugustin 🙏

PR #2062, opened 2023-07-30
Add filename to code snippets in the documentation

Kludex:


mkdocs.yml
```diff
@@ -16,6 +16,9 @@ theme:
       toggle:
         icon: "material/lightbulb-outline"
         name: "Switch to light mode"
+  features:
+    - content.code.copy
```
Kludex, line 20: This adds the copy bottom, which is missing.

michaeloliverx (approved):

PR #2073, opened 2023-08-14
Update scope structure in docs to match what is actually returned by the server

kaypee90:
# Summary
This PR addresses the previous discussion #2072 

The documentation specifies that the scope returns headers of structure a list of lists but the actual implementation returns a list of tuples. This PR is to address that inconsistency.

# Checklist

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -343,9 +343,9 @@ An incoming HTTP request might have a connection `scope` like this:
     'method': 'GET',
     'path': '/',
     'headers': [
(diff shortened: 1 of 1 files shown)


Kludex:
The documentation is mentioning the request (client) sending to the server, so in any case, it's correct.

That said, to be more consistent with the rest of code examples, and the fact that we use tuples internally, I'll accept this. 

Thanks @kaypee90 :)

Kludex (approved):

PR #2084, opened 2023-08-30
change should_reload to reload

RheagalFire:
# Summary 

Starting the server using below
```
config = {
"api.main:app",
      host=HOST,
      port=PORT,
      log_level=log_level,
      reload=reload,
      reload_dirs=reload_dirs,
      use_colors=use_colors,
      log_config=log_config,
}
server.run()
```
This doesn't start the server with the reload flag even when set to true that is because in the `main.py` it gets initiated by the should_reload variable that is not declared anywhere, we should check instead for the config.reload variable.

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -577,7 +577,7 @@ def run(
         )
         sys.exit(1)
 
-    if config.should_reload:
+    if config.reload:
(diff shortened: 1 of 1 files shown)


Kludex:
> should_reload variable that is not declared anywhere

It is.

PR #2094, opened 2023-09-06
UvicornWorker: respect the gunicorn graceful shutdown timeout

BrendanGraham14:
# Summary

This change propagates Gunicorn's `graceful_timeout` parameter to `Uvicorn`'s `timeout_graceful_shutdown` configuration when running `Uvicorn` as a Gunicorn worker process.

# Checklist

- Not sure if a test is necessary here? In any case it looks like `UvicornWorker` doesn't have any tests currently.

### DIFF
--- a/uvicorn/workers.py
+++ b/uvicorn/workers.py
@@ -40,6 +40,7 @@ class UvicornWorker(Worker):
             "callback_notify": self.callback_notify,
             "limit_max_requests": self.max_requests,
             "forwarded_allow_ips": self.cfg.forwarded_allow_ips,
+            "timeout_graceful_shutdown": self.cfg.graceful_timeout,
         }
 
         if self.cfg.is_ssl:


Kludex:
I don't think they are equivalent.

BrendanGraham14:
@Kludex what's the difference?

Gunicorn: https://docs.gunicorn.org/en/stable/signals.html#master-process
> TERM: Graceful shutdown. Waits for workers to finish their current requests up to the [graceful_timeout](https://docs.gunicorn.org/en/stable/settings.html#graceful-timeout).
> 
> `graceful_timeout`
> After receiving a restart signal, workers have this much time to finish serving requests. Workers still alive after the timeout (starting from the receipt of the restart signal) are force killed.

Uvicorn: https://www.uvicorn.org/settings/#timeouts
> --timeout-graceful-shutdown <int> - Maximum number of seconds to wait for graceful shutdown. After this timeout, the server will start terminating requests.

Kludex:
I think the details are on the PR that introduced it. But I think it's because the one from Uvicorn is related to the ASGI event (shutdown).

BrendanGraham14:
Hm, but they both apply to the graceful termination of workers/tasks in the event of a signal to shutdown. Is there any reason to be concerned about making the Uvicorn worker aware of the graceful shutdown timeout?

Kludex:
We are deprecating the `workers` module within Uvicorn. PR welcome on https://github.com/Kludex/uvicorn-worker.

PR #2106, opened 2023-09-20
Allow setting APP with envvar UVICORN_APP

aceat64:
# Summary

Allows configuring the `APP` setting using the environment variable `UVICORN_APP`. Currently environment variables can be used for all other settings thanks to click's `auto_envvar_prefix`, but this applies only to options and not to arguments.

# Checklist

I'm not sure if a new test is needed for this change. Additionally, I believe no documentation changes are needed. The docs currently state that env vars can be used for configuration but doesn't call out the caveat that `APP` can not.

### DIFF
--- a/scripts/check
+++ b/scripts/check
@@ -10,7 +10,7 @@ export SOURCE_FILES="uvicorn tests"
 set -x
 
 ./scripts/sync-version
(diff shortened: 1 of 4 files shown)


Kludex (approved):

Kludex:
Thanks @aceat64 :)

PR #2107, opened 2023-09-21
:bug: Fixed --reload-dir option not working as expected.

bybatkhuu:
# --reload-dir specified, but default dir(cwd) still been watching

Not working as cli help explained:

txt
```
uvicorn --help:
--reload-dir PATH   Set reload directories explicitly, instead
```

So, I tried to fix it by counting `self.reload_dirs`, and if it's not provided by arguments, then it will add the default current directory as the watching directory.

Previous discussion: <https://github.com/encode/uvicorn/discussions/1833>

And also: <https://github.com/encode/uvicorn/issues/1326>

# Checklist

### DIFF
--- a/uvicorn/supervisors/watchfilesreload.py
+++ b/uvicorn/supervisors/watchfilesreload.py
@@ -68,7 +68,7 @@ class WatchFilesReload(BaseReload):
         for directory in config.reload_dirs:
             if Path.cwd() not in directory.parents:
                 self.reload_dirs.append(directory)
-        if Path.cwd() not in self.reload_dirs:
+        if (len(self.reload_dirs) == 0) and (Path.cwd() not in self.reload_dirs):
             self.reload_dirs.append(Path.cwd())
 
         self.watch_filter = FileFilter(config)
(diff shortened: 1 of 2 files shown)


iudeen:
This solution looks fine for me on a quick glance. I'll see if I can review this properly before giving my verdict. Also, Fix the tests please :)

imdoroshenko:
This issue can block the usage of the reload feature in some cases. I work on a CLI tool that should start the server from a host directory where **a lot** FS operations happen every few seconds. In my case, `reload-dir` targets the correct directory with the server source code, but those lines cause constant reload because `Path.cwd()` is added to the `reload_dirs` no matter what. 

edit:
This issue is also hard to detect because logs print directories it plans to watch before it adds `Path.cwd()` there. 

```
INFO:     Will watch for changes in these directories: ['directories form --reload-dir']
```

akefirad:
This is a bit suboptimal DX. Any help needed with this fix? I can fix the tests if @bybatkhuu doesn't have time. LMK.

bybatkhuu:
Sorry, I don't have time to understand tests and fix for this case. @akefirad, of course. Thank you~

loukkyy:
anyone working on this? Would be good to add this fix :)

Kludex:
Thanks for the contribution, and apologies for the long delay on this. This has been addressed by #2598, so I'm closing this one.

PR #2108, opened 2023-09-21
websockets.client replaced with httpx.AsyncClient for websocket tests

likecodingloveproblems:
# Summary

# Checklist


Kludex:
I'm sorry for making you lose time. I think we should continue with the current setup.

PR #2130, opened 2023-10-14
use asyncio.run(..., loop_factory) to avoid asyncio.set_event_loop_policy

graingert:
# Summary

set_event_loop_policy will be deprecated in python 3.13 and uvloop.EventLoopPolicy is deprecated already, this backports and uses the preferred `loop_factory` keyword to configure asyncio.run
# Checklist

### DIFF
--- a/uvicorn/loops/asyncio.py
+++ b/uvicorn/loops/asyncio.py
@@ -1,10 +1,17 @@
+from __future__ import annotations
+
 import asyncio
 import logging
 import sys
+from collections.abc import Callable
+from typing import TypeVar
+
+_T = TypeVar("_T")
 
 logger = logging.getLogger("uvicorn.error")
 
 
-def asyncio_setup(use_subprocess: bool = False) -> None:
-    if sys.platform == "win32" and use_subprocess:
-        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())  # pragma: full coverage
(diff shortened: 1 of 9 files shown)


uvicorn/loops/uvloop.py
```diff
@@ -1,7 +1,12 @@
+from __future__ import annotations
+
 import asyncio
+from collections.abc import Callable
 
 import uvloop
 
 
-def uvloop_setup(use_subprocess: bool = False) -> None:
-    asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
```
graingert, line 7: uvloop.EventLoopPolicy is deprecated

uvicorn/loops/asyncio.py
```diff
@@ -1,10 +1,19 @@
...
+from typing import TypeVar
+
+_T = TypeVar("_T")
 
 logger = logging.getLogger("uvicorn.error")
 
 
-def asyncio_setup(use_subprocess: bool = False) -> None:
-    if sys.platform == "win32" and use_subprocess:
-        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```
graingert: set_event_loop_policy will be deprecated in python 3.13

graingert:
This still needs a doc change

graingert:
actually this doesn't need a doc change as `--loop` never allowed a custom loop setup function

Kludex (approved):

PR #2129, opened 2023-10-14
Support Python 3.12

Kludex:

PR #2134, opened 2023-10-20
Bump uvloop to v0.18.0

Ale-Cas:
# Summary

This PR bumps uvloop to v0.18.0 to try and address the mypy error in feat/support-3.12.
Related:
https://github.com/encode/uvicorn/pull/2129#issuecomment-1763485380

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,9 +9,7 @@ description = "The lightning-fast ASGI server."
 readme = "README.md"
 license = "BSD-3-Clause"
 requires-python = ">=3.8"
-authors = [
-    { name = "Tom Christie", email = "tom@tomchristie.com" },
-]
+authors = [{ name = "Tom Christie", email = "tom@tomchristie.com" }]
 classifiers = [
     "Development Status :: 4 - Beta",
     "Environment :: Web Environment",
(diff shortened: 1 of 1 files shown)


Kludex:
That PR is already using the latest uvloop, that's exactly why it's failing.

Ale-Cas:
You're right I thought it was using 0.17 but I double-checked the run and it's already 0.18, I'm trying to understand what would make mypy see EventLoopPolicy since it's in the uvloop __init__ but I guess that's not related to uvicorn so I'll close this, sorry for the spam 🙏🏻

Kludex:
np. Thanks for closing it. 👍

graingert:
@Ale-Cas they added an `__init__.pyi` with most of the members removed this hides EventLoopPolicy from mypy. The issue is fixed upstream pending a release this weekend: https://github.com/MagicStack/uvloop/issues/575#issuecomment-1771305350

Ale-Cas:
Great! thank you both!
I was thinking something like this workaround / temp solution would make mypy happy:
```python
def uvloop_setup(use_subprocess: bool = False) -> None:
    if hasattr(uvloop, "EventLoopPolicy"):
```
and maybe not related but what's the reason for the unused use_subprocess variable there? is it just for compatibility with the loop_setup from asyncio?

PR #2145, opened 2023-11-01
Fix Ctrl+C hang on Python 3.12 when connections are open

jcheng5:
# Summary

The asyncio.Server.wait_closed() method was a no-op in versions of Python earlier than 3.12. The intention of this method was to block until all existing connections are closed, but due to a bug in its implementation, it wouldn't actually wait. This bug was fixed in Python 3.12, which exposed uvicorn's dependence on the buggy behavior: the implementation of uvicorn.Server.shutdown() called wait_closed() before asking existing connections to close. As a result, attempting to stop a Uvicorn server with an open connection would result in a blocked process, with further attempts to Ctrl+C having no effect.

See discussion at https://github.com/encode/uvicorn/discussions/2122

# Checklist

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -268,8 +268,6 @@ class Server:
             server.close()
         for sock in sockets or []:
             sock.close()
-        for server in self.servers:
-            await server.wait_closed()
 
         # Request shutdown on all existing connections.
         for connection in list(self.server_state.connections):
@@ -310,6 +308,20 @@ class Server:
             msg = "Waiting for background tasks to complete. (CTRL+C to force quit)"
             logger.info(msg)
             while self.server_state.tasks and not self.force_exit:
+                await asyncio.sleep(0.1)
+
+        # Wait for servers to close. They won't do so until all connections are
+        # closed, which we've already waited for above, so this should be quick.
+        servers_closed = asyncio.gather(
+            *[server.wait_closed() for server in self.servers]
+        )
+        # Give the servers_closed future a chance to complete so we don't
+        # spuriously log about this operation.
+        await asyncio.sleep(0.1)
+        if not servers_closed.done() and not self.force_exit:
+            msg = "Waiting for servers to close. (CTRL+C to force quit)"
+            logger.info(msg)
+            while not servers_closed.done() and not self.force_exit:
                 await asyncio.sleep(0.1)
 
     def install_signal_handlers(self) -> None:
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -310,20 +310,8 @@ class Server:
             while self.server_state.tasks and not self.force_exit:
                 await asyncio.sleep(0.1)
 
-        # Wait for servers to close. They won't do so until all connections are
-        # closed, which we've already waited for above, so this should be quick.
-        servers_closed = asyncio.gather(
-            *[server.wait_closed() for server in self.servers]
-        )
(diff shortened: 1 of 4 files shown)


zanieb (approved):
LGTM! Thanks for contributing this fix

zanieb:
Did you test this with Python 3.12?

jcheng5:
@zanieb Yes, and I ran pytest on 3.12. Are there other tests I should run?

Edit: This was on Mac. I'll try on Windows as well

jcheng5:
Works well on Windows, _except_ that on the combination of [Windows Terminal](https://learn.microsoft.com/en-us/windows/terminal/) and PowerShell, Uvicorn doesn't respond to Ctrl+C or Ctrl+Break at all, even with much older versions of Uvicorn. No problem with:

* Windows Terminal + cmd
* Windows console host + cmd
* Windows console host + PowerShell

In all of those instances, the behavior is what I expect--hang when exiting with open connections without this fix, clean exit with this fix.

jcheng5:
On the combination of Windows Terminal (v1.18.2822.0) and PowerShell, I can't even use Ctrl+C to stop `py -m http.server`. So I guess whatever is causing that is completely unrelated to Uvicorn and something I should figure out about my own configuration.

Update: Restarted Windows Terminal and now WT + PS work as expected, just like the other configurations. 🤷‍♂️

zanieb:
Nice :)

uvicorn/server.py
```diff
@@ -312,6 +310,20 @@ async def _wait_tasks_to_complete(self) -> None:
...
+            *[server.wait_closed() for server in self.servers]
+        )
+        # Give the servers_closed future a chance to complete so we don't
+        # spuriously log about this operation.
+        await asyncio.sleep(0.1)
+        if not servers_closed.done() and not self.force_exit:
+            msg = "Waiting for servers to close. (CTRL+C to force quit)"
+            logger.info(msg)
+            while not servers_closed.done() and not self.force_exit:
+                await asyncio.sleep(0.1)
```
Kludex: We should be using `TaskGroup` for higher Python versions.

uvicorn/server.py
```diff
@@ -312,6 +310,20 @@ async def _wait_tasks_to_complete(self) -> None:
...
+
```
Kludex: Can't we just do this?
```suggestion
        for server in self.servers:
            await server.wait_closed()
```

zanieb: This matches the other implementations in this function.

jcheng5: The (only) problem with that way @Kludex is if something goes wrong and `wait_closed` blocks again, hitting Ctrl+C again doesn't do anything. The way I wrote it is a lot more complex but preserves not only the Ctrl+C behavior, but also, showing the message iff it takes more than 100ms or whatever--just like the other operations in this function.

Kludex: What can go wrong on `wait_closed`?

jcheng5: Probably nothing as of today. But this PR is necessary precisely because `wait_closed` changed behavior.

That being said, I don't feel strongly about this at all. If you prefer the simpler lines I'm absolutely fine with it.

Kludex: Please change it.

zanieb (approved):

jcheng5:
Thank you @zanieb and @Kludex! Our users will appreciate it!

PR #2155, opened 2023-11-19
Raise IOError if disconnected on `send()`

Kludex:
- [ ] Starlette PR that catch `IOError`.
- [ ] Tests.
- [ ] Bump the ASGI specification.

Ref.: https://github.com/django/asgiref/issues/66


Kludex:
I need help to add tests on the HTTP protocols.

PR #2158, opened 2023-11-21
Properly annotate `uvicorn.run()`

Kludex:
I know I closed this https://github.com/encode/uvicorn/issues/1855, but I don't think another type check for the `run()` function.

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -9,7 +9,6 @@ import ssl
 import sys
 from pathlib import Path
 from typing import (
-    TYPE_CHECKING,
     Any,
     Awaitable,
     Callable,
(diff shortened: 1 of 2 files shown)


uvicorn/config.py
```diff
@@ -31,6 +32,11 @@
 from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
 from uvicorn.middleware.wsgi import WSGIMiddleware
 
+if TYPE_CHECKING:
+    PathLike = os.PathLike[str]
+else:
+    PathLike = os.PathLike
```
abersheeran: Why do we need to write this? 🤔

Kludex: Because `os.PathLike` is not a generic at runtime in previous Python versions.

Ref.: https://github.com/python/mypy/pull/5833/files

abersheeran: In baize I write this, mypy or pylance works fine. 

```
        directory: Union[str, "os.PathLike[str]"],
```

Kludex: That's better. Thanks.

agronholm: Even better would be to use `from __future__ import annotations`.

Kludex: Yes, PR welcome with that.

abersheeran (approved):

PR #2174, opened 2023-12-03
Create `http_protocol_cls` fixture

Kludex:
- Related to https://github.com/encode/uvicorn/pull/2049

### DIFF
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -4,7 +4,6 @@ import httpx
 import pytest
 import websockets.client
 
-from tests.protocols.test_http import HTTP_PROTOCOLS
 from tests.response import Response
 from tests.utils import run_server
 from uvicorn._types import ASGIReceiveCallable, ASGISendCallable, Scope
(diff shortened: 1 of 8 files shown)


uvicorn/_types.py
Kludex, line 1: Complying with the wording on the spec: https://asgi.readthedocs.io/en/latest/specs/www.html

tests/middleware/test_proxy_headers.py
```diff
@@ -114,10 +115,9 @@ async def test_proxy_headers_invalid_x_forwarded_for() -> None:
 
 
 @pytest.mark.anyio
-@pytest.mark.parametrize("http_protocol_cls", HTTP_PROTOCOLS)
 async def test_proxy_headers_websocket_x_forwarded_proto(
     ws_protocol_cls: "Type[WSProtocol | WebSocketProtocol]",
-    http_protocol_cls,
+    http_protocol_cls: "Type[H11Protocol | HttpToolsProtocol]",
```
frankie567, line 120: Would it be worth to add a type alias for this? It would avoid having to import `H11Protocol` and `HttpToolsProtocol` in each test.

Probably the same goes for `Type[WSProtocol | WebSocketProtocol]`.

frankie567 (approved):
LGTM 👍 Using a parametrized fixture is really smart and elegant in this context!

tests/middleware/test_proxy_headers.py
abersheeran: Yes, I also think adding a type alias is more elegant (and easier to maintain later).

iudeen (approved):
Very clean and consistent.

Kludex:
I'm going to add the type alias on a future PR. I tried to do it, but then I found myself changing a lot of other types. 😅

Issue #2166, opened 2023-11-29
`--reload-include` doesn't work with hidden files e.g. `--reload-include .env`


michaeloliverx:
### Initial Checks

- [X] I confirm this was discussed, and the maintainers suggest I open an issue.
- [X] I'm aware that if I created this issue without a discussion, it may be closed without a response.

### Discussion Link

```Text
https://github.com/encode/uvicorn/discussions/1705
```

### Description

The `--reload-include` CLI flag doesn't work when specifying a hidden file directly e.g.
```sh
uvicorn src.main:app --reload --reload-include .env
```

### Python, Uvicorn & OS Version

```Text
Running uvicorn 0.23.2 with CPython 3.11.5 on Darwin
```

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/uvicorn/issues/2166">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/2166/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/uvicorn/issues/2166/pledge.svg">
</picture>
</a>

Kludex:
@michaeloliverx Is this done?

adithya04dev:
its still not reloading after the .env file changes

enigma:
@adithya04dev and who else might be having a similar issue: you need to have `watchfiles` installed for `--reload-include` to work.

eg: `uv run --with watchfiles python -m uvicorn ... --reload-include '.env'`

You'll then see:

```
INFO:     Started reloader process [107] using WatchFiles
INFO:     Started server process [109]
WARNING:  WatchFiles detected changes in '.env'. Reloading...
```


See https://www.uvicorn.org/settings/#reloading-with-watchfiles

PR #2176, opened 2023-12-04
bug: allow explicit hidden file paths in `--reload-include`

michaeloliverx:
# Summary

Relates to https://github.com/encode/uvicorn/issues/2166

# Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -124,6 +124,9 @@ def reload_directory_structure(tmp_path_factory: pytest.TempPathFactory):
     │       └── sub.py
     ├── ext
     │   └── ext.jpg
+    ├── .dotted
+    ├── .dotted_dir
+    │   └── file.txt
     └── main.py
     """
     root = tmp_path_factory.mktemp("reload_directory")
--- a/uvicorn/supervisors/watchfilesreload.py
+++ b/uvicorn/supervisors/watchfilesreload.py
@@ -43,6 +43,9 @@ class FileFilter:
     def __call__(self, path: Path) -> bool:
         for include_pattern in self.includes:
             if path.match(include_pattern):
+                if str(path).endswith(include_pattern):
+                    return True
+
                 for exclude_dir in self.exclude_dirs:
                     if exclude_dir in path.parents:
                         return False
--- a/uvicorn/supervisors/watchfilesreload.py
+++ b/uvicorn/supervisors/watchfilesreload.py
@@ -19,7 +19,7 @@ class FileFilter:
         self.includes.extend(config.reload_includes)
         self.includes = list(set(self.includes))
 
-        default_excludes = [".*", ".py[cod]", ".sw.*", "~*"]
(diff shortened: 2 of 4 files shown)


tests/conftest.py
```diff
@@ -124,6 +124,9 @@ def reload_directory_structure(tmp_path_factory: pytest.TempPathFactory):
     │       └── sub.py
     ├── ext
     │   └── ext.jpg
+    ├── .dotted
+    ├── .dotted_dir
+    │   └── file.txt
```
michaeloliverx, line 129: The tree in the docstring was out of date with the implementation.

uvicorn/supervisors/watchfilesreload.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, config: Config):
         self.includes.extend(config.reload_includes)
         self.includes = list(set(self.includes))
 
-        default_excludes = [".*", ".py[cod]", ".sw.*", "~*"]
+        default_excludes = [".py[cod]", ".sw.*", "~*"]
```
michaeloliverx: I am not fully aware of the implications of this change but it is much simpler than trying to construct a dynamic glob pattern based on `config.reload_includes`.

lovelydinosaur: > I am not fully aware of the implications of this change

I assume the original behaviour of *not* scanning hidden folders is intentional.

michaeloliverx: Agreed, cache directories such as `.mypy_cache` can contain a large number of files and directories which we probably want to avoid scanning.

Do you have any other ideas?

lovelydinosaur: It looks to me like it's intendedd that `should_watch_dir` would already correctly handle this? I guess I'd start by doing some debugging there to figure out if/why that's not the case. 🤷‍♂️

uvicorn/supervisors/watchfilesreload.py
```diff
@@ -43,6 +43,9 @@ def __init__(self, config: Config):
     def __call__(self, path: Path) -> bool:
         for include_pattern in self.includes:
             if path.match(include_pattern):
+                if str(path).endswith(include_pattern):
+                    return True
```
michaeloliverx, line 47: Here we check if the path ends with the include pattern, should handle relative and absolute cases.

uvicorn/supervisors/watchfilesreload.py
michaeloliverx: The default exclude of `.*` will match **all** hidden directories regardless of what is in the include patterns.

I modified it to handle if the include pattern appears exactly at the end of a path then to include it.

Does the added tests provide enough cover here?

lovelydinosaur (approved):

adambenali:
@michaeloliverx does this work for glob patterns? e.g. `--reload-include '.env*'`

PR #2179, opened 2023-12-08
Support to change the options in the SSL context, for the test environment to simulate the server with different tls version protocols.

TheBetterSolution:
Support to change the options in the SSL context, for the test environment to simulate the server with different tls version protocols.

# Summary

Support to change the options in the SSL context, for the test environment to simulate the server with different tls version protocols.

My team hope to use uvicorn server as the test server to test the case for the different tls protocols deployed on server, for example:
1. Server only enabled tls1.2
2. Server only enabled tls1.3
3. Server enabled tls1.2 and tls 1.3.
3. Server enabled tls1.1 and tls 1.2.
Use the ssl_versions can't control the tls versions precisely.
I think that's common requirement to test the server's support capability for tls protocols.

I want to confirm whether the idea of this patch is allowed first, thanks.


TheBetterSolution:
We have a product that can configure the tls version explicitly, we hope the tls version can also be configured on the server explicitly for testing.

I think other professional softwares also needs this feature.

Thanks.

Kludex:
- This is duplicated of https://github.com/encode/uvicorn/pull/1692.

We agreed to have a non-CLI parameter to configure the `SSLContext`.

PR #2183, opened 2023-12-15
New multiprocess manager

abersheeran:
# Summary

About https://github.com/encode/uvicorn/discussions/2164.

The multiprocess manager introduced by this PR includes process keep-alive and process hung detection. It also imitates gunicorn and uses hup, ttin, and ttou signals to control child processes.

# Checklist

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -186,7 +186,7 @@ Uvicorn includes a `--workers` option that allows you to run multiple worker pro
 $ uvicorn main:app --workers 4
 ```
 
-Unlike gunicorn, uvicorn does not use pre-fork, but uses [`spawn`](https://docs.python.org/zh-cn/3/library/multiprocessing.html#contexts-and-start-methods), which allows uvicorn's multiprocess manager to still work well on Windows.
+Unlike gunicorn, uvicorn does not use pre-fork, but uses [`spawn`](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods), which allows uvicorn's multiprocess manager to still work well on Windows.
 
 The default process manager monitors the status of child processes and automatically restarts child processes that die unexpectedly. Not only that, it will also monitor the status of the child process through the pipeline. When the child process is accidentally stuck, the corresponding child process will be killed through an unstoppable system signal or interface.
 
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -178,7 +178,23 @@ Running Uvicorn using a process manager ensures that you can run multiple proces
 
 A process manager will handle the socket setup, start-up multiple server processes, monitor process aliveness, and listen for signals to provide for processes restarts, shutdowns, or dialing up and down the number of running processes.
 
-Uvicorn provides a lightweight way to run multiple worker processes, for example `--workers 4`, but does not provide any process monitoring.
+### Built-in
+
+Uvicorn includes a `--workers` option that allows you to run multiple worker processes.
+
+```bash
+$ uvicorn main:app --workers 4
+```
+
+Unlike gunicorn, uvicorn does not use pre-fork, but uses [`spwan`](https://docs.python.org/zh-cn/3/library/multiprocessing.html#contexts-and-start-methods), which allows uvicorn's multi-process manager to still work well on Windows.
+
+The default process manager monitors the status of child processes and automatically restarts child processes that die unexpectedly. Not only that, it will also monitor the status of the child process through the pipeline. When the child process is accidentally stuck, the corresponding child process will be killed through an unstoppable system signal or interface.
+
+You can also manage child processes by sending specific signals to the main process. (Not supported on Windows.)
+
+- `SIGHUP`: Work processeses are graceful restarted one after another. If you update the code, the new worker process will use the new code.
+- `SIGTTIN`: Increase the number of worker processes by one.
+- `SIGTTOU`: Decrease the number of worker processes by one.
 
 ### Gunicorn
 
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -186,7 +186,7 @@ Uvicorn includes a `--workers` option that allows you to run multiple worker pro
 $ uvicorn main:app --workers 4
 ```
 
-Unlike gunicorn, uvicorn does not use pre-fork, but uses [`spwan`](https://docs.python.org/zh-cn/3/library/multiprocessing.html#contexts-and-start-methods), which allows uvicorn's multi-process manager to still work well on Windows.
+Unlike gunicorn, uvicorn does not use pre-fork, but uses [`spawn`](https://docs.python.org/zh-cn/3/library/multiprocessing.html#contexts-and-start-methods), which allows uvicorn's multiprocess manager to still work well on Windows.
 
 The default process manager monitors the status of child processes and automatically restarts child processes that die unexpectedly. Not only that, it will also monitor the status of the child process through the pipeline. When the child process is accidentally stuck, the corresponding child process will be killed through an unstoppable system signal or interface.
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,7 +100,7 @@ omit = [
 
 [tool.coverage.report]
 precision = 2
-fail_under = 96.83
+fail_under = 96.82
 show_missing = true
 skip_covered = true
 exclude_lines = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,7 +103,7 @@ omit = [
 
 [tool.coverage.report]
 precision = 2
-fail_under = 98.13
+fail_under = 98.35
 show_missing = true
 skip_covered = true
 exclude_lines = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,7 +100,7 @@ omit = [
 
 [tool.coverage.report]
 precision = 2
-fail_under = 96.95
+fail_under = 96.92
 show_missing = true
 skip_covered = true
 exclude_lines = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,7 +100,7 @@ omit = [
 
 [tool.coverage.report]
 precision = 2
-fail_under = 98.35
+fail_under = 96.95
 show_missing = true
 skip_covered = true
 exclude_lines = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,7 +115,11 @@ exclude_lines = [
 ]
 
 [tool.coverage.coverage_conditional_plugin.omit]
-"sys_platform == 'win32'" = ["uvicorn/loops/uvloop.py", "uvicorn/supervisors/multiprocess.py"]
+"sys_platform == 'win32'" = [
+    "uvicorn/loops/uvloop.py",
+    "uvicorn/supervisors/multiprocess.py",
+    "tests/supervisors/test_multiprocess.py",
+]
 "sys_platform != 'win32'" = ["uvicorn/loops/asyncio.py"]
 
 [tool.coverage.coverage_conditional_plugin.rules]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,7 +101,7 @@ omit = [
 
 [tool.coverage.report]
 precision = 2
-fail_under = 96.82
+fail_under = 98.13
 show_missing = true
 skip_covered = true
 exclude_lines = [
@@ -113,7 +113,7 @@ exclude_lines = [
 ]
 
 [tool.coverage.coverage_conditional_plugin.omit]
-"sys_platform == 'win32'" = ["uvicorn/loops/uvloop.py"]
+"sys_platform == 'win32'" = ["uvicorn/loops/uvloop.py", "uvicorn/supervisors/multiprocess.py"]
 "sys_platform != 'win32'" = ["uvicorn/loops/asyncio.py"]
 
 [tool.coverage.coverage_conditional_plugin.rules]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,7 +100,7 @@ omit = [
 
 [tool.coverage.report]
 precision = 2
-fail_under = 96.92
+fail_under = 96.83
 show_missing = true
 skip_covered = true
 exclude_lines = [
--- a/tests/supervisors/test_multiprocess.py
+++ b/tests/supervisors/test_multiprocess.py
@@ -33,7 +33,7 @@ def new_console_in_windows(test_function: Callable[[], Any]):
                 "-c",
                 f"from {module} import {name}; {name}.__wrapped__()",
             ],
-            creationflags=subprocess.CREATE_NO_WINDOW,
+            creationflags=subprocess.CREATE_NO_WINDOW,  # type: ignore
         )
 
     return new_function
--- a/tests/supervisors/test_multiprocess.py
+++ b/tests/supervisors/test_multiprocess.py
@@ -56,8 +56,32 @@ def test_multiprocess_health_check() -> None:
     supervisor.join_all()
 
 
-# Test is skipped because windows does not support SIGHUP
-@pytest.mark.skipif(not hasattr(signal, "SIGHUP"), reason="SIGHUP is not supported on Windows")
+def test_multiprocess_sigterm() -> None:
+    """
+    Ensure that the SIGTERM signal is handled as expected.
+    """
+    config = Config(app=app, workers=2)
+    supervisor = Multiprocess(config, target=run, sockets=[])
+    threading.Thread(target=supervisor.run, daemon=True).start()
+    time.sleep(1)
+    supervisor.signal_queue.append(signal.SIGTERM)
+    supervisor.join_all()
+
+
+@pytest.mark.skipif(not hasattr(signal, "SIGBREAK"), reason="platform unsupports SIGBREAK")
+def test_multiprocess_sigbreak() -> None:
+    """
+    Ensure that the SIGBREAK signal is handled as expected.
+    """
+    config = Config(app=app, workers=2)
+    supervisor = Multiprocess(config, target=run, sockets=[])
+    threading.Thread(target=supervisor.run, daemon=True).start()
+    time.sleep(1)
+    supervisor.signal_queue.append(signal.SIGBREAK)
+    supervisor.join_all()
+
+
+@pytest.mark.skipif(not hasattr(signal, "SIGHUP"), reason="platform unsupports SIGHUP")
 def test_multiprocess_sighup() -> None:
     """
     Ensure that the SIGHUP signal is handled as expected.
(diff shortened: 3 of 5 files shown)


uvicorn/supervisors/multiprocess.py
```diff
@@ -30,45 +109,129 @@ def __init__(
...
+            del self.processes[idx]
+            process = Process(self.config, self.target, self.sockets)
+            process.start()
+            self.processes.append(process)
+
+    def run(self) -> None:
+        message = "Started parent process [{}]".format(os.getpid())
+        color_message = "Started parent process [{}]".format(
+            click.style(str(os.getpid()), fg="cyan", bold=True)
+        )
```
baseplate-admin: Hi, i have a question. Is there a reason this is not using `f-strings`?

pyproject.toml
```diff
@@ -100,7 +100,7 @@ omit = [
 
 [tool.coverage.report]
 precision = 2
-fail_under = 98.35
+fail_under = 96.82
```
Kludex: I'm not willing to introduce this feature without testing it.

Kludex (changes requested):

pyproject.toml
abersheeran: Yes, I just want to see your attitude towards this PR before adding the test. Because I remember you have implemented something similar before.

Kludex: Ah, then let me review and get back to you.

uvicorn/supervisors/multiprocess.py
```diff
@@ -30,45 +109,129 @@ def __init__(
...
+            self.should_exit.set()
+        else:
+            self.terminate_all()
+
+    def handle_break(self) -> None:
+        logger.info("Received SIGBREAK, exiting")
+        if not self.should_exit.is_set():
+            self.should_exit.set()
+        else:
+            self.terminate_all()
```
Kludex: Instead of using the style:
```py
signal.signal(signal.SIGINT, lambda sig, frame: self.handle_int())
```
Can we use:
```py
def handle_exit(sig: Signal, frame: FrameType) -> None:
    logger.info("Received %s, exiting", sig)
    if not self.should_exit.is_set():
        self.should_exit.set()
    else:
        self.terminate_all()

signal.signal(signal.SIGINT, self.handle_exit)
```

uvicorn/supervisors/multiprocess.py
```diff
@@ -30,45 +109,129 @@ def __init__(
...
+
+    def handle_hup(self) -> None:
+        logger.info("Received SIGHUP, restarting processes")
+        self.restart_all()
```
Kludex: How do we intend to use this with the "dev mode" (`--reload`)?

uvicorn/supervisors/multiprocess.py
```diff
@@ -30,45 +109,129 @@ def __init__(
...
+
+    def handle_ttou(self) -> None:
+        logger.info("Received SIGTTOU, decreasing processes")
+        if self.processes_num <= 1:
+            logger.info("Cannot decrease processes any more")
+            return
+        self.processes_num -= 1
+        process = self.processes.pop()
+        process.terminate()
+        process.join()
```
Kludex: I'm fine with adding those, but is there any literature about them that we can link on docstrings?

uvicorn/supervisors/multiprocess.py
```diff
@@ -30,45 +109,129 @@ def __init__(
...
+    def run(self) -> None:
+        message = "Started parent process [{}]".format(os.getpid())
+        color_message = "Started parent process [{}]".format(
+            click.style(str(os.getpid()), fg="cyan", bold=True)
+        )
+        logger.info(message, extra={"color_message": color_message})
+
+        self.init_processes()
+
+        while not self.should_exit.wait(0.5):
```
Kludex, line 151: I'm fine with this interval. Is it the same as gunicorn uses?

Kludex (changes requested):
## Checklist
- [x] Increase coverage for the new multiprocess manager to 100% (if tests are slow, run on a separate job in the CI, and only if the `multiprocess.py` is changed).
- [ ] Docstrings.
- [x] Documentation (I can help here).
- [x] Answer the question: how this should interact with `--reload`?
- [x] Move `UvicornWorker` out from Uvicorn's source code (I'll do it).

Kludex:
@abersheeran I'm fully available to move this forward. 👍

Kludex:
Pinging @ahopkins to review the initial steps since he offered helped on #2164. 🙏

uvicorn/supervisors/multiprocess.py
abersheeran: Improvements to dev mode may be something that needs to be addressed in the next PR, in this PR I just want to focus on multi-process management. [There is a usage in zibai](https://github.com/abersheeran/zibai/blob/e4822efea7199d38d86c74319ca6ecd1da83c03c/src/zibai/multiprocess.py#L261-L278). If this PR passes, maybe we can refactor reload.

uvicorn/supervisors/multiprocess.py
abersheeran: https://docs.gunicorn.org/en/latest/signals.html#master-process

uvicorn/supervisors/multiprocess.py
abersheeran: I don't know, but this value doesn't affect anything.

uvicorn/supervisors/multiprocess.py
abersheeran: But writing it this way is not convenient for testing, building a Frame is a bit troublesome. We do not use sig and frame in the signal processing function.

Kludex:
Ok. 👍 

We can move forward with the checklist above 🙏

uvicorn/supervisors/multiprocess.py
```diff
@@ -4,24 +4,101 @@
...
+            return True
+        return False
+
+    def pong(self) -> None:
+        self.child_conn.recv()
+        self.child_conn.send(b"pong")
+
+    def always_pong(self) -> None:
+        while True:
+            self.pong()
```
ahopkins, line 50: I would test this out and see what this might do for CPU if the pipe gets flooded. Might need some throttle in there.

uvicorn/supervisors/multiprocess.py
```diff
@@ -32,39 +109,125 @@ def __init__(
...
 
-        message = f"Stopping parent process [{str(self.pid)}]"
-        color_message = "Stopping parent process [{}]".format(click.style(str(self.pid), fg="cyan", bold=True))
+    def restart_all(self) -> None:
+        for idx, process in enumerate(tuple(self.processes)):
+            process.terminate()
+            process.join()
+            del self.processes[idx]
+            process = Process(self.config, self.target, self.sockets)
+            process.start()
```
ahopkins: Are you confirmed this is the behavior that you want? Shutdown first, then start the new processes with a potential gap in coverage? This was the initial pattern we went for with Sanic, but we received requests and ran into use cases where it was not desired and therefore offered it as configurable: https://github.com/sanic-org/sanic/blob/main/sanic/worker/process.py#L109

uvicorn/supervisors/multiprocess.py
ahopkins: Mocks are wonderful :wink: 

I also would second the idea about collapsing. And if testing is a concern, then wrap them one abstraction layer higher. Personally, I would not let that dictate the API, regardless it doesn't negate the idea that it could all be a single method. That would then fix the maintainability issue that you have even now with the inconsistent logging.

uvicorn/supervisors/multiprocess.py
```diff
@@ -32,39 +109,125 @@ def __init__(
...
+    def handle_hup(self) -> None:
+        logger.info("Received SIGHUP, restarting processes")
+        self.restart_all()
+
+    def handle_ttin(self) -> None:
+        logger.info("Received SIGTTIN, increasing processes")
+        self.processes_num += 1
+        process = Process(self.config, self.target, self.sockets)
+        process.start()
+        self.processes.append(process)
```
ahopkins, line 213: This pattern of `Process` instantiation, `start`, and `append` is scattered in several locations. This will IMO become an issue over time as features are built on top of this. I'd suggest centralizing that somehow in a nicer pattern that allows more flexibility.

ahopkins (approved):
I see what the PR is doing, but I somewhat fail to understand the _why_ and the goals. Is the pinging the ultimate feature add here?

Kludex:
> I see what the PR is doing, but I somewhat fail to understand the _why_ and the goals. Is the pinging the ultimate feature add here?

The goal is to not suggest people to use Gunicorn anymore.

uvicorn/supervisors/multiprocess.py
abersheeran: pong blocks until a packet is received. So there won't be much CPU consumption here.

uvicorn/supervisors/multiprocess.py
abersheeran: What problems might this gap cause?

uvicorn/supervisors/multiprocess.py
abersheeran: Excessive abstraction can cause more problems. I don't think a unified approach is needed until many of the new features you talk about are added. At least in my opinion, multi-process managers only have these functions at most (see gunicorn)

uvicorn/supervisors/multiprocess.py
abersheeran: https://github.com/benoitc/gunicorn/blob/88fc4a43152039c28096c8ba3eeadb3fbaa4aff9/gunicorn/arbiter.py#L245-L285 I prefer gunicorn the same way. We don't need a frame here, and there is no need to add testing trouble to ourselves. Of course we can use mocks, but why add unnecessary mocks?

uvicorn/supervisors/multiprocess.py
abersheeran: https://github.com/encode/uvicorn/blob/4a503d84fa8703d7534d810bb10b3a0b0e6e1a39/uvicorn/main.py#L563-L573

By the way, currently uvicorn's reload cannot work with workers. So in this PR, we can ignore this issue for now. We can easily add this feature later, and I want this PR to be as small as possible.

abersheeran:
Next I will submit tests and documentation. If anyone can help, please commit directly to the branch of this PR... This will be a long and painful process. /(ㄒoㄒ)/~~

tests/supervisors/test_multiprocess.py
```diff
@@ -13,9 +18,11 @@ async def app(scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable
 
 
 def run(sockets: list[socket.socket] | None) -> None:
-    pass  # pragma: no cover
+    while True:
+        time.sleep(1)
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="In Windows, Ctrl+C/Ctrl+Break will sent to the parent process.")
```
abersheeran: Can someone help me solve this problem? Or should I ignore the coverage here?

abersheeran: https://github.com/abersheeran/zibai/blob/0d962e0979ba26df809da030469fb7bc2500ce12/tests/test_multiprocess.py#L19 

I am confident that I have found the solution. The Qingming Festival holiday in China is approaching, and I will complete it.

docs/deployment.md
```diff
@@ -178,7 +178,23 @@ Running Uvicorn using a process manager ensures that you can run multiple proces
...
-Uvicorn provides a lightweight way to run multiple worker processes, for example `--workers 4`, but does not provide any process monitoring.
+### Built-in
+
+Uvicorn includes a `--workers` option that allows you to run multiple worker processes.
+
+```bash
+$ uvicorn main:app --workers 4
+```
+
+Unlike gunicorn, uvicorn does not use pre-fork, but uses [`spwan`](https://docs.python.org/zh-cn/3/library/multiprocessing.html#contexts-and-start-methods), which allows uvicorn's multi-process manager to still work well on Windows.
```
Kludex:
```suggestion
Unlike gunicorn, uvicorn does not use pre-fork, but uses [`spawn`](https://docs.python.org/zh-cn/3/library/multiprocessing.html#contexts-and-start-methods), which allows uvicorn's multiprocess manager to still work well on Windows.
```

uvicorn/supervisors/multiprocess.py
```diff
@@ -32,39 +110,108 @@ def __init__(
...
+
+    def handle_signals(self) -> None:
+        for sig in tuple(self.signal_queue):
+            self.signal_queue.remove(sig)
+            sig_name = SIGNALS[sig]
+            sig_handler = getattr(self, f"handle_{sig_name.lower()}", None)
+            if sig_handler is not None:
+                sig_handler()
+            else:  # pragma: no cover
+                logger.info(f"Received signal [{sig_name}], but nothing to do")
```
Kludex: I think `debug` is fine for _ignored_ signals.
```suggestion
                logger.debug(f"Received signal {sig_name}, but no handler is defined for it.")
```

Mainly suggesting this because of what I see locally when I resize my terminal window:
```bash
INFO:     Received signal [WINCH], but nothing to do
INFO:     Received signal [WINCH], but nothing to do
INFO:     Received signal [WINCH], but nothing to do
INFO:     Received signal [WINCH], but nothing to do
```

uvicorn/supervisors/multiprocess.py
```diff
@@ -32,39 +110,108 @@ def __init__(
...
+
+    def handle_ttin(self) -> None:  # pragma: py-win32
+        logger.info("Received SIGTTIN, increasing processes")
+        self.processes_num += 1
+        process = Process(self.config, self.target, self.sockets)
+        process.start()
+        self.processes.append(process)
+
+    def handle_ttou(self) -> None:  # pragma: py-win32
+        logger.info("Received SIGTTOU, decreasing processes")
```
Kludex:
```suggestion
        logger.info("Received SIGTTOU, decreasing number of processes.")
```

uvicorn/supervisors/multiprocess.py
```diff
@@ -32,39 +110,108 @@ def __init__(
...
+        if self.processes_num <= 1:
+            logger.info("Cannot decrease processes any more")
```
Kludex:
```suggestion
            logger.info("Already reached one process, cannot decrease the number of processes anymore.")
```

uvicorn/supervisors/multiprocess.py
```diff
@@ -32,39 +110,108 @@ def __init__(
...
+    def handle_break(self) -> None:  # pragma: py-not-win32
+        logger.info("Received SIGBREAK, exiting")
+        self.should_exit.set()
+
+    def handle_hup(self) -> None:  # pragma: py-win32
+        logger.info("Received SIGHUP, restarting processes")
+        self.restart_all()
+
+    def handle_ttin(self) -> None:  # pragma: py-win32
+        logger.info("Received SIGTTIN, increasing processes")
```
Kludex:
```suggestion
        logger.info("Received SIGTTIN, increasing the number of processes.")
```

uvicorn/supervisors/multiprocess.py
```diff
@@ -32,39 +110,108 @@ def __init__(
...
+    def handle_term(self) -> None:
+        logger.info("Received SIGTERM, exiting")
+        self.should_exit.set()
+
+    def handle_break(self) -> None:  # pragma: py-not-win32
+        logger.info("Received SIGBREAK, exiting")
+        self.should_exit.set()
+
+    def handle_hup(self) -> None:  # pragma: py-win32
+        logger.info("Received SIGHUP, restarting processes")
```
Kludex:
```suggestion
        logger.info("Received SIGHUP, restarting processes.")
```

uvicorn/supervisors/multiprocess.py
```diff
@@ -32,39 +110,108 @@ def __init__(
...
+    def handle_int(self) -> None:
+        logger.info("Received SIGINT, exiting")
+        self.should_exit.set()
+
+    def handle_term(self) -> None:
+        logger.info("Received SIGTERM, exiting")
+        self.should_exit.set()
+
+    def handle_break(self) -> None:  # pragma: py-not-win32
+        logger.info("Received SIGBREAK, exiting")
```
Kludex:
```suggestion
        logger.info("Received SIGBREAK, exiting.")
```

uvicorn/supervisors/multiprocess.py
```diff
@@ -32,39 +110,108 @@ def __init__(
...
+                sig_handler()
+            else:  # pragma: no cover
+                logger.info(f"Received signal [{sig_name}], but nothing to do")
+
+    def handle_int(self) -> None:
+        logger.info("Received SIGINT, exiting")
+        self.should_exit.set()
+
+    def handle_term(self) -> None:
+        logger.info("Received SIGTERM, exiting")
```
Kludex:
```suggestion
        logger.info("Received SIGTERM, exiting.")
```

uvicorn/supervisors/multiprocess.py
```diff
@@ -32,39 +110,108 @@ def __init__(
...
+            self.signal_queue.remove(sig)
+            sig_name = SIGNALS[sig]
+            sig_handler = getattr(self, f"handle_{sig_name.lower()}", None)
+            if sig_handler is not None:
+                sig_handler()
+            else:  # pragma: no cover
+                logger.info(f"Received signal [{sig_name}], but nothing to do")
+
+    def handle_int(self) -> None:
+        logger.info("Received SIGINT, exiting")
```
Kludex:
```suggestion
        logger.info("Received SIGINT, exiting.")
```

pyproject.toml
```diff
@@ -101,7 +101,7 @@ omit = [
 
 [tool.coverage.report]
 precision = 2
-fail_under = 98.35
+fail_under = 98.13
```
Kludex: I don't want to decrease this, can we add pragmas where needed?

Kludex (changes requested):
I think we need to change the logs to print the PID, otherwise it gets confusing:

```bash
❯ uvicorn main:app --workers 2
INFO:     Started parent process [9954]
INFO:     Started child process [9956]
INFO:     Started child process [9957]
INFO:     Started server process [9956]
INFO:     Started server process [9957]
```

Kludex:
Also, if I press `CTRL + C` on `uvicorn main:app --workers 2`, it doesn't work as expected.

abersheeran:
> Also, if I press `CTRL + C` on `uvicorn main:app --workers 2`, it doesn't work as expected.

Can you try the latest commit version? This should solve the problem.

docs/deployment.md
```diff
@@ -178,7 +178,23 @@ Running Uvicorn using a process manager ensures that you can run multiple proces
...
-Uvicorn provides a lightweight way to run multiple worker processes, for example `--workers 4`, but does not provide any process monitoring.
+### Built-in
+
+Uvicorn includes a `--workers` option that allows you to run multiple worker processes.
+
+```bash
+$ uvicorn main:app --workers 4
+```
+
+Unlike gunicorn, uvicorn does not use pre-fork, but uses [`spawn`](https://docs.python.org/zh-cn/3/library/multiprocessing.html#contexts-and-start-methods), which allows uvicorn's multiprocess manager to still work well on Windows.
```
baseplate-admin: `https://docs.python.org/zh-cn/3/library/multiprocessing.html#contexts-and-start-methods` -> `https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods`

baseplate-admin (changes requested):
Please use the English version of the documentation instead of chinese version

pyproject.toml
Kludex: @abersheeran can you check what's missing here? I'm already fine with the implementation.

abersheeran: > @abersheeran can you check what's missing here? I'm already fine with the implementation.

I will solve it tonight. I'm glad to hear that you like this implementation.

Kludex:
I think the PID should be shown in the `"uvicorn.access"` logger (only when `workers > 1`).

Is it normal for the child processes to be alive after I kill the manager process?

abersheeran:
> I think the PID should be shown in the `"uvicorn.access"` logger (only when `workers > 1`).

Here I just reused the previous logger. If we want to refactor the logs, I think it can be done in another PR.

> Is it normal for the child processes to be alive after I kill the manager process?

No, normally only killing the manager process through the SIGKILL signal will result in the survival of child processes (this is due to system signal mechanism reasons, and it also exists in all other multi-process managers), under other circumstances, child processes should be killed.

pyproject.toml
```diff
@@ -115,7 +115,11 @@ exclude_lines = [
 ]
 
 [tool.coverage.coverage_conditional_plugin.omit]
-"sys_platform == 'win32'" = ["uvicorn/loops/uvloop.py"]
+"sys_platform == 'win32'" = [
+    "uvicorn/loops/uvloop.py",
+    "uvicorn/supervisors/multiprocess.py",
+    "tests/supervisors/test_multiprocess.py",
```
Kludex, line 121: Are you doing this just to not add a lot of pragmas?

abersheeran: Yes, because of the signaling mechanism on windows, although we can test part of it, coverage can't catch it.

Kludex (approved):

Kludex:
Thanks @abersheeran 🙏

uwu-420:
I know that some use SIGHUP to signal an app to reload its configuration ( https://en.wikipedia.org/wiki/Signal_(IPC)#SIGHUP ) but imo the only correct interpretation of this signal is to shutdown. After all it's the signal that a process receives when the terminal, from which it has been started, is closed. @abersheeran @Kludex I think the handling of SIGHUP should be exactly the same as for SIGINT and SIGTERM, i.e. shut down the app gracefully. Or am I missing something?

Related discussion https://github.com/encode/uvicorn/discussions/1869

MarkofChina:
is this feature stable enough for production now? I've been using gunicorn all the way.

PR #2218, opened 2024-01-17
Raise `Disconnect` on `send()` when client disconnected

Kludex:
- Comply with https://github.com/django/asgiref/pull/431.

### DIFF
--- a/uvicorn/protocols/utils.py
+++ b/uvicorn/protocols/utils.py
@@ -6,6 +6,10 @@ import urllib.parse
 from uvicorn._types import WWWScope
 
 
+class Disconnected(IOError):
+    ...
+
+
 def get_remote_addr(transport: asyncio.Transport) -> tuple[str, int] | None:
     socket_info = transport.get_extra_info("socket")
     if socket_info is not None:
--- a/uvicorn/protocols/utils.py
+++ b/uvicorn/protocols/utils.py
@@ -6,7 +6,7 @@ import urllib.parse
 from uvicorn._types import WWWScope
 
 
-class Disconnected(IOError):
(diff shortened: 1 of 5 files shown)


uvicorn/protocols/utils.py
```diff
@@ -6,6 +6,10 @@
 from uvicorn._types import WWWScope
 
 
+class Disconnected(IOError):
```
adriangb: Maybe ClientDisconnected?

adriangb (commented):
What about http?

Kludex:
> What about http?

I don't know yet, but shouldn't be a blocker for this PR.

adriangb:
Ok just wanted to know it will be handled in a future PR

adriangb (approved):

Kludex:
> Ok just wanted to know it will be handled in a future PR

I have personal plans to work on this on the next days, but I don't want to promise. 👀

adriangb:
No need to promise. I just wanted to make sure you're not proposing that it doesn't get implemented by anyone at any time (i.e. this PR is all we're going to do).

uvicorn/protocols/utils.py
Kludex: done

uvicorn/protocols/utils.py
```diff
@@ -6,6 +6,10 @@
 from uvicorn._types import WWWScope
 
 
+class ClientDisconnected(IOError):
```
graingert, line 9: IOError is an alias for OSError on python 3

adriangb: Ouch I think we put `IOError` in the spec...

PR #2220, opened 2024-01-20
Raise `ClientDisconnected` on `send()` when client disconnected

Kludex:
~We need to clarify this on https://github.com/django/asgiref/commit/3f0147dc79bae052113b31a02db4f2bdfffc7d4e before merging it.~

👆 Clarified.

Also, tests are missing here.

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -4,7 +4,7 @@ import logging
 import socket
 import threading
 import time
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, overload
 
 import pytest
 
@@ -16,6 +16,7 @@ from uvicorn.lifespan.off import LifespanOff
 from uvicorn.lifespan.on import LifespanOn
 from uvicorn.main import ServerState
 from uvicorn.protocols.http.h11_impl import H11Protocol
+from uvicorn.protocols.utils import ClientDisconnected
 
 try:
     from uvicorn.protocols.http.httptools_impl import HttpToolsProtocol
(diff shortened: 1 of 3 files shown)


uvicorn/protocols/http/httptools_impl.py
```diff
@@ -414,11 +414,13 @@ def __init__(
...
 
     # ASGI exception wrapper
-    async def run_asgi(self, app: "ASGI3Application") -> None:
+    async def run_asgi(self, app: ASGI3Application) -> None:
         try:
             result = await app(  # type: ignore[func-returns-value]
                 self.scope, self.receive, self.send
             )
+        except ClientDisconnected:
+            ...
```
Kludex: Do I need to call the `self.transport.close()`?

uvicorn/protocols/http/h11_impl.py
```diff
@@ -412,6 +413,8 @@ async def run_asgi(self, app: "ASGI3Application") -> None:
             result = await app(  # type: ignore[func-returns-value]
                 self.scope, self.receive, self.send
             )
+        except ClientDisconnected:
+            ...
```
Kludex, line 417: Do I need to call `self.transport.close()`?

uvicorn/protocols/http/httptools_impl.py
Kludex: Well... The `connection_lost` callback is called, and the transport is closed there.

Kludex: Actually, not adding the `self.transport.close()` means that we maintain the same behavior as before (which is better).

Kludex:
I'm going to add a test before merging, but review welcome.

uvicorn/protocols/http/h11_impl.py
adriangb:
```suggestion
            pass
```

I think is more common in this context?

adriangb (commented):
Looks good, just missing a test

tests/protocols/test_http.py
```diff
@@ -629,9 +620,26 @@ async def app(scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable
 
 
 @pytest.mark.anyio
-async def test_early_response(
-    http_protocol_cls: HTTPProtocol,
-):
+async def test_disconnect_on_send(http_protocol_cls: HTTPProtocol) -> None:
```
Kludex, line 603: Test added.

adriangb (approved):

PR #2225, opened 2024-01-21
Change CTRL_C_EVENT to the SIGINT signal, as PyCharm does not handle CTRL_C_EVENT signals.

liuyiyangwang:
# Summary

Unfortunately, starting from uvicorn 0.21.1, when developing with PyCharm, reload cannot terminate the process properly.
Change CTRL_C_EVENT to the SIGINT signal, as PyCharm does not handle CTRL_C_EVENT signals.

`os.kill(self.process.pid, signal.CTRL_C_EVENT)`  -- >>  `os.kill(self.process.pid, signal.SIGINT)` 

I am certain that this is an issue with PyCharm. Unfortunately, they haven't responded to my feedback. I hope the author can make a compromise and merge this pull request, as it addresses a significant usability issue.

### DIFF
--- a/uvicorn/supervisors/basereload.py
+++ b/uvicorn/supervisors/basereload.py
@@ -90,7 +90,7 @@ class BaseReload:
         if sys.platform == "win32":  # pragma: py-not-win32
             self.is_restarting = True
             assert self.process.pid is not None
-            os.kill(self.process.pid, signal.CTRL_C_EVENT)
+            os.kill(self.process.pid, signal.SIGINT)
         else:  # pragma: py-win32
             self.process.terminate()
         self.process.join()


abersheeran:
https://docs.python.org/3/library/os.html#os.kill

On Windows, os.kill(self.process.pid, signal.SIGINT) does not send the Ctrl+C signal, but uses TerminateProcess to kill the process directly.

liuyiyangwang:
> https://docs.python.org/3/library/os.html#os.kill
> 
> On Windows, os.kill(self.process.pid, signal.SIGINT) does not send the Ctrl+C signal, but uses TerminateProcess to kill the process directly.

I am well aware of the purpose of signal.SIGINT. It's because of a bug in PyCharm that I'm using signal.SIGINT to terminate the process.

Kludex:
If the bug is in PyCharm, why fixing it here?

liuyiyangwang:
> If the bug is in PyCharm, why fixing it here?


The issues raised in PyCharm have been unresolved for six months now, with no response received to date. This problem is severely affecting PyCharm users' ability to use Uvicorn with FastAPI. Due to this, we hope the author will consider compromising.

Kludex:
Sorry, but this doesn't make sense.

This is a volunteer driven project, which is getting $2.7k dollars in donation a month. PyCharm is the product of JetBrains, which is valuated at $7 billion, which doesn't even sponsor this organization.

In any case, since we've confirmed the problem is on their side, let them fix it.

liuyiyangwang:
> Sorry, but this doesn't make sense.
> 
> This is a volunteer driven project, which is getting $2.7k dollars in donation a month. PyCharm is the product of JetBrains, which is valuated at $7 billion, which doesn't even sponsor this organization.
> 
> In any case, since we've confirmed the problem is on their side, let them fix it.


It's been over half a year, and we haven't received any response from them. Although I understand that this is an issue that PyCharm should address, the problems it has caused have already impacted many people. There have been numerous reports about this issue both in our project and in FastAPI. (⋟﹏⋞)

PR #2229, opened 2024-01-24
Downgrade mkdocs-material

Kludex:
I need this to fix the docs...

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,4 +29,4 @@ watchgod==0.8.2
 
 # Documentation
 mkdocs==1.5.3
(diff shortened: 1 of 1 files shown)


frankie567 (approved):
Go!

lovelydinosaur:
Do we know how this regression was introduced in `mkdocs` or `mkdocs-material`?
Is there an open issue?

Kludex:
No, there's no issue, and I didn't investigate it.

I also can't reproduce locally with uvicorn.

(an issue would be nice)

PR #1611, opened 2022-08-22
Improve ProxyHeadersMiddleware

pypae:
**Summary**
This PR addresses multiple issues mentioned in https://github.com/encode/uvicorn/issues/1068 to improve the `ProxyHeadersMiddleware`.
- :bug: Fix the `host` for requests from clients running on the proxy server itself. (The main issue)
- :bug: Fallback to host that was already set for empty `x-forwarded-for` headers. ([Mentioned](https://github.com/encode/uvicorn/issues/1068#issuecomment-855371576) by @b0g3r)
- ✨ Also allow to specify IP ranges in CIDR notation as trusted hosts. ([Mentioned](https://github.com/encode/uvicorn/issues/1068#issuecomment-1004813267) by @jalaziz) This greatly simplifies deployments on docker swarm / kubernetes, where the reverse proxy might have a dynamic IP.

Issue #1974, opened 2023-05-10
`ProxyHeadersMiddleware` not using the `X-Forwarded-Port` header


fiannac:
### Discussed in https://github.com/encode/uvicorn/discussions/1948

<div type='discussions-op-text'>

<sup>Originally posted by **fiannac** April 21, 2023</sup>
From the documentation:
--proxy-headers / --no-proxy-headers   Enable/Disable X-Forwarded-Proto, X-Forwarded-For, X-Forwarded-Port to populate remote address info.

But I've noticed that, when the flag is set, the `X-Forwarded-Port` is not actually used by the application, and the source port is always set to 0.
This has to do with how the `ProxyHeadersMiddleware' is implemented, which does not match the behavior described in the docs.</div>

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/uvicorn/issues/1974">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/1974/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/uvicorn/issues/1974/pledge.svg">
</picture>
</a>

Kludex:
The discussion is already created.

Kludex:
Is there any server implementation that uses `X-Forwarded-Port`?

victor-torres:
@Kludex https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html

RussianCow:
My understanding is that, without support for `X-Forwarded-Port`, there is no way to host a Uvicorn server on a non-standard port if it's behind a reverse proxy such as nginx. @Kludex Is there a workaround for this issue? Otherwise, I'm happy to contribute a PR (although #2473 should be sufficient if you remove the flag).

keatonLiu:
I finally wrote a middleware to solve this problem temporarily:
```python
from uvicorn._types import ASGI3Application, ASGIReceiveCallable, Scope, ASGISendCallable


class ProxyHeadersMiddlewareWithPort:
    def __init__(self, app: ASGI3Application) -> None:
        self.app = app

[...]

PR #2231, opened 2024-01-25
Improve ProxyHeadersMiddleware

nhairs:
# Foreword
This PR is a continuation of #1611 building upon the existing work of @pypae

# Summary

This PR addresses multiple issues mentioned in #1068 to improve the `ProxyHeadersMiddleware`.

* 🐛 Fix the `host` for requests from clients running on the proxy server itself. (The main issue)
* 🐛 Fallback to host that was already set for empty `x-forwarded-for` headers. ([Mentioned](https://github.com/encode/uvicorn/issues/1068#issuecomment-855371576) by @b0g3r)
* ✨ Also allow to specify IP Networks as trusted hosts. ([Mentioned](https://github.com/encode/uvicorn/issues/1068#issuecomment-1004813267) by @jalaziz) This greatly simplifies deployments on docker swarm / kubernetes, where the reverse proxy might have a dynamic IP.
  - This includes support for IPv6 Address / Networks

# Checklist

  - I believe that the `_TrustedHost` test cases are adequate for testing that IPv6 is supported without also passing values through the `ProxyHeaderMiddleware` test cases as well.

# Test plan

Advanced testing using the following NGINX config which can be tweaked to test different combinations of proxies.

```nginx
server {
    server_name proxy1;

    listen 127.0.0.1:80;
    listen [::1]:80;
    listen unix:/tmp/proxy1.sock;

    location / {
        proxy_set_header Host $http_host;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection $connection_upgrade;
        proxy_redirect off;
        proxy_buffering off;
        proxy_pass http://uvicorn;
    }
}


server {
    server_name proxy2;

    listen 127.0.100.1:80;
    listen [::1]:8080;

    # https://blog.davidsierra.dev/posts/connect-nginxs-through-sockets/
    listen unix:/tmp/proxy2.sock;

    location / {
        proxy_set_header Host $http_host;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection $connection_upgrade;
        proxy_redirect off;
        proxy_buffering off;
        proxy_pass http://unix:/tmp/proxy1.sock;
    }
}

map $http_upgrade $connection_upgrade {
    default upgrade;
    '' close;
}

upstream uvicorn {
    server 127.0.0.1:9000;
    server unix:/tmp/uvicorn.sock;
}
```


nhairs:
@Kludex - this is mostly done

A few questions though:

q1: I've noticed that in the README we state the project is 3.8+ though the pyproject.yaml is for 3.7+

Should we do anything about this?

~q2: I'm getting a bunch of lint errors from elsewhere in the package, should I fix these up in the same PR?~

Turns out I should have run `./scripts/install` first :upside_down_face: 

image

~q3: I don't know enough about Uvicorn to understand what I should be doing here for this test case (in such a way that it fixes the lint errors)~

Nevermind fixed it myself.

image

docs/deployment.md
```diff
@@ -255,7 +255,7 @@ Using Nginx as a proxy in front of your Uvicorn processes may not be necessary,
 In managed environments such as `Heroku`, you won't typically need to configure Nginx, as your server processes will already be running behind load balancing proxies.
 
 The recommended configuration for proxying from Nginx is to use a UNIX domain socket between Nginx and whatever the process manager that is being used to run Uvicorn.
-Note that when doing this you will need to run Uvicorn with `--forwarded-allow-ips='*'` to ensure that the domain socket is trusted as a source from which to proxy headers.
+Note that when doing this you will need to run Uvicorn with `--forwarded-allow-ips='/path/to/socket'` to ensure that the domain socket is trusted as a source from which to proxy headers.
```
nhairs: I've not actually tested this, but from all my reading I'm pretty sure this is how it works...

tests/middleware/test_proxy_headers.py
```diff
@@ -142,3 +432,24 @@ async def websocket_app(scope, receive, send):
...
             data = await websocket.recv()
             assert data == "wss://1.2.3.4:0"
+
+
+@pytest.mark.anyio
+async def test_proxy_headers_empty_x_forwarded_for() -> None:
+    # fallback to the default behavior if x-forwarded-for is an empty list
+    # https://github.com/encode/uvicorn/issues/1068#issuecomment-855371576
+    app_with_middleware = ProxyHeadersMiddleware(app, trusted_hosts="*")
+    transport = httpx.ASGITransport(app=app_with_middleware, client=("1.2.3.4", 8080))
```
nhairs, line 442: This has the following typing error but I don't know how I should be fixing it.

```console
tests/middleware/test_proxy_headers.py:442: error: Argument "app" to "ASGITransport" has incompatible type "ProxyHeadersMiddleware"; expected "Callable[[dict[str, Any], Callable[[], Awaitable[dict[str, Any]]], Callable[[dict[str, Any]], Coroutine[None, None, None]]], Coroutine[None, None, None]]"  [arg-type]
tests/middleware/test_proxy_headers.py:442: note: "ProxyHeadersMiddleware.__call__" has type "Callable[[Arg(HTTPScope | WebSocketScope | LifespanScope, 'scope'), Arg(Callable[[], Awaitable[HTTPRequestEvent | HTTPDisconnectEvent | WebSocketConnectEvent | _WebSocketReceiveEventBytes | _WebSocketReceiveEventText | WebSocketDisconnectEvent | LifespanStartupEvent | LifespanShutdownEvent]], 'receive'), Arg(Callable[[HTTPResponseStartEvent | HTTPResponseBodyEvent | HTTPResponseTrailersEvent | HTTPServerPushEvent | HTTPDisconnectEvent | WebSocketAcceptEvent | _WebSocketSendEventBytes | _WebSocketSendEventText | WebSocketResponseStartEvent | WebSocketResponseBodyEvent | WebSocketCloseEvent | LifespanStartupCompleteEvent | LifespanStartupFailedEvent | LifespanShutdownCompleteEvent | LifespanShutdownFailedEvent], Awaitable[None]], 'send')], Coroutine[Any, Any, None]]"
```

nhairs:
~I still need to add some more test cases~

I believe this is ready for review @Kludex

docs/deployment.md
```diff
@@ -337,3 +340,44 @@ $ gunicorn --keyfile=./key.pem --certfile=./cert.pem -k uvicorn.workers.UvicornW
...
+
+### UNIX Domain Sockets (UDS)
+
+Although it is common for UNIX Domain Sockets to be used for communicating between various HTTP servers, they can mess with some of the expected received values as they will be various non-address strings or missing values.
+
+For example:
+- when NGINX itself is running behind a UDS it will add the literal `unix:` as the client in the `X-Forwarded-For` header.
+- When Uvicorn is running behind a UDS the initial client will be `None`.
+
+In the case of Uvicorn, if it run using `--uds`, then it will automatically trust `None` as the initial client.
```
nhairs, line 373: So when I originally implemented this this seemed like the logical behaviour, especially to avoid people trusting all hosts (as was previously suggested by the documentation).

But I'm wondering if instead of making it default behaviour we pass it as a flag as well?

Ref: `config.py:501`

docs/settings.md
```diff
@@ -90,7 +90,7 @@ Note that WSGI mode always disables WebSocket support, as it is not supported by
 * `--root-path <str>` - Set the ASGI `root_path` for applications submounted below a given URL path.
 * `--proxy-headers` / `--no-proxy-headers` - Enable/Disable X-Forwarded-Proto, X-Forwarded-For, X-Forwarded-Port to populate remote address info. Defaults to enabled, but is restricted to only trusting
 connecting IPs in the `forwarded-allow-ips` configuration.
-* `--forwarded-allow-ips` <comma-separated-list> Comma separated list of IPs to trust with proxy headers. Defaults to the `$FORWARDED_ALLOW_IPS` environment variable if available, or '127.0.0.1'. A wildcard '*' means always trust.
+* `--forwarded-allow-ips` <comma-separated-list> Comma separated list of IP Addresses, IP Networks, or literals (e.g. UNIX Socket path) to trust with proxy headers. Defaults to the `$FORWARDED_ALLOW_IPS` environment variable if available, or '127.0.0.1'. The literal `'*'` means trust everything.
 * `--server-header` / `--no-server-header` - Enable/Disable default `Server` header.
```
nhairs, line 94: TODO: Add `--forwarded-trust-literals` here since isn't auto-generated.

Kludex:
> q1: I've noticed that in the README we state the project is 3.8+ though the pyproject.yaml is for 3.7+

I don't see that... 🤔

Kludex:
@nhairs You are adding much more stuff on top of the previous PR. Can we do this step by step?

The way I propose:
1. Apply the comments from the previous PR (even your comments if you think they make sense), and open a PR with only that.
2. Create new PRs with the other changes from this PR.

nhairs:
> > q1: I've noticed that in the README we state the project is 3.8+ though the pyproject.yaml is for 3.7+
> 
> I don't see that... 🤔

Ah never-mind, I suspect it's because of the original PR being so old and referencing the older code against the current website :facepalm:.

nhairs:
> @nhairs You are adding much more stuff on top of the previous PR. Can we do this step by step?
>
> ...

Are you able to be more specific about what parts you're concerned about?

If we exclude update related tests / docs, I think we are left with the following:

A: supporting both IP version 4 and 6 (which is in the comments I left on the original PR).

I don't think it makes sense to separate generic support versus v4 only support.

The direct support of IP addresses instead of just networks is simply an optimisation / guard rail for users - strictly speaking you could remove the specialised handling code and the user experience would be the same. I chose to keep it because "justifying the PR" could be complicated when I could just keep it apart of the large changes occurring.

B: Better Unix socket handling.

Whilst I could pull this into a separate PR, I'd prefer to fix this at the same time. The current recommended solution of trusting everything could very easily introduce vulnerabilities to user's applications. I'd rather not tell users to aim a gun at their foot.

C: There's some other miscellaneous fixes referenced in the original PR that could be fixed at the same time.

Rather than introducing bugs that we'll either immediately replace with the refactor this PR does or immediately submit a PR for doesn't make sense to me.

If you have a different view of "what the chunks are" I'm happy to discuss :slightly_smiling_face:

----------

Related to your question in #2237 perhaps we are looking some of this the wrong way - specifically trying to auto-magically support all proxy-header use cases from the command line. Perhaps it would be better to deprecate the command-line support (that is probably best described as a stop-gap measure) and instead provide a more fleshed out set of proxy middle-ware that user's can configure themselves before wrapping their application in it.

If this were the case it's arguable that such "middleware" might be better abstracted into it's own package so that any ASGI compatible project can re-use the logic without needing to implement it itself (if there's one thing I've learnt looking into these headers its that properly handling them is non-trivial). Or even generic enough that it can provide ASGI, WSGI, or raw handlers.

Kludex:
Please remove the new introduced cli args and params to the middleware from this PR.

nhairs:
> Please remove the new introduced cli args and params to the middleware from this PR.

This is done

uvicorn/middleware/proxy_headers.py
```diff
@@ -1,14 +1,5 @@
-"""
-This middleware can be used when a known proxy is fronting the application,
-and is trusted to be properly setting the `X-Forwarded-Proto` and
-`X-Forwarded-For` headers with the connecting client information.
-
-Modifies the `client` and `scheme` information so that they reference
-the connecting client, rather that the connecting proxy.
-
-https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#Proxies
-"""
```
Kludex, line 10: Why did you remove this?

nhairs: I moved this from being module level documentation to class level documentation on `ProxyHeadersMiddleware`.

I believe this is the more appropriate location for this documentation as:

- This module now contains multiple classes.
- The module doesn't modify the request, the `ProxyHeadersMiddleware` does.
- In the future we may want to have different middleware classes for different use-cases.
  - For example we may want to have a different class for handling the `Forwarded` header - we'd likely want a different class because servers that trust the `Forwarded` header will likely want to ignore the `X-Forwarded-*` headers. Same applies if we wanted to add support for the `X-Real-IP` header that sometimes gets used.

uvicorn/config.py
```diff
@@ -493,7 +493,8 @@ def load(self) -> None:
             self.loaded_app = MessageLoggerMiddleware(self.loaded_app)
         if self.proxy_headers:
             self.loaded_app = ProxyHeadersMiddleware(
-                self.loaded_app, trusted_hosts=self.forwarded_allow_ips
+                self.loaded_app,
+                trusted_hosts=self.forwarded_allow_ips,
```
```suggestion
                self.loaded_app, trusted_hosts=self.forwarded_allow_ips
```

uvicorn/middleware/proxy_headers.py
```diff
@@ -20,40 +11,129 @@
...
+    """
+
+    def __init__(
+        self,
+        app: "ASGI3Application",
+        trusted_hosts: Union[List[str], str] = "127.0.0.1",
+    ) -> None:
+        self.app = app
+        self.trusted_hosts = _TrustedHosts(trusted_hosts)
+        return
```
Kludex:
```suggestion
```

docs/deployment.md
```diff
@@ -337,3 +336,37 @@ $ gunicorn --keyfile=./key.pem --certfile=./cert.pem -k uvicorn.workers.UvicornW
...
+
+### UNIX Domain Sockets (UDS)
+
+Although it is common for UNIX Domain Sockets to be used for communicating between various HTTP servers, they can mess with some of the expected received values as they will be various non-address strings or missing values.
+
+For example:
+- when NGINX itself is running behind a UDS it will add the literal `unix:` as the client in the `X-Forwarded-For` header.
+- When Uvicorn is running behind a UDS the initial client will be `None`.
+
+
```
Kludex, line 378:
```suggestion
```

uvicorn/middleware/proxy_headers.py
```diff
@@ -20,40 +11,129 @@
...
+
+    This middleware can be used when a known proxy is fronting the application,
+    and is trusted to be properly setting the `X-Forwarded-Proto` and
+    `X-Forwarded-For` headers with the connecting client information.
+
+    Modifies the `client` and `scheme` information so that they reference
+    the connecting client, rather that the connecting proxy.
+
+    References:
+
```
Kludex:
```suggestion
```

uvicorn/middleware/proxy_headers.py
```diff
@@ -70,15 +150,22 @@ async def __call__(
...
+                        # See: https://github.com/encode/uvicorn/issues/1068
+
+                        # Unless we can relaibly use x-forwarded-port (see below) then
+                        # we will not have any port information so set it to 0.
+                        port = 0
+                        scope["client"] = (host, port)
+
+                    # if b"x-forwarded-port" in headers:
+                    #     ...
+                    # https://github.com/encode/uvicorn/issues/1974
```
Kludex: - I've reopened https://github.com/encode/uvicorn/issues/1974, and I'll accept a follow-up PR.
```suggestion
                        # We've lost the connecting client's port information by now,
                        # so only include the host.
                        port = 0
                        scope["client"] = (host, port)
```

Kludex (changes requested):
@nhairs Would you mind rebasing it, and checking if the `proxy_headers.py` file is fully covered by tests? I'll continue my review as soon as that is done.

I want to clean up the PRs we have here.

nhairs:
@Kludex - master has been merged in

uvicorn/config.py
Kludex: ?

uvicorn/middleware/proxy_headers.py
```diff
@@ -1,69 +1,150 @@
...
                 if b"x-forwarded-proto" in headers:
-                    # Determine if the incoming request was http or https based on
-                    # the X-Forwarded-Proto header.
                     x_forwarded_proto = headers[b"x-forwarded-proto"].decode("latin1").strip()
-                    if scope["type"] == "websocket":
-                        scope["scheme"] = x_forwarded_proto.replace("http", "ws")
-                    else:
-                        scope["scheme"] = x_forwarded_proto
+
+                    if x_forwarded_proto in {"http", "https", "ws", "wss"}:
```
Kludex, line 130: Why do we need this conditional?

tests/middleware/test_proxy_headers.py
```diff
@@ -19,96 +20,442 @@
...
+
+    Args:
+        trusted_hosts: trusted_hosts for proxy middleware
+        client: transport client to use
+    """
+
+    app = cast(
+        httpx._transports.asgi._ASGIApp,
+        ProxyHeadersMiddleware(default_app, trusted_hosts),
+    )
```
Kludex: Please remove the cast.

Kludex (commented):
I'll finish the review in a few days. I'm a bit AFK those days.

## Checklist for myself
- [ ] Check if tests makes sense, and think about the `make_*` utility functions (are they noisy or good?)

uvicorn/middleware/proxy_headers.py
nhairs: To prevent garbage being passed through.

E.g. (from the tests) if you pass through `X-forwarded-proto: asdf` it will be ignored.

uvicorn/config.py
nhairs: I missed this when I went through before, also apparently ruff didn't think it needed changing either...

uvicorn/middleware/proxy_headers.py
Kludex: I mean, we know that the `x-forwarded-proto` will assume those values anyway. Is this check saving us from something?

tests/middleware/test_proxy_headers.py
nhairs: I am casting to prevent the following error:
image


I couldn't figure out an easier way to prevent it

uvicorn/middleware/proxy_headers.py
nhairs: There is nothing to guarantee that it will not receive garbage.

tests/middleware/test_proxy_headers.py
Kludex: you can ignore that with `# type: ignore`, it's ok.

tests/middleware/test_proxy_headers.py
```diff
@@ -19,96 +20,439 @@
...
+        proto: forwarded proto value
+    """
+    headers: dict[str, str] = {}
+    if for_ is not None:
+        headers["X-Forwarded-For"] = for_
+
+    if proto is not None:
+        headers["X-Forwarded-Proto"] = proto
+
+    return headers
```
Kludex: Can you remove this, please?

Kludex (changes requested):

tests/middleware/test_proxy_headers.py
Kludex: Create the headers on each test.

nhairs: At this point it feels like you're requesting changes because stylistically you'd write it differently.

Having had to check all the test cases against the new logic and writing new tests where needed, these helper functions definitely helped me write more robust test cases.

Kludex: I think those helpers make it harder for me to read each individual test.

nhairs:
@Kludex - just making sure you got a notification that I updated this for requested changes.

easybe:
What is the status here? I am interested in this (or more precisely #2237) and am willing to help if needed.

nhairs:
> What is the status here? I am interested in this (or more precisely #2237) and am willing to help if needed.

AFAIAA I'm waiting on @Kludex to finish his review.

uvicorn/middleware/proxy_headers.py
```diff
@@ -1,70 +1,150 @@
...
+        #   something that isn't an IP Address e.g. a unix socket.
+
+        if not self.always_trust:
+            if isinstance(trusted_hosts, str):
+                trusted_hosts = _parse_raw_hosts(trusted_hosts)
+
+            for host in trusted_hosts:
+                # Note: because we always convert invalid IP types to literals it
+                # is not possible for the user to know they provided a malformed IP
+                # type - this may lead to unexpected / difficult to debug behaviour.
```
Kludex, line 42: Right. Any suggestions?

Kludex:
I'm not able to push to this branch, I'm not sure why.

I have mainly nitpicks that I'd like to push.

Kludex:
@nhairs Beautiful work.

I think this is the best PR I've ever seen:

- really descriptive description;
- clean implementation;
- meaningful in-code comments;
- update docs;
- and great test coverage.

I'm sorry for the delay on the review here.
 
I hope the delay on the review doesn't demotivate you to open more PRs here. 🙏 

---

I have some docs fixes, typing improvements, and nitpicks that I'll address in https://github.com/encode/uvicorn/pull/2468, since I can't push changes to this branch.

PR #2233, opened 2024-01-26
Fix nav override for newer version of Mkdocs Material

frankie567:
Following #2229, I've investigated why the navigation bar was hidden in newer version of Mkdocs Material. The `partials/nav.html` have slightly changed, so we needed to update our own override.

Now it displays correctly!


Kludex (approved):

Kludex:
Thanks @frankie567 ❤️

PR #2246, opened 2024-02-08
Fixed default to use stdout instead of stderr

iksinskiyt:
# Summary

Hey, I wanted to change that for the longest time. It makes literally no sense  to print every single line of output as stderr instead of stdout, and I belive that this change will improve every single programmer's mental health as well as the visibility inside of terminal. :> For example in pyCharm stderr is defaultly rendered as red, so it actually conflicts with the config that has errors and critical errors in red*. Thanks for your time! ❤

# Checklist

- [✔] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [❌] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [❌] I've updated the documentation accordingly.

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -80,7 +80,7 @@ LOGGING_CONFIG: dict[str, Any] = {
         "default": {
             "formatter": "default",
             "class": "logging.StreamHandler",
-            "stream": "ext://sys.stderr",
+            "stream": "ext://sys.stdout",
         },
         "access": {
             "formatter": "access",


zanieb:
I believe it is proper behavior for log information from a program to go to standard error by default — this is basically the purpose of standard error. I understand that this may be frustrating in e.g. PyCharm where it's rendered differently but perhaps you should consider reconfiguring the logger in your application or redirecting the stream if that's the concern?

Just my 2-cents, thank you for taking the time to contribute to the project!

iksinskiyt:
> I believe it is proper behavior for log information from a program to go to standard error by default — this is basically the purpose of standard error. I understand that this may be frustrating in e.g. PyCharm where it's rendered differently but perhaps you should consider reconfiguring the logger in your application or redirecting the stream if that's the concern?
> 
> Just my 2-cents, thank you for taking the time to contribute to the project!

I understand your approach, I want y'all to know that this commit came out spontaneously as me and my team were working on a project. Back to the topic I belive that using stderr for debug info is not necessary as uvicorn does not have a debug/console layer and a application layer - so it doesn't benefit from the natural separation of those output streams(I might be totally wrong here). Researching this issue further I've found out that it's rather hard to change your terminal's behaviour and prople have been struggling with similar issues. What I'm trying to say is - there is no easy solution, since everydevelopment environment is different, there might be IDEs that doesn't allow the user to temper with how stdout and stderr are handled and doesn't allow replacing it's built in terminal with thr system one (Yes I'm yapping at pyCharm). To sum up things - it was more of a meme commit than anything, but I'm grateful that I could ahare my concerns regarding this topic. For anyone reading this - have a great day 🥰

Kludex:
Please create a discussion with resources to prove your point.

PR #2258, opened 2024-02-29
Use `X-Forwarded-Proto` for WebSockets scheme when the proxy provides it

aadnehovda:
# Summary

Fix for https://github.com/encode/uvicorn/pull/2043 which set the scheme to `ws` when the proxy passes `wss` in `X-Forwarded-Proto`. Traefik sets the proto to `ws`/`wss` for websockets. https://github.com/traefik/traefik/blob/c1ef7429771104e79f2e87b236b21495cb5765f0/pkg/middlewares/forwardedheaders/forwarded_header.go#L148-L153

(please note the code is not tested)

# Checklist


Kludex (approved):

Kludex (approved):

Kludex:
Thanks @aadnehovda 👍

PR #2265, opened 2024-03-05
Add CIDR optionality to `ProxyHeadersMiddleware`

b-rowan:
# Summary

Added the ability to pass CIDRs to `ProxyHeadersMiddleware` using the stdlib `ipaddress` library.  This should fix this comment - https://github.com/encode/uvicorn/issues/1068#issuecomment-1004813267
# Checklist


Kludex:
This was fixed some time ago.

PR #2278, opened 2024-03-19
Add release notes to the documentation

Kludex:


docs/release-notes.md
```diff
@@ -0,0 +1 @@
+--8<-- "CHANGELOG.md"
```
lovelydinosaur, line 1: 🤔

Kludex: With this we have the CHANGELOG on the root level, and we reference it from the documentation, and we don't duplicate the content.

Kludex:
It's this kind of PR that I don't think it makes sense for me to wait, because I want to do more... I want to be granular on the PRs merged, and if I keep working on improvements, I would need to rebase on top of this branch, and get a list of PRs, which will take longer to be approved, and it will be more work to rebase...

This is a comment based on what we talked about "merging some PRs without review" @tomchristie

frankie567 (approved):
Looks sensible to me, the include pattern is clever to show the changelog in Mkdocs.

PR #2297, opened 2024-04-05
tidied up #1815

MadsRC:
# Summary

This PR tidies up the old PR #1815.

The code is a carbon copy, except for some added testing and some lint and type hint fixes.

I could not find any API documentation to update, and opted to not mention anything about ssl_context for the CLI documentation, given that I ripped out the CLI portion of the PR. Please correct me if I'm mistaken and API documentation does exist (that needs to be updated).

# Checklist


Kludex:
Pipeline is failing.

We need to think about the different ways we can implement this.

MadsRC:
The reason for the broken pipeline was that GitHub had an outage the same day I pushed. I thought I had added a comment asking for the pipeline to be retried, but alas I can't see that I did.

Anyways, the recent update to master allowed me to update this branch and re-trigger the pipeline.

One of the tests is failing for 3.9 on MacOS latest - I'm not sure if it is a fluke/flaky test or if it is related to these changes - I will investigate.

Would it be possible to ask for that pipeline to be re-run?

MadsRC:
Thank you to who ever re-ran the pipeline. It looks like it was a flaky test. The new failure is on Windows, Py 3.10 for code coverage, which `0.01` point from being acceptable. I'll see what I can do about getting a but more unit tests in to bump that to acceptable levels ;)

MadsRC:
After running the tests several times locally, I can see that some of the tests are flaky, giving different coverage results. I'll try and see if I can't add more SSL tests to make up for that fact, but I'm quite unfamiliar with this codebase.

MadsRC:
@Kludex, this PR's still marked as "awaiting author" - May I ask what it is we are waiting for, so that I can provide it?

There was a mention of re-assessing how we went about implementing the change, but I'm not sure if that was because of the failing pipeline, or if it was an ask for me to try and add the functionality in some other way?

PR #2305, opened 2024-04-14
Add Python 3.13 to test suite

Kludex:


Kludex:
As every year, we need to wait for httptools and uvloop.

Let's reopen when they support 3.13.

PR #2309, opened 2024-04-17
fix: stop server on "lifespan.shutdown.failed"

### LINKED ISSUE #2308: Should server shutdown after receiving "lifespan.shutdown.failed"?
### Discussed in https://github.com/encode/uvicorn/discussions/2298

<sup>Originally posted by **peterschutt** April  5, 2024</sup>
If the shutdown failure message isn't initiated by the app receiving a "lifespan.shutdown" from uvicorn, then the app continues to run after the "lifespan.shutdown.failure" message is received.

Reproducer:

```py
import asyncio
import contextlib
from collections.abc import AsyncIterator

import anyio
import uvicorn
from starlette.responses import JSONResponse


async def homepage(_) -> JSONResponse:
    return JSONResponse({'hello': 'world'})


async def sleep_and_raise() -> None:
    await asyncio.sleep(1)
    raise RuntimeError("An error occurred")


@contextlib.asynccontextmanager
async def lifespan(_: Starlette) -> AsyncIterator[None]:
    async with contextlib.AsyncExitStack() as stack:
        tg = await stack.enter_async_context(anyio.create_task_group())
        tg.start_soon(sleep_and_raise)
        yield


app = Starlette(debug=True, routes=[Route('/', homepage)], lifespan=lifespan)


if __name__ == "__main__":
    uvicorn.run(app, lifespan="on")
```

After the error occurs in the lifespan task, the app continues to serve:

image

Given that apps like starlette and litestar encourage use of the `lifespan` context for orchestration of things that should have a lifespan equivalent to the application object, then I think it would make sense for the app to stop if something has failed within that lifespan after the app has sent "startup.complete" but before the server has sent "shutdown" to the app.

The [spec][1] seems to agree:

> If a server sees this it should log/print the message provided and then terminate.

[1]: https://asgi.readthedocs.io/en/latest/specs/lifespan.html#shutdown-failed-send-event</div>

> [!IMPORTANT]

<a href="https://polar.sh/encode/uvicorn/issues/2308">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/2308/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

peterschutt:
# Summary

Stop the server if a "lifespan.shutdown.failure" message is received from the application after "lifespan.startup.complete" has been sent by the application, but before the server has sent "lifespan.shutdown".

Applications typically use a context manager around the asgi lifespan protocol to manage resources that should share the same lifecycle as the web server. Where a managed resource encounters an error before the server has initiated shutdown, app frameworks such as starlette and litestar will send a `lifespan.shutdown.failed` message.

E.g, in [starlette](https://github.com/encode/starlette/blob/eb76cae6fdb6c1b0bfcace17d0dec946fe767f84/starlette/routing.py#L723-L750):

```py
    async def lifespan(self, scope: Scope, receive: Receive, send: Send) -> None:
        started = False
        app: typing.Any = scope.get("app")
        await receive()
        try:
            async with self.lifespan_context(app) as maybe_state:
                ...
        except BaseException:
            exc_text = traceback.format_exc()
            if started:
                await send({"type": "lifespan.shutdown.failed", "message": exc_text})
            else:
                await send({"type": "lifespan.startup.failed", "message": exc_text})
            raise
        else:
```

This PR gives the application a way to tell the server that something critical has failed internally, and that it should shut down.

Closes #2308

# Checklist

### DIFF
--- a/uvicorn/lifespan/on.py
+++ b/uvicorn/lifespan/on.py
@@ -71,7 +71,6 @@ class LifespanOn:
 
         if self.shutdown_failed or (self.error_occured and self.config.lifespan == "on"):
             self.logger.error("Application shutdown failed. Exiting.")
-            self.should_exit = True
         else:
             self.logger.info("Application shutdown complete.")
 
@@ -130,6 +129,7 @@ class LifespanOn:
             assert not self.shutdown_event.is_set(), STATE_TRANSITION_ERROR
             self.shutdown_event.set()
             self.shutdown_failed = True
+            self.should_exit = True
             if message.get("message"):
                 self.logger.error(message["message"])
 
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -252,7 +252,7 @@ class Server:
             return True
         if self.config.limit_max_requests is not None:
             return self.server_state.total_requests >= self.config.limit_max_requests
-        return False
+        return self.lifespan.should_exit
 
     async def shutdown(self, sockets: list[socket.socket] | None = None) -> None:
         logger.info("Shutting down")
--- a/tests/test_lifespan.py
+++ b/tests/test_lifespan.py
@@ -262,3 +262,32 @@ def test_lifespan_state():
     loop = asyncio.new_event_loop()
     loop.run_until_complete(test())
     loop.close()
+
+
+@pytest.mark.parametrize("mode", ("auto", "on"))
+@pytest.mark.parametrize("raise_exception", (True, False))
+def test_lifespan_receive_shutdown_failed_before_shutdown(mode, raise_exception, caplog):
+    async def app(scope, receive, send):
+        message = await receive()
+        assert message["type"] == "lifespan.startup"
+        await send({"type": "lifespan.startup.complete"})
+        await send({"type": "lifespan.shutdown.failed", "message": "the lifespan event failed"})
+
+        if raise_exception:
+            # App should be able to re-raise an exception caught in lifespan context.
+            raise RuntimeError()
+
+    async def test():
+        config = Config(app=app, lifespan=mode)
+        lifespan = LifespanOn(config)
+        await lifespan.startup()
+        assert lifespan.shutdown_failed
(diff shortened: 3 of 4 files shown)


uvicorn/lifespan/on.py
```diff
@@ -71,7 +71,6 @@ async def shutdown(self) -> None:
 
         if self.shutdown_failed or (self.error_occured and self.config.lifespan == "on"):
             self.logger.error("Application shutdown failed. Exiting.")
-            self.should_exit = True
```
euri10, line 74: I played with this line removal in the shutdown method a little, as it surprised me at first, but it seems to have no effect whatsoever on tests ie having the line or not doesnt change any outcome.

This would advocate for removing entirely the line (provided that this is tested ofc), that said to stay on the safe side would there be a case where we'd still need it ? It's used here and a sys.exit is performed in that case:

https://github.com/encode/uvicorn/blob/0efd3835da6dcc713f74aadf7b52779d0d1fa17d/uvicorn/server.py#L169-L171

wdyt ?

uvicorn/server.py
```diff
@@ -252,7 +252,7 @@ async def on_tick(self, counter: int) -> bool:
             return True
         if self.config.limit_max_requests is not None:
             return self.server_state.total_requests >= self.config.limit_max_requests
-        return False
+        return self.lifespan.should_exit
```
euri10, line 255:
```suggestion
        # Determine if we should exit.
        if self.should_exit or self.lifespan.should_exit:
            return True
        if self.config.limit_max_requests is not None:
            return self.server_state.total_requests >= self.config.limit_max_requests
        return False
```

would that suggestion be marginally better ? ie we test for exit in `on_tick` on both the server `should_exit` flag and the lifespan one instead of delaying it at the end ? either way is ok to me though

Kludex:
I've explained on the issue why this is not an issue.

Issue #2294, opened 2024-03-30
Unexcepted behavior while reloading


TahaAhvazi:
### Initial Checks

- [x] I'm aware that if I created this issue without a discussion, it may be closed without a response.

### Description

Hello, in the uvicorn version **0.29.0** which is the latest version on 3/30/2024 when I wrote my code updated it, and saved the changes, I saw unexpected information in the terminal that I mentioned below in the code example, where this error is not occurring in the previous versions of uvicorn like:0.27.1

### Example Code

```Python
INFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [3608]
Process SpawnProcess-2:
Traceback (most recent call last):
  File "C:\Users\AppData\Local\Programs\Python\Python310\lib\multiprocessing\process.py", line 314, in _bootstrap
    self.run()
  File "C:\Users\AppData\Local\Programs\Python\Python310\lib\multiprocessing\process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\backend\env\lib\site-packages\uvicorn\_subprocess.py", line 78, in subprocess_started
    target(sockets=sockets)
  File "C:\-backend\env\lib\site-packages\uvicorn\server.py", line 65, in run
    return asyncio.run(self.serve(sockets=sockets))
  File "C:\Users\AppData\Local\Programs\Python\Python310\lib\asyncio\runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "C:\Users\AppData\Local\Programs\Python\Python310\lib\asyncio\base_events.py", line 633, in run_until_complete
    self.run_forever()
  File "C:\Users\AppData\Local\Programs\Python\Python310\lib\asyncio\base_events.py", line 600, in run_forever
    self._run_once()
  File "C:\Users\AppData\Local\Programs\Python\Python310\lib\asyncio\base_events.py", line 1896, in _run_once
    handle._run()
  File "C:\UsersAppData\Local\Programs\Python\Python310\lib\asyncio\events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "C:\-backend\env\lib\site-packages\uvicorn\server.py", line 68, in serve
    with self.capture_signals():
  File "C:\Users\AppData\Local\Programs\Python\Python310\lib\contextlib.py", line 142, in __exit__
    next(self.gen)
[...]

PR #2314, opened 2024-04-25
Remove forced `raise_signal` calls

### LINKED ISSUE #2289: Child Processes Not Terminating with Uvicorn 0.29.0
### Discussed in https://github.com/encode/uvicorn/discussions/2281

<sup>Originally posted by **karimkalimu** March 21, 2024</sup>
Description
I recently updated to Uvicorn 0.29.0 and encountered a significant issue where child processes spawned by my FastAPI application are not being terminated upon a graceful shutdown. This behavior seems to be specific to the latest version of Uvicorn, as everything worked as expected prior to the update. Despite trying various configurations and explicitly using the --timeout-graceful-shutdown option, these child processes persist, becoming orphaned and reassigned to PID 1 once the main Uvicorn process is killed.

Environment
Uvicorn Version: 0.29.0
Operating System: CentOS 7
FastAPI Version: 0.100.0
Python Version: 3.10.13
Expected Behavior
I would expect all child processes, especially those flagged with daemon=True, to terminate alongside the main process during a graceful shutdown.

Actual Behavior
Instead, the child processes do not terminate as anticipated. They become orphaned and are reassigned to PID 1, persisting beyond the termination of the Uvicorn process.

Steps to Reproduce
Launch a Uvicorn server hosting a FastAPI application that spawns child processes.
Initiate a graceful shutdown of the Uvicorn server.
Notice that the child processes fail to terminate and are instead reassigned to PID 1.
Additional Context
This issue was not present in earlier versions of Uvicorn.
The observed behavior was consistent across two separate systems, both running CentOS 7.
All implicated child processes were configured with daemon=True.
Seeking Insights and Suggestions
Has anyone else encountered this issue with the most recent update of Uvicorn? I'm curious to know if this is an identified bug, if there are any known workarounds, or if perhaps there's a configuration detail I might be overlooking. Additionally, if this behavior indicates a potential bug introduced in Uvicorn 0.29.0, I'm hoping this discussion can aid in identifying and rectifying the issue.

Any insights, advice, or shared experiences would be greatly appreciated. Thank you in advance for your help!</div>

> [!IMPORTANT]

<a href="https://polar.sh/encode/uvicorn/issues/2289">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/2289/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

kleschenko:
# Summary
This change removes the `raise_signal` call which breaks default signal handling for subprocesses.

Closes https://github.com/encode/uvicorn/issues/2289 (and possibly https://github.com/encode/uvicorn/issues/2294)

# Checklist

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1,36 +1,13 @@
 from __future__ import annotations
 
 import asyncio
-import contextlib
 import signal
 import sys
-from typing import Generator
 
 import pytest
 
 from uvicorn.config import Config
 from uvicorn.server import Server
-
-
-# asyncio does NOT allow raising in signal handlers, so to detect
-# raised signals raised a mutable `witness` receives the signal
-@contextlib.contextmanager
-def capture_signal_sync(sig: signal.Signals) -> Generator[list[int], None, None]:
-    """Replace `sig` handling with a normal exception via `signal"""
-    witness: list[int] = []
-    original_handler = signal.signal(sig, lambda signum, frame: witness.append(signum))
-    yield witness
-    signal.signal(sig, original_handler)
-
-
-@contextlib.contextmanager
-def capture_signal_async(sig: signal.Signals) -> Generator[list[int], None, None]:  # pragma: py-win32
-    """Replace `sig` handling with a normal exception via `asyncio"""
-    witness: list[int] = []
-    original_handler = signal.getsignal(sig)
-    asyncio.get_running_loop().add_signal_handler(sig, witness.append, sig)
-    yield witness
-    signal.signal(sig, original_handler)
 
 
 async def dummy_app(scope, receive, send):  # pragma: py-win32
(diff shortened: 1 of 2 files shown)


Kludex:
Well... Then the goal of the previous PR is just ignored...

kleschenko:
@Kludex I've just tested the snippet from the original issue https://github.com/encode/uvicorn/issues/1579 and it works as expected for me with this change

Kludex:
Does this makes sense @maxfischer2781 ?

maxfischer2781:
I honestly don't know how the `raise_signal` is supposed to affect subprocesses in the first place; there is no clear repro for #2289 so I cannot say what is the actual issue.

#2294 should *definitely not* be fixed by removing the raised signal. Yes, the server properly killing itself will lead to a traceback but if the traceback is undesired then only the traceback should be suppressed, not the proper killing.

maxfischer2781:
@Kludex I've created #2317 instead to fix the issues I could reproduce.

PR #2322, opened 2024-05-01
Replace % string formatting with f-string

dolfinus:
# Summary

Update `string % subst` usage with f-string. For logging use `logger.method(string, subst)` instead of explicit message formatting.

# Checklist


baseplate-admin:
I did a [comment](https://github.com/encode/uvicorn/pull/2183#discussion_r1451371549) on a issue, i don't think this is prioritized at this moment :

dolfinus:
That comment just asks "why not f-string", and then closed without any discussion or describing a reason why not.

baseplate-admin:
> then closed without any discussion

The thing is i didn't close the comment.

>  describing a reason why not.

Precisely why i dont think that this formatting issue is not high priority

PR #2324, opened 2024-05-06
Add `reason` support on WebSocketDisconnectEvent

frankie567:
# Summary

In the ASGI specification repo, we recently talked about the [lack of support for the `reason` field in the `WebSocketDisconnectEvent`](https://github.com/django/asgiref/issues/459). This will be likely [added to the spec soon](https://github.com/django/asgiref/pull/462).

This PR adds support for this new addition. It doesn't need to be merged before the spec is officially modified, but we're already prepared here 😄

# Checklist

- [ ] ~I've updated the documentation accordingly.~ *Don't think it's necessary for this change*


Kludex (approved):

surick (approved):

frankie567:
Now that the [ASGI spec change has been merged](https://github.com/django/asgiref/pull/462), I think we can go forward with this 😄

PR #2345, opened 2024-05-21
Allow horizontal tabs `\t` in response header values

jhominal:
# Summary

This PR allows for `\t` to be returned in response header values by uvicorn when `httptools` are in use.
As explained in #2344, both the letter of the current spec (RFC 9210), and the `h11` implementation, allow tab characters to be present in HTTP header values.

# Checklist

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -251,6 +251,25 @@ async def test_get_request(http_protocol_cls: HTTPProtocol):
     assert b"Hello, world" in protocol.transport.buffer
 
 
+@pytest.mark.parametrize(
+    "char",
+    [
+        pytest.param("c", id="allow_ascii_letter"),
+        pytest.param("\t", id="allow_tab"),
+        pytest.param(" ", id="allow_space"),
+        pytest.param("µ", id="allow_non_ascii_char"),
+    ],
+)
+async def test_header_value_allowed_characters(http_protocol_cls: HTTPProtocol, char: str):
+    app = Response("Hello, world", media_type="text/plain", headers={"key": f"<{char}>"})
+    protocol = get_connected_protocol(app, http_protocol_cls)
+    protocol.data_received(SIMPLE_GET_REQUEST)
+    await protocol.loop.run_one()
+    assert b"HTTP/1.1 200 OK" in protocol.transport.buffer
+    assert (b"\r\nkey: <" + char.encode() + b">\r\n") in protocol.transport.buffer
+    assert b"Hello, world" in protocol.transport.buffer
+
+
 @pytest.mark.parametrize("path", ["/", "/?foo", "/?foo=bar", "/?foo=bar&baz=1"])
 async def test_request_logging(path: str, http_protocol_cls: HTTPProtocol, caplog: pytest.LogCaptureFixture):
     get_request_with_query_string = b"\r\n".join(
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -492,6 +492,17 @@ async def test_partial_response_returned(http_protocol_cls: HTTPProtocol):
     assert protocol.transport.is_closing()
 
 
+async def test_response_header_splitting(http_protocol_cls: HTTPProtocol):
+    app = Response(b"", headers={"key": "value\r\nCookie: smuggled=value"})
+
+    protocol = get_connected_protocol(app, http_protocol_cls)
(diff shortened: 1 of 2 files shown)


jhominal:
The main question I have is whether I should be updating the CHANGELOG.md now or if that will be on the person who will be preparing the release to do that.

Kludex:
> The main question I have is whether I should be updating the CHANGELOG.md now or if that will be on the person who will be preparing the release to do that.

I'll do it.

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -26,7 +26,7 @@
 from uvicorn.server import ServerState
 
 HEADER_RE = re.compile(b'[\x00-\x1f\x7f()<>@,;:[]={} \t\\"]')
```
Kludex, line 28: Should the name allow it as well, or is this good?

lovelydinosaur: Hmmm... https://github.com/python-hyper/h11/blob/master/h11/_abnf.py#L21

jhominal: I suspect that, in the case of header names, the above regular expression does not match anything that is actually valid…

I will take another look at the new HTTP RFCs

jhominal:
Another question @Kludex @tomchristie : would you be opposed to adding, in the exception messages, a repr of the invalid value? (note: h11 does that)

uvicorn/protocols/http/httptools_impl.py
jhominal: Sorry for the late comment.

In short:

RFC 9110 has the following relevant bits:

In [Section 5.1 Field names](https://datatracker.ietf.org/doc/html/rfc9110#section-5.1):
```
  field-name     = token
```

And in [Section 5.6.1 Tokens](https://datatracker.ietf.org/doc/html/rfc9110#section-5.6.2):
```
  token          = 1*tchar

  tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                 / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                 / DIGIT / ALPHA
                 ; any VCHAR, except delimiters
```

 * A check of the `tchar` whitelist from RFC 9110 shows that it does not overlap at all with the blacklist of the `HEADER_RE` regular expression. Thus, there is no reason for there to be an issue where someone's valid header name would be blocked by this check;
 * Only 2 ASCII characters are not covered by either the whitelist from RFC 9110, or the `HEADER_RE` blacklist: `/` and `?`. I can add them to the `HEADER_RE` regex if you would like to;
 * In principle, any byte above 0x80 would be forbidden by the standard too. Should they also be added to `HEADER_RE`?

jhominal: In other words, I think that the `HEADER_RE` regular expression for header names is fine, in that it does not block any standard-valid input.

tests/protocols/test_http.py
```diff
@@ -492,6 +511,17 @@ async def app(scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable
...
 
+async def test_response_header_splitting(http_protocol_cls: HTTPProtocol):
+    app = Response(b"", headers={"key": "value\r\nCookie: smuggled=value"})
+
+    protocol = get_connected_protocol(app, http_protocol_cls)
+    protocol.data_received(SIMPLE_GET_REQUEST)
+    await protocol.loop.run_one()
+    assert b"HTTP/1.1 500 Internal Server Error" not in protocol.transport.buffer
+    assert b"\r\nCookie: smuggled=value\r\n" not in protocol.transport.buffer
+    assert protocol.transport.is_closing()
```
Kludex, line 522: What's this test about? This was already passing before.

jhominal: While working on this PR, I did not find a test that checked that invalid response header values were blocked by uvicorn (preventing, among other issues, header response splitting).

Thus, this test fulfills two purposes for me:

1. It blocks me from implementing "allow horizontal tabs" by removing the regular expression check, either wilfully or inadvertently;
2. It confirms that there is no need to go and change the `h11` implementation for header splitting issues;

In other words, I believed that there was no other test that was covering that bit of functionality, so I decided to add one.

Kludex (approved):
Always great to review your work @jhominal :)

Kludex:
If you want, we can make a release right away. Please create a bump PR, if you want.

Also, feel free to merge this.

PR #2360, opened 2024-06-06
fix: upgrade is not websocket and dependencies are installed, should not warning

vvanglro:
# Summary
When I upgraded uvicorn to standard, I found that I encountered a lot of `Unsupported upgrade request.` prompts in the logs, so I debugged locally and found that h11 and httptools handle upgrades with different logic. I think I should keep the same logic.
If the upgrade field is not equal to websocket in h11, the `self._should_upgrade_to_ws()` method will not run, while httptools is different.

# Checklist

### DIFF
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -143,13 +143,15 @@ class HttpToolsProtocol(asyncio.Protocol):
 
     def _should_upgrade_to_ws(self) -> bool:
         if self.ws_protocol_class is None:
-            if self.config.ws == "auto":
-                msg = "Unsupported upgrade request."
-                self.logger.warning(msg)
-                msg = "No supported WebSocket library detected. Please use \"pip install 'uvicorn[standard]'\", or install 'websockets' or 'wsproto' manually."  # noqa: E501
-                self.logger.warning(msg)
             return False
         return True
+
+    def _unsupported_upgrade_warning(self) -> None:
+        msg = "Unsupported upgrade request."
+        self.logger.warning(msg)
+        if not self._should_upgrade_to_ws():
+            msg = "No supported WebSocket library detected. Please use \"pip install 'uvicorn[standard]'\", or install 'websockets' or 'wsproto' manually."  # noqa: E501
+            self.logger.warning(msg)
 
     def _should_upgrade(self) -> bool:
         upgrade = self._get_upgrade()
@@ -170,6 +172,8 @@ class HttpToolsProtocol(asyncio.Protocol):
         except httptools.HttpParserUpgrade:
             if self._should_upgrade():
                 self.handle_websocket_upgrade()
+            else:
+                self._unsupported_upgrade_warning()
 
     def handle_websocket_upgrade(self) -> None:
         if self.logger.level <= TRACE_LOG_LEVEL:
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -147,17 +147,14 @@ class HttpToolsProtocol(asyncio.Protocol):
         return True
 
     def _unsupported_upgrade_warning(self) -> None:
-        msg = "Unsupported upgrade request."
-        self.logger.warning(msg)
+        self.logger.warning("Unsupported upgrade request.")
         if not self._should_upgrade_to_ws():
             msg = "No supported WebSocket library detected. Please use \"pip install 'uvicorn[standard]'\", or install 'websockets' or 'wsproto' manually."  # noqa: E501
             self.logger.warning(msg)
 
     def _should_upgrade(self) -> bool:
         upgrade = self._get_upgrade()
-        if upgrade == b"websocket" and self._should_upgrade_to_ws():
-            return True
-        return False
+        return upgrade == b"websocket" and self._should_upgrade_to_ws()
 
     def data_received(self, data: bytes) -> None:
         self._unset_keepalive_if_required()
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -142,6 +142,8 @@ class HttpToolsProtocol(asyncio.Protocol):
         return None
 
     def _should_upgrade_to_ws(self, upgrade: bytes | None) -> bool:
+        if upgrade != b"websocket":
+            return False
         if upgrade == b"websocket" and self.ws_protocol_class is not None:
             return True
         if self.config.ws == "auto":
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -141,21 +141,21 @@ class HttpToolsProtocol(asyncio.Protocol):
             return upgrade
         return None
 
-    def _should_upgrade_to_ws(self, upgrade: bytes | None) -> bool:
-        if upgrade != b"websocket":
+    def _should_upgrade_to_ws(self) -> bool:
+        if self.ws_protocol_class is None:
+            if self.config.ws == "auto":
+                msg = "Unsupported upgrade request."
+                self.logger.warning(msg)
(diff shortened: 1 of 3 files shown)


vvanglro:
@Kludex Hey, would you look at that?

Kludex:
Can you create a table of expected behavior you see?

Also, while checking this... I've noticed that this is printed multiple times with httptools:

```bash
WARNING:  Unsupported upgrade request.
WARNING:  No supported WebSocket library detected. Please use "pip install 'uvicorn[standard]'", or install 'websockets' or 'wsproto' manually.
WARNING:  Unsupported upgrade request.
WARNING:  No supported WebSocket library detected. Please use "pip install 'uvicorn[standard]'", or install 'websockets' or 'wsproto' manually.
WARNING:  Unsupported upgrade request.
WARNING:  No supported WebSocket library detected. Please use "pip install 'uvicorn[standard]'", or install 'websockets' or 'wsproto' manually.
```

Does this also happens for you?

vvanglro:
I have created a table to describe the expected behavior and the actual behavior I observed:

| Protocol  | Expected Behavior                        | Actual Behavior                                               |
| --------- | ---------------------------------------- | ------------------------------------------------------------- |
| h11       | Ignore upgrade requests not equal to websocket | Works as expected, no warnings                                 |
| httptools | Ignore upgrade requests not equal to websocket | Multiple warnings: "Unsupported upgrade request"              |

yes, I also encountered the warnings multiple times with httptools.

Kludex:
Is this PR fixing the multiple warnings? I tried it, and it didn't change that. I may have done something wrong?

vvanglro:
How did you test it?  
Run the code below without this PR, then you will see multiple warnings when httptools is selected.
```py
from fastapi import FastAPI

app = FastAPI()


@app.get("/")
async def root():
    return {"message": "Hello, World!"}


if __name__ == '__main__':
    import uvicorn
    # uvicorn.run(app, http="h11")
    uvicorn.run(app, http="httptools")
```

```py
import httpx


async def a():
    headers = {"Connection": "upgrade", "Upgrade": "123"}
        response = await client.get('http://0.0.0.0:8000', headers=headers)
        print(response.text)


import asyncio
asyncio.run(a())
```

Kludex:
I'll get back to you in some hours. I'm away.

Kludex:
I'm using [wscat](https://github.com/websockets/wscat).

I'd like to see this working against:

```py
from starlette.routing import WebSocketRoute


async def route(websocket: WebSocket) -> None:
    await websocket.accept()
    await websocket.send_text("Hello, world!")
    await websocket.close()


app = Starlette(routes=[WebSocketRoute("/ws", route)])
```

Running with `uvicorn main:app --http httptools`.

Kludex:
Ah, got it.

The `_should_upgrade` is called multiple times in the httptools one.

vvanglro:
If you are using the standard ws client to request it, no warning will be generated.
If you use http and bring the upgrade upgrade protocol field in the header. Then a warning will be generated.

image

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -141,19 +141,21 @@ def _get_upgrade(self) -> bytes | None:
...
+                msg = "Unsupported upgrade request."
+                self.logger.warning(msg)
+                msg = "No supported WebSocket library detected. Please use \"pip install 'uvicorn[standard]'\", or install 'websockets' or 'wsproto' manually."  # noqa: E501
+                self.logger.warning(msg)
+            return False
+        return True
 
     def _should_upgrade(self) -> bool:
         upgrade = self._get_upgrade()
-        return self._should_upgrade_to_ws(upgrade)
```
vvanglro, line 156: The biggest difference between h11 and httptools is that h11 strictly determines that upgrade is equal to websocket, whereas httptools does not, which leads to warnings when upgrade is not a websocket.

Kludex:
Are you saying that your PR just doesn't solve the multiple warnings on the scenario which should really warn?

vvanglro:
There seems to be two issues now, there should only be one warning prompt if a user receives a request to upgrade ws when they have not chosen to install a websocket dependency.

When a user installs a websocket dependency and receives a request with an incorrect value in the upgrade field, multiple warnings are generated, which is incorrect because I already have the dependency installed, this PR is to fix that.

vvanglro:
Only one warning is generated now.

And the warnings are friendly, if the dependency is installed and the upgrade is not supported, then only prompted the upgrade is not supported. If the dependency is not installed and the upgrade is not supported, then both are prompted.

urec56:
Hey. What the solution of this problem now, guys? I have it on my fastapi app for some time.
image

And it starts like 
```Dokerfile
CMD ["gunicorn", "app.main:app", "--workers", "1", "--worker-class", "uvicorn.workers.UvicornWorker", \
 	"--bind=0.0.0.0:8000", "--max-requests", "100", "--max-requests-jitter", "50"]
```
My libs:
```
h11==0.14.0
httptools==0.6.1
uvicorn==0.26.0
uvloop==0.19.0
websockets==12.0
```

vvanglro:
@urec56 This comment is the solution to this PR: https://github.com/encode/uvicorn/pull/2360#issuecomment-2242088779

If you can reproduce the problem, you can try installing this PR to verify that it solves the problem.
```sh
pip install git+https://github.com/vvanglro/uvicorn.git
```

The expected result is that you will only receive: `[WARNING] Unsupported upgrade request.`

urec56:
Ok, I'll try it with this fix and reply here in a few days

urec56:
So, I tried this fix and it looks like it solved the problem. Thank you, guys

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -141,19 +141,23 @@ def _get_upgrade(self) -> bytes | None:
...
             msg = "No supported WebSocket library detected. Please use \"pip install 'uvicorn[standard]'\", or install 'websockets' or 'wsproto' manually."  # noqa: E501
             self.logger.warning(msg)
-        return False
 
     def _should_upgrade(self) -> bool:
         upgrade = self._get_upgrade()
-        return self._should_upgrade_to_ws(upgrade)
+        if upgrade == b"websocket" and self._should_upgrade_to_ws():
+            return True
+        return False
```
Kludex:
```suggestion
        return upgrade == b"websocket" and self._should_upgrade_to_ws()
```

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -141,19 +141,23 @@ def _get_upgrade(self) -> bytes | None:
...
-            msg = "Unsupported upgrade request."
-            self.logger.warning(msg)
+    def _should_upgrade_to_ws(self) -> bool:
+        if self.ws_protocol_class is None:
+            return False
+        return True
+
+    def _unsupported_upgrade_warning(self) -> None:
+        msg = "Unsupported upgrade request."
+        self.logger.warning(msg)
```
Kludex:
```suggestion
        self.logger.warning("Unsupported upgrade request.")
```

Kludex:
I'll check this over the weekend. Sorry the delay.

vvanglro:
@Kludex Maybe there should be a timeout for the test action?
```diff
jobs:
  tests:
    name: "Python ${{ matrix.python-version }} ${{ matrix.os }}"
    runs-on: "${{ matrix.os }}"
+   timeout-minutes: 30
```

Kludex:
Yes. Some process is hanging. Do you want to create a PR for it?

vvanglro:
> Yes. Some process is hanging. Do you want to create a PR for it?

I'll do.

Kludex (approved):
Thanks @vvanglro 🙏

urec56:
In what release of uvicorn we can download this fix?

Kludex:
> In what release of uvicorn we can download this fix?

0.30.6

urec56:
Thank you

bashkirtsevich:
After upgrade to version `uvicorn==0.30.6` I have same issues as in `0.30.4` described [here](https://github.com/encode/uvicorn/pull/2375#issuecomment-2262789960).

After invoking `curl http://0.0.0.0:8000` server opens the connection and still silent infinite (no any responses).

PR #2367, opened 2024-06-13
fix: Avoid CancelledError from being propagated to lifespan's receive()

bellini666:
# Summary

Currently `LifeSpanOn.main()` is excepting `BaseException`, which also catches `CancelledError`.

`CancelledError` is used by `asyncio` to signal that a task has been cancelled and is not an error condition.

This PR changes the behavior to catch `CancelledError` separately and log it as a "Lifespan task cancelled" instead of letting the error propagate.

This is a traceback that I have seen some times on sentry for a project that I work on.

Screenshot 2024-06-13 at 19 08 19

# Checklist

### DIFF
--- a/uvicorn/lifespan/on.py
+++ b/uvicorn/lifespan/on.py
@@ -84,6 +84,11 @@ class LifespanOn:
                 "state": self.state,
             }
             await app(scope, self.receive, self.send)
+        except asyncio.CancelledError:
+            # Lifespan task was cancelled, likely due to the server shutting down.
+            # This is not an error that should be handled by BaseException above,
+            # so we just log it and exit.
+            self.logger.info("Lifespan task cancelled.")
         except BaseException as exc:
             self.asgi = None
             self.error_occured = True
--- a/tests/test_lifespan.py
+++ b/tests/test_lifespan.py
@@ -1,4 +1,5 @@
 import asyncio
+from unittest import mock
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


Kludex:
What problem are you trying to solve? Can you give me a real example?

uvicorn/lifespan/on.py
```diff
@@ -84,6 +84,11 @@ async def main(self) -> None:
                 "state": self.state,
             }
             await app(scope, self.receive, self.send)
+        except asyncio.CancelledError:
+            # Lifespan task was cancelled, likely due to the server shutting down.
+            # This is not an error that should be handled by BaseException above,
+            # so we just log it and exit.
+            self.logger.info("Lifespan task cancelled.")
```
maxfischer2781, line 91: [`CancelledError` should generally be re-raised](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError), in order to propagate cancellation up the call chain or to the to-be-cancelled task.
```suggestion
            self.logger.info("Lifespan task cancelled.")
            raise
```

bellini666:
Hey @Kludex,

So, it is not actually a real issue that would cause the application to misbehave or anything like that, but just a noisy error that will appear on your log and sentry (which is my case) but it is actually ok.

It is very rare (occurred 3 times in the past 30 days for me). But looking at the traceback (https://parade-ai.sentry.io/share/issue/4797523b6aa9458c927599ef0abf35a5/) and the logging below, it happened at the same second that uvicorn started, so my theory is that something canceled it during its startup process in a moment that the code was not expecting to be canceled.

Screenshot 2024-06-14 at 11 08 12

While replying here I tried to dig a bit deeper into the traceback and just noticed something that I completely missed yesterday:

Screenshot 2024-06-14 at 11 30 44

starlette is actually the one that is doing something "wrong" in here. For any `BaseException` it will send either a `"lifespan.shutdown.failed"` or a `"lifespan.startup.failed"`, which is usually what we want, but for `CancelledError` that is not the case, it is just that the loop is asking the task to stop and uses that exception as a way to allow the code to except it and do some cleanup work.

Maybe the proper fix would be doing something like this on starlette?

```python
        except asyncio.CancelledError:
            # CancelledError means this task was asked to terminate and it is not actually
            # an error. We want it to behave just like the `else` below, but reraise
            # it to make sure it gets propagated up to the call chain
            raise
        except BaseException:
            exc_text = traceback.format_exc()
            if started:
                await send({"type": "lifespan.shutdown.failed", "message": exc_text})
            else:
                await send({"type": "lifespan.startup.failed", "message": exc_text})
            raise
        else:
```

Not totally sure about the `await send({"type": "lifespan.shutdown.complete"})` (makes sense to me, but you know more), but not handling it as an error seems to me to be correct (but again, you know more =P)

btw. Nice to have meet you at pycon italia! :)

uvicorn/lifespan/on.py
bellini666: You are correct. Initially I went with the reraise, but the issue persisted, so I naively assumed it was already the top of the chain and that not propagating it would be fine.

I think the main reason is because the issue is not actually here but on starllete (I commented about in https://github.com/encode/uvicorn/pull/2367#issuecomment-2168196065)

bellini666:
@Kludex let me know if you agree with my conclusion in my previous comment, in which case I'm going to close this and propose a PR at starlette instead :)

sherpya:
got same problem and same solution, I'm using fastapi + crawlee with playwright, if I spawn the browser then ctrl+c
causes the error, it happens with uvloop and asyncio event loop

so what is the proposed solution?

PR #2371, opened 2024-06-25
feat: Improve process termination logic in multiprocess manager

abersheeran:
# Summary

About https://github.com/encode/uvicorn/discussions/2369

In our cluster, we accidentally discovered the zombie process. We found the reason. Uvicorn's new process manager will JOIN child processes one by one after sending all exit signals. When the previous child process does not exit for a long time, the subsequent child processes cannot be JOIN.

I noticed that Uvicorn has an inherent shutdown timeout, which would be nice if we could use it with a multiprocessor.

The reason why we don't use terminate&join sequentially is to kill all processes faster, as mentioned in this PR https://github.com/encode/uvicorn/pull/2010

# Checklist

### DIFF
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -92,7 +92,7 @@ class Process:
 
     def join(self, timeout: float | None = None) -> None:
         logger.info(f"Waiting for child process [{self.process.pid}]")
-        self.process.join(join_timeout)
+        self.process.join(timeout)
         # Timeout, kill the process
         while self.process.exitcode is None:
             self.process.kill()
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -90,9 +90,13 @@ class Process:
         # In Unix, the method will send SIGKILL to the process.
         self.process.kill()
 
-    def join(self) -> None:
+    def join(self, join_timeout: float | None = None) -> None:
         logger.info(f"Waiting for child process [{self.process.pid}]")
-        self.process.join()
+        self.process.join(join_timeout)
+        # Timeout, kill the process
+        while self.process.exitcode is None:
+            self.process.kill()
+            self.process.join(1)
 
     @property
     def pid(self) -> int | None:
@@ -131,7 +135,7 @@ class Multiprocess:
 
     def join_all(self) -> None:
         for process in self.processes:
-            process.join()
+            process.join(self.config.timeout_graceful_shutdown)
 
     def restart_all(self) -> None:
         for idx, process in enumerate(tuple(self.processes)):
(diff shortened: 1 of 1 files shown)


abersheeran:
I will add unit tests later, but I have no experience in designing a process that will hang, so if anyone can help, it would be greatly appreciated.

Kludex:
> I will add unit tests later, but I have no experience in designing a process that will hang, so if anyone can help, it would be greatly appreciated.

I'll try to check this over the weekend. Sorry the delay.

For next time, your PRs always have preference @abersheeran , please ping me if I take long.

Kludex:
I'm not sure if we should use the same `timeout_graceful_shutdown` 🤔

abersheeran:
Do you have any new ideas? We really need a configurable timeout here.

uvicorn/supervisors/multiprocess.py
```diff
@@ -90,9 +90,13 @@ def kill(self) -> None:
...
 
-    def join(self) -> None:
+    def join(self, join_timeout: float | None = None) -> None:
         logger.info(f"Waiting for child process [{self.process.pid}]")
-        self.process.join()
+        self.process.join(join_timeout)
+        # Timeout, kill the process
+        while self.process.exitcode is None:
+            self.process.kill()
+            self.process.join(1)
```
Kludex: Why do we have a `join(1)` here?

uvicorn/supervisors/multiprocess.py
```diff
@@ -90,9 +90,13 @@ def kill(self) -> None:
         # In Unix, the method will send SIGKILL to the process.
         self.process.kill()
 
-    def join(self) -> None:
+    def join(self, join_timeout: float | None = None) -> None:
```
Kludex: Can we call it only `timeout`?

```suggestion
    def join(self, timeout: float | None = None) -> None:
```

uvicorn/supervisors/multiprocess.py
abersheeran: Wait for the kill command to take effect. If it does not take effect within 1 second, send the kill command again.

uvicorn/supervisors/multiprocess.py
```diff
@@ -90,9 +90,13 @@ def kill(self) -> None:
...
         self.process.kill()
 
-    def join(self) -> None:
+    def join(self, timeout: float | None = None) -> None:
         logger.info(f"Waiting for child process [{self.process.pid}]")
-        self.process.join()
+        self.process.join(timeout)
+        # Timeout, kill the process
+        while self.process.exitcode is None:
+            self.process.kill()
```
abersheeran, line 98: The reason why CI failed is that this is not covered by the test. But I don't know how to design a process that will be 100% stuck.

Kludex: It's okay to add the pragma here.

Youngkingman:
Sorry to bother. The zombie process issue still exists in the mainstream versions of Uvicorn. Do you have any plans to fix this? We're rolling out services using Uvicorn at scale, and we've detected a large number of zombie processes in our k8s environment

PR #2373, opened 2024-06-28
Iterate subprocesses in-place

maxfischer2781:
# Summary

This PR fixes a subtle [iterate-remove bug](https://stackoverflow.com/questions/6260089/strange-result-when-removing-item-from-a-list-while-iterating-over-it-in-python) when checking subprocesses. Previously, subprocesses were deleted by index, which would shift the remaining processes and their index to go out of sync. This caused the wrong process to be deleted if more than one died in a cycle.  See #2372.

Instead of being deleted-appended, process are now replaced inplace to ensure positions and thus iteration is stable. This also makes copying the process list superfluous, so I removed this where appropriate.

# Checklist

### DIFF
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -134,7 +134,7 @@ class Multiprocess:
             process.join()
 
     def restart_all(self) -> None:
-        for idx, process in enumerate(tuple(self.processes)):
+        for idx, process in enumerate(self.processes):
             process.terminate()
             process.join()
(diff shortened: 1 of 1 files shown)


abersheeran (approved):

abersheeran:
This is definitely an oversight on my part.

Kludex:
If someone creates a release PR (changelog + version change), I can make a release.

vvanglro:
> If someone creates a release PR (changelog + version change), I can make a release.

Done. #2380

PR #2384, opened 2024-07-05
Suppress interrupt indication on CLI and programmatic usage

maxfischer2781:
# Summary

This PR suppresses `KeyboardInterrupt` on CLI and programmatic usage, restoring the non-`async` behaviour from before #1600.

# Checklist


Kludex:
This makes more sense to me.

Kludex (approved):

PR #2386, opened 2024-07-11
Use the non-deprecated worker in the documentation

vanschelven:


Kludex:
I want to remove Gunicorn from our docs. I'll pass on this.

Thanks anyway.

vanschelven:
> I want to remove Gunicorn from our docs.

Which is fair, but makes me wonder "is there anything I should know about the uvicorn/gunicorn setup?"

Kludex:
> Which is fair, but makes me wonder "is there anything I should know about the uvicorn/gunicorn setup?"

I mean, there's a need for restructuring the docs a bit, but I wouldn't say that you **need** to know something that is not already there.

Btw, you live in a beautiful city. 😎 👍

PR #2391, opened 2024-07-16
Allow passing a custom loop setup function instead of None

gnir-work:
# Summary
Support passing a custom IOLoop setup function in order to support running `uvicorn` in custom ioloop implementation other that the default and `uvloop`.

For example this feature will allow using [this monitored IOLoop](https://github.com/gnir-work/monitored-ioloop) inside `uvicorn` applications :)

Link to discussions thread - https://github.com/encode/uvicorn/discussions/2256

# Checklist

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -57,7 +57,11 @@ Options:
   --workers INTEGER               Number of worker processes. Defaults to the
                                   $WEB_CONCURRENCY environment variable if
                                   available, or 1. Not valid with --reload.
-  --loop [auto|asyncio|uvloop]    Event loop implementation.  [default: auto]
+  --loop [custom|auto|asyncio|uvloop]
+                                  Event loop implementation.  [default: auto]
+  --loop-setup TEXT               Import path to event loop setup function.,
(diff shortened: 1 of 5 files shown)


Kludex:
@gnir-work I don't want to have 2 similar parameters. I'd prefer us to go with this approach: https://github.com/encode/uvicorn/pull/2130/files (which turns the `--loop` into your `--loop-setup`.

If you want to move this forward, the best approach would be to continue that PR - fixing the pipeline, etc.

hxnir:
No problem :)
Just a quick question - How should I go about it? Should I fork the `graingert:uvicorn` repo and open another PR?
I can also update the current diff to be in the same spirit as the [PR you mentioned](https://github.com/encode/uvicorn/pull/2130/files)

Kludex:
I would add a remote on his repo, get the branch, and push to your fork, and then create a PR.

gnir-work:
Closing this PR in favor of https://github.com/encode/uvicorn/pull/2435.
Will update it from draft to open as soon as everything is passing :)

PR #2397, opened 2024-07-25
add timeout_worker_is_alive as an option for main.run.

socratescli:
# Summary

The option lets users set the timeout for judging if a process is dead in multiprocessing mode(i.e., workers > 1).
When the module calls uvicorn.run has some heavy load like importing a lot of modules, the is_alive check in supervisors.multiprocessing may be blocked. When this happens, the default 5 seconds timeout may cause the misjudgement of a process and kill it.
Issues like https://github.com/encode/uvicorn/discussions/2396 , https://stackoverflow.com/questions/78680384/during-uvicorn-startup-child-process-dies-in-kubernetes-cluster may be caused by the case.

# Checklist

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -280,6 +280,12 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     default=None,
     help="Maximum number of seconds to wait for graceful shutdown.",
 )
+@click.option(
+    "--timeout-worker-is-alive",
+    type=float,
+    default=5,
+    help="Maximum number of seconds to wait for judging if a worker process is alive.",
+)
 @click.option("--ssl-keyfile", type=str, default=None, help="SSL key file", show_default=True)
 @click.option(
     "--ssl-certfile",
@@ -394,6 +400,7 @@ def main(
     limit_max_requests: int,
     timeout_keep_alive: int,
     timeout_graceful_shutdown: int | None,
+    timeout_worker_is_alive: float,
     ssl_keyfile: str,
     ssl_certfile: str,
     ssl_keyfile_password: str,
@@ -443,6 +450,7 @@ def main(
         limit_max_requests=limit_max_requests,
         timeout_keep_alive=timeout_keep_alive,
         timeout_graceful_shutdown=timeout_graceful_shutdown,
+        timeout_worker_is_alive=timeout_worker_is_alive,
         ssl_keyfile=ssl_keyfile,
         ssl_certfile=ssl_certfile,
         ssl_keyfile_password=ssl_keyfile_password,
@@ -495,6 +503,7 @@ def run(
     limit_max_requests: int | None = None,
     timeout_keep_alive: int = 5,
     timeout_graceful_shutdown: int | None = None,
+    timeout_worker_is_alive: float = 5,
     ssl_keyfile: str | None = None,
     ssl_certfile: str | os.PathLike[str] | None = None,
     ssl_keyfile_password: str | None = None,
@@ -547,6 +556,7 @@ def run(
         limit_max_requests=limit_max_requests,
         timeout_keep_alive=timeout_keep_alive,
         timeout_graceful_shutdown=timeout_graceful_shutdown,
+        timeout_worker_is_alive=timeout_worker_is_alive,
         ssl_keyfile=ssl_keyfile,
         ssl_certfile=ssl_certfile,
         ssl_keyfile_password=ssl_keyfile_password,
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -112,6 +112,9 @@ Options:
   --timeout-graceful-shutdown INTEGER
                                   Maximum number of seconds to wait for
                                   graceful shutdown.
+  --timeout-worker-is-alive FLOAT
+                                  Maximum number of seconds to wait for
+                                  judging if a worker process is alive.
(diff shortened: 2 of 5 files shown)


abersheeran:
This PR does fix some edge cases. Please fix the CI failures so we can merge it.

socratescli:
> This PR does fix some edge cases. Please fix the CI failures so we can merge it.

checks passed, please go ahead

Kludex:
Defaulting to a low number doesn't seem right.

xconverge:
> Defaulting to a low number doesn't seem right.

It reads to me to be defaulting to the exact number that is already being used to preserve behavior I think?

https://github.com/encode/uvicorn/blob/master/uvicorn/supervisors/multiprocess.py#L65

alpha5611331:
Hi @abersheeran,
I urgently need this feature.
During my project development, in case of multiple workers, child processes are dying every 5 seconds.
I discovered that this PR resolves the issue.
Please merge this PR as soon as possible.
🙏🙏🙏

alpha5611331:
I've tried almost Uvicorn alternatives including Gunicorn, Hypercorn, Daphne, Sanic, Tornado, Waitress.
But I love Uvicorn most.
Unfortunately, worker alive timeout is hardcoded.
If only this is fixed, Uvicorn is almost PERFECT.
Please merge this PR ASAP.

abersheeran (approved):
Good job.

uvicorn/main.py
```diff
@@ -282,6 +282,12 @@ def print_version(ctx: click.Context, param: click.Parameter, value: bool) -> No
     default=None,
     help="Maximum number of seconds to wait for graceful shutdown.",
 )
+@click.option(
+    "--timeout-worker-is-alive",
+    type=float,
+    default=5,
```
Kludex, line 288: I'm not sure this is wanted as default value?

Kludex (changes requested):

uvicorn/main.py
Kludex: Ah, I've mentioned before: https://github.com/encode/uvicorn/pull/2397#issuecomment-2353586324

abersheeran: Yes, it's a good default value for most programs. Only those with heavy loads require larger values.

abersheeran: Additionally, the default timeout for a single request in httpx is also 5 seconds.

alpha5611331: Failed some checks. Can I fix it?

abersheeran: Yes, please fix it.

Kludex: There's an unrelated issue on the pipeline.

lightbeem3296: Thank you. I'll.
Excuse me, I'm not familiar with the forked repository contribution.
I think, I need to fork `socrateslee:feature/timeout_worker_is_alive`, but I'm not sure how to do it.
Could you please explain for me?

alpha5611331: Hi @Kludex , 
I think, it's not only current branch issue.
Your master branch doesn't pass `scripts/test` too.
Could you please checkout?
image

Kludex: > Yes, it's a good default value for most programs. Only those with heavy loads require larger values.

Well... Those with heavy loads will have to change this. Which is a bothersome for them.

mabrowning: Its the existing default. 5s is indeed too low, but there is not anyway to currently change it without this PR :)

I resorted to monkey-patching `uvicorn.supervisors.multiprocess.Process.is_alive` to set a different value.

Petr-Siegl: Yes, the 5s timeout is the current default and it is too low for heavy loads but there is no way to change this currently.

With this PR merged we can change it at least. It's fine to keep the defaults but we need some way to adjust this in needed cases.

These seem related to this:
https://github.com/encode/uvicorn/issues/2506
https://github.com/encode/uvicorn/discussions/2450

racinmat: @Kludex What default would you like to see? The 5s is the current behavior, so this default value which you correctly describe as low keeps the behavior the same, while allowing others to change it. 
Would it make more sense for you to change the default behavior from what is now in the released versions?

bharatmshrtr: Hello @mabrowning 
If i degrade uvicorn to 0.29.0 then will i be able to get rid of error where worker exits in between of activity or is there any other alternative to which we can fall upon. This is causing trouble for us as worker performing activity suddenly exits and process is stuck in between

UDPSOCKET:
Any status update on this? 

Thank you.

stijndehaes:
We are using Airflow 3 and are noticing issues because of the currently hardcoded 5s, sometimes Airflows fails to start up because of it.
This is currently mitigated by setting the workers to 1, but is not a good solution.

See: https://github.com/stackabletech/airflow-operator/issues/641

@Kludex can I ask you to have a look at this PR again, as the impact is quite high for Airflow users

lfrancke:
> See: [stackabletech/airflow-operator#641](https://github.com/stackabletech/airflow-operator/issues/641)

We (@stackabletech) sponsored @Kludex to look at this and get it merged. Thank you for following up on it and getting it merged so fast. @stijndehaes we'll include it in our next version.

PR #2405, opened 2024-08-01
📝 Add docs and attribution for creator and maintainer

tiangolo:
# Summary

📝 Add docs and attribution for creator and maintainer.

I think attribution is very well deserved and it would be nice to add a bit of visibility, to highlight the people behind Uvicorn. :bow: 

image

# Checklist


lovelydinosaur (approved):
@tiangolo I think this looks fantastic. ✨
Thanks so much for the suggestion. ☺️

Kludex:
I know we talked about recognition, and how I felt about the projects I maintain, but this didn't make me feel better.

I'll close this.

PR #2408, opened 2024-08-01
Fix 0.30.4 issue with connection close header

theyashl:
# Summary
When `connection: close` header is passed to uvicorn server, `RequestResponseCycle`'s `receive` method goes into waiting state. Resulting in request timeout or explicit connection close action from the client.
This PR will fix the issue.
Detailed RCA: https://github.com/encode/uvicorn/pull/2375#issuecomment-2263997261

# Checklist

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -1011,6 +1011,7 @@ async def test_return_close_header(http_protocol_cls: HTTPProtocol):
 
 async def test_close_connection_with_multiple_requests(http_protocol_cls: HTTPProtocol):
     app = Response("Hello, world", media_type="text/plain")
+
     protocol = get_connected_protocol(app, http_protocol_cls)
     protocol.data_received(REQUEST_AFTER_CONNECTION_CLOSE)
     await protocol.loop.run_one()
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -61,7 +61,7 @@ SIMPLE_POST_REQUEST = b"\r\n".join(
 
 CONNECTION_CLOSE_REQUEST = b"\r\n".join([b"GET / HTTP/1.1", b"Host: example.org", b"Connection: close", b"", b""])
 
-CONNECTION_CLOSE_POST_REQUEST_WITH_BODY = b"\r\n".join(
+CONNECTION_CLOSE_POST_REQUEST = b"\r\n".join(
     [
         b"POST / HTTP/1.1",
         b"Host: example.org",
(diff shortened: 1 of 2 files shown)


Kludex (changes requested):
The previous test was good, no need to modify it.

Your test is not representative, is it? Can we have a test with a POST request and `connection: close` header?

theyashl:
Hi @Kludex ,
Previous test case was relying on creating an app using test `Response` class. Which did not had `await receive()` statement.

While pre-existing `test_post_request` test case also have implemented `app` definition explicitly.

Kludex:
> While pre-existing test_post_request test case also have implemented app definition explicitly.

Yeah, but it's a POST request, and it's reading the body. But that test doesn't have the `connection: close`.

> Previous test case was relying on creating an app using test Response class. Which did not had await receive() statement.

It's a valid test anyway.

theyashl:
Ahh got your point. So I can just implement a POST test case with `Connection: close` header.

Should I include that test case too? While reverting the current one back to the original state.

Kludex:
> Ahh got your point. So I can just implement a POST test case with Connection: close header.

I mean, that's the case you want to really handle here, right? If yes, then the answer is yes.

> Should I include that test case too? While reverting the current one back to the original state.

Yes, please.

tests/protocols/test_http.py
```diff
@@ -61,6 +61,22 @@
...
+        b"Host: example.org",
+        b"Connection: close",
+        b"Content-Type: application/json",
+        b"Content-Length: 18",
+        b"",
+        b"{'hello': 'world'}",
+    ]
+)
+
+CONNECTION_CLOSE_POST_REQUEST_WITHOUT_BODY = b"\r\n".join(
```
Kludex: You can remove this one. The above case is fine.

tests/protocols/test_http.py
```diff
@@ -998,8 +1014,8 @@ async def test_return_close_header(http_protocol_cls: HTTPProtocol):
 
 
 async def test_close_connection_with_multiple_requests(http_protocol_cls: HTTPProtocol):
-    app = Response("Hello, world", media_type="text/plain")
-
+    response_content = b"Hello, world"
```
Kludex: Don't add a diff here please.

Kludex (changes requested):

tests/protocols/test_http.py
theyashl: Remove this case of POST without body. Kept POST with body and connection close header

tests/protocols/test_http.py
theyashl: removed diffs

Kludex (approved):

Konano:
I was wondering why the newly installed uvicorn wasn't responding. Turns out, the problem was here!

Although I managed to pinpoint the issue myself, it looks like it's already been fixed.

This bug has a significant impact, so I hope a new version can be released soon. Thanks!

Kludex:
This is fixed in 0.30.5.

theyashl:
Sorry for the problem caused due to the previous fix. :)

Kludex:
> Sorry for the problem caused due to the previous fix. :)

Nah, it's fine. :)

PR #2410, opened 2024-08-03
Added logging for startup failures.

mingregister:
# Summary

Added logging for startup failures.

We deploy our Uvicorn app in Kubernetes, and we send our logs to the Aliyun log service. However, we've noticed that when the startup fails (e.g: due to a forgotten module import), the error logs are not visible in the Aliyun log service. As a result:
  1. We are forced to log into the Kubernetes node to determine what happened.
  2. Furthermore, when the startup fails, the Kubernetes pod is restarted, leading to the loss of logs, which increases the difficulty of troubleshooting.

Therefore, I've added logging to capture errors when the startup fails.

# Checklist

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -65,7 +65,7 @@ class Server:
         try:
             self.config.setup_event_loop()
             return asyncio.run(self.serve(sockets=sockets))
-        except Exception as e:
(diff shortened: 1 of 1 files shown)


Kludex:
Please create a discussion, and present the current traceback.

PR #2420, opened 2024-08-09
fix: stop processing when requests reaches max request limit

vvanglro:
# Summary
https://github.com/encode/uvicorn/discussions/2412

# Checklist

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -890,23 +890,6 @@ async def test_invalid_http_request(
     assert b"Invalid HTTP request received." in protocol.transport.buffer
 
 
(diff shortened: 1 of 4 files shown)


vvanglro:
Strangely, running this newly added use case on its own succeeds every time, and occasionally fails when run with all the use cases.

vvanglro:
Ah, I understand. Since the total number of requests is accumulated after the response is completed, it seems that it is impossible to accurately limit the maximum number of requests at present.

```py
def on_response_complete(self) -> None:
    # Callback for pipelined HTTP requests to be started.
    self.server_state.total_requests += 1
```

PR #2424, opened 2024-08-13
differentiate %2f and / in url path

link89:
# Summary
differentiate  %2F and / when unquote path

https://github.com/encode/uvicorn/discussions/2423

# Checklist

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -113,7 +113,7 @@ FINISH_POST_REQUEST = b'{"hello": "world"}'
 
 HTTP10_GET_REQUEST = b"\r\n".join([b"GET / HTTP/1.0", b"Host: example.org", b"", b""])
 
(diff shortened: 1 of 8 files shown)


Kludex:
Why urllib doesn't have this logic then?

link89:
I guess that's because how to encode/decode path is not part of URI standard but a W3C recommendation. 

https://www.w3.org/Addressing/URL/4_URI_Recommentations.html

I try to find a method to allow me to ignore specific %xx, for example `unquote(path, ignore=('%2f', '%2F'))` but with no luck.

I have try another asgi engine `hypercorn` which do follow the URI recommentations and there are quite a lot of related issue about this.

Kludex:
> I have try another asgi engine hypercorn which do follow the URI recommentations and there are quite a lot of related issue about this.

Can you point me out to those issues?

link89:
Sure, how about discuss in https://github.com/encode/uvicorn/discussions/2423

Kludex:
Given the comment on https://github.com/encode/uvicorn/discussions/2423#discussioncomment-10323966, I'll be closing this PR.

PR #2426, opened 2024-08-13
ci: timeout for test suite runs to 30 minutes

vvanglro:
# Summary

# Checklist


Kludex (approved):

Kludex:
We do have a problem on our test suite, but this helps to not waste workers hours - even if they are free for open source projects, we all live in this world. 🙏

PR #2430, opened 2024-08-14
feat(server): log warning when max request limit is exceeded

vvanglro:
# Summary
ref: https://github.com/encode/uvicorn/discussions/2412

We can't exactly limit the number of requests https://github.com/encode/uvicorn/pull/2420, so at least we can have a friendly reminder log after meeting the limit.

# Checklist

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -1,18 +1,14 @@
 from __future__ import annotations
 
-import asyncio
 import logging
 import socket
 import threading
 import time
 from typing import TYPE_CHECKING, Any
 
-import httpx
 import pytest
 
 from tests.response import Response
-from tests.test_default_headers import app
-from tests.utils import run_server
 from uvicorn import Server
 from uvicorn._types import ASGIApplication, ASGIReceiveCallable, ASGISendCallable, Scope
 from uvicorn.config import WS_PROTOCOLS, Config
@@ -1096,19 +1092,6 @@ async def test_lifespan_state(http_protocol_cls: HTTPProtocol):
         assert b"Hi!" in protocol.transport.buffer
 
     assert not expected_states  # consumed
-
-
-async def test_request_than_limit_max_requests_warn_log(
-    unused_tcp_port: int, http_protocol_cls: HTTPProtocol, caplog: pytest.LogCaptureFixture
-):
-    caplog.set_level(logging.WARNING, logger="uvicorn.error")
-    config = Config(app=app, limit_max_requests=1, port=unused_tcp_port, http=http_protocol_cls)
-    async with run_server(config):
-        async with httpx.AsyncClient() as client:
-            tasks = [client.get(f"http://127.0.0.1:{unused_tcp_port}") for _ in range(3)]
-            responses = await asyncio.gather(*tasks)
-            assert len(responses) == 3
-    assert "Exceeded max request limit ending process." in caplog.text
 
 
 async def test_header_upgrade_is_not_websocket_depend_installed(
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -25,7 +25,7 @@ try:
     from uvicorn.protocols.http.httptools_impl import HttpToolsProtocol
 
     skip_if_no_httptools = pytest.mark.skipif(False, reason="httptools is installed")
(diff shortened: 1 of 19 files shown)


tests/protocols/test_http.py
```diff
@@ -1094,6 +1098,19 @@ async def app(scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable
...
+    unused_tcp_port: int, http_protocol_cls: HTTPProtocol, caplog: pytest.LogCaptureFixture
+):
+    caplog.set_level(logging.WARNING, logger="uvicorn.error")
+    config = Config(app=app, limit_max_requests=1, port=unused_tcp_port, http=http_protocol_cls)
+    async with run_server(config):
+        async with httpx.AsyncClient() as client:
+            tasks = [client.get(f"http://127.0.0.1:{unused_tcp_port}") for _ in range(3)]
+            responses = await asyncio.gather(*tasks)
+            assert len(responses) == 3
+    assert "Exceeded max request limit ending process." in caplog.text
```
Kludex: This doesn't feel like the right place for this test. We are testing the config/server behavior. Can you find a better place for it, please?

vvanglro: Yes, moved to config.

tests/test_config.py
```diff
@@ -545,3 +556,16 @@ def test_warn_when_using_reload_and_workers(caplog: pytest.LogCaptureFixture) ->
...
+    unused_tcp_port: int, http_protocol_cls: type[H11Protocol | HttpToolsProtocol], caplog: pytest.LogCaptureFixture
+):
+    caplog.set_level(logging.WARNING, logger="uvicorn.error")
+    config = Config(app=app, limit_max_requests=1, port=unused_tcp_port, http=http_protocol_cls)
+    async with run_server(config):
+        async with httpx.AsyncClient() as client:
+            tasks = [client.get(f"http://127.0.0.1:{unused_tcp_port}") for _ in range(2)]
+            responses = await asyncio.gather(*tasks)
+            assert len(responses) == 2
+    assert f"Maximum request limit of {config.limit_max_requests} exceeded. Terminating process." in caplog.text
```
Kludex: @vvanglro I don't want to be pedantic... But I want to follow our current structures. I don't think we want to add the `run_server` function on this test file.

Is there another way to test this? Or maybe we should move this to elsewhere?

vvanglro: Maybe it should be moved to the `test_server.py`, because that's what's in the test `server.py`.

If possible, I will commit.

tests/test_server.py
```diff
@@ -65,3 +78,16 @@ async def interrupt_running(srv: Server):
...
+    unused_tcp_port: int, http_protocol_cls: type[H11Protocol | HttpToolsProtocol], caplog: pytest.LogCaptureFixture
+):
+    caplog.set_level(logging.WARNING, logger="uvicorn.error")
+    config = Config(app=app, limit_max_requests=1, port=unused_tcp_port, http=http_protocol_cls)
+    async with run_server(config):
+        async with httpx.AsyncClient() as client:
+            tasks = [client.get(f"http://127.0.0.1:{unused_tcp_port}") for _ in range(2)]
+            responses = await asyncio.gather(*tasks)
+            assert len(responses) == 2
+    assert f"Maximum request limit of {config.limit_max_requests} exceeded. Terminating process." in caplog.text
```
Kludex:
```suggestion
    assert f"Maximum request limit of 1 exceeded. Terminating process." in caplog.text
```

tests/test_config.py
vvanglro: If it's just for test coverage, then it can be.
```py
async def test_request_than_limit_max_requests_warn_log(
    unused_tcp_port: int, http_protocol_cls: type[H11Protocol | HttpToolsProtocol], caplog: pytest.LogCaptureFixture
):
    caplog.set_level(logging.WARNING, logger="uvicorn.error")
    config = Config(app=app, limit_max_requests=0, port=unused_tcp_port, http=http_protocol_cls)
    await server.on_tick(1)
    assert f"Maximum request limit of {config.limit_max_requests} exceeded. Terminating process." in caplog.text
```

Kludex (approved):

PR #2431, opened 2024-08-14
test(supervisors): fix occasional assertion failures and hangs

vvanglro:
# Summary
It's not entirely clear why the assertion fails here. `assert not process.is_alive()` 

Make sure that test resources can be cleaned up anyway.

# Checklist


vvanglro:
Now there are two problems, one is that the process hangs when assertion fails(This PR solves), and the second is that it hangs on execution.

related:
    https://stackoverflow.com/questions/74114409/flask-hangs-when-run-as-a-subprocess-alongside-pytest
    https://github.com/pytest-dev/pytest/issues/10965
    
I create a Windows environment, but it's hard to reproduce.🙃

tests/supervisors/test_multiprocess.py
```diff
@@ -75,6 +76,7 @@ def test_multiprocess_run() -> None:
     supervisor.join_all()
 
 
+@pytest.mark.flaky(reruns=3, reruns_delay=2)
```
vvanglro: @Kludex For unstable test cases, how about add a retry?

Kludex: No. We should make them stable.

vvanglro: This feels a bit difficult, because the service is started with a thread. And the run method involves multiple waits.

https://github.com/encode/uvicorn/blob/master/uvicorn/supervisors/multiprocess.py#L151-L153

Switching execution between threads, as well as killed processes, all affect test stability and assertions.

vvanglro:
Ah, It seems to make sense now. `process.kill()` process does not end immediately, This is why failure is occasionally asserted here `assert not process.is_alive()`.  

Successful cases are all executed for more than 7s, that is, when the ping is executed, the process is really closed, and then the timeout returns false.

I add a timeout of 1 second to the assertion here, and the total execution time is more than 3 s, indicating that the ping method is taken here.
image

The same.

image

Kludex:
See? No need for retry. 😌👍

Thanks for your time investigating this. 🙏

vvanglro:
https://github.com/encode/uvicorn/blob/master/uvicorn/supervisors/multiprocess.py#L65-L69
Should the built-in is_alive() be preferred here?

that is:
```py
def is_alive(self, timeout: float = 5) -> bool:
    if self.process.is_alive():
        return True  # pragma: full coverage

    return self.ping(timeout)
```

tests/supervisors/test_multiprocess.py
```diff
@@ -83,15 +84,18 @@ def test_multiprocess_health_check() -> None:
     config = Config(app=app, workers=2)
     supervisor = Multiprocess(config, target=run, sockets=[])
     threading.Thread(target=supervisor.run, daemon=True).start()
-    time.sleep(1)
+    time.sleep(1)  # ensure server is up
```
Kludex:
```suggestion
    # Ensure server is up.
    time.sleep(1)
```

tests/supervisors/test_multiprocess.py
```diff
@@ -34,6 +34,7 @@ def new_function():
                 "-c",
                 f"from {module} import {name}; {name}.__wrapped__()",
             ],
+            timeout=10,
```
Kludex: Why do we need the timeout?

vvanglro: Oh, this was added earlier when I was investigating why it was hanging. No need for it now, I'll remove it later

Kludex (dismissed):

Kludex:
Thanks.

vvanglro:
We can't control it with precision, so we might as well retry.

Just make sure the dead process is eventually started successfully.

Kludex:
@abersheeran suggestions to solve this flaky test?

vvanglro:
Very rare, Run it many times before encountering. It looks like the restart of the process is delayed?

image

vvanglro:
Ah, does the unit test code have to be statistically covered too?

abersheeran:
> @abersheeran suggestions to solve this flaky test?

I think using while to check is not a good idea in unit tests. To be honest, I think it would be better to just kill all processes in the tree externally in case of weird scheduling errors.

vvanglro:
@Kludex I think the issue is resolved in the latest commit.

Kludex (approved):
Sorry the delay!

vvanglro:
🤦 This use case seems very unstable under Windows, and perhaps increasing the retry count is the correct way to suppress it.

abersheeran (approved):

vvanglro:
Using a retry decorator might be better, as I have observed that some test cases are also unstable, such as ws cases, subsequently, it can be considered to use this decorator for these unstable use cases.

vvanglro:
Looks like it's been resolved.

PR #2433, opened 2024-08-15
Clean-shutdown the server even in early shutdown case

kichanyurd:
# Summary

When running the following code (pytest):
```python
import asyncio
from random import random
from typing import AsyncIterator
from fastapi import FastAPI
from fastapi.responses import JSONResponse
import uvicorn


@asynccontextmanager
async def run_server(app: FastAPI) -> AsyncIterator[None]:
    config = uvicorn.Config(app=app, port=8089)
    yield
    server.should_exit = True
    await task


def rng_app() -> FastAPI:
    app = FastAPI()

    @app.get("/random_number")
    async def get_random_number(max: int) -> JSONResponse:
        return JSONResponse({"number": int(random() * max)})

    return app


async def test_this() -> None:
    async with run_server(rng_app()):
        pass
```

I'm getting the following error:
```
task: <Task pending name='Task-3' coro=<LifespanOn.main() running at /workspaces/emcie/server/.venv/lib/python3.12/site-packages/uvicorn/lifespan/on.py:86> wait_for=<Future pending cb=[Task.task_wakeup()]>>
  File "/home/vscode/.pyenv/versions/3.12.4/lib/python3.12/asyncio/queues.py", line 158, in get
    await getter
GeneratorExit

  File "/workspaces/emcie/server/.venv/lib/python3.12/site-packages/starlette/routing.py", line 743, in lifespan
    await receive()
  File "/workspaces/emcie/server/.venv/lib/python3.12/site-packages/uvicorn/lifespan/on.py", line 137, in receive
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vscode/.pyenv/versions/3.12.4/lib/python3.12/asyncio/queues.py", line 160, in get
    getter.cancel()  # Just in case getter is not done yet.
    ^^^^^^^^^^^^^^^
  File "/home/vscode/.pyenv/versions/3.12.4/lib/python3.12/asyncio/base_events.py", line 795, in call_soon
    self._check_closed()
  File "/home/vscode/.pyenv/versions/3.12.4/lib/python3.12/asyncio/base_events.py", line 541, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

This PR fixes the issue by making sure that a graceful shutdown (including that of the lifespan class) is always awaited before returning from the `serve()` function.

Please note that I didn't add a test as this meant incorporating Starlette/FastAPI in the tests as a (heavy) dependency.
# Checklist

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -82,10 +82,13 @@ class Server:
         logger.info(message, process_id, extra={"color_message": color_message})
 
         await self.startup(sockets=sockets)
-        if self.should_exit:
-            return
-        await self.main_loop()
-        await self.shutdown(sockets=sockets)
+
+        try:
+            if self.should_exit:
+                return
+            await self.main_loop()
+        finally:
+            await self.shutdown(sockets=sockets)
 
         message = "Finished server process [%d]"
         color_message = "Finished server process [" + click.style("%d", fg="cyan") + "]"


Kludex:
How do you have so many thumbs up here?

kichanyurd:
@Kludex It's currently affecting our test suite so some people helped call attention to this :)

Kludex:
> @Kludex It's currently affecting our test suite so some people helped call attention to this :)

I see. Maybe instead of wasting your coworkers time, you can sponsor me. :)

Also, why do you run uvicorn on your test suite?

kichanyurd:
@Kludex I'm already sponsoring them anyway :) But happy to make a contribution!

I'm running it as part of an E2E test suite for our client/server setup. We have a complex server CLI that internally runs uvicorn by importing the library, and in these tests we automatically verify the behavior between the server CLI and the client CLI.

kichanyurd:
Hey @Kludex, I reckon you're quite busy over there. Is there anything I can do to facilitate the process in some way?

Kludex:
Well, it looks like you just don't have an event loop fixture? https://github.com/fastapi/fastapi/issues/5692#issuecomment-1328105554

Please create a discussion. This is not a bug.

kichanyurd:
> Well, it looks like you just don't have an event loop fixture? [fastapi/fastapi#5692 (comment)](https://github.com/fastapi/fastapi/issues/5692#issuecomment-1328105554)
> 
> Please create a discussion. This is not a bug.

@Kludex we actually have an event loop running, and all of our tests are async tests which are using it. The uvicorn server does run and function correctly, aside from the shutdown issue, which it seems is due to not awaiting on all started tasks before the process is finished, leading to asyncio errors at the end of the process's lifetime.

Also attaching a link here to the discussion I previously opened on the topic: https://github.com/encode/uvicorn/discussions/2434

PR #2439, opened 2024-08-20
Add anyio to requirements.txt

mtelka:
# Summary

The `anyio` package is directly needed to run tests so it should be explicitly listed in the `requirements.txt` file.
See also #2436.

# Checklist


Kludex:
AnyIO is a dependency of 3 packages here, no need to explicit add all transitive dependencies we have...

PR #2443, opened 2024-08-20
refactor(config): reduce config load and increase priority

vvanglro:
# Summary
#2440 

# Checklist


vvanglro:
All right, 😑  that doesn't work because some objects in fastapi can't be serialized, so they have to be loaded after the process starts.  https://github.com/fastapi/fastapi/discussions/9078

Although the tests passed, they couldn't really start up.

image

Kludex:
It's not only on FastAPI.

PR #2444, opened 2024-08-20
feat(config): ensure that the app loads correctly before the service starts

vvanglro:
# Summary

https://github.com/encode/uvicorn/discussions/2440
https://github.com/encode/uvicorn/pull/2443

Extract the parts that will cause the process to exit and perform a check import before starting the service.

# Checklist


Kludex:
Closed by https://github.com/Kludex/uvicorn/pull/2919

Sorry the delay here.

PR #2472, opened 2024-09-29
add socket-load-balance flag

graingert:
# Summary

Fixes https://github.com/encode/uvicorn/discussions/2467

# Checklist

### DIFF
--- a/tests/supervisors/test_multiprocess.py
+++ b/tests/supervisors/test_multiprocess.py
@@ -205,7 +205,6 @@ async def lb_app(
 @pytest.mark.skipif(
     not ((sys.platform == "linux" and hasattr(socket, "SO_REUSEPORT")) or hasattr(socket, "SO_REUSEPORT_LB")),
     reason="unsupported",
-    strict=True,
     raises=RuntimeError,
 )
 def test_multiprocess_socket_balance() -> None:
--- a/tests/supervisors/test_multiprocess.py
+++ b/tests/supervisors/test_multiprocess.py
@@ -1,17 +1,21 @@
 from __future__ import annotations
 
 import functools
+import multiprocessing
 import os
 import signal
 import socket
+import sys
 import threading
 import time
-from typing import Any, Callable
+from typing import Any, Callable, TypeVar
 
+import httpx
 import pytest
 
 from uvicorn import Config
 from uvicorn._types import ASGIReceiveCallable, ASGISendCallable, Scope
+from uvicorn.server import Server
 from uvicorn.supervisors import Multiprocess
 from uvicorn.supervisors.multiprocess import Process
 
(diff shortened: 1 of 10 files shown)


uvicorn/_subprocess.py
```diff
@@ -9,19 +9,57 @@
...
 spawn = multiprocessing.get_context("spawn")
 
+class SocketSharePickle:
+    def __init__(self, sock: socket.socket):
+        self._sock =  sock
+
+    def get(self) -> socket.socket:
+        return self._sock
+
+if (sys.platform == "linux" and hasattr(socket, "SO_REUSEPORT")) or hasattr(socket, "SO_REUSEPORT_LB"):
```
graingert: Needs https://github.com/python/cpython/issues/124761

Kludex (commented):
Why do we need to add a new argument i.e. `socket-load-balance`? Can't this behavior be the default? I'd assume people already thinks that's the default.

graingert:
There's a tradeoff where the default unfair load balancing favours lower latency and the SO_REUSEPORT(_LB) approach favours an even load between processes 

I suspect most people do want the fair load balancing, however I think it should bed in a bit before we suddenly turn it on by default for everyone.

tests/supervisors/test_multiprocess.py
```diff
@@ -169,3 +174,71 @@ def test_multiprocess_sigttou() -> None:
...
+@pytest.mark.skipif(
+    not ((sys.platform == "linux" and hasattr(socket, "SO_REUSEPORT")) or hasattr(socket, "SO_REUSEPORT_LB")),
+    reason="unsupported",
+)
+def test_multiprocess_socket_balance() -> None:  # pragma: py-darwin pragma: py-win32
+    with multiprocessing.Manager() as m:
+        started = m.Event()
+        d = m.dict()
+        app = functools.partial(lb_app, d, started)
+        config = Config(app=app, workers=2, socket_load_balance=True, port=0, interface="asgi3")
```
graingert, line 219: this needs a test for two independent calls to Multiprocess(...).run() running on the same port:

```python
def test_bind_to_used_port():
    port = ephemeral_port_reserve.reserve() 
    config = Config(app=app, workers=2, socket_load_balance=True, port=port, interface="asgi3")
    with multiprocessing.get_context("spawn").Pool(max_workers=1) as pool:
        try:
            f1 = pool.apply_async(uvicorn.main, ...)
            # wait for started
            with pytest.raises(OSError, match="already bound"):
                uvicorn.main(...)
        finally: pool.terminate()
```

uvicorn/_subprocess.py
```diff
@@ -18,10 +18,42 @@
...
+    def __init__(self, sock: socket.socket):
+        self._sock = sock
+
+    def get(self) -> socket.socket:
+        return self._sock
+
+
+class SocketShareRebind:
+    def __init__(self, sock: socket.socket):
+        if not (sys.platform == "linux" and hasattr(socket, "SO_REUSEPORT")) or hasattr(socket, "SO_REUSEPORT_LB"):
```
vvanglro, line 31: Wouldn't it be better to put the logic here in `config.bind_socket`?
Just add it to the else, uds and fd don't seem to support SO_REUSEPORT.

graingert: hmm I'm not sure about this, I want to make sure the SO_REUSEPORT flag isn't set before the socket is bound so we still fail if another uvicorn is using this port

vvanglro: How about this?
https://github.com/vvanglro/uvicorn/commit/f2b103712c4fde78fa6f49c9e9f552ca9f30aa18

papa99do:
@graingert , do you have any updates on this work? It would be great to support socket LB for a simpler deployment that can leverage multiple CPU cores more effectively without a dependency on another process manager. Is there any blockers? How can I help? Cheers.

graingert:
It's feature complete, it just need a decision on how coverage for subprocesses will be managed, eg pytest-cov?

wb7777:
When we use `uds` and `socket_load_balance == True` on linux we get error:
```python
 File "/python3.13/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/python3.13/site-packages/uvicorn/_subprocess.py", line 117, in subprocess_started
    target(sockets=[s.get() for s in sockets])
                    ~~~~~^^
  File "/python3.13/site-packages/uvicorn/_subprocess.py", line 46, in get
    sock.bind(self._sockname)
    ~~~~~~~~~^^^^^^^^^^^^^^^^
OSError: [Errno 98] Address already in use
```

methane:
The method of binding a new socket in the child process and setting SO_REUSEPORT does not work well for graceful restarts. Each worker process has its own backlog, so the new and old processes have different backlogs, which can lead to errors being returned for new connections that are still in the backlog.

To achieve a complete graceful restart, it is necessary to create a listen socket with SO_REUSEPORT set in the parent process for the number of workers and pass it to the child processes. By passing that listen socket to the new process during a graceful restart, the old and new processes can share the backlog, allowing the new worker to handle unaccepted connections.

I recommend using a method where each worker listens on a different port rather than using SO_REUSEPORT. For example, if you are using nginx as a reverse proxy, when using SO_REUSEPORT, nginx will not know which connection to send requests to for proper load balancing. Nginx will simply reuse the last connection that became keepalive in a LIFO manner.
The implementation method is almost the same as SO_REUSEPORT when listening on multiple ports. The master process creates listen sockets for the number of workers and passes them to the worker processes. The only difference is whether to bind the same address and set SO_REUSEPORT or to bind a different address.

PR #2473, opened 2024-09-30
Create setting/flag to enable/disable X-Forwarded-Port

victor-torres:
# Summary

Create a new setting/flag to enable/disable X-Forwarded-Port to populate remote address info.

This flag is disabled by default. Command line options are: 
- `--forwarded-port`
- `--no-forwarded-port`

# Checklist


victor-torres:
This pull request is addressing the issue described in #1974

victor-torres:
There's a failing test case in `tests/supervisors/test_multiprocess.py` that I don't think is related to this PR. 
It fails on Python 3.9 running on macOS.

I think this PR is ready for review. Let me know if I can help with anything else or if there's anything that could be improved here :D

Kludex:
Do we need to add a new flag?

victor-torres:
I don't mind considering X-Forwarded-Port (if available) by default if already checking for X-Forwarded-For. 

My decision to include a new flag was to avoid changing the default behaviour, something that some maintainers and some users usually ask for. 

But I can totally remove the flag and make it be considered by default +1 

Just let me know.

Kludex:
Let's take a step back here, and continue the discussion on https://github.com/encode/uvicorn/issues/1974#issuecomment-2401602267, please.

Kludex:
I'll be closing this until we have a decision on the issue.

I don't think we want this to be a parameter anyway.

PR #2479, opened 2024-10-01
docs: fix configuration option for trusting forwarded IPs

gunter1020:
# Summary

fix configuration option for trusting forwarded IPs

# Checklist


Kludex (approved):
Thanks! 🙏

Issue #2477, opened 2024-09-30
--forwarded-allow-ips '*' broken in combination with gunicorn


Kludex:
### Discussed in https://github.com/encode/uvicorn/discussions/2475

<div type='discussions-op-text'>

<sup>Originally posted by **andreas-sch-b** September 30, 2024</sup>
I'm running uvicorn in combination with gunicorn like this:

`gunicorn -k uvicorn.workers.UvicornWorker --forwarded-allow-ips "*" ....`

This worked well with the 0.30.6, however it's broken with the 0.31.0 of uvicorn. The issue got introduced with the PR [1]. The problem is, that the check for the wildcard changed. In the old version there was a check like `"*" in trusted_hosts`. The new code now checks for `trusted_hosts == "*"` which causes the problem.

[1] https://github.com/encode/uvicorn/pull/2468</div>

fsecada01:
Confirmed. This happened to me.

theyashl:
Turns out gunicorn is parsing `forwarded_allow_ips` command line option and then putting its values into a list before handing it over to uvicorn's worker as part of configs. Refer to `ForwarderHeaders` class from gunicorn's `gunicorn/config.py` file. This class is handling cli `--forwarder-headers`, with a validator called `validate_string_to_list`.

This validator puts all the comma-separated values into a list which are received with the given parameter. i.e. `--forwarder-headers`.

So basically the check `_TrustedHosts` should have is `trusted_hosts == ["*"]` instead of `trusted_hosts == "*"`. But this would be completely `gunicorn` specific change. So we would have to make sure that this checks should work fine with both uvicorn as a standalone application and with gunicorn integration. We can make it as one of the following: `trusted_hosts == ["*"] or trusted_hosts == "*"` / `trusted_hosts in (["*"], "*")` / `"*" in trusted_hosts`.

Kludex:
- Closed on https://github.com/encode/uvicorn/pull/2480

PR #2480, opened 2024-10-02
Add support for `[*]` in trusted hosts

theyashl:
# Summary

Fixed trusted host parsing for always trust case.
This PR will fix allowing all networks upon passing `--forwarded-allow-ips` as `*` from gunicorn with uvicorn as a worker class.

The fix ensures that gunicorn integration and uvicorn as a standalone application will work fine for storing and validating trusted networks.

Complete RCA is present on issue: #2477 

# Checklist

### DIFF
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -65,7 +65,7 @@ _TRUSTED_IPv6_ADDRESSES = [
     "::11.22.33.44",  # This is a dual address
 ]
 _TRUSTED_IPv6_NETWORKS = "2001:db8:abcd:0012::0/64"
-_TRUSTED_LITERALS = "some-literal , unix:///foo/bar  ,  /foo/bar"
+_TRUSTED_LITERALS = "some-literal , unix:///foo/bar  ,  /foo/bar, garba*gewith*"
 
 
 @pytest.mark.parametrize(
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -56,6 +56,7 @@ def make_httpx_client(
 # of the _TrustedHosts.__init__ method.
 _TRUSTED_NOTHING: list[str] = []
 _TRUSTED_EVERYTHING = "*"
+_TRUSTED_EVERYTHING_LIST = ["*"]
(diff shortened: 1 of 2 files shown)


theyashl:
@Kludex requesting your review and inputs for this PR :)

uvicorn/middleware/proxy_headers.py
```diff
@@ -68,7 +68,8 @@ class _TrustedHosts:
     """Container for trusted hosts and networks"""
 
     def __init__(self, trusted_hosts: list[str] | str) -> None:
-        self.always_trust: bool = trusted_hosts == "*"
+        # gunicorn sends trusted hosts as a parsed list of strings while uvicorn forwards as is provided by input
```
Kludex: I don't want references to `gunicorn` anymore.
```suggestion
```
This change makes sense on the uvicorn context as well.

tests/middleware/test_proxy_headers.py
```diff
@@ -65,7 +65,7 @@ def make_httpx_client(
     "::11.22.33.44",  # This is a dual address
 ]
 _TRUSTED_IPv6_NETWORKS = "2001:db8:abcd:0012::0/64"
-_TRUSTED_LITERALS = "some-literal , unix:///foo/bar  ,  /foo/bar"
+_TRUSTED_LITERALS = "some-literal , unix:///foo/bar  ,  /foo/bar, garba*gewith*"
```
Kludex, line 69: We probably have an assertive test case with `*`, can you add `[*]` as well?

Kludex:
Please change the title as well.

tests/middleware/test_proxy_headers.py
theyashl: Added test case with trust host using `["*"]`, It's similar to the existing trust everything so I did not add every address as an input.

uvicorn/middleware/proxy_headers.py
theyashl: Got that. But just to clarify on uvicorn context, uvicorn does not send trusted hosts as `["*"]` to `_TrustedHosts` in both cases as a Command line program or as an explicit application parameter.

So that's why I thought that gunicorn is the only one who is sending that value after passing it in a lits.

Kludex (approved):
Thanks. :)

PR #2484, opened 2024-10-14
Enhance multiprocess manager's log data

przlada:
# Summary

This PR proposes to extend the multiprocess manager's "Child process died" log with additional information, specifically the exit code and the name of the signal that terminated the process, similarly as `gunicorn` is reporting. This change will provide more context for troubleshooting issues related to unexpected process termination, particularly in scenarios such as processes being terminated due to "out of memory" errors.

The additional information will be added as `extra` values in the log.

# Checklist

### DIFF
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -98,6 +98,10 @@ class Process:
     def pid(self) -> int | None:
         return self.process.pid
 
+    @property
+    def exitcode(self) -> int | None:
+        return self.process.exitcode
+
 
 class Multiprocess:
     def __init__(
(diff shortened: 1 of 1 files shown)


uvicorn/supervisors/multiprocess.py
```diff
@@ -167,13 +171,20 @@ def keep_subprocess_alive(self) -> None:
...
             if self.should_exit.is_set():
                 return  # pragma: full coverage
 
-            logger.info(f"Child process [{process.pid}] died")
+            signal_name = None
+            if exitcode is not None and exitcode < 0:
+                try:
+                    signal_name = signal.Signals(-exitcode).name
+                except ValueError:  # pragma: no cover
+                    pass
```
Kludex, line 186: When do we have this?

Kludex (changes requested):
Would you mind adding a test for it?

Issue #571, opened 2020-02-10
Support Binding to Multiple Ports


bartboy011:
Hello there! I'd like to use Uvicorn for an application I'm working on, however, my use-case requires the ability to bind the application to multiple ports. Looking through the source code, this doesn't seem possible, but I'm curious if a workaround has been discussed before, and/or if this is on the roadmap?

lovelydinosaur:
Not currently planned. Out of interest, why is binding to multiple ports necessary/useful here?

lovelydinosaur:
Actually now I think about it, that’s not quite true, you can do it using Gunicorn with the Uvicorn worker.

bartboy011:
@tomchristie thank you! We'll look into that.

The use case is that the app will be deployed in a kubernetes cluster - the main port uses a TLS connection, but the kubelet doesn't have access to the TLS certs so it can't connect to the primary port to conduct a health check. So, effectively the use-case for a second port binding is to allow for one port to be for web traffic while the other is used for health checks by Kubernetes.

jeet-parekh:
@tomchristie, adding my use case here.

I was trying to run plotly dash, which is tightly integrated with flask. I wanted to add SSL and HTTPS redirect without adding nginx to the stack.

This is what I can do with hypercorn.

```python
from hypercorn.middleware import AsyncioWSGIMiddleware, HTTPToHTTPSRedirectMiddleware
app_server = HTTPToHTTPSRedirectMiddleware(AsyncioWSGIMiddleware(dash_app.server), "127.0.0.1:443")
```

And to run it,

```
hypercorn --certfile mycertfile --keyfile mykeyfile --bind 0.0.0.0:443 --insecure-bind 0.0.0.0:80 main:app_server
```

This was my similar attempt with fastapi+uvicorn

```python
from fastapi import FastAPI
from fastapi.middleware.httpsredirect import HTTPSRedirectMiddleware
from fastapi.middleware.wsgi import WSGIMiddleware

app_server = FastAPI()
app_server.add_middleware(HTTPSRedirectMiddleware)
app_server.mount("/", WSGIMiddleware(dash_app.server))
```

And to run it,

```
uvicorn --ssl-keyfile ../throughput.ai.key --ssl-certfile ../STAR_throughput_ai.crt --port 443 main:app_server
```

But now it won't accept connections on port 80.

[...]

PR #2486, opened 2024-10-17
feat: supports setting multiple hosts

vvanglro:
# Summary
On the realization of this discussion
https://github.com/encode/uvicorn/discussions/1529

# Checklist


docs/deployment.md
```diff
@@ -31,8 +31,7 @@ $ uvicorn --help
 Usage: uvicorn [OPTIONS] APP
 
 Options:
-  --host TEXT                     Bind socket to this host.  [default:
-                                  127.0.0.1]
+  --host TEXT                     Bind socket to this host.
```
vvanglro: Can't recognize cli multiple types of default values?

vvanglro:
Not sure why that test case is failing, doesn't seem to have anything to do with this PR. The local tests have always passed.

Both PRs failed. #2444 #2431  

The last few commits haven't involved changes to failed use cases either, so why is that?🤔

docs/deployment.md
vvanglro: I missed `show_default`.

vvanglro:
Well, the investigation revealed that httptools changed its behavior in 0.6.3.
https://github.com/MagicStack/httptools/commit/560bd9eabf121de6eb2c7476e18becb02dcb8867
#2375 
```sh
pip install --upgrade httptools==0.6.3
```
```py
import httptools

    [
        b"GET / HTTP/1.1",
        b"Host: example.org",
        b"Connection: close",
        b"",
        b"",
        b"GET / HTTP/1.1",
        b"Host: example.org",
        b"",
        b"",
    ]
)

parser = httptools.HttpRequestParser("a")
# Adding this line works fine.
# parser.set_dangerous_leniencies(lenient_data_after_close=True)
print(parser.feed_data(REQUEST_AFTER_CONNECTION_CLOSE))
```
output:
```sh
print(parser.feed_data(REQUEST_AFTER_CONNECTION_CLOSE))
  File "httptools/parser/parser.pyx", line 259, in httptools.parser.parser.HttpParser.feed_data
httptools.parser.errors.HttpParserError: Data after `Connection: close`
```

Kludex:
Is it a problem on uvicorn or on httptools?

vvanglro:
> Is it a problem on uvicorn or on httptools?

```py
    [
        b"GET / HTTP/1.1",
        b"Host: example.org",
        b"Connection: close",
        b"",
        b"",
        b"GET / HTTP/1.1",
        b"Host: example.org",
        b"",
        b"",
    ]
)
```

If such a message exists in a real environment, it's a problem with uvicorn.

Kludex:
I don't understand. Is it because of a security issue that was fixed?

vvanglro:
https://github.com/MagicStack/httptools/commit/560bd9eabf121de6eb2c7476e18becb02dcb8867

Here are the details of this commit:
```docs
  Expose leniency flags via the new `set_dangerous_leniencies` parser
  method if somebody needs to opt into the old vulnerable behavior.
  ```
  
  It looks like the error was passed to the caller, maybe it hadn't been before.
  
  I'm not a cython expert.

**Comment by vvanglro** [comment]
Oh, I see, now using a more standardized and strict parsing method. That's why the above message, when processed, reported an error. After the connection is closed, no messages should be sent after it.

**Comment by Kludex** [comment]
> Oh, I see, now using a more standardized and strict parsing method. That's why the above message, when processed, reported an error. After the connection is closed, no messages should be sent after it.

So uvicorn should not send anything to the parser after, right?

**Comment by vvanglro** [comment]
> So uvicorn should not send anything to the parser after, right?

yes.

PyCharm 2024-10-17 22 16 17

```diff

class HttpToolsProtocol(asyncio.Protocol):
    def __init__(
        self,
        config: Config,
        server_state: ServerState,
        app_state: dict[str, Any],
        _loop: asyncio.AbstractEventLoop | None = None,
    ) -> None:
        if not config.loaded:
            config.load()

        self.config = config
        self.app = config.loaded_app
        self.loop = _loop or asyncio.get_event_loop()
        self.logger = logging.getLogger("uvicorn.error")
        self.access_logger = logging.getLogger("uvicorn.access")
        self.access_log = self.access_logger.hasHandlers()
        self.parser = httptools.HttpRequestParser(self)
++      self.parser.set_dangerous_leniencies(lenient_data_after_close=True)

```

It seems we can only add one line of code to handle this.
After adding this line of code, the response before the connection is closed can be processed normally. 
Messages after the connection is closed will not be parsed, That is, the headers value in the figure, only resolved to the pre-close.

vvanglro:
#2238 

This is wrong behavior, right?
image

that's right?
image

Kludex:
I don't know, I'd have to read some resources before answering that.

vvanglro:
We'll take the next step when we figure it out.

Kludex:
@vvanglro What do you feel we need to do to support both this PR and this issue https://github.com/Kludex/uvicorn/issues/571?

Sorry the delay here.

vvanglro:
> @vvanglro What do you feel we need to do to support both this PR and this issue #571?

This problem requires multiple ports, and this PR was originally designed with multiple hosts, so if we want to solve these two problems, then I suggest adding `-bind` to gradually replace the current `host` `port`mode.

Similar to https://github.com/pgjones/hypercorn/blob/main/src/hypercorn/__main__.py#L45, But our bind only supports host:port, and uds and fd are still separate. Avoid the burden of complicated analysis.

Kludex:
Let's go with https://github.com/Kludex/uvicorn/pull/2818. Thanks for this and sorry the delay.

PR #2503, opened 2024-11-05
Make logger names configurable

raf-moreau:
I'm using uvicorn to start, programmaticly, multiple servers.
But logs were always uvicorn.error or uvicorn.access.
I need to make a difference regarding the source server emitting the log.

I added an optional parameter in Config constructor to allow overriding the loggers names.
And i patched the code to use the logger coming from Config instead of using global objects.

So, as an example, adding this parameter
logger_mappings = { "general": "mylogger.general", "access": "mylogger.access", "asgi": "mylogger.asgi", } 
to the Config constructor shall configure the logging in the correct loggers.

I did hack a bit the Config object to maintain maximum compatibility with existing code.

PS: If you find my code is ugly, i agree and i'm willing to improve it.
PS2: Forgive my english

### DIFF
--- a/tests/middleware/test_logging.py
+++ b/tests/middleware/test_logging.py
@@ -29,6 +29,11 @@ if typing.TYPE_CHECKING:
     WSProtocol: TypeAlias = "type[WebSocketProtocol | _WSProtocol]"
 
 pytestmark = pytest.mark.anyio
+my_logger_mappings = {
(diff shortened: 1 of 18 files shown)


Kludex:
Can you please create a discussion?

We do have an issue regarding logger names, but I don't think that's what you expect to change? https://github.com/encode/uvicorn/issues/562

PR #2508, opened 2024-11-11
trust_number_of_proxies to ProxyHeadersMiddleware

mzealey:
# Summary

The motivation for this is if you know you are behind eg loadbalancer + reverse-proxy nginx, but their IPs are not sensibly known (eg they change), you can configure the middleware to just skip the known number of them before applying any other heuristics. This should be backwards-compatible with previous versions of the middleware. This makes it easier to port code from https://werkzeug.palletsprojects.com/en/stable/middleware/proxy_fix/

Adds a new setting of `trusted_number_of_proxies` to the
`ProxyHeadersMiddleware` that allows the user to specify the number of
proxies that are trusted to be present in the request and not otherwise
verified.

# Checklist

### DIFF
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -39,15 +39,17 @@ async def default_app(scope: Scope, receive: ASGIReceiveCallable, send: ASGISend
 def make_httpx_client(
     trusted_hosts: str | list[str],
     client: tuple[str, int] = ("127.0.0.1", 123),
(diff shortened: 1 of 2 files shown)


mzealey:
(the tests currently seem to be failing due to the tests on master branch being broken?)

Kludex:
Please create a discussion first.

mzealey:
Opened in https://github.com/encode/uvicorn/discussions/2519

PR #2516, opened 2024-11-20
Update `publish.yaml` with latest PyPI recommendations

Kludex:


.github/workflows/publish.yml
```diff
@@ -6,24 +6,97 @@ on:
...
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.inspect_package.outputs.version }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.8"
```
Kludex:
```suggestion
      - uses: actions/setup-python@v5
        with:
          python-version: "3.12"
```

PR #2528, opened 2024-12-04
Fixes import of uvicorn server. (encode/uvicorn#2527)

tback:
# Summary

This fixes a wrong import as discussed in #2527. Impact should be minimal.

# Checklist


Kludex (approved):

Kludex:
Thanks.

PR #2529, opened 2024-12-05
Unique Worker ID

whilenot-dev:
# Summary

Following [this discussion](https://github.com/encode/uvicorn/discussions/2459), this is a draft to inject a unique worker ID into the app state of an application. Feedback welcome!

The worker ID should be consistent across restarts and provide the missing building block for idempotent  startup- and shutdown routines in the lifespan of a worker.

# Checklist


whilenot-dev:
Before I'll tackle the test coverage I'd like to have some feedback first:
1. Should I limit the worker ID injection to multiprocess-mode only?
1. Should I document the injection more like a warning? Can the injection be considered a bad sursprise?
1. How about the simplicity, @Kludex? :D

PR #2533, opened 2024-12-13
Update run in main.py to handle case when port is not int

shashstormer:
# Summary

when port is passed as string it will try converting to integer

Then it checks if the port is int and raises ValueError when port is not integer

# Checklist

### DIFF
--- a/uvicorn/main.py
+++ b/uvicorn/main.py
@@ -512,7 +512,13 @@ def run(
 ) -> None:
     if app_dir is not None:
         sys.path.insert(0, app_dir)
-
+    if isinstance(port, str):
+        try:
+            port = int(port)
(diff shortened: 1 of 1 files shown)


shashstormer:
SKIPPED [1] tests/supervisors/test_multiprocess.py:110: platform unsupports SIGBREAK
================== 825 passed, 1 skipped in 82.58s (0:01:22) ===================

Is this causing < 100% coverage?

Kludex:
This was brought several times, but I'm still not convinced this is a good addition. We don't validate any field when running programmatically, but we do have proper type hints, and when running via CLI the parsing is already in place.

Nonetheless, thanks for the PR.

PR #2537, opened 2024-12-14
Speed up test suite with pytest-xdist

Kludex:
We created the `unused_tcp_port` fixture some time ago to enable `pytest-xdist`.

The test suite is too slow right now. I'll give this a try.


Kludex:
There seems to be some flakiness in the reloader tests on Windows. 😢

tests/supervisors/test_reload.py
```diff
@@ -23,11 +22,8 @@
     WatchFilesReload = None  # type: ignore[misc,assignment]
 
 
-# TODO: Investigate why this is flaky on MacOS M1.
-skip_if_m1 = pytest.mark.skipif(
-    sys.platform == "darwin" and platform.processor() == "arm",
-    reason="Flaky on MacOS M1",
-)
+# TODO: Investigate why this is flaky on MacOS, and Windows.
+skip_non_linux = pytest.mark.skipif(sys.platform in ("darwin", "win32"), reason="Flaky on Windows and MacOS M1")
```
Kludex:
```suggestion
skip_non_linux = pytest.mark.skipif(sys.platform in ("darwin", "win32"), reason="Flaky on Windows and MacOS")
```

PR #2540, opened 2024-12-14
Add `WebSocketsSansIOProtocol`

Kludex:


Kludex:
@aaugustin Do you have availability to review this in the next days?

Kludex:
Seems like I need to drop Python 3.8 to make my life easier. 👀

aaugustin:
I'm making a note to look into it tomorrow. If I don't find time tomorrow, then I'll do over the Christmas break.

requirements.txt
```diff
@@ -7,7 +7,8 @@ h11 @ git+https://github.com/python-hyper/h11.git@master
 # Explicit optionals
 a2wsgi==1.10.7
 wsproto==1.2.0
-websockets==13.1
+websockets==13.1; python_version < '3.9'
+websockets==14.1; python_version >= '3.9'
```
Kludex: Let's bump this in another PR.
```suggestion
websockets==13.1
```

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,405 @@
...
+            return
+        self.handle_events()
+
+    def handle_events(self) -> None:
+        for event in self.conn.events_received():
+            if isinstance(event, Request):
+                self.handle_connect(event)
+            if isinstance(event, Frame):
+                if event.opcode == Opcode.CONT:
+                    self.handle_cont(event)
```
Kludex: Recommendation on how to test this? The other implementations don't have the continuation frame 🤔

tests/middleware/test_logging.py
```diff
@@ -104,7 +106,7 @@ async def websocket_app(scope: Scope, receive: ASGIReceiveCallable, send: ASGISe
             elif message["type"] == "websocket.disconnect":
                 break
 
-    async def open_connection(url):
+    async def open_connection(url: str):
         async with websockets.client.connect(url) as websocket:
             return websocket.open
```
aaugustin: `return websocket.state is websockets.protocol.State.OPEN`

Cf. https://websockets.readthedocs.io/en/stable/howto/upgrade.html#open-and-closed

uvicorn/protocols/websockets/websockets_sansio_impl.py
aaugustin: This happens when a message is fragmented. See https://datatracker.ietf.org/doc/html/rfc6455#section-5.4 for context. In short:

- the first frame is TEXT or BINARY
- every subsequent frame is CONT
- the last frame has the FIN bit set

This is correctly implemented in `handle_cont`. As far as I can tell, ASGI doesn't support fragmented WebSocket messages so you have to buffer and reassemble.

---

For comparison, websockets provides both options — buffer and reassemble the message (if you don't care) or receive it frame by frame (if you need streaming). If you're curious, it's implemented [here](https://github.com/python-websockets/websockets/blob/main/src/websockets/asyncio/messages.py).

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,405 @@
+from __future__ import annotations
+
+import asyncio
+import logging
+from asyncio.transports import BaseTransport, Transport
+from http import HTTPStatus
+from typing import Any, Literal, cast
+from urllib.parse import unquote
+
+from websockets import InvalidState
```
aaugustin: `from websockets.exceptions import InvalidState`

aaugustin: Rationale: I recommend that you avoid shurtcut imports going forwards.

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,405 @@
...
+        # WebSocket state
+        self.queue: asyncio.Queue[ASGIReceiveEvent] = asyncio.Queue()
+        self.handshake_initiated = False
+        self.handshake_complete = False
+        self.close_sent = False
+        self.initial_response: tuple[int, list[tuple[str, str]], bytes] | None = None
+
+        extensions = []
+        if self.config.ws_per_message_deflate:
+            extensions = [ServerPerMessageDeflateFactory()]
```
aaugustin: Default settings lead to high-memory usage with marginal performance benefits e.g. 316KiB RAM per connection instead of 64KiB for an additional 10% compression in a basic benchmark.

FYI websockets defaults to:

```python
            ServerPerMessageDeflateFactory(
                server_max_window_bits=12,
                client_max_window_bits=12,
                compress_settings={"memLevel": 5},
            )
```

See [here](https://websockets.readthedocs.io/en/stable/topics/compression.html) for much more than you ever wanted to know.

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,405 @@
...
+        self.conn = ServerProtocol(
+            extensions=extensions,
+            max_size=self.config.ws_max_size,
+            logger=logging.getLogger("uvicorn.error"),
+        )
```
aaugustin, line 93: Are you checking the Origin header elsewhere in uvicorn? If not, supporting it in the config would be good, as it helps defend against Cross-Site WebSocket Hijacking attacks. (I have forgotten about the details of this attack.)

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,405 @@
...
+        self.client = get_remote_addr(transport)
+        self.scheme = "wss" if is_ssl(transport) else "ws"
+
+        if self.logger.level <= TRACE_LOG_LEVEL:
+            prefix = "%s:%d - " % self.client if self.client else ""
+            self.logger.log(TRACE_LOG_LEVEL, "%sWebSocket connection made", prefix)
+
+    def connection_lost(self, exc: Exception | None) -> None:
+        code = 1005 if self.handshake_complete else 1006
+        self.queue.put_nowait({"type": "websocket.disconnect", "code": code})
```
aaugustin, line 117: In a normal closing handshake scenario where you generate a `"websocket.disconnect"` event when you receive a close frame, it looks like this will generete a second `"websocket.disconnect"` event when the TCP connection terminates, which sounds incorrect. Ideally, you'd generate that event only if you haven't generated one already.

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,405 @@
...
+        if self.logger.level <= TRACE_LOG_LEVEL:
+            prefix = "%s:%d - " % self.client if self.client else ""
+            self.logger.log(TRACE_LOG_LEVEL, "%sWebSocket connection lost", prefix)
+
+        self.handshake_complete = True
+        if exc is None:
+            self.transport.close()
+
+    def eof_received(self) -> None:
+        pass
```
aaugustin, line 129: I would expect `self.conn.receive_eof()` here.

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,405 @@
...
+        self.bytes = event.data
+        self.curr_msg_data_type = "bytes"
+        if event.fin:
+            self.send_receive_event_to_app()
+
+    def send_receive_event_to_app(self) -> None:
+        data_type = self.curr_msg_data_type
+        msg: WebSocketReceiveEvent
+        if data_type == "text":
+            msg = {"type": "websocket.receive", data_type: self.bytes.decode()}
```
aaugustin: You may want to handle the case when `decode()` raises an exception. In that case, websockets closes the connection with code 1007 (invalid data).

Invalid UTF-8 is relatively uncommon so you could choose to ignore it. I forgot this case in the new asyncio implementation and it took some time to detect the bug.

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,405 @@
...
+        else:
+            msg = {"type": "websocket.receive", data_type: self.bytes}
+        self.queue.put_nowait(msg)
+        if not self.read_paused:
+            self.read_paused = True
+            self.transport.pause_reading()
+
+    def handle_ping(self, event: Frame) -> None:
+        output = self.conn.data_to_send()
+        self.transport.write(b"".join(output))
```
aaugustin, line 243: I would suggest that you always do this in `handle_events()` rather than only in `handle_xxx()` methods where you believe that websockets sent an automatic response. I'm not sure how much of a different it will make in practice but it feels cleaner.

aaugustin (commented):
I'm not very familiar with ASGI & uvicorn so I focused my review on the WebSocket protocol.

Your usage of `websockets` is straightforward.

I noticed that:

- You aren't detecting the case when `data_to_send()` returns the empty bytestring to tell you to half-close the connection; that could be OK if you're confident that you're closing the transport in every scenario where the connection is terminating.
- You aren't using the `close_expected` API to check whether you should close the TCP connection; that may be harmless on the server side because [the server is expected to initiate the TCP closing handshake](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.1).

ilyakamens:
Hi! I'm just leaving a comment to say that I'm running into the issue this fixes, and I'm greatly anticipating this being merged. Let me know if there's anything I can do to help.

uvicorn/protocols/websockets/websockets_sansio_impl.py
Kludex: So is the `ws_per_message_deflate` setting useless?

uvicorn/protocols/websockets/websockets_sansio_impl.py
Kludex: I think ASGI allows sends multiple `"websocket.disconnect"`. I'll confirm.

uvicorn/protocols/websockets/websockets_sansio_impl.py
Kludex: Does it make a difference? None of the other protocols have this implemented.

uvicorn/protocols/websockets/websockets_sansio_impl.py
aaugustin: Compression is useful. However, I recommend to adjust the settings like I showed, because it's likely to lead to better performance in typical use cases (lots of connections receiving frequent, small messages).

uvicorn/protocols/websockets/websockets_sansio_impl.py
aaugustin: Since you are only implementing a server, you can get away without it, as the server is expected to close the TCP connection first.

Including it could cause websockets to send a close frame or log a message when losing connectivity, which could make troubleshooting easier vs. the connection gets garbage collected and you don't know what happened.

uvicorn/protocols/websockets/websockets_sansio_impl.py
Kludex: Just to be clear here, you are not suggesting a specific window/memory level - you're just saying that it would be smart to change it, right? Since the normal websocket server would receive a lot of connections with small messages.

I see this paragraph on the docs:
> websockets defaults to Window Bits = 12 and Memory Level = 5 to stay away from Window Bits = 10 or Memory Level = 3 where performance craters, raising doubts on what could happen at Window Bits = 11 and Memory Level = 4 on a different corpus.

Is your point that I should move the default to window bits = 11, and memory level = 4?

uvicorn/protocols/websockets/websockets_sansio_impl.py
Kludex: We do not check "Origin" in uvicorn. I'll read about it.

uvicorn/protocols/websockets/websockets_sansio_impl.py
aaugustin: My point is that you should not default to window bits = 15 and memory level = 9 — which would be the default if you do nothing special.

I don't have a strong opinion on 12/5 vs. 11/4 (or any other combo in the 11–13 and 4–6 ranges). Whatever default you choose is going to work better for some users and worse from others depending on the contents of their messages, on the latency and throughput of their network, and on their CPU and RAM costs.

If you want a reco, I stand by the default in websockets (12/5) for servers. No one ever suggested something different. While that doesn't prove that it's optimal, at least it proves that it doesn't cause major trouble.

uvicorn/protocols/websockets/websockets_sansio_impl.py
aaugustin: FWIW I don't think that's a release blocker. For everyone running behind a reverse proxy (Cloudflare, Cloudfront, etc.), the reverse proxy does it.

uvicorn/protocols/websockets/websockets_sansio_impl.py
Kludex: Thanks. 👍

FeodorFitsner:
Is there any hope this PR could be finished/merged? 😅

Kludex:
> Is there any hope this PR could be finished/merged? 😅

It's going to happen at some point. No ETA. Not much needed to finish this PR tho.

---

Funny. I am at PyCon SG and I heard a talk about flet yesterday. Nice stuff! 😁

I don't understand why your project has both hypercorn and Uvicorn as dependencies tho.

FeodorFitsner:
Ah, great! :)

> I don't understand why your project has both hypercorn and Uvicorn as dependencies tho.

Good catch, it was probably added during the testing of some options. Flet uses Uvicorn as a built-in web server and it's awesome.

uvicorn/protocols/websockets/websockets_sansio_impl.py
Kludex: You mean those 2 lines, right?

Which other method besides this one we would be able to do it tho? The others seem to close the transport conditionally.

Kludex:
@aaugustin Pipeline is passing. Do you have time to give a last look? (also, sorry for the delay)

Things that I skip in this PR:
1. We are not testing `CONT` frames.
2. We are not testing when there's a parser issue.

I'd like to have those 2 points tested, but I don't want to spend more time in this to have this merged.

---

I also think that WebSockets merits more documentation in Uvicorn.

---

Besides that, I checked all the comments above, and none of them seem to block this PR to get merged (correct me if wrong @aaugustin).

aaugustin:
I will have a look this week-end.

tests/middleware/test_logging.py
```diff
@@ -105,9 +105,9 @@ async def websocket_app(scope: Scope, receive: ASGIReceiveCallable, send: ASGISe
             elif message["type"] == "websocket.disconnect":
                 break
 
-    async def open_connection(url):
+    async def open_connection(url: str):
         async with websockets.client.connect(url) as websocket:
```
aaugustin, line 109: You should use `websockets.asyncio.client.connect` and adjust imports accordingly to avoid the deprecation warning.

Then you don't have to add `"ignore: websockets.client.connect is deprecated.*:DeprecationWarning"` in pyproject.toml.

uvicorn/protocols/websockets/websockets_sansio_impl.py
```diff
@@ -0,0 +1,413 @@
...
+
+    def handle_bytes(self, event: Frame) -> None:
+        self.bytes = event.data
+        self.curr_msg_data_type = "bytes"
+        if event.fin:
+            self.send_receive_event_to_app()
+
+    def send_receive_event_to_app(self) -> None:
+        if self.curr_msg_data_type == "text":
+            self.queue.put_nowait({"type": "websocket.receive", "text": self.bytes.decode()})
```
aaugustin: If you receive invalid UTF-8, `self.bytes.decode()` will raise `UnicodeDecodeError`; should it be handled?

Perhaps it's fine to just let the exception bubble up; I have no idea how error handling works in uvicorn.

aaugustin (approved):
I don't see any blockers for merging this PR.

aaugustin:
> We are not testing CONT frames.

They're uncommon in real life and your implementation is simple enough that it's probably right :-)

If you want to do a quick manual test, you can run this:

```python
import asyncio
from websockets.asyncio.client import connect

async def fragmented_hello():
    async with connect("<address of echo server built with uvicorn>") as websocket:
        await websocket.send(["Hell", "o!"])
        await websocket.recv()

if __name__ == "__main__":
    asyncio.run(hello())
```

> We are not testing when there's a parser issue.

The parser is hardened and quite hard to crash. oss-fuzz has been trying to crash if for several years now (https://github.com/python-websockets/websockets/blob/main/fuzzing/fuzz_websocket_parser.py) and hasn't crashed it yet.

tests/middleware/test_logging.py
Kludex: I've tried this. I think this will need to be a next step... It seems is not just replacing it with the new structures. The behavior itself seems to change in some tests.

uvicorn/protocols/websockets/websockets_sansio_impl.py
Kludex: "Fatal error on transport TCPTransport"

I guess I do need to handle. 😆

PR #2541, opened 2024-12-14
Bump websockets on `requirements.txt`

Kludex:


tests/protocols/test_websocket.py
```diff
@@ -692,7 +690,7 @@ async def get_subprotocol(url: str):
 
 
 MAX_WS_BYTES = 1024 * 1024 * 16
-MAX_WS_BYTES_PLUS1 = MAX_WS_BYTES + 1
+MAX_WS_BYTES_PLUS1 = MAX_WS_BYTES + 10
```
Kludex, line 693: @aaugustin Do you know why I had to add 10 instead of 1 here? 🤔 Did something changed on the legacy server protocol in this regard?

requirements.txt
```diff
@@ -7,7 +7,7 @@ h11 @ git+https://github.com/python-hyper/h11.git@master
 # Explicit optionals
 a2wsgi==1.10.7
 wsproto==1.2.0
-websockets==13.1
+websockets==14.1
```
Kludex, line 10: Maybe we can drop Python 3.8 first, and then move this to 14.1.

tests/protocols/test_websocket.py
aaugustin: That doesn't right a bell :-(

PR #2435, opened 2024-08-18
Support custom IOLOOPs

gnir-work:
# Summary

This is merge of two PR's
1. https://github.com/encode/uvicorn/pull/2391
Support passing a custom IOLoop setup function in order to support running `uvicorn` in custom ioloop implementation other that the default and `uvloop`.
For example this feature will allow using [this monitored IOLoop](https://github.com/gnir-work/monitored-ioloop) inside `uvicorn` applications :)
Link to discussions thread - https://github.com/encode/uvicorn/discussions/2256

2. https://github.com/encode/uvicorn/pull/2130
`set_event_loop_policy` will be deprecated in python 3.13 and `uvloop.EventLoopPolicy` is deprecated already, this backports and uses the preferred `loop_factory` keyword to configure `asyncio.run`

# Checklist

- [x] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.

PR #2548, opened 2024-12-30
test on 3.14

graingert:
…licy

# Summary

# Checklist


graingert:
I'll take this out of draft when https://github.com/encode/uvicorn/pull/2435 is in

QuLogic:
#2435 is in now... Also 3.14 is now in RC phase, so it would be a good time to get any issues sorted out before it goes final.

Kludex:
Closing this in favor of #2723

PR #2550, opened 2025-01-01
Move changelog to release notes

Kludex:


LincolnPuzey:
@Kludex deleting the `CHANGELOG.md` file from the `master` branch has broken the Changelog link for recent (all?) PyPI releases.
e.g. Go here https://pypi.org/project/uvicorn/ -> click "Changelog" -> get 404

I was about to open an issue before I saw this PR.

Not saying anything needs to change, I think this will be mostly fixed by the next release. Just an FYI really.

Kludex:
I think we need to add the file back, and say that it was moved there.

Kludex:
I created a symlink, but it doesn't redirect on GitHub 😢 

Ref.: https://github.com/encode/uvicorn/blob/master/CHANGELOG.md

I'm not sure this is enough...

PR #2558, opened 2025-01-08
Add Winloop To the list of loop setups and optional-dependencies

Vizonex:
# Summary

I added winloop as another option for loop setups since winloop is a fork of uvloop that I made almost 2 years ago for using
uvloop with windows. Even though I've made a tutorial in my own readme on how to use winloop with uvicorn and fastapi, I thought it would be time to actually integrate my library to this one. However I am questioned about how we go about making a test and also how I should go about updating the documentation for this new addition that I have added to this library so I have those two things on the checklist unchecked, other than that I am excited to see where this will go in the future as well as seeing how people will go about using this in the future. 

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ standard = [
     "python-dotenv>=0.13",
     "PyYAML>=5.1",
     "uvloop>=0.14.0,!=0.15.0,!=0.15.1; sys_platform != 'win32' and (sys_platform != 'cygwin' and platform_python_implementation != 'PyPy')",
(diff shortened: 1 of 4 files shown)


graingert:
I think it's best to wait for this PR: https://github.com/encode/uvicorn/pull/2435 which will then work with `--loop=winloop:new_event_loop`

Kludex:
Thanks for the PR @Vizonex , and sorry the delay!

I'll go with documenting the usage on https://github.com/Kludex/uvicorn/pull/2725, so I'll be closing this PR.

---

Sorry the delay.

PR #2561, opened 2025-01-12
Add socket path to `scope["server"]`

Kludex:
I think we can refactor a bit what is here, but this should solve https://github.com/encode/uvicorn/discussions/2559.


nggit:
I haven't tried it yet but those seem like the right changes, which are safe to do without much refactoring.

Only minor (subjective):
- `/tmp` is global and it's better to use a unique prefix for files like `uvicorn-` or `uvicorn/`, otherwise create `test.sock` in the current directory - Edit: But that's okay, if `MockSocket` thing doesn't create a real file
- `if` rather than `elif`.

PR #2571, opened 2025-02-05
fix asgi spec violation causing crash in lilya

devkral:
# Summary

# Checklist

Detail:

The asgi spec only specifies that the headers are an iterable. Uvicorn assumes that the headers in scope are a list. This causes a crash at least in lilya and can cause hard to debug bugs because of exhausted iterators

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -465,8 +465,11 @@ class RequestResponseCycle:
 
             status = message["status"]
             headers = self.default_headers + list(message.get("headers", []))
+            # we need to make sure that the headers are a list, the spec requires only an iterator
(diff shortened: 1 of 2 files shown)


Kludex:
I don't think it assumes that.

```py
if (1, 1) in iter([(1, 1), (2, 2), (3, 3)]):
    print('It works with iterators...')
```

Please post some reference to the issue.

devkral:
it works *once*:
```
foo = iter([(1, 1), (2, 2), (3, 3)])
if (1, 1) in foo:
  print("works 1")
if (1, 1) in foo:
  print("works 2")
```

The problem are the iterator is exhausted. I referenced already a framework which has bugs because of this behavior.

Note: I introduced the change which caused the bug.

PR #2583, opened 2025-02-13
relax reload_dirs to allow files, and stop adding cwd when reload_dirs is set

adhami3310:
# Summary

Based on the discussion on https://github.com/encode/uvicorn/discussions/1833 the behavior currently done by a uvicorn is intuitive and most likely will lead to bugs by downstream users. It's also the case that watchfiles can watch individual files, as such it's a manufactured constraint by uvicorn is very arbitrary. Some other ASGIs just has `reload_paths`, that handles files just fine. I didn't change the variable name to not break backwards compatibility, but the former change can be seen as breaking, even if the existing behavior can be considered as a bug. 

# Checklist

I'm not familiar with the larger project, so I will be holding off on documentation and tests until it's verified that the project has interest in the change getting merged.

### DIFF
--- a/uvicorn/supervisors/watchfilesreload.py
+++ b/uvicorn/supervisors/watchfilesreload.py
@@ -63,10 +63,7 @@ class WatchFilesReload(BaseReload):
         self.reloader_name = "WatchFiles"
         self.reload_dirs = []
         for directory in config.reload_dirs:
-            if Path.cwd() not in directory.parents:
-                self.reload_dirs.append(directory)
-        if Path.cwd() not in self.reload_dirs:
-            self.reload_dirs.append(Path.cwd())
+            self.reload_dirs.append(directory)
 
         self.watch_filter = FileFilter(config)
         self.watcher = watch(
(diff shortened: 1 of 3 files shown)


stinovlas (approved):
Looks good to me! I bumped into the issue with `--reload-dir` not behaving like documentation states and this should fix it. Hopefully some maintainer will pick it up soon :-).

uvicorn/supervisors/watchfilesreload.py
```diff
@@ -63,10 +63,7 @@ def __init__(
         self.reloader_name = "WatchFiles"
         self.reload_dirs = []
         for directory in config.reload_dirs:
-            if Path.cwd() not in directory.parents:
-                self.reload_dirs.append(directory)
-        if Path.cwd() not in self.reload_dirs:
-            self.reload_dirs.append(Path.cwd())
+            self.reload_dirs.append(directory)
```
ateijelo, line 66: maybe just do `self.reload_dirs = config.reload_dirs` ?

PR #2586, opened 2025-02-21
Asgiref tls extension proposal

eirikhex:
# Summary

Simplification of #1119 according to the discussion in django/asgiref#466 
A proof of concept for a change in the asgi spec for the tls extension where
`client_cert_name `is removed because it's not the role of the ASGI web server to parse X.509 certificate. The content of the field is a part of the `client_cert_chain`, and should be extracted from there. (as shown in the test in this PR)
`client_cert_error ` is removed as it was impossible to fill with any useful information.
`tls_version `and `cipher_suite ` is filled with string returned by [ssl.SSLSocket.version()](https://docs.python.org/3.13/library/ssl.html#ssl.SSLSocket.cipher) and [ssl.SSLSocket.cipher()](https://docs.python.org/3.13/library/ssl.html#ssl.SSLSocket.version)

# Checklist


uvicorn/protocols/http/httptools_impl.py
```diff
@@ -106,6 +125,9 @@ def connection_made(  # type: ignore[override]
         self.client = get_remote_addr(transport)
         self.scheme = "https" if is_ssl(transport) else "http"
 
+        if self.config.is_ssl:
+            self.tls = get_tls_info(transport, self.config)
```
Kludex: We don't need to do this here. We also don't need the `tls` attribute.

We can just pass it below:

```py
        if self.config.is_ssl:
            self.scope["extensions"]["tls"] = get_tls_info(self.transport, self.config)
```

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -21,8 +26,21 @@
...
 from uvicorn.config import Config
 from uvicorn.logging import TRACE_LOG_LEVEL
-from uvicorn.protocols.http.flow_control import CLOSE_HEADER, HIGH_WATER_LIMIT, FlowControl, service_unavailable
-from uvicorn.protocols.utils import get_client_addr, get_local_addr, get_path_with_query_string, get_remote_addr, is_ssl
+from uvicorn.protocols.http.flow_control import (
+    CLOSE_HEADER,
+    HIGH_WATER_LIMIT,
+    FlowControl,
+    service_unavailable,
+)
```
Kludex: Revert this, please.

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -7,7 +7,12 @@
 import urllib
 from asyncio.events import TimerHandle
 from collections import deque
-from typing import Any, Callable, Literal, cast
+from typing import (
+    Any,
+    Callable,
+    Literal,
+    cast,
+)
```
Kludex: Revert.

uvicorn/config.py
```diff
@@ -260,6 +260,7 @@ def __init__(
         self.callback_notify = callback_notify
         self.ssl_keyfile = ssl_keyfile
         self.ssl_certfile = ssl_certfile
+        self.ssl_cert_pem: str | None = None
```
Kludex, line 263: Why did you add this? Seems out of scope?

uvicorn/protocols/utils.py
```diff
@@ -54,3 +57,44 @@ def get_path_with_query_string(scope: WWWScope) -> str:
     if scope["query_string"]:
         path_with_query_string = "{}?{}".format(path_with_query_string, scope["query_string"].decode("ascii"))
     return path_with_query_string
+
+
+class TLSInfo(TypedDict, total=False):
+    server_cert: str | None
+    client_cert_chain: list[str]
+    tls_version: str | None
+    cipher_suite: str | None
```
Kludex: This needs to go in the `_types` module, and be added to the `extensions` key.

uvicorn/protocols/utils.py
```diff
@@ -54,3 +57,44 @@ def get_path_with_query_string(scope: WWWScope) -> str:
...
+
+
+def get_tls_info(transport: asyncio.Transport, server_config: Config) -> TLSInfo:
```
Kludex:
```suggestion
def get_tls_info(transport: asyncio.Transport, config: Config) -> TLSInfo:
```

Kludex (changes requested):

uvicorn/config.py
eirikhex: Was a part of the original PR. Not used, so it is now removed.

eirikhex: This was added in order to serve read and store the server cert pem one place. This need to be passed to the scope, and the file should not be read on every request.

should this be placed somewhere else?

uvicorn/protocols/utils.py
eirikhex: Bit unsure about how that works. This is my attempt to fix it: https://github.com/encode/uvicorn/pull/2586/commits/3186a0dbeee35909df81ced9d44b27ab961c8c7f

lattwood (commented):
Would it make sense for this to be optional/configurable so there's 0 overhead if you don't want this information?

PR #2590, opened 2025-03-05
Avoid test dependency on typing_extensions

carlwgeorge:
# Summary

TypedDict was added to the standard library typing module in Python 3.8. Since this project already requires at least Python 3.9, we can rely on this being available.

Previous discussed in https://github.com/encode/uvicorn/discussions/2589.

# Checklist

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -9,7 +9,6 @@ import pytest
 import websockets
 import websockets.client
 import websockets.exceptions
-from typing_extensions import TypedDict
 from websockets.extensions.permessage_deflate import ClientPerMessageDeflateFactory
 from websockets.typing import Subprotocol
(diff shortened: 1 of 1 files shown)


tests/protocols/test_websocket.py
```diff
@@ -9,7 +9,7 @@
 import websockets
 import websockets.client
 import websockets.exceptions
-from typing_extensions import TypedDict
+from typing import TypedDict
```
Kludex: I think you need to use `typing.TypedDict` instead of changing the import.

carlwgeorge: Ah I see, typing is already imported further up.  Will adjust.

Kludex (approved):

Kludex:
Thanks.

PR #2598, opened 2025-03-14
Don't include `cwd()` when non-empty `--reload-dirs` is passed

stinovlas:
# Summary

This PR fixes the behavior of `--reload-dir` to match the documentation. The documentation states:

```
  --reload-dir PATH               Set reload directories explicitly, instead
```

also:

> `--reload-dir <path>` - Specify which directories to watch for python file changes. May be used multiple times. If unused, then by default the whole current directory will be watched. If you are running programmatically use `reload_dirs=[]` and pass a list of strings.

However, the current implementation always adds `Path.cwd()` to list of watched directories. This is undesirable and contradicts the documentation. `Path.cwd()` should only be added as a default when no `--reload-dir` is specified.

# Checklist

- [x]  No documentation update is required.

### DIFF
--- a/tests/supervisors/test_reload.py
+++ b/tests/supervisors/test_reload.py
@@ -306,24 +306,21 @@ class TestBaseReload:
 
 
 @pytest.mark.skipif(WatchFilesReload is None, reason="watchfiles not available")
-def test_should_watch_one_dir_cwd(mocker: MockerFixture, reload_directory_structure: Path):
+def test_should_watch_cwd(mocker: MockerFixture, reload_directory_structure: Path):
     mock_watch = mocker.patch("uvicorn.supervisors.watchfilesreload.watch")
-    app_dir = reload_directory_structure / "app"
-    app_first_dir = reload_directory_structure / "app_first"
 
-    with as_cwd(reload_directory_structure):
-        config = Config(
-            app="tests.test_config:asgi_app",
-            reload=True,
-            reload_dirs=[str(app_dir), str(app_first_dir)],
-        )
-        WatchFilesReload(config, target=run, sockets=[])
-        mock_watch.assert_called_once()
-        assert mock_watch.call_args[0] == (Path.cwd(),)
+    config = Config(
+        app="tests.test_config:asgi_app",
+        reload=True,
+        reload_dirs=[],
+    )
+    WatchFilesReload(config, target=run, sockets=[])
+    mock_watch.assert_called_once()
+    assert mock_watch.call_args[0] == (Path.cwd(),)
 
 
 @pytest.mark.skipif(WatchFilesReload is None, reason="watchfiles not available")
-def test_should_watch_separate_dirs_outside_cwd(mocker: MockerFixture, reload_directory_structure: Path):
+def test_should_watch_multiple_dirs(mocker: MockerFixture, reload_directory_structure: Path):
     mock_watch = mocker.patch("uvicorn.supervisors.watchfilesreload.watch")
     app_dir = reload_directory_structure / "app"
     app_first_dir = reload_directory_structure / "app_first"
@@ -337,7 +334,6 @@ def test_should_watch_separate_dirs_outside_cwd(mocker: MockerFixture, reload_di
     assert set(mock_watch.call_args[0]) == {
         app_dir,
         app_first_dir,
-        Path.cwd(),
     }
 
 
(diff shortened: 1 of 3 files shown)


uvicorn/supervisors/watchfilesreload.py
```diff
@@ -63,10 +63,7 @@ def __init__(
         self.reloader_name = "WatchFiles"
         self.reload_dirs = []
         for directory in config.reload_dirs:
-            if Path.cwd() not in directory.parents:
-                self.reload_dirs.append(directory)
-        if Path.cwd() not in self.reload_dirs:
-            self.reload_dirs.append(Path.cwd())
```
Ale-Cas, line 69: I think that in order to correctly match the documentation you still need to add the cwd when config.reload_dirs is empty, so the code that you removed can be replaced with: 
```python
if not config.reload_dirs:
    self.reload_dirs.append(Path.cwd())
```

adhami3310:
i tried doing so in https://github.com/encode/uvicorn/pull/2583 but it seems the maintainer has better things to do :/ since then i moved to a different library that handled reload dir in a sensible way

Kludex:
> i tried doing so in #2583 but it seems the maintainer has better things to do :/ since then i moved to a different library that handled reload dir in a sensible way

Pretty rude considering your company leverages the contributions I've made to the ecosystem.

adhami3310:
> Pretty rude considering your company leverages the contributions I've made to the ecosystem.

Everyone appreciates your contributions! We're also trying to contribute back :)

If you're busy with other things, that's ok. Everyone's been there, I've been there. I think, however, one should *aim* to communicate with their community on the timescale they expect things to happen. If the community finds an issue in the project they want to fix, and they offer the fix to you, it's (I'm not going to say common) decency to acknowledge them. In the same way you expect people to appreciate your contributions make them feel like their contributions are appreciated.

I talk in ideas of course, who among us is perfect.

uvicorn/supervisors/watchfilesreload.py
stinovlas: That's what I had there earlier, but it seems that `Path.cwd()` is added somewhere else in the chain, this function never actually receives empty `config.reload_dirs`, which is why I wasn't able to test this in any sensible way.

Ale-Cas: You're right, it's probably just added when initializing the config in that case. As a minor thing, I would still change this code a little bit to use `Path.cwd()` instead of `Path(os.getcwd())` so that it's also easier to search where that is added in the chain.

Screenshot 2025-03-21 at 13 20 57

stinovlas: Sure, I wanted to keep the change minimal and isolated, but changing `Path(os.getcwd())` to `Path.cwd()` won't hurt anything, so I applied the suggestion.

tests/supervisors/test_reload.py
```diff
@@ -306,24 +306,21 @@ def test_watchfiles_no_changes(self) -> None:
...
-            reload_dirs=[str(app_dir), str(app_first_dir)],
-        )
-        WatchFilesReload(config, target=run, sockets=[])
-        mock_watch.assert_called_once()
-        assert mock_watch.call_args[0] == (Path.cwd(),)
+    config = Config(
+        app="tests.test_config:asgi_app",
+        reload=True,
+        reload_dirs=[],
+    )
```
Kludex:
```suggestion
    config = Config(app="tests.test_config:asgi_app", reload=True, reload_dirs=[])
```

Kludex:
Hmmm... What about the scenario when you run `uvicorn main:app --reload --reload-dirs jinja-templates`?

If you change the `main.py`, it will not reload. I understand that we don't want `Path.cwd()`, but maybe we want the module's app directory to be included? e.g. on `app.main:app`, we'd watch what is inside `app`.

I can see how confusing it can be for people that, for example, just include a folder with jinja files into `--reload-dirs`.

stinovlas:
> Hmmm... What about the scenario when you run `uvicorn main:app --reload --reload-dirs jinja-templates`?
> 
> If you change the `main.py`, it will not reload. I understand that we don't want `Path.cwd()`, but maybe we want the module's app directory to be included? e.g. on `app.main:app`, we'd watch what is inside `app`.
> 
> I can see how confusing it can be for people that, for example, just include a folder with jinja files into `--reload-dirs`.

If it were up to me, I wouldn't automagically add anything to the watchlist. If you specify `--reload-dirs`, you're already overriding the default behavior. IMHO, explicit is better than implicit.

stinovlas:
One more point is that I tried to make this PR as small as possible and not introduce any behavior changes apart from fixing the implementation to match the docs. I can see some arguments for adding the application dir (as well as some arguments against it), but I think that adding the application dir should be out of scope of this particular PR.

Kludex (approved):

Issue #2000, opened 2023-06-08
Slow reload on Windows 10 using PyCharm


Kludex:
### Discussed in https://github.com/encode/uvicorn/discussions/1977

<div type='discussions-op-text'>

<sup>Originally posted by **adhihargo** May 14, 2023</sup>
Uvicorn used to work fine in my dev laptop running Windows 10, but now every time there's a modified change, it stuck in this line:

`WARNING:  WatchFiles detected changes in 'main.py'. Reloading...`

... for nearly 5 minutes before actually reloading the files (printing `Shutting down` in the next line, etc.). During that time, the API server part works normally but serving the older version before file modification..

Downgraded to v0.21.1, and reloading works instantly again.</div>

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/uvicorn/issues/2000">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/2000/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/uvicorn/issues/2000/pledge.svg">
</picture>
</a>

3to1null:
I think issue lies somewhere here: https://github.com/encode/uvicorn/blob/7d4eab20ac0cf0fb07b3afaa9ca35a2ff09fb499/uvicorn/supervisors/basereload.py#L87-L99

With the changes introduced in #1584, it will hang for multiple minutes on the process.join() call. If I remove the changes, so the function looks like:
```
    def restart(self) -> None:
        self.process.terminate()
        self.process.join()
        self.process = get_subprocess(
            config=self.config, target=self.target, sockets=self.sockets
        )
        self.process.start()
```
The reload is pretty much instant, like it was on earlier versions.

3to1null:
I can only reproduce this issue when the program is run using Pycharm. If I run it directly (either via `uvicorn.run("app:app", reload=True)` in Python, or with the `uvicorn app:app --reload` from the terminal), this issue is **not** present.

While it would be very nice to have this working again, I am not sure if this is something that should be fixed in Uvicorn (if this is even possible).

[...]

PR #2604, opened 2025-03-28
Flush the stdout buffer on Windows to trigger reload

jamesh1999:
# Summary

I'm raising this in response to #2000 as it's a fix I've been using locally. Through my experimentation, I have found that printing something after the Ctrl+C event is sent, allows the terminal running uvicorn to process the event, causing it to reload immediately.

I explain this in #2000 and there is already a reply saying it has resolved the issue for somebody else.

Tested on Windows 11 with:
- PyCharm (built-in terminal)
- VS Code (built-in terminal)
- Windows Terminal app

All of these now reliably reload for me. However I still have issues in CI running headlessly. I've started #2609 to discuss a fix for this too.

# Checklist

Don't believe the last two points are necessary here but feel free to correct me.

### DIFF
--- a/uvicorn/supervisors/basereload.py
+++ b/uvicorn/supervisors/basereload.py
@@ -90,7 +90,10 @@ class BaseReload:
             self.is_restarting = True
             assert self.process.pid is not None
             os.kill(self.process.pid, signal.CTRL_C_EVENT)
-            print("")  # This helps trigger windows to respond to the Ctrl+C event
+
+            # This is a workaround to ensure the Ctrl+C event is processed
+            sys.stdout.write(" ") # This has to be a non-empty string
+            sys.stdout.flush()
         else:  # pragma: py-win32
             self.process.terminate()
         self.process.join()
--- a/uvicorn/supervisors/basereload.py
+++ b/uvicorn/supervisors/basereload.py
@@ -90,7 +90,7 @@ class BaseReload:
             self.is_restarting = True
             assert self.process.pid is not None
             os.kill(self.process.pid, signal.CTRL_C_EVENT)
-            print("") # This helps trigger windows to respond to the Ctrl+C event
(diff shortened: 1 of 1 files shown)


YousefBawaliz:
That seems to fix it. huge thanks man! 
this issue has been bugging me for way too long now. Almost gave up FastAPI because of it.

let's hope it's merged as fast as possible.

uvicorn/supervisors/basereload.py
```diff
@@ -90,6 +90,7 @@ def restart(self) -> None:
             self.is_restarting = True
             assert self.process.pid is not None
             os.kill(self.process.pid, signal.CTRL_C_EVENT)
+            print("")  # This helps trigger windows to respond to the Ctrl+C event
```
Kludex: Can you try to use `sys.stdout.flush()` instead?

Kludex (changes requested):

uvicorn/supervisors/basereload.py
jamesh1999: So I did give this a go before and `sys.stdout.flush()` on its own doesn't seem to be enough. Based on Copilot's review, I also tried the following:
- `sys.stdout.write("")` then `sys.stdout.flush()` - No success here, it still doesn't trigger a reload.
- `sys.stdout.write("\n")` then `sys.stdout.flush()` - This is closer to `print("")` and does work. But at that point, I'm not sure it's any better than just using `print("")`

jamesh1999: Looks like `sys.stdout.write(" ")` then `sys.stdout.flush()` works too actually. That's a little less intrusive to the logs than printing a blank line and possibly a little bit clearer at showing what's required to get a reload to happen (something needs to be printed & the buffer need to be flushed)

Kludex (approved):

PR #2627, opened 2025-05-07
Fix unfavorable logger name uvicorn.error

tejaschauhan369:
# Summary

# Checklist

### DIFF
--- a/tests/middleware/test_logging.py
+++ b/tests/middleware/test_logging.py
@@ -79,12 +79,12 @@ async def test_trace_logging_on_http_protocol(http_protocol_cls, caplog, logging
         log_config=logging_config,
         port=unused_tcp_port,
     )
-    with caplog_for_logger(caplog, "uvicorn.error"):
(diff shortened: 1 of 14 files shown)


Kludex:
That's not the way to move forward on this. It will break a lot of projects.

tejaschauhangit:
Hey Marcelo,
Please suggest the correct way, I want to contribute

Kludex:
The thing is, if we merge this, there will be a lot of people complaining because I broke everybody's logging setup.

We need a way to move towards this idea in a way that doesn't break people's code. I think the idea is to move towards v1, and then on v2 change this.

tejaschauhangit:
FYI: I haven't removed the existing logger "uvicorn.error" [code diff](https://github.com/encode/uvicorn/pull/2627/files#diff-6982a5e270879281f9182303a9da9776148857b46325afb230e50ef8eb3264c9R95),
I've added a new logger "uvicorn.log" and used it,
Do you still think it will break?

PR #2642, opened 2025-06-01
Add notes about 0.34.3 and Google Analytics

Kludex:
I forgot to make the 0.34.3 release, so this PR adds more notes about what was merged after that.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -56,6 +56,11 @@ nav:
   - Contributing: contributing.md
   - Sponsorship: sponsorship.md
 
+extra:
+  analytics:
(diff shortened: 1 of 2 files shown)


mkdocs.yml
```diff
@@ -56,6 +56,11 @@ nav:
   - Contributing: contributing.md
   - Sponsorship: sponsorship.md
 
+extra:
+  analytics:
+    provider: google
+    property: G-KTS6TXPD85
+
```
Kludex, line 63: I'm introducing Google Analytics, but happy to replace it anyone can educate me.

PR #2658, opened 2025-07-02
Allow to provide importable string in `--http`, `--ws` and `--loop`

Kludex:
Now it's possible to pass strings for those options:

```bash
uvicorn --loop rloop:new_event_loop main:app
```

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -480,7 +480,7 @@ class Config:
             try:
                 return import_from_string(self.loop)
             except ImportFromStringError as exc:
-                logger.exception("Error loading custom loop setup function. %s" % exc)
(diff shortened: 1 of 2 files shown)


uvicorn/config.py
```diff
@@ -480,7 +480,7 @@ def get_loop_factory(self) -> Callable[[], asyncio.AbstractEventLoop] | None:
             try:
                 return import_from_string(self.loop)
             except ImportFromStringError as exc:
-                logger.error("Error loading custom loop setup function. %s" % exc)
+                logger.exception("Error loading custom loop setup function. %s" % exc)
```
```suggestion
                logger.error("Error loading custom loop setup function. %s" % exc)
```

PR #2680, opened 2025-08-28
Add parameter `--timeout-process-probing` for computationally intensive tasks

lddfym:
# Summary

When the child process is a computationally intensive task, the heartbeat thread (pong) of the child process will be preempted of its CPU time slice, resulting in process heartbeat failure and restart. Therefore, it is necessary to increase the heartbeat timeout between processes.

# Checklist


Kludex:
Is there any discussion for this?

PR #2695, opened 2025-09-13
ci: publish docs on `workflow_dispatch`

Kludex:


scripts/publish
Kludex, line 1: This is not needed. We publish only via CI.

PR #2707, opened 2025-09-22
Add `--limit-max-requests-jitter` to stagger worker restarts

jonashaag:
# Summary

Implement `--max-requests-jitter` that works the same way as in Gunicorn (https://docs.gunicorn.org/en/stable/settings.html#max-requests-jitter).

Ref: https://github.com/Kludex/uvicorn/issues/574

*Help wanted:* The test I've added seems to be flaky: Sometimes `caplog.txt == ""` and I don't understand why.

# Checklist


jonashaag:
@Kludex I think this ready to review.

The failing test seems flaky (last time I ran it, a different test failed)

kaxil:
I would love to use this feature in Airflow to avoid all workers restarting at the same time. @Kludex - would appreciate a review on this one, please

Kludex (approved):

Kludex:
Thanks! :)

PR #2713, opened 2025-09-24
Update pyproject.toml for PEP639 compliance

NGANAMODEIJunior:
# Summary

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,8 @@ name = "uvicorn"
 dynamic = ["version"]
 description = "The lightning-fast ASGI server."
 readme = "README.md"
-license = "BSD-3-Clause"
+license = { text = "BSD-3-Clause" }
+licence-files = ["LICENCE", "COPYING"]
 requires-python = ">=3.9"
 authors = [
     { name = "Tom Christie", email = "tom@tomchristie.com" },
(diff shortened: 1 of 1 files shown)


NGANAMODEIJunior:
I removed the License :: classifiers and added license + license-files according to PEP639::

pyproject.toml
```diff
@@ -7,7 +7,8 @@ name = "uvicorn"
 dynamic = ["version"]
 description = "The lightning-fast ASGI server."
 readme = "README.md"
-license = "BSD-3-Clause"
+license = { text = "BSD-3-Clause" }
+licence-files = ["LICENCE", "COPYING"]
```
Kludex: Where did you get the COPYING file from?

pyproject.toml
```diff
@@ -7,7 +7,8 @@ name = "uvicorn"
 dynamic = ["version"]
 description = "The lightning-fast ASGI server."
 readme = "README.md"
-license = "BSD-3-Clause"
+license = { text = "BSD-3-Clause" }
```
Kludex: The string is valid. I don't think we need to replace it.

pyproject.toml
Kludex: Also, it's `license-files`, not `licence-files`.

And the file is `LICENSE.md`.

Kludex (approved):

PR #2729, opened 2025-10-15
Add first-class support for `winloop` 

CNSeniorious000:
# Summary

[Winloop 0.3.0](https://github.com/Vizonex/Winloop/releases/tag/v0.3.0) just shipped with Python 3.14 support; I’ve tested it with some semi-large FastAPI apps on Windows and they ran fine. Since it claims to be in sync with `uvloop`, I believe it should be stable enough to serve as a drop-in replacement for `asyncio`. Additionally, having `uvicorn` use it directly could help uncover potential issues with `winloop`.

This PR adds first-class winloop support and includes it in the `[standard]` extras.

# Checklist

### DIFF
--- a/docs/concepts/event-loop.md
+++ b/docs/concepts/event-loop.md
@@ -3,15 +3,15 @@
 Uvicorn provides two event loop implementations that you can choose from using the [`--loop`](../settings.md#implementation) option:
 
 ```bash
-uvicorn main:app --loop <auto|asyncio|uvloop>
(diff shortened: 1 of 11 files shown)


Kludex:
Thanks for the contribution! We already document how to use `winloop` via the custom event loop interface (`--loop winloop:new_event_loop`), so it's usable today without any changes to uvicorn.

I'd rather not add the extra complexity here - the custom loop interface exists exactly for cases like this.

Issue #2167, opened 2023-11-30
ContextVars pollution when uvicorn installed without [standard] extensions


Kludex:
### Discussed in https://github.com/encode/uvicorn/discussions/2044

<div type='discussions-op-text'>

<sup>Originally posted by **vladyslav-burylov** July 12, 2023</sup>
Hi team, we have discovered a weird ContextVars behaviour when uvicorn being installed without [standard] extensions.

Can you please take a look - should this be considered as a bug or maybe you can help to explain why it happens?

# Summary
- Since uvicorn creates async task for each specific request, it is expected that any context vars set while processing would be isolated within a request scope
- However, above works well with 'uvicorn[standard]' edition only

# Workarounds
- Use `uvicorn[standard]` instead of plain `uvicorn` (seems to be a good option)
- Use ASGI `scope` to store request-bound data (not always possible: for example opentelemetry cloud tracing middleware stores current trace span inside context var and this logic cannot be changed)

# Steps to reproduce

## Hardware/OS/tools
- MacBook Pro M1
- macOS Ventura 13.3.1
- Python 3.10.9 installed via pyenv
- uvicorn 0.22.0

## Prepare environment
```sh
mkdir test && cd test
pyenv install 3.10.9 # unless already installed
pyenv local 3.10.9
python --version # expected to print 3.10.9
python -m venv test
source ./test/bin/activate
pip install requests
pip install 'uvicorn==0.22.0'
```

### client.py
```python
import time
import uuid
import requests

print("")
print("Starting...")


def main():
    # Everything is good when payload is small
    huge_payload = ""
    for x in range(10000):
        huge_payload += f"{str(uuid.uuid4())}\n"

    # Send request each 1 second reusing single connection (without connection reuse it works well)
    with requests.Session() as session:
        while True:
            try:
                response = session.post('http://127.0.0.1:8002/test', huge_payload)
                print(response.text) # server will be responding with either 'OK' or 'FAIL'

                # In case if server reports 'FAIL' - we reproduced the issue and are OK to exit
                if response.text == "FAIL":
                    break

            except KeyboardInterrupt:
                break

            except BaseException as ex:
                print(str(ex))

            time.sleep(0.1)

main()
```

### server.py
```python
import uuid
import uvicorn
from contextvars import ContextVar

# This var expected to store some request-bound value. Here 'request_id' as example.
request_id: ContextVar[str] = ContextVar("request_id", default="")


[...]

PR #2742, opened 2025-10-29
explicitly start ASGI run with empty context

pmeier:
# Summary

One possible solution for #2167.

# Checklist

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -1181,18 +1181,15 @@ async def test_no_contextvars_pollution(http_protocol_cls: type[HTTPProtocol]):
     # body has to be larger than HIGH_WATER_LIMIT to trigger a reading pause on the main thread
     # and a resumption inside the ASGI task
     large_body = b"a" * (HIGH_WATER_LIMIT + 1)
-    large_request = (
-        b"\r\n".join(
-            [
-                b"POST /large-body HTTP/1.1",
-                b"Host: example.org",
-                b"Content-Type: application/octet-stream",
-                f"Content-Length: {len(large_body)}".encode(),
-                b"",
-                b"",
-            ]
-        )
-        + large_body
+    large_request = b"\r\n".join(
+        [
+            b"POST /large-body HTTP/1.1",
+            b"Host: example.org",
+            b"Content-Type: application/octet-stream",
+            f"Content-Length: {len(large_body)}".encode(),
+            b"",
+            large_body,
+        ]
     )
 
     await assert_empty_initial_context(large_request)
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import contextvars
 import json
 import logging
@@ -1129,7 +1130,37 @@ async def test_header_upgrade_is_websocket_depend_not_installed(
     assert b"Hello, world" in protocol.transport.buffer
 
 
-async def test_no_contextvars_pollution(http_protocol_cls: type[HTTPProtocol]):
+@contextlib.asynccontextmanager
+async def server(*, app, port, http_protocol_cls):
+    config = Config(app=app, port=port, loop="asyncio", http=http_protocol_cls)
+    server = Server(config=config)
+    task = asyncio.create_task(server.serve())
+
+    while not server.started:
+        await asyncio.sleep(0.01)
+
+    reader, writer = await asyncio.open_connection("127.0.0.1", port)
+
+    async def extract_json_body(request: bytes):
+        writer.write(request)
+        await writer.drain()
+
+        status, *headers = (await reader.readuntil(b"\r\n\r\n")).split(b"\r\n")[:-2]
+        assert status == b"HTTP/1.1 200 OK"
+
+        content_length = next(int(h.split(b":", 1)[1]) for h in headers if h.lower().startswith(b"content-length:"))
+        return json.loads(await reader.readexactly(content_length))
+
+    try:
+        yield extract_json_body
+    finally:
+        writer.close()
+        await writer.wait_closed()
+        server.should_exit = True
+        await task
+
+
+async def test_no_contextvars_pollution(http_protocol_cls: type[HTTPProtocol], unused_tcp_port: int):
     # non-regression test for https://github.com/encode/uvicorn/issues/2167
 
     default_contextvars = {c.name for c in contextvars.copy_context().keys()}
@@ -1165,19 +1196,6 @@ async def test_no_contextvars_pollution(http_protocol_cls: type[HTTPProtocol]):
         )
         await send({"type": "http.response.body", "body": body})
 
-    protocol = get_connected_protocol(app, http_protocol_cls)
-
-    async def assert_empty_initial_context(request: bytes):
-        try:
-            protocol.data_received(request)
-            await protocol.loop.run_one()
-
-            assert b"HTTP/1.1 200 OK" in protocol.transport.buffer
-            response_body = protocol.transport.buffer.split(b"\r\n\r\n", 1)[1]
-            assert json.loads(response_body) == {}
-        finally:
-            protocol.transport.clear_buffer()
-
     # body has to be larger than HIGH_WATER_LIMIT to trigger a reading pause on the main thread
     # and a resumption inside the ASGI task
     large_body = b"a" * (HIGH_WATER_LIMIT + 1)
@@ -1192,5 +1210,6 @@ async def test_no_contextvars_pollution(http_protocol_cls: type[HTTPProtocol]):
         ]
     )
 
-    await assert_empty_initial_context(large_request)
-    await assert_empty_initial_context(SIMPLE_GET_REQUEST)
+    async with server(app=app, http_protocol_cls=http_protocol_cls, port=unused_tcp_port) as extract_json_body:
+        assert await extract_json_body(large_request) == {}
+        assert await extract_json_body(SIMPLE_GET_REQUEST) == {}
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import asyncio
+import contextvars
+import json
 import logging
 import socket
 import threading
@@ -15,6 +17,7 @@ from uvicorn._types import ASGIApplication, ASGIReceiveCallable, ASGISendCallabl
 from uvicorn.config import WS_PROTOCOLS, Config
 from uvicorn.lifespan.off import LifespanOff
 from uvicorn.lifespan.on import LifespanOn
+from uvicorn.protocols.http.flow_control import HIGH_WATER_LIMIT
 from uvicorn.protocols.http.h11_impl import H11Protocol
 from uvicorn.server import ServerState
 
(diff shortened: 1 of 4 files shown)


Kludex:
This does make sense, but do you know why it happens?

pmeier:
~I think so, yes. `asyncio` does not clean up the context when starting a new task. This is why they added the `context` keyword argument in 3.11 to make this simpler. The [documentation](https://docs.python.org/3.11/library/asyncio-task.html#asyncio.create_task) states~

~> The current context copy is created when no context is provided.~

~So far I only have a not-so-minimal reproducer: https://github.com/pmeier/asyncio-contextvars-pollution. Using the patch in this PR, the pollution is resolved.~

~How exactly it happens in the first place, I don't know yet. My goal is to get to a reproducer that just depends on `uvicorn` that I hopefully can put into a unittest. This could serve as testing ground to find the underlying reason.~

~Finally, my working hypothesis as for why this is not happening by `uvloop` is that they are not a drop-in replacement for this specific behavior of `asyncio`. [On the surface](https://github.com/MagicStack/uvloop/blob/74f4c96d3fc5281b1820491d2568de771ea7851b/uvloop/loop.pyx#L1428-L1443) it seems that they are replicating it correctly, but I haven't dug any deeper just yet.~

@Kludex I was wrong. This is a bug in `asyncio`. See https://github.com/Kludex/uvicorn/issues/2167#issuecomment-3481001395 for details.

pmeier:
Ping @Kludex

Kludex:
We need to fix it in httptools as well, and do you think you can create a test that reproduces the issue?

pmeier:
@Kludex 

> We need to fix it in httptools as well

Can you explain that? Are you talking about

https://github.com/Kludex/uvicorn/blob/8ae0bcbecb0a655789abf0c2dd4200848fc68a30/pyproject.toml#L41-L44

https://github.com/MagicStack/httptools

> do you think you can create a test that reproduces the issue?

Yeah, likely. I haven't written one, because the solution is not the only one that is possible to fix the issue. See https://github.com/Kludex/uvicorn/issues/2167#issuecomment-3481001395.

Can you decide if you want to go with the solution before I put in more work?

Kludex:
> Can you decide if you want to go with the solution before I put in more work?

It's a bit unfortunate we need to do it here, but I guess it's fine.

---

> Can you explain that? Are you talking about

We have a `h11_impl.py` and `httptools_impl.py` that has analogous logic.

pmeier:
@Kludex 

> It's a bit unfortunate we need to do it here, but I guess it's fine.

There is an open PR python/cpython#141158 that'll hopefully fix this for Python >=3.15. But until that there is no way around doing this in `uvicorn` if we don't want to tell users to use `uvloop` by default.

> can create a test that reproduces the issue

I've added a test that should reproduce the issue. However, it also won't pass right now I don't know why yet. Trying my fix on reproducers outside of the test suite works just fine. Thus, my working assumption is that one of the [mock components ](https://github.com/Kludex/uvicorn/blob/8ae0bcbecb0a655789abf0c2dd4200848fc68a30/tests/protocols/test_http.py#L270-L283) does not behave like the real component would. I'll try to find out which one.

tests/protocols/test_http.py
```diff
@@ -1124,3 +1128,88 @@ async def test_header_upgrade_is_websocket_depend_not_installed(
     assert msg in caplog.text
     assert b"HTTP/1.1 200 OK" in protocol.transport.buffer
     assert b"Hello, world" in protocol.transport.buffer
+
+
+@contextlib.asynccontextmanager
+async def server(*, app, port, http_protocol_cls):
```
pmeier: Naming TBD. Just wanted to get feedback first if the test architecture is ok.

pmeier (commented):
@Kludex I opted to avoid using the mock components, because in order to reproduce the issue, we'd need to implement the buggy behavior for them. Instead, the regression test now uses the outer `asyncio` loop.

The CI for d099a87ff3669e92a62c7bc2413767c7d2295d32 (just the test without the changes to the task handling) shows nicely that

1. The regression test correctly fails if we do not start the ASGI task with an empty context
2. Windows is not affected by this bug (see https://github.com/Kludex/uvicorn/issues/2167#issuecomment-2155060271)

pmeier:
Ping @Kludex

uvicorn/protocols/http/h11_impl.py
```diff
@@ -247,7 +248,12 @@ def handle_events(self) -> None:
                     message_event=asyncio.Event(),
                     on_response=self.on_response_complete,
                 )
-                task = self.loop.create_task(self.cycle.run_asgi(app))
+                # For the asyncio loop, we need to explicitly start with an empty context
+                # as it can be polluted from previous ASGI runs.
+                # See https://github.com/python/cpython/issues/140947 for details.
+                task = contextvars.Context().run(self.loop.create_task, self.cycle.run_asgi(app))
+                # TODO: Replace the line above with the line below for Python >= 3.11
+                # task = self.loop.create_task(self.cycle.run_asgi(app), context=contextvars.Context())
```
Kludex, line 256: Currently, the lifespan is a sibling task, but that's not the case in Hypercorn. It may be the case in the future that we refactor the server to make the lifespan task a parent of the whole process instead of a sibling task. Which means that the context would need to come from there. We still don't want the context to leak between sibling tasks as it's currently happening in asyncio...

Since this is going to be fixed in Python 3.15, can we have a note about it? I would like to revert the `context=` when it lands, or at least remember that we can do it if we decide to make the lifespan task a parent from this one.

tests/protocols/test_http.py
```diff
@@ -1124,3 +1128,88 @@ async def test_header_upgrade_is_websocket_depend_not_installed(
...
+    try:
+        yield extract_json_body
+    finally:
+        writer.close()
+        await writer.wait_closed()
+        server.should_exit = True
+        await task
+
+
+async def test_no_contextvars_pollution(http_protocol_cls: type[HTTPProtocol], unused_tcp_port: int):
```
```suggestion
async def test_no_contextvars_pollution_asyncio(http_protocol_cls: type[HTTPProtocol], unused_tcp_port: int):
```

tests/protocols/test_http.py
```suggestion
async def server(*, app: ASGIApplication, port: int, http_protocol_cls: type[HTTPProtocol]):
```

Kludex: You should use `run_server` from `utils.py`.

tests/protocols/test_http.py
Kludex: I would like to remove this test when 3.15 reaches EOF, given it's the event loop's job to make sure this is the case. 😎

Can we have a mark to not run this on >=3.15? 🙏

Kludex:
> Ping @Kludex

Sorry. I was on a short vacation.

uvicorn/protocols/http/h11_impl.py
pmeier: I'm aware of python/cpython#141158, but since this is not merged or ready to be merged, do we already want to assume that 3.15 will fix this?

tests/protocols/test_http.py
pmeier: We can depending on the answer of https://github.com/Kludex/uvicorn/pull/2742#discussion_r2627147586.

marctc:
hey @Kludex @pmeier I dont have much expertise on asyncio stuff, but is this a temporary fix or the path this project will go? I'm asking because I'm working on implementing automatic [context propagation for python asyncio](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/966) with eBPF and I found out that this PR breaks the use case of FasAPI+uvicorn. The conflicting bit is where  "we need to explicitly start with an empty context". We are using eBPF to track the execution of `context_run` and `context_create` from CPython to create connected spans for a distributed trace and from version 0.39 of uvicorn tests are failing.

thanks!

pmeier:
> is this a temporary fix or the path this project will go

I can't answer for the project / @Kludex, but IMO this is just a temporary fix. This is only required, because `asyncio` is leaking context vars out of tasks.

I'm interested in your use case though: are you setting a "global" context var, i.e. not on a per-request basis, and by starting with an empty context this global context is hidden?

marctc:
Hey @pmeier thanks for quick response!

> I'm interested in your use case though: are you setting a "global" context var, i.e. not on a per-request basis, and by starting with an empty context this global context is hidden?

We hook `context_run` (which is called when `Context.run()` executes) and `context_new_from_vars` (when a new context is created). When a request comes in, we capture the Python context pointer and store it in a BPF map, associating it with the trace. When that same request makes an outgoing HTTP call, we look up the current context pointer to find the parent trace and link the spans together.

Before 0.39, this worked because the request handler inherited whatever context was active. The context pointer we captured at request start was the same one (or a child of it) when the handler made HTTP calls.
After 0.39, with `contextvars.Context().run(...),` each request gets a brand new empty context. So the context pointer we see inside the request handler has no relationship to anything we tracked before - it's a fresh object created just for that request. We can't follow the parent chain because there isn't one.

So we're not setting a "global" contextvar. Instead, we intercept the C functions in libpython that manage context objects and track the raw memory addresses (pointers) of those `PyContext` objects.

Kludex:
I think we should make lifespan a parent task, and then use the context from there.

---

We can also make this fix only for `asyncio`, and leave uvloop out of it, since it's what people will use in production.

marctc:
Hey thanks also for quick response!

> We can also make this fix only for `asyncio`, and leave uvloop out of it, since it's what people will use in production.

Unless I'm doing something wrong or I missunderstood your message, it's also broken for uvicorn + uvloop. I never tested with uvloop before and I had to tweak the code, but it works for versions prior 0.39.

Kludex:
> Hey thanks also for quick response!
> 
> > We can also make this fix only for `asyncio`, and leave uvloop out of it, since it's what people will use in production.
> 
> Unless I'm doing something wrong or I missunderstood your message, it's also broken for uvicorn + uvloop. I never tested with uvloop before and I had to tweak the code, but it works for versions prior 0.39.

What I meant is to "revert" this PR for uvloop, since the problem this PR tried to solve was not an issue with uvloop.

Kludex:
By "revert" here I mean wrap the line that runs the app, check what event loop is running, and apply a behavior accordingly.

marctc:
I see, I understand now, thanks for clarifying!

tubarao312:
Hi, this fix has actually broken some metaprogramming features I was relying on - I was accessing a single variable for multiple uvicorn apps in my monorepo and then changing the variable depending on the context it was in. This context is no longer accessible inside ASGI function calls. Any idea what I should do?

Pinging @Kludex since this is a rather old thread.

Also I realize this might seem like an insane use of context vars but from my point of view it made a lot of sense that I could use a FastAPI lifespan to inject whatever I wanted into the context. Classic Hyrum's Law 😅:
> With a sufficient number of users of an API, it does not matter what you promise in the contract: all observable behaviours of your system will be depended on by somebody.

Kludex:
@pmeier I know it's just one user complaining (for now), but I think it's a fair point. I'm inclined to revert the changes (given that I've seen progress on the CPython side for 3.15).

> [...] seem like an insane use of context vars

It does, but my reasoning was that the current bug in CPython (and the non-predictability behavior) overweights that.

pmeier:
Yeah, I get it. TBH when I started this I was 90%+ sure that this was a bug in `uvicorn` rather than CPython. And after I was finally able to reproduce it with just CPython, I just wanted "closure".

But I guess your reasoning to revert is sound. We just need to make sure to also push this to downstream projects. For example, I originally stumbled over this bug when instrumenting with opentelemetry. And after my fix here was merged and released, I posted there that the bug reports can be closed: https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2852#issuecomment-3696670058

Kludex:
I've reverted this PR and merged https://github.com/Kludex/uvicorn/pull/2912.

To have this behavior, you need to opt-in with `--reset-contextvars`.

---

I'm sorry for the delay. I'll post a message on related closed issues.

PR #2760, opened 2025-11-12
Allow handling proxy headers from AF_UNIX sockets

bemoody:
# Summary

The `--forwarded-allow-ips` option allows uvicorn to identify "trusted" proxies by their IP address.  However, if a proxy is connecting to uvicorn via an AF_UNIX socket (using `uvicorn --uds`), then it has no IP address.  We still need a way to identify that the proxy is trusted in this case.

To allow middleware and applications to reliably tell that they are talking to an AF_UNIX peer, the ASGI spec says that `scope["server"]` should be set to `[path, None]`.  (This was added by ASGI *spec_version* 2.2; earlier versions didn't mention Unix sockets.  uvicorn claims to support *spec_version* 2.3.)

Then, in uvicorn's built-in middleware, treat an AF_UNIX peer as trusted if `--forwarded-allow-ips` includes the literal string `unix:`.

Discussion thread: https://github.com/Kludex/uvicorn/discussions/2743

# Checklist


bemoody:
Added a couple of fixes for type-checking.  The thoroughness of the CI in this project is impressive; kudos.

bemoody:
Note that the first patch here is equivalent in essence to https://github.com/Kludex/uvicorn/pull/2561 .  I didn't see that pull request and would be happy to rebase on top of that if you prefer.

PR #2776, opened 2025-12-25
Fix typo: error_occured -> error_occurred

t-kawasumi:
# Summary

Fix typo in variable name: `error_occured` → `error_occurred` in `LifespanOn` class.

## Changes
- `uvicorn/lifespan/on.py`: Fixed attribute name
- `tests/test_lifespan.py`: Updated test assertions

## Note
This changes a public attribute name on `LifespanOn` class.
I searched for external usage but couldn't find any code outside of uvicorn that accesses this attribute. However, please let me know if you have any concerns about backward compatibility.

# Checklist

### DIFF
--- a/tests/test_lifespan.py
+++ b/tests/test_lifespan.py
@@ -98,7 +98,7 @@ def test_lifespan_auto_with_error():
         lifespan = LifespanOn(config)
 
         await lifespan.startup()
-        assert lifespan.error_occured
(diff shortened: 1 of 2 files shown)


Kludex (approved):

Kludex:
This is not a breaking change, it's okay.

I think I put this fix around 10x in other PRs and never merged them. 😅 

Thanks!

PR #2783, opened 2026-01-08
Adds support for the `http.response.trailers` ASGI event.

### LINKED ISSUE #1596: Add support for trailers
### Discussed in https://github.com/encode/uvicorn/discussions/1593

<sup>Originally posted by **sam-kleiner** August  4, 2022</sup>
I would like to add support for [HTTP trailers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer). 

This is already supported in the [h11 library](https://github.com/python-hyper/h11/blob/master/h11/_events.py#L340) so adding this is fairly easy.

I wrote a [POC with tests](https://github.com/encode/uvicorn/compare/master...sam-kleiner:uvicorn:trailers)

Uses for this in include returning hashes for validation, expire info, errors during stream, etc.

Here is the output (using my code) of a curl command getting streaming json where an error happens before all json is sent to the client

```
$ curl -v -N http://localhost:8000 --raw
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET / HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.79.1
> Accept: */*
>
< HTTP/1.1 200 OK
< date: Wed, 03 Aug 2022 22:37:47 GMT
< server: uvicorn
< content-type: application/json
< trailer: X-Error
< Transfer-Encoding: chunked
<
c
{"message":[
8
"bEHWA",
8
"xuyBF",
8
"14vBZ",
8
"f4Jyp",
8
"uadVV",
8
"oTbyL",
8
"aZcZr",
0
x-error: Exception message here

* Connection #0 to host localhost left intact
```
</div>

willjschmitt:
# Summary

This borrows from the initial implementation by @Kludex as an extension library in
https://github.com/Kludex/uvicorn-extensions/tree/main/src/python/uvicorn-trailers/uvicorn_trailers but also implements it for h11 with some slight refactoring in approach.

This implements the trailers ASGI extension. This refactors the http implementations slightly into a more state-machine-like flow, since this now introduces a few more state transitions that are optional and/or could be continued. I maintained the original state attributes, since they were public, for backwards compatibility, but they could be removed if desired, since nothing in `uvicorn` strictly depends upon them that I
  can tell.

I generally made new attributes private, which is maybe stylistically-inconsistent, so happy to revert that if desired.

Fixes #1596 

# Checklist

- [ x] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [ x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ x] I've updated the documentation accordingly.

### DIFF
--- a/docs/concepts/asgi.md
+++ b/docs/concepts/asgi.md
@@ -171,7 +171,48 @@ async def app(scope, receive, send):
     })
 ```
 
----
+### Sending trailers
+
+You can send trailing headers by indicating adding `"trailers": True` to the
+`http.response.start` message and sending `http.response.trailers` messages 
+after the body is completer.
+
+```python
+import asyncio
+
+
+async def app(scope, receive, send):
+    """
+    Send an HTTP response with headers and trailers.
+    """
+    await send({
+        'type': 'http.response.start',
+        'status': 200,
+        'headers': [
+            [b'content-type', b'text/plain'],
+        ],
+        "trailers": True
+    })
+    await send({
+        'type': 'http.response.body',
+        'body': b'Hello, world!',
+    })
+    await send({
+        'type': 'http.response.trailers',
+        'body': b'',
+        'headers': [
+            [b'x-app-status', b'1'],
+        ],
+        'more_trailers': True
+    })
+    await send({
+        'type': 'http.response.trailers',
+        'body': b'',
+        'headers': [
+            [b'x-app-details', b'failed'],
+        ]
+    })
+```
 
 ## Why ASGI?
 
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -1118,3 +1118,60 @@ async def test_header_upgrade_is_websocket_depend_not_installed(
     assert msg in caplog.text
     assert b"HTTP/1.1 200 OK" in protocol.transport.buffer
     assert b"Hello, world" in protocol.transport.buffer
+
+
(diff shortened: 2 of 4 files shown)


willjschmitt:
This is to implement the trailers feature in `uvicorn` itself (#1596)

willjschmitt:
Something that isn't supported here (or in the original impl, I think), and it's not clear if it should be, is if trailers should be sendable even if the last message was `http.response.body` and `"more_body": True`. This would be useful in exception handling, where the trailers are related to exception details

Kludex:
> Something that isn't supported here (or in the original impl, I think), and it's not clear if it should be, is if trailers should be sendable even if the last message was `http.response.body` and `"more_body": True`. This would be useful in exception handling, where the trailers are related to exception details

No, it should not. You should use proper exception handling for it.

Kludex:
> This implements the trailers ASGI extension. This refactors the http implementations slightly into a more state-machine-like flow, since this now introduces a few more state transitions that are optional and/or could be continued. I maintained the original state attributes, since they were public, for backwards compatibility, but they could be removed if desired, since nothing in uvicorn strictly depends upon them that I
can tell.

I would prefer to have a simpler setup, would it be possible to remove those?

PR #2786, opened 2026-01-10
Accept PathLike log_config, normalize log_level case, raise helpful PyYAML error

nuno-andre:
# Summary

This PR improves the following aspects of the logging configuration:
- Allows passing the path as a `os.PathLike` (e.g. `pathlib.Path`) object.
- Provides further information when PyYAML is not available.
- Normalizes logging level strings to lowercase, enabling the use of Python’s `logging` uppercase literals.

# Checklist


nuno-andre:
Those failed checks seem unrelated to this PR.

Kludex:
I would prefer to have each change in its own PR, but I think all changes here make sense. I'll be open those, and co-author you @nuno-andre .

Thanks for the PR. :)

PR #2793, opened 2026-01-26
Support HTTP/2

Kludex:


pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
 [project.optional-dependencies]
 standard = [
     "colorama>=0.4; sys_platform == 'win32'",
+    "h2>=4.0.0",
```
Kludex: Let's bump this already, 4.0.0 is from 2020. Let's bump it to 4.2.0.

Kludex:
@codex review

tests/protocols/test_http.py
```diff
@@ -1210,6 +1248,71 @@ async def test_header_upgrade_is_not_websocket_depend_installed(
...
+    ]
+)
+
+
+@skip_if_no_h2
+async def test_alpn_h2_upgrade(http_protocol_cls: type[HTTPProtocol]):
+    """Test that ALPN h2 negotiation switches to H2Protocol."""
+    app = Response("Hello, world", media_type="text/plain")
+    protocol = get_connected_protocol(app, http_protocol_cls, alpn_protocol="h2", http2=True)
+    assert isinstance(protocol.transport.get_protocol(), H2Protocol)
```
Kludex, line 1130: Isn't this duplicated?

tests/protocols/test_http2.py
```diff
@@ -0,0 +1,789 @@
...
+    def resume_reading(self):  # pragma: no cover
+        self.read_paused = False
+
+    def is_closing(self):
+        return self.closed
+
+    def clear_buffer(self):
+        self.buffer = b""
+
+    def set_protocol(self, protocol: asyncio.Protocol): ...
```
Kludex: We shouldn't duplicate those. Maybe we should have a helper module in the tests e.g. http_utils.py

Kludex:
Addressed the open review comments and ran several rounds of `codex review --base main -c model_reasoning_effort=high`. Highlights from the new commits on top of `82c3add`:

- **h2c upgrade validation** - reject the upgrade (no 101) when `HTTP2-Settings` is missing, duplicated, empty, fails strict base64url validation, fails `H2Connection.initiate_upgrade_connection` (catches `InvalidSettingsValueError`), or when the request carries a body.
- **Request scope** - drop duplicate `Host` headers when `:authority` is present, fall back to `Host` when `:authority` is absent, accumulate tokens across multiple `Connection` header fields, forward bytes the peer shipped after the upgrade request to the new H2Protocol.
- **Response correctness** - preserve ASGI header bytes verbatim (no Latin-1 -> UTF-8 round-trip), strip the HTTP/2 hop-by-hop set (`transfer-encoding`, `keep-alive`, `upgrade`, `proxy-connection`, `te` != `trailers`), validate Content-Length and reject 1xx/204/304 + HEAD bodies.
- **Lifecycle** - `H2Protocol.shutdown` now refuses new streams while draining and wakes any cycle blocked in `receive`. `on_response_complete` runs the keep-alive / pending-shutdown housekeeping after deferred flow-controlled cleanup. Stream resets clear pending bytes and resume backpressure. Idle PING/SETTINGS frames re-arm the keep-alive timer.
- **Backpressure** - track an HTTP/2-specific `h2_write_paused` `asyncio.Event` independent of the transport `pause_writing`. The cycle's `send` races the drain against a per-cycle `disconnect_event` so resets/disconnects unblock paused tasks. Reads only resume once every active stream's body buffer is below the high-water mark.
- **Tests** - moved the duplicated `MockTransport` / `MockLoop` / fixtures into `tests/protocols/http_utils.py`, added parametrised malformed-h2c-upgrade tests, and covered the new shutdown / backpressure / header / Content-Length paths. 1078 tests, 100% coverage, ruff + mypy clean.

Pyproject already has `h2>=4.2.0`.

## AI Disclaimer

PR #2799, opened 2026-01-30
Revert "explicitly start ASGI run with empty context (#2742)"

Kludex:


uvicorn/protocols/http/httptools_impl.py
```diff
@@ -289,12 +288,7 @@ def on_headers_complete(self) -> None:
...
             # Standard case - start processing the request.
-            # For the asyncio loop, we need to explicitly start with an empty context
-            # as it can be polluted from previous ASGI runs.
-            # See https://github.com/python/cpython/issues/140947 for details.
-            task = contextvars.Context().run(self.loop.create_task, self.cycle.run_asgi(app))
-            # TODO: Replace the line above with the line below for Python >= 3.11
-            # task = self.loop.create_task(self.cycle.run_asgi(app), context=contextvars.Context())
+            task = self.loop.create_task(self.cycle.run_asgi(app))
             task.add_done_callback(self.tasks.discard)
             self.tasks.add(task)
```
nggit, line 293: Hello. Not meant to do a review. But I see this order can be improved since the context of discussion is memory leak.
```
            task = self.loop.create_task(self.cycle.run_asgi(app))
            task.add_done_callback(self.tasks.discard)
            <-- worst case: what if the task done early right here
            self.tasks.add(task)  # will never be removed
```

Kludex: > what if the task done early right here

For example?

---

This PR didn't change this tho

nggit:
```python
            task = self.loop.create_task(self.cycle.run_asgi(app))
            task.add_done_callback(self.tasks.discard)

            task.cancel()
            await asyncio.sleep(10)
            <-- worst case: what if the task done early right here

            self.tasks.add(task)  # will never be removed, discard was called before this

```
I can't provide the meaningful example. It's probably just my paranoid. But asyncio the example agreed to "add_done_callback" after the task was already inside "set()" https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

> This PR didn't change this tho

Yes. I just wonder why it was written that way.

PR #2802, opened 2026-02-01
Update sdist include list in pyproject.toml

iudeen:
# Summary

This pull request makes a minor adjustment to the build configuration in `pyproject.toml`. 

Removed `/requirements.txt` from the `include` list for the source distribution build target in `pyproject.toml` as the project no longer uses it
# Checklist


Kludex (approved):

Kludex:
Thanks! :)

PR #2808, opened 2026-02-05
added possibility to run server on uds socket AND host/port pair

vprysiaz:
# Summary
this change is aimed to modify server startup function so it allows to listed incoming request both on uds socket and host/port pair. I saw many project when this is required. for example when Nginx is configured to redirect incoming request to uvicorn serving e.g. fastAPI via UDS socket and at the same time to serve host/port for internal microservices. 
this change is running on one of my projects for a while without any issues.  

# Checklist

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -156,9 +156,9 @@ class Server:
             os.chmod(config.uds, uds_perms)
             assert server.sockets is not None  # mypy
             listeners = server.sockets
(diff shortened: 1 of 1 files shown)


Kludex:
I'm more inclined to go with https://github.com/Kludex/uvicorn/pull/2486, but adapting that PR to accept `--bind`.

I'm working on it.

PR #2810, opened 2026-02-10
Use string for `SERVER_PORT` and `stderr` for `wsgi.errors` per PEP3333

veeceey:
## Summary

Two issues in `build_environ()` violated the WSGI spec ([PEP 3333](https://peps.python.org/pep-3333/)):

- **`SERVER_PORT` was set to an integer instead of a string.** PEP 3333 requires all CGI-derived environ variables to be native strings. This causes `TypeError` in any WSGI application that performs string operations on `SERVER_PORT` (e.g. `SERVER_NAME + ":" + SERVER_PORT` in URL reconstruction). Changed to `str(server[1])`.

- **`wsgi.errors` was set to `sys.stdout` instead of `sys.stderr`.** PEP 3333 specifies `wsgi.errors` as "an output stream to which error output text can be written." The standard convention is `sys.stderr`. Using `sys.stdout` causes WSGI error messages to mix with regular program output, which can corrupt structured output and confuse log aggregators.

## Changes

- `uvicorn/middleware/wsgi.py`: Convert `SERVER_PORT` to string, use `sys.stderr` for `wsgi.errors`
- `tests/middleware/test_wsgi.py`: Add three new tests verifying PEP 3333 compliance

## Test plan

- [x] All existing WSGI tests pass
- [x] New tests verify `SERVER_PORT` is a string (with both explicit and default server)
- [x] New test verifies `wsgi.errors` is `sys.stderr`
- [x] Full test suite passes (432 passed, 20 skipped for missing optional deps)
- [x] Linting passes (`ruff check`)


Kludex:
I was going to merge it, but actually... The wsgi implementation is already deprecated. I'd prefer to leave it as is.

Thanks anyway.

PR #2818, opened 2026-02-15
Add `--bind` / `-b` option for multiple socket bindings

Kludex:
Add a `--bind` / `-b` option for binding to multiple addresses simultaneously. Accepts `HOST:PORT`, `[HOST]:PORT` (IPv6), `unix:PATH`, and `fd://NUM` formats. Mutually exclusive with `--host`/`--port`/`--uds`/`--fd`, fully backwards compatible.

Supersedes #2486, which changed `--host` to accept multiple values - that approach altered the type signature of `host` from `str` to `list[str]`, breaking the existing API. A dedicated `--bind` option avoids that, and also covers mixed binding types (e.g. TCP + UDS) that multi-host alone can't express. This follows the pattern established by Gunicorn and Hypercorn.

Closes #2808 as well, which hardcoded UDS + host/port together rather than offering a general mechanism.


Kludex:
@vvanglro since you've worked on https://github.com/Kludex/uvicorn/pull/2486, what do you think about this PR?

vvanglro (approved):

vvanglro:
LGTM!

Kludex:
> LGTM!

Thanks for the review!

vprysiaz:
are you about to merge this PR? I see it have been reviewed a while ago

PR #2819, opened 2026-02-15
Use `context=` kwarg on `loop.create_task` for Python 3.11+

Kludex:
## Summary
- On Python 3.11+, pass `context=contextvars.Context()` directly to `loop.create_task` instead of using `Context().run(loop.create_task, ...)`. This was already noted as a TODO.
- Falls back to `Context().run()` on Python 3.10.

## Context
Investigating #2794. Could not reproduce the memory leak on Python 3.13.11/Linux with ALB-like traffic patterns (50k requests, abrupt closes, concurrent connections), but this change avoids the `Context.run()` indirection which may behave differently under certain workloads.

## Test plan
- [ ] All existing HTTP protocol tests pass (128/128)
- [ ] Reporter from #2794 tests with this branch to check if it resolves their memory leak

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -226,7 +226,7 @@ class MockLoop:
         self._tasks: list[asyncio.Task[Any]] = []
         self._later: list[MockTimerHandle] = []
 
-    def create_task(self, coroutine: Any) -> Any:
+    def create_task(self, coroutine: Any, **kwargs: Any) -> Any:
         self._tasks.insert(0, coroutine)
         return MockTask()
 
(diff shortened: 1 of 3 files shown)


tests/protocols/test_http.py
```diff
@@ -226,7 +226,7 @@ def __init__(self):
         self._tasks: list[asyncio.Task[Any]] = []
         self._later: list[MockTimerHandle] = []
 
-    def create_task(self, coroutine: Any) -> Any:
+    def create_task(self, coroutine: Any, **kwargs: Any) -> Any:
```
```suggestion
    def create_task(self, coroutine: Any) -> Any:
```

Kludex: I need to revert this.

PR #2822, opened 2026-02-17
Fix X-Forwarded-For not properly handling ports

### LINKED ISSUE #2789: X-Forwarded-For doesn't properly handle ports
### Initial Checks

### Discussion Link

```Text

```

### Description

Resulting entries like:
```
1.1.1.1:1024:0
```

### Example Code

```Python

```

```Text
```

takeda:
# Summary

Fixes https://github.com/Kludex/uvicorn/issues/2789

X-Forwarded-For did not account for setups when port was being provided. This patch updates the code to understand the format containing port.

# Checklist

- [N/A] I've updated the documentation accordingly.

### DIFF
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import ipaddress
 from typing import TYPE_CHECKING
 
 import httpx
@@ -30,6 +31,8 @@ async def default_app(scope: Scope, receive: ASGIReceiveCallable, send: ASGISend
         client_addr = "NONE"  # pragma: no cover
     else:
         host, port = client
+        if ipaddress.ip_address(host).version == 6:
+            host = f"[{host}]"
         client_addr = f"{host}:{port}"
 
     response = Response(f"{scheme}://{client_addr}", media_type="text/plain")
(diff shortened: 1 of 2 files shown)


hrv-dys (commented):
This is the regex-based alternative to #2823 for the same X-Forwarded-For port parsing problem.

Notes on this approach:

1. **Compiled regex patterns** — `_IPV6_HOST_PORT` and `_IPV4_HOST_PORT` as module-level `Final` constants compiled once. For high-throughput servers where every request goes through this middleware, compiled regex is efficient.

2. **`_get_host_port()`** — the walrus operator pattern `if (match := _IPV6_HOST_PORT.match(entry)) or (match := _IPV4_HOST_PORT.match(entry))` is compact. Tries IPv6 first (bracketed), falls back to IPv4. Returns `(ip, port_int)` or `(entry, 0)` if no match. Clean.

3. **API change: `get_trusted_client_host` → `get_trusted_client_host_port`** — returns `tuple[str, int]` instead of `str`. This is a cleaner contract since the method now extracts both host and port in one pass, avoiding the double-parse that #2823 does (once in `get_trusted_client_host` for trust checking, once in the middleware for the client tuple).

4. **The test app wraps IPv6 in brackets** for display (`f"[{host}]"`) which is the standard representation. That's a nice touch for test readability.

5. **Compared to #2823**: This PR changes the internal API (`get_trusted_client_host_port`), which is cleaner but a bigger diff. PR #2823 keeps the API surface the same and adds parsing at the call sites. Both approaches are correct and pass all 245 tests. If this method is only used internally (which it is — it's on `_TrustedHosts`), the API change is fine.

6. **One concern with the regex approach**: `_IPV4_HOST_PORT` is `r"(?P<ip>[\d.]+):(?P<port>\d+)"` — this would also match malformed inputs like `999.999.999.999:80`. The trust checking later handles validation (the ipaddress module will reject bad IPs), so this isn't a security issue, but it's worth noting that the regex is intentionally permissive and relies on downstream validation.

All 245 tests pass (8 skipped, pre-existing).

PR #2826, opened 2026-02-20
Ensure WSGI SERVER_PORT is a string per PEP 3333

bysiber:
The WSGI middleware sets `SERVER_PORT` directly from the ASGI scope's server tuple:

```python
```

`server[1]` is `int | None` in the ASGI spec, but PEP 3333 requires all CGI variables to be strings. WSGI apps or frameworks that do string operations on `SERVER_PORT` (e.g., concatenation) would get a `TypeError`.

Additionally, when serving over a Unix domain socket, the ASGI scope can have `server = ("path", None)`, which would set `SERVER_PORT = None`.

This converts the port to a string, falling back to an empty string when the port is `None`.


Kludex:
This module is already deprecated.

PR #2833, opened 2026-03-01
Add `ssl_context_factory` parameter for custom SSL contexts

### LINKED ISSUE #806: Support custom SSL Context

Br1an67:
# Summary

Add an `ssl_context_factory` parameter to `Config` and `uvicorn.run()` that accepts a callable returning an `ssl.SSLContext`. This allows users to provide fully customized SSL contexts (e.g., mutual TLS, custom ciphers, certificate pinning) without being limited to the existing `ssl_*` parameters.

The factory approach was suggested by @Kludex in https://github.com/encode/uvicorn/issues/806#issuecomment-1307413547 and solves the `SSLContext` pickling issue when using `--reload` or `--workers > 1`, since the factory callable is picklable while `SSLContext` objects are not.

**Usage:**
```python
import ssl
import uvicorn

def ssl_context_factory() -> ssl.SSLContext:
    ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
    ctx.load_cert_chain("cert.pem", "key.pem")
    return ctx

uvicorn.run("app:app", ssl_context_factory=ssl_context_factory, reload=True)
```

When `ssl_context_factory` is provided, it takes priority over individual `ssl_*` parameters.

Closes #806

# Checklist


hrv-dys (commented):
Clean addition. Having `ssl_context_factory` alongside the existing `ssl_certfile`/`ssl_keyfile` parameters gives users the flexibility to construct their own SSL contexts without uvicorn having to grow more and more TLS knobs.

Notes:

1. **The `is_ssl` property update** — `bool(self.ssl_keyfile or self.ssl_certfile or self.ssl_context_factory)` is correct. This ensures the server binds with TLS when a factory is provided, even without explicit cert/key paths.

2. **Load order in `config.load()`** — checking `ssl_context_factory is not None` first, before falling through to the cert/key path, means the factory takes precedence. This is the right priority — if someone provides both, the factory should win (they presumably know what they're doing). Though you might want to emit a warning if both are provided, to avoid silent surprises.

3. **The factory is called exactly once during `config.load()`** — the result is stored in `self.ssl`. This means the SSL context is created at startup, not per-connection. For most use cases (e.g., loading a cert chain), this is correct. If someone needed per-connection context rotation (e.g., for cert renewal without restart), they'd need a different mechanism, but that's out of scope here.

4. **The test** creates a standard `ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)`, loads a cert chain, and verifies the server responds with 204. Simple and effective.

5. **No CLI flag** — the factory is only available via the programmatic API (`uvicorn.run()`), not via CLI. This makes sense since you can't pass a callable on the command line.

All 5 SSL tests pass.

Kludex:
- Closed by https://github.com/Kludex/uvicorn/pull/2920

Thanks for the PR.

PR #2839, opened 2026-03-09
fix: allow Ctrl+C to interrupt hanging lifespan#2752

theseriff:
# Summary

reopen #2752

# Checklist

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -12,7 +12,7 @@ import socket
 import sys
 import threading
 import time
-from collections.abc import Generator, Sequence
+from collections.abc import Coroutine, Generator, Sequence
 from email.utils import formatdate
 from types import FrameType
 from typing import TYPE_CHECKING, TypeAlias
@@ -62,6 +62,7 @@ class Server:
         self.should_exit = False
         self.force_exit = False
         self.last_notified = 0.0
+        self.exit_event = asyncio.Event()
 
         self._captured_signals: list[int] = []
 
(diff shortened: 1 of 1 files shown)


hrv-dys (commented):
This solves a genuine pain point. If the lifespan handler hangs (broken app, stuck dependency, etc.), a second Ctrl+C should force-exit instead of also hanging.

The approach is solid — wrapping lifespan calls in `_wait_safe_lifespan()` with an `asyncio.Event` that gets set from the signal handler via `call_soon_threadsafe`. A few observations:

1. **`call_soon_threadsafe` in `handle_exit`** — this is correct because signal handlers run in the main thread but the event loop may be running in a different context. `loop.call_soon_threadsafe(self.exit_event.set)` ensures the event is set safely without races.

2. **The `finally` block in `_wait_safe_lifespan`** properly cancels whichever task didn't finish first. If the exit event fires, the lifespan task gets cancelled; otherwise the waiter is cleaned up. This prevents leaked tasks.

3. **One thing to consider**: when the lifespan task gets cancelled mid-shutdown, the app's `lifespan.shutdown` handler may not have completed cleanup (e.g., closing database pools, flushing buffers). This is inherent to a force-exit though — the user explicitly chose to interrupt, so partial cleanup is the expected tradeoff. Worth documenting in the commit message or a comment if it isn't already.

4. **The `pragma: no cover` on the `waiter_task.done()` branch** — this makes sense since it's hard to reliably trigger a second SIGINT in a test, but it's the more critical code path. If there's a way to test it (maybe by directly calling `exit_event.set()` instead of sending a signal), that'd strengthen confidence.

One thing this codebase has historically been careful about: lifespan shutdown failures don't automatically propagate to the server as an exit signal. This PR doesn't change that contract — it just provides an escape hatch when the lifespan call itself never returns. That's the right scope.

Tests all pass (16 lifespan + 10 main = 26/26).

PR #2840, opened 2026-03-11
Add Pathsend extension support

Archmonger:
# Summary

Currently a draft PR, so linting, type-checking, and coverage aren't expected to be 100% yet. Awaiting maintainer(s) input before proceeding to finalizing.

Demonstrates the level of effort required to implement pathsend support (which uses `loop.sendfile`).

# Checklist

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -2,10 +2,12 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import os
 import socket
 import threading
 import time
 from collections.abc import Callable
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, TypeAlias
 
 import pytest
@@ -225,6 +227,7 @@ class MockLoop:
     def __init__(self):
         self._tasks: list[asyncio.Task[Any]] = []
         self._later: list[MockTimerHandle] = []
+        self.sendfile_calls: list[dict[str, Any]] = []
 
     def create_task(self, coroutine: Any) -> Any:
         self._tasks.insert(0, coroutine)
@@ -237,6 +240,27 @@ class MockLoop:
 
     async def run_one(self):
         return await self._tasks.pop()
+
+    async def sendfile(
+        self,
+        transport: MockTransport,
+        file: Any,
+        offset: int = 0,
+        count: int | None = None,
+        *,
+        fallback: bool = True,
+    ) -> int:
+        self.sendfile_calls.append(
+            {
+                "offset": offset,
+                "count": count,
+                "fallback": fallback,
+            }
+        )
+        file.seek(offset)
+        data = file.read() if count is None else file.read(count)
+        transport.write(data)
+        return len(data)
 
     def run_later(self, with_delay: float) -> None:
         later: list[MockTimerHandle] = []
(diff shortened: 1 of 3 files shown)


hrv-dys (commented):
This is a well-structured implementation of the `http.response.pathsend` ASGI extension. Using `loop.sendfile()` with `fallback=False` to leverage `os.sendfile` is the right approach for zero-copy file serving.

Detailed notes:

1. **Scope extension advertisement** — only advertising `http.response.pathsend` when both `os.sendfile` and `loop.sendfile` exist (`self.supports_pathsend`) is correct. This avoids promising capabilities the platform can't deliver. The test `test_pathsend_scope_extension_advertised` confirms the app can detect this.

2. **Security: `os.path.isabs(path)` check** — this rejects relative paths, which is important since the ASGI app controls the path string. However, an absolute path can still contain traversal sequences like `/var/www/../../etc/passwd`. The ASGI spec likely leaves path validation to the framework (e.g., Starlette), but it might be worth documenting that uvicorn trusts the app to provide a safe path.

3. **Content-Length validation** — `if file_size != expected_content_length` is a good safety check. Without this, a mismatch would cause the client to either hang (waiting for more bytes) or receive a truncated response.

4. **The mutual exclusion between `response.body` and `response.pathsend`** — tracked via `response_body_sent` and `response_pathsend_sent` flags with clear error messages. This prevents mixing, which could produce garbled responses.

5. **HEAD request handling** — `if self.scope["method"] == "HEAD": return 0` skips the actual sendfile. Correct, since HEAD responses must not include a body.

6. **Chunked encoding rejection** — `if self.chunked_encoding: raise RuntimeError("Path send requires a 'Content-Length' header")`. This is important — `sendfile` writes raw bytes to the transport and doesn't know about chunked framing. If chunked encoding were active, the client would see raw file bytes without chunk headers and fail to parse the response. This aligns with a known constraint in this codebase: always provide content-length when the app provides raw body data, rather than relying on chunked transfer encoding.

7. **httptools-only for now** — since h11 manages its own state machine and doesn't expose the raw transport for sendfile, limiting this to httptools makes sense. Worth noting in the PR description if it isn't already.

8. **The `MockLoop.sendfile` addition in tests** — clean mock that records calls and verifies `fallback=False`. Captures the essential behavior without needing real file I/O through the event loop.

All 131 HTTP tests pass (127 passed, 4 pre-existing skips). The 3 new pathsend tests pass cleanly.

Kludex:
@Archmonger Is the performance here worth the complexity?

Archmonger:
The performance benefit is largely going to depend on the loop's implementation of `loop.sendfile`, but it has a lot of potential for significant speed gains.

Within `loop.sendfile` it could offload the sendfile syscall onto a separate thread (similar to the WSGI's FileWrapper on gunicorn). This would achieve performance several orders of magnitude faster than Python file chunking. 

For context, my overarching goal here is to make it possible to serve files directly within ASGI in a high-bandwidth setting (e.g. video streaming) without mandating usage of external services such as Nginx, especially for household self-hosted applications.

To push things along, I will also be PRing [os-native loop.sendfile support](https://github.com/MagicStack/uvloop/issues/228) to uvloop. At a glance it also seems possible to extend this implementation to `zerocopysend`, but I don't believe it's beneficial to maintain two interfaces that would ultimately serve the same purpose.

Archmonger:
Just checking in - should I put effort into pushing this PR past the draft stage, or should I close this out?

Archmonger:
Closing due to lack of response from maintainers. Will assume there is not an interest in this feature, unless told otherwise.

Kludex:
A bit of patience please.

For reference, https://github.com/Kludex/zuvloop supports `sendfile`.

PR #2854, opened 2026-03-17
use  `bytearray` where possible

albertedwardson:
# Summary

I was planning to make the changes that were made in #2845, but I abandoned them because I was too lazy to write benchmarks and defend my changes.

Let's see what codspeed will show 

# Checklist

### DIFF
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -189,33 +189,30 @@ class HttpToolsProtocol(asyncio.Protocol):
 
         self.connections.discard(self)
         method = self.scope["method"].encode()
-        output = [method, b" ", self.url, b" HTTP/1.1\r\n"]
+        output = bytearray(method + b" " + self.url + b" HTTP/1.1\r\n")
         for name, value in self.scope["headers"]:
-            output += [name, b": ", value, b"\r\n"]
-        output.append(b"\r\n")
+            output += name + b": " + value + b"\r\n"
+        output += b"\r\n"
         protocol = self.ws_protocol_class(  # type: ignore[call-arg, misc]
             config=self.config,
             server_state=self.server_state,
             app_state=self.app_state,
         )
         protocol.connection_made(self.transport)
-        protocol.data_received(b"".join(output))
+        protocol.data_received(bytes(output))
         self.transport.set_protocol(protocol)
 
     def send_400_response(self, msg: str) -> None:
-        content = [STATUS_LINE[400]]
+        content = bytearray(STATUS_LINE[400])
         for name, value in self.server_state.default_headers:
-            content.extend([name, b": ", value, b"\r\n"])  # pragma: full coverage
-        content.extend(
-            [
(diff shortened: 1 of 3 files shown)


albertedwardson:
well, in theory, it will also reduce memory usage a tiny bit

uvicorn/protocols/http/httptools_impl.py
```diff
@@ -189,33 +189,30 @@ def handle_websocket_upgrade(self) -> None:
 
         self.connections.discard(self)
         method = self.scope["method"].encode()
-        output = [method, b" ", self.url, b" HTTP/1.1\r\n"]
+        output = bytearray(method + b" " + self.url + b" HTTP/1.1\r\n")
         for name, value in self.scope["headers"]:
-            output += [name, b": ", value, b"\r\n"]
-        output.append(b"\r\n")
+            output += name + b": " + value + b"\r\n"
```
albertedwardson, line 194: in toy like microbenchmarks += is about 20% faster than extend (on cpython 3.14.2)

hrv-dys (commented):
Replacing `list[bytes]` + `b"".join()` with `bytearray` accumulation is a nice follow-up to the request body accumulation change in #2845. Fewer intermediate allocations, and `bytearray.__iadd__` avoids creating new list entries.

Observations:

1. **The `send_500_response` changes in `websockets_impl.py` and `httptools_impl.py`** — using `bytearray()` with implicit bytes literal concatenation is clean. The final `self.transport.write(content)` works because `transport.write` accepts any bytes-like object.

2. **The `handle_websocket_upgrade` changes** — `bytearray(event.method + b" " + event.target + b" HTTP/1.1\r\n")` followed by `+=` for each header, then `protocol.data_received(bytes(output))` — the final `bytes()` conversion is necessary since the websocket protocol's `data_received` may not accept `bytearray` directly. Good attention to detail there.

3. **The response header construction in `httptools_impl.py`** (`send()` method) — this is the hottest path since it runs for every response. Switching from `content = [STATUS_LINE[status_code]]` + `content.extend(...)` + `b"".join(content)` to `content = bytearray(STATUS_LINE[status_code])` + `content += ...` avoids the join step entirely. The `transport.write(content)` at the end works since asyncio transports accept bytearray.

4. **Chunked encoding body writes** — `bytearray(b"%x\r\n" % len(body) + body + b"\r\n")` creates one allocation instead of a 3-element list. For large bodies this should be measurable.

One thing this codebase has historically been careful about: always providing `content-length` rather than relying on chunked encoding for WebSocket responses. The `send_500_response` changes maintain that — the content-length header is still computed and included. 

Tests pass cleanly (pre-existing websockets deprecation failures only).

Kludex:
@albertedwardson I'll close this one as well. Again, if we can have numbers to sustain, and it's meaningful, I'm happy to merge this kind of PR, otherwise I don't think we should merge this.

Sorry if this is disappointing.

albertedwardson:
yeah, no problem, totally understand

PR #2855, opened 2026-03-17
don't call `typing.cast` in runtime

albertedwardson:
# Summary

`typing.cast` is a python function and it does nothing more than adds a few extra nanoseconds in runtime

the solution is ugly, but it saves a bit resources

I hope CPython will optimize this somehow in the future.

this workaround once significantly improved performance: https://github.com/pylint-dev/astroid/pull/2500

# Checklist

### DIFF
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -5,7 +5,7 @@ import contextvars
 import http
 import logging
 from collections.abc import Callable
-from typing import Any, Literal, cast
+from typing import TYPE_CHECKING, Any, Literal, cast
 from urllib.parse import unquote
 
 import h11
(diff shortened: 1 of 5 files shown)


hrv-dys (commented):
Makes sense. `typing.cast()` is a no-op at runtime in terms of type safety — it just returns its argument — but it still has function call overhead. Wrapping them in `if TYPE_CHECKING` removes that cost from the hot path.

A couple of notes:

1. **The `websockets_impl.py` change for `accepted_subprotocol`** is slightly different from the others — it needs an `else` branch to handle the non-TYPE_CHECKING case since the assignment is semantically important (not just a narrowing). That's handled correctly here with the `else: self.accepted_subprotocol = message.get("subprotocol")`.

2. **The `httptools_impl.py` change for `body`** splits the assignment into two lines under `TYPE_CHECKING`, which is the right approach since `message.get("body", b"")` returns the value in both branches.

3. This is a lot of churn across 5 files for what's ultimately a micro-optimization. The performance win per call is tiny (one function call + one type lookup), but given these run in the per-request hot path, it's a reasonable tradeoff. Would be interesting to see if this shows up in the CodSpeed benchmarks that were recently added.

Tests pass — the 2 failures in websocket tests are pre-existing (websockets library deprecation of `InvalidStatusCode`, not related to this PR).

albertedwardson:
@hrv-dys forwarding response from llm as a code review is not really helpful

Kludex:
I don't think this is worth it... But maybe this is: https://github.com/Kludex/uvicorn/pull/2875

Kludex:
But thanks for the PR!

PR #2856, opened 2026-03-17
use `__slots__` where possible

albertedwardson:
# Summary

this breaks weakrefs, but I believe the objects in question shouldn't be weakref`ed anyway

# Checklist

### DIFF
--- a/uvicorn/protocols/http/flow_control.py
+++ b/uvicorn/protocols/http/flow_control.py
@@ -1,4 +1,5 @@
 import asyncio
+from dataclasses import dataclass, field
 
 from uvicorn._types import ASGIReceiveCallable, ASGISendCallable, Scope
 
@@ -7,12 +8,14 @@ CLOSE_HEADER = (b"connection", b"close")
 HIGH_WATER_LIMIT = 65536
 
 
+@dataclass(slots=True, repr=False, eq=False)
 class FlowControl:
-    def __init__(self, transport: asyncio.Transport) -> None:
-        self._transport = transport
-        self.read_paused = False
-        self.write_paused = False
-        self._is_writable_event = asyncio.Event()
+    _transport: asyncio.Transport
+    read_paused: bool = False
+    write_paused: bool = False
+    _is_writable_event: asyncio.Event = field(default_factory=asyncio.Event)
+
+    def __post_init__(self) -> None:
         self._is_writable_event.set()
 
     async def drain(self) -> None:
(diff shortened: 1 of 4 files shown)


hrv-dys (commented):
The `__slots__` additions make sense for the protocol classes — these are instantiated per-connection, so reducing per-instance memory overhead adds up under load.

A few things:

1. **ServerState as a `@dataclass(slots=True)`** — clean conversion. The `field(default_factory=...)` for the mutable defaults (set, list) is correct and avoids the classic mutable-default-argument footgun.

2. **`FlowControl` as a dataclass** — the `__post_init__` to call `self._is_writable_event.set()` is a nice touch. However, I notice `repr=False, eq=False` on both dataclasses — is that to avoid accidentally comparing or printing these in hot paths? Might be worth a brief comment for future readers.

3. **The `limit_max_requests` change from `@functools.cached_property` to `__init__` assignment** — this is necessary because `__slots__` and `cached_property` don't play well together (cached_property needs `__dict__`). The rename to `_limit_max_requests()` as a regular method called in `__init__` is a reasonable workaround. Just want to flag: this changes the laziness — it's now computed eagerly at construction time. In practice this shouldn't matter since `Server.__init__` is called once, but it's a behavioral change worth noting.

4. **H11Protocol and HttpToolsProtocol slots lists** — I spot-checked these against the `__init__` methods and they look complete. If a future PR adds an attribute and forgets to update `__slots__`, it'll silently fall through to a per-instance `__dict__` (unless I'm wrong and `asyncio.Protocol` doesn't define `__dict__`?). Actually, since the base `asyncio.Protocol` doesn't use `__slots__`, subclasses still get `__dict__`. So the slots here act as a performance hint rather than a strict contract. That's fine, just worth being aware of.

Tests pass with only pre-existing failures (websockets deprecation warnings unrelated to this PR).

albertedwardson:
@hrv-dys unfortunately llm didn't catch that my changes with `@dataclass` changes the `__init__` signature

uvicorn/protocols/http/flow_control.py
```diff
@@ -7,12 +8,14 @@
...
 
+@dataclass(slots=True, repr=False, eq=False)
 class FlowControl:
-    def __init__(self, transport: asyncio.Transport) -> None:
-        self._transport = transport
-        self.read_paused = False
-        self.write_paused = False
-        self._is_writable_event = asyncio.Event()
+    _transport: asyncio.Transport
+    read_paused: bool = False
```
albertedwardson, line 14: init=False or don't use dataclasses at all

albertedwardson:
can be closed in favour of #2697 that already introduces slots

or maybe just deffered until the mentioned one will be merged

Kludex:
@albertedwardson I think if we can find meaningful proof of performance improvement then it's worth it, otherwise I would prefer to close it for the time being. Thanks for the push tho.

Issue #2722, opened 2025-10-09
Bug report candidate: Uvicorn drops body with Upgrade: h2c + chunked"


Kludex:
### Discussed in https://github.com/Kludex/uvicorn/discussions/2637

<div type='discussions-op-text'>

<sup>Originally posted by **jinho7** May 15, 2025</sup>
## Summary

Hi maintainers! I'm reporting a low-level parsing issue that likely affects Uvicorn users under certain HTTP/1.1 upgrade edge cases.

In short: when a client (e.g., Java RestClient) sends an HTTP/1.1 request with:

Upgrade: h2c

Transfer-Encoding: chunked

...Uvicorn (via httptools_impl) ignores the upgrade (correctly), but the request body is never parsed or passed to the app.

## Observed Behavior

Example request:

```
POST / HTTP/1.1
Host: localhost
Upgrade: h2c
Connection: Upgrade, HTTP2-Settings
Transfer-Encoding: chunked
Content-Type: application/json

3\r\nabc\r\n0\r\n\r\n
```

Log output:
```
WARNING: Unsupported upgrade request.
INFO: No request body
WARNING: Invalid HTTP request received.
```

But the request is valid and should fallback to HTTP/1.1 parsing per RFC 7230 Section 6.7.

If I route the request through ngrok or remove Upgrade: h2c, everything works fine. Body is delivered to the ASGI app.

## Root Cause

I've traced the issue and confirmed that:

httptools.HttpRequestParser sets upgrade = True

Uvicorn does not upgrade (e.g., to WebSocket)

But because the upgrade wasn't accepted, on_body() never fires

The result is: the parser never emits the body events.

Upstream discussion and patch suggestion (at httptools):
https://github.com/MagicStack/httptools/issues/124

## Proposal

Once httptools exposes a method like resume_after_upgrade(), I suggest Uvicorn update:

```python
def on_headers_complete(self):
    if self.upgrade and self.upgrade.lower() != b"websocket":
        self.parser.resume_after_upgrade()  # <-- resume HTTP/1.1 body parsing
```

This would allow Uvicorn to properly recover and parse chunked request bodies.

## Why this matters

Java's HttpClient sends Upgrade: h2c by default

Spring Boot RestClient uses chunked transfer by default for object bodies

Uvicorn fails silently unless proxy (e.g., nginx, ngrok) strips headers

This causes confusing bugs in real-world scenarios, despite all parties behaving correctly by spec.

[...]

PR #2857, opened 2026-03-18
Fix request body hang after rejected HTTP Upgrade in httptools

hrv-dys:
## Summary

When httptools receives an HTTP/1.1 request with an unsupported `Upgrade`
header (e.g., `Upgrade: h2c`), it raises `HttpParserUpgrade`. If the upgrade
is rejected by `_should_upgrade()`, the warning is logged but the request
cycle is left in a broken state — `on_message_complete` never fires because
the parser is stuck in "upgrade" mode, so `more_body` remains True and the
ASGI app hangs forever waiting for body data that will never arrive.

The h11 implementation doesn't have this problem because it checks
`_should_upgrade()` within its normal event processing loop without
raising exceptions.

This fix manually signals message completion after a rejected upgrade,
allowing the ASGI app to process the request normally. Per RFC 7230, a server
MAY ignore an Upgrade header it doesn't support and process the request as-is.

**Belief that guided this fix**: don't assume an HTTP/1.1 request with an
Upgrade header will always result in a protocol switch — Uvicorn may reject
the upgrade, leaving the request body unparsed if the parser is not explicitly
resumed (#2722).

## Test plan
- [x] `tests/protocols/test_http.py` passes
- [ ] Manual test with `Upgrade: h2c` header to verify the app receives the request

### DIFF
--- a/uvicorn/protocols/http/httptools_impl.py
+++ b/uvicorn/protocols/http/httptools_impl.py
@@ -181,6 +181,9 @@ class HttpToolsProtocol(asyncio.Protocol):
                 self.handle_websocket_upgrade()
             else:
                 self._unsupported_upgrade_warning()
+                if self.cycle is not None:
+                    self.cycle.more_body = False
+                    self.cycle.message_event.set()
 
     def handle_websocket_upgrade(self) -> None:
         if self.logger.level <= TRACE_LOG_LEVEL:


themavik (commented):
test review body minimal

themavik (commented):
Clearing more_body and signaling message_event after a rejected upgrade unblocks RequestResponseCycle.receive() in httptools_impl. nit: If h11_impl can reject Upgrade the same way, mirror the guard there so both parsers behave the same.

PR #2859, opened 2026-03-18
Use native `context` parameter for `create_task` on Python 3.11+

hrv-dys:
## Summary

Both HTTP protocol implementations (`h11_impl.py` and `httptools_impl.py`) use
`contextvars.Context().run(loop.create_task, ...)` to start ASGI tasks with a
fresh context. Python 3.11 added a `context=` parameter to `create_task()`,
which avoids the extra indirection through `Context.run()`.

This has been a known TODO in the codebase for a while. Under high-concurrency
workloads, the `Context().run()` wrapper adds a small but measurable overhead
per request compared to the native kwarg, since it has to set up and tear down
the context activation around the call.

The change uses `sys.version_info` to branch at runtime — 3.11+ gets the native
kwarg, older versions keep the existing behavior. Coverage pragmas follow the
existing convention in `_types.py` (`py-lt-311` / `py-gte-311` on the branch
lines).

**Bonus**: Also fixes a pre-existing coverage flake on Python 3.14t in
`watchfilesreload.py:44-45` where the `exclude_dir in path.parents` branch
wasn't consistently hit during free-threaded test runs.

**Belief that guided this fix**: the project's own history notes that
`Context().run()` for task creation on Python 3.11+ introduces unnecessary
indirection that may behave differently under high-concurrency workloads
compared to the native `context=` kwarg (#2819).

## Test plan
- [x] `tests/protocols/test_http.py` passes (both h11 and httptools backends)
- [ ] Verify on Python 3.10 and 3.11+ (CI will cover this)


themavik (commented):
Using create_task(..., context=...) on 3.11+ fixes the polluted contextvars issue without the Context.run dance. nit: The pragma labels read backwards—the >=3.11 branch is tagged py-lt-311.

uvicorn/supervisors/watchfilesreload.py
```diff
@@ -41,7 +41,7 @@ def __call__(self, path: Path) -> bool:
                     return True  # pragma: full coverage
 
                 for exclude_dir in self.exclude_dirs:
-                    if exclude_dir in path.parents:
+                    if exclude_dir in path.parents:  # pragma: full coverage
```
Kludex:
```suggestion
                    if exclude_dir in path.parents:
```

Kludex (approved):

PR #2869, opened 2026-03-22
fix: properly handle port in X-Forwarded-For header (#2789)

### LINKED ISSUE #2789: X-Forwarded-For doesn't properly handle ports

r266-tech:
Fixes #2789

## Problem
The proxy headers middleware ignores ports in X-Forwarded-For and always appends ":0". For example, if the header contains "1.1.1.1:1024", the resulting client is ("1.1.1.1:1024", 0) instead of ("1.1.1.1", 1024).

## Solution
Added a `_parse_host_port()` helper that correctly extracts host and port from X-Forwarded-For entries, handling:
- **IPv4 with port**: `1.2.3.4:8080` → `("1.2.3.4", 8080)`
- **IPv4 without port**: `1.2.3.4` → `("1.2.3.4", 0)` (original behavior)
- **IPv6 without port**: `::1` → `("::1", 0)`
- **IPv6 with port**: `[::1]:8080` → `("::1", 8080)`

## Tests
Added `test_proxy_headers_port_in_forwarded_for` with 5 parametrized cases covering the above scenarios. All 253 proxy header tests pass.

### DIFF
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -493,3 +493,31 @@ async def test_proxy_headers_empty_x_forwarded_for() -> None:
         response = await client.get("/", headers=headers)
     assert response.status_code == 200
     assert response.text == "https://127.0.0.1:123"
+
+
+@pytest.mark.anyio
+@pytest.mark.parametrize(
+    ("forwarded_for", "expected"),
+    [
+        # IPv4 with port
+        ("1.2.3.4:1024", "https://1.2.3.4:1024"),
+        # IPv4 without port (original behavior)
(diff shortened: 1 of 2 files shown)


themavik (commented):
Host/port split stops X-Forwarded-For values like 1.2.3.4:8080 from turning into bogus host:port:0. nit: Brackets are stripped for IPv6; worth scanning other proxy tests for whether client scope or URL text should stay bracketed for ::1.

PR #2884, opened 2026-04-03
Implement keepalive for websockets-sansio

MrNaif2018:
# Summary

`--ws-ping-interval` and `--ws-ping-timeout` are ignored by the `websockets-sansio` implementation. Without pings, idle connections get killed by proxies (nginx, caddy, etc).

The legacy `websockets` implementation inherits keepalive from the library.
See https://github.com/python-websockets/websockets/blob/ea164d2fe0cb699dd52d28bfbe98165fb35cb13c/src/websockets/legacy/protocol.py#L1213

The sans-I/O `ServerProtocol` has `send_ping()` but nobody was calling it.

So I ported `keepalive()`, `handle_pong()`, and the start/stop lifecycle from `websockets.asyncio.connection.Connection` into `WebSocketsSansIOProtocol`. 
https://github.com/python-websockets/websockets/blob/ea164d2fe0cb699dd52d28bfbe98165fb35cb13c/src/websockets/asyncio/connection.py#L808
Two new tests connect via raw TCP to verify ping frames are sent and that the server closes with 1011 on pong timeout. Both fail on sansio before this patch, pass after.
I'm not sure whether this is the right approach though, but I felt that websockets has the most mature implementation for it. But probably in sans-io case we need to indeed implement keepalive ourselves.
Fixes #2883 (there is a reproduction script in it too)

cc @aaugustin Sorry for the ping, I saw you review the original websockets-sansio PR. When using sansio implementation, are we supposed to implement keepalive pings ourselves? I just don't know if porting it from `websockets.asyncio.connection.Connection` is the best way or there is some logic we can re-use here already implemented. Thank you!

# Checklist

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -95,8 +95,8 @@ Using Uvicorn with watchfiles will enable the following options (which are other
 * `--ws <str>` - Set the WebSockets protocol implementation. Either of the `websockets` and `wsproto` packages are supported. There are two versions of `websockets` supported: `websockets` and `websockets-sansio`. Use `'none'` to ignore all websocket requests. **Options:** *'auto', 'none', 'websockets', 'websockets-sansio', 'wsproto'.* **Default:** *'auto'*.
 * `--ws-max-size <int>` - Set the WebSockets max message size, in bytes. Only available with the `websockets` protocol. **Default:** *16777216* (16 MB).
 * `--ws-max-queue <int>` - Set the maximum length of the WebSocket incoming message queue. Only available with the `websockets` protocol. **Default:** *32*.
-* `--ws-ping-interval <float>` - Set the WebSockets ping interval, in seconds. Only available with the `websockets` protocol. **Default:** *20.0*.
-* `--ws-ping-timeout <float>` - Set the WebSockets ping timeout, in seconds. Only available with the `websockets` protocol. **Default:** *20.0*.
+* `--ws-ping-interval <float>` - Set the WebSockets ping interval, in seconds. Available with the `websockets` and `websockets-sansio` protocols. **Default:** *20.0*.
+* `--ws-ping-timeout <float>` - Set the WebSockets ping timeout, in seconds. Available with the `websockets` and `websockets-sansio` protocols. **Default:** *20.0*.
 * `--ws-per-message-deflate <bool>` - Enable/disable WebSocket per-message-deflate compression. Only available with the `websockets` protocol. **Default:** *True*.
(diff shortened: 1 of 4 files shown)


aaugustin:
Yes, I confirm that the Sans-I/O layer cannot implement the keepalive Indeed, a Sans-I/O layer doesn't have access to an event loop or even to a concept of time, making it impossible to schedule events.

Porting from `websockets.asyncio.connection.Connection.keepalive` is the best approach. That's how websockets does it internally to support different I/O paradigms.

aaugustin:
The code appears to be in the right ballpark. (Disclaimer: I don't know much about uvicorn's architecture.)

I'm less confident about the tests. `raw_ws_handshake` looks like a LLM reinventing the wheel. (Same disclaimer: I only looked at the diff and I'm not familiar with the context.)

MrNaif2018:
Yeah, the idea with that method was that I can test that after websocket connection got established, we receive PING frames successfully
Otherwise it is hard to verify that ping frames are there and websocket connection won't be dropped by e.g. nginx, I didn't find a better way to test this (what I did when testing on uvicorn manually, I created some kind of a simple proxy which closes idle tcp connection: https://github.com/Kludex/uvicorn/discussions/2883#discussion-9818798. But because uvicorn tests are kind of low level, I probably can't test it this way. And this helper function is simpler in code than that proxy with idle detection functions)

Something like websockets.client.connect(..., ping_interval=None) probably won't help me test this
(also I noticed websockets.client.connect is probably deprecated, but it seems like existing tests use it)

aaugustin:
Yes, we have `websockets.connect` (shortcut for convenience) or `websockets.asyncio.client.connect` (explicit path) now but not more `websockets.client.connect` — [details here](https://websockets.readthedocs.io/en/stable/howto/upgrade.html#import-paths).

aaugustin:
For tests I recommend the explicit import path.

Kludex:
@aaugustin if you have time, can you give me your thoughts on https://github.com/Kludex/uvicorn/pull/2888? 🙏

I've implemented that as an alternative. Please ignore the test suite, I haven't checked it yet (Claude worked on it). I'm interested in having feedback on the implementation itself.

aaugustin:
Done!

Kludex:
@MrNaif2018 I'll go with https://github.com/Kludex/uvicorn/pull/2888 instead. I think the callback approach is easier to understand, and it's more aligned with the rest of the code source.

Thanks for this PR tho, without it, I wouldn't have investigated it properly.

PR #2889, opened 2026-04-05
Add docstrings to Server class and its public methods

mastash3ff:
# Summary

Add docstrings to `uvicorn/server.py` to make the server lifecycle easier to
understand for new contributors and API users.  No logic, behaviour, or
algorithms are changed.

**Items documented:**

- `Server` – class-level summary describing its role
- `Server.limit_max_requests` – explains the jitter logic and `None` semantics
- `Server.run()` – synchronous entry point description
- `Server.serve()` – async entry point and signal-capture relationship
- `Server.main_loop()` – tick rate and delegation to `on_tick`
- `Server.on_tick()` – housekeeping cadence, header refresh, callback notify, and return value semantics
- `Server.shutdown()` – graceful shutdown sequence summary
- `Server.capture_signals()` – context manager contract, thread restriction, signal re-raise behaviour
- `Server.handle_exit()` – double-Ctrl+C / force-exit logic

# Checklist

### DIFF
--- a/uvicorn/server.py
+++ b/uvicorn/server.py
@@ -54,6 +54,13 @@ class ServerState:
 
 
 class Server:
+    """
+    ASGI server that manages the full lifecycle of a single server process.
+
+    Handles startup, the main request loop, graceful shutdown, and OS signal
+    capture.  Intended to be used directly or driven by a reload supervisor.
+    """
+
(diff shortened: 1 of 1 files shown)


Kludex:
Thanks, but I'd prefer to not merge this for now. Maybe at some point.

PR #2927, opened 2026-04-30
Treat `fd=0` as a valid file descriptor with reload/workers

eltoder:
# Summary

Passing a socket as stdin (`--fd=0`) currently does not work in modes that use subprocesses (i.e. with `--reload` and `--workers`), because `config.fd` is checked for truthiness instead of for `None`.

# Checklist

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -6,6 +6,7 @@ import json
 import logging
 import os
 import socket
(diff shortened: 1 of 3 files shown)


Kludex (approved):

Kludex:
Thanks!

PR #2928, opened 2026-04-30
Support RFC 7239 `Forwarded` header and `X-Forwarded-Host`

### LINKED ISSUE #965: Support X-Forwarded-Host in proxy-headers
### Checklist

When using uvicorn, with starlette URLs with url_for are generated with bad host, when application is behind proxy and proxy sets original host to X-Forwarded-Host. 

https://github.com/encode/starlette/issues/604

Currently uvicorn uses only X-Forwarded-Proto, X-Forwarded-For, X-Forwarded-Port to populate scope with remote information. But there is X-Forwarded-Host header, which should be used for populating server information in scope.

### LINKED ISSUE #2237: Add support for Fowarded header (RFC 7239)
### Discussed in https://github.com/encode/uvicorn/discussions/2236

<sup>Originally posted by **nhairs** January 28, 2024</sup>
We should probably support the `Forwarded` header which was standardised in [RFC 7239](https://datatracker.ietf.org/doc/html/rfc7239).

One of the benefits of this header over `X-Forwarded-*` headers is that it supports including the port of the client rather than just the address.

See also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded</div>

> [!IMPORTANT]

<a href="https://polar.sh/encode/uvicorn/issues/2237">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/uvicorn/issues/2237/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

Kludex:
Closes #2237. Also closes #965 (rolls in the `X-Forwarded-Host` work that was paused as #2922).

## Summary

- New `--proxy-headers` mode option to choose between the existing `X-Forwarded-*` family (default `x-forwarded`) and the standardized `Forwarded` header from [RFC 7239](https://datatracker.ietf.org/doc/html/rfc7239) (new `forwarded`).
- Modes are mutually exclusive: when one is selected, the other family is ignored entirely. Silent fallback would let a client smuggle values via whichever family the proxy is not configured for.
- The default `x-forwarded` mode also now reads `X-Forwarded-Host` and rewrites the request `Host` header. `scope["server"]` is intentionally left untouched per the ASGI spec.
- Trust gating is unchanged: `forwarded_allow_ips` (default `127.0.0.1`) - headers are honored only when the connecting peer is in the allowlist.

## CLI is backwards compatible

| Invocation | Behavior |
|---|---|
| (default) | `x-forwarded` mode, enabled |
| `--proxy-headers` | `x-forwarded` mode, enabled (matches the prior boolean) |
| `--no-proxy-headers` | disabled |
| `--proxy-headers x-forwarded` | new explicit form |
| `--proxy-headers forwarded` | RFC 7239 mode |

## Implementation notes

- `ProxyHeadersMiddleware` is a thin dispatcher; per-mode logic lives in `_XForwardedParser` and `_RFC7239Parser` (gzip-style pattern).
- RFC 7239 parser handles quoted-strings, backslash escapes, IPv6 brackets, and rejects forwarded-elements with duplicate parameters (smuggling defense).
- Hop selection mirrors `X-Forwarded-For`: rightmost-untrusted with leftmost fallback when all hops are trusted. `for=unknown` and `_obfuscated` placeholders are filtered before the walk.
- `_join_field` correctly handles repeated headers (RFC 7230 §3.2.2). For single-valued headers (`X-Forwarded-Proto`, `X-Forwarded-Host`), `_last_field` returns the rightmost value - this matches the prior `dict(scope["headers"])` last-wins behavior and is the correct direction when a trusted proxy appends.

## Test plan

- [x] `tests/middleware/test_proxy_headers.py` - 41 new tests covering RFC 7239 parsing, hop selection, mode mutual exclusion, multi-header join, IPv6, placeholder filtering, duplicate-param smuggling defense, programmatic Config wiring.
- [x] Existing `X-Forwarded-For/-Proto` tests unchanged - no behavior regression.
- [x] `mypy uvicorn` clean.
- [x] `ruff check .` clean.

## AI Disclaimer


ptrovatelli:
>> _last_field returns the rightmost value - this matches the prior dict(scope["headers"]) last-wins behavior and is the correct direction when a trusted proxy appends.

It is a tricky choice. last-win behavior is better for security but can break things when multiple reverse proxies are used. See this example https://github.com/Kludex/uvicorn/discussions/2310#discussioncomment-16768131

It should be made apparent in the doc that if multiple reverse proxies are used, only the first one should define Forwarded header (other proxies should be configured with ["IfNone" behavior for example for HAProxy](https://docs.redhat.com/en/documentation/openshift_container_platform/4.8/html/networking/configuring-ingress#nw-ne-openshift-ingress_configuring-ingress)). Default value for HAProxy is "append" which creates the issues described
FYI the other solutions i've seen around is the definition of a "max number of trusted proxies" so that each proxy can add values and the application server will go up the chain only to a max number of header definition to avoid header poisoning by the end user

PR #2944, opened 2026-05-21
Ignore duplicate forwarding headers in `ProxyHeadersMiddleware`

Kludex:
## Summary

`ProxyHeadersMiddleware` converted `scope["headers"]` to a `dict` before reading `X-Forwarded-For` and `X-Forwarded-Proto`. ASGI preserves the headers as a list of `(name, value)` tuples, and duplicate names are allowed - passing such a list to `dict()` silently keeps only the last value.

When duplicate forwarding headers reached the middleware, only the last value was used to populate `scope["client"]` and `scope["scheme"]`, even though the previous proxy in the chain may have appended its own value while passing a client-supplied duplicate through. The result was ambiguous and surprising.

## Changes

- Iterate `scope["headers"]` once, collecting each forwarding header into its own list.
- Only consume `X-Forwarded-For` / `X-Forwarded-Proto` when exactly one copy is present. If more than one copy is seen, the middleware ignores the forwarded value and leaves `scope["client"]` / `scope["scheme"]` untouched.
- Added two regression tests driving the middleware directly with crafted ASGI scopes.

## Test plan

- [x] `uv run pytest tests/middleware/test_proxy_headers.py` - 263 passed
- [x] `uv run ruff check uvicorn/middleware/proxy_headers.py tests/middleware/test_proxy_headers.py`
- [x] `uv run ruff format --check uvicorn/middleware/proxy_headers.py tests/middleware/test_proxy_headers.py`
- [x] `uv run mypy uvicorn/middleware/proxy_headers.py`
- [x] Confirmed the new regression tests fail on `main` without the middleware change

## AI Disclaimer

### DIFF
--- a/tests/middleware/test_proxy_headers.py
+++ b/tests/middleware/test_proxy_headers.py
@@ -10,7 +10,7 @@ import websockets.client
 
 from tests.response import Response
 from tests.utils import run_server
-from uvicorn._types import ASGIReceiveCallable, ASGIReceiveEvent, ASGISendCallable, ASGISendEvent, Scope
+from uvicorn._types import ASGIReceiveCallable, ASGIReceiveEvent, ASGISendCallable, ASGISendEvent, HTTPScope, Scope
 from uvicorn.config import Config
 from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware, _TrustedHosts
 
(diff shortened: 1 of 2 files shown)


dtczest:
This change caused our service to no longer trust its proxies. I've opened a discussion here:
https://github.com/Kludex/uvicorn/discussions/2968

PR #2946, opened 2026-05-21
Align IPv6 socket binding across worker modes

### LINKED ISSUE #2945: Inconsistent IPv6 dual-stack behavior between single-worker and multi-worker modes
## Summary

## Steps to Reproduce

```bash

```

## Root Cause

```python
```

```python
    create_protocol,
    host=config.host,
    port=config.port,
    ...
)
```

```python
if af == socket.AF_INET6:
```

## Impact

## Suggested Fix

## Workaround

```python
import socket
import uvicorn

sock.bind(("::", 8080))

```

## Environment

- uvicorn 0.34.3
- Python 3.13

puneetdixit200:
# Summary

Pre-bound IPv6 sockets used by reload and worker modes now set `IPV6_V6ONLY` before binding, matching the single-worker `asyncio.create_server()` path and avoiding worker-count-dependent dual-stack behavior.

Closes #2945

# Tests

- `UV_CACHE_DIR=$PWD/.uv-cache uv run pytest tests/test_config.py::test_socket_bind_ipv6_only -q`
- `UV_CACHE_DIR=$PWD/.uv-cache uv run pytest tests/test_config.py -q`
- `UV_CACHE_DIR=$PWD/.uv-cache scripts/check`
- `UV_CACHE_DIR=$PWD/.uv-cache scripts/test`
- `git diff --check`

# Checklist

- [x] Documentation was not updated because this fixes internal socket setup consistency without adding a new user-facing option.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -18,7 +18,7 @@ import yaml
 from pytest_mock import MockerFixture
 
 from tests.custom_loop_utils import CustomLoop
-from tests.utils import as_cwd, get_asyncio_default_loop_per_os
+from tests.utils import as_cwd, get_asyncio_default_loop_per_os, has_ipv6
 from uvicorn._types import ASGIApplication, ASGIReceiveCallable, ASGISendCallable, Environ, Scope, StartResponse
 from uvicorn.config import Config, LoopFactoryType
 from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
@@ -260,17 +260,6 @@ def test_socket_bind() -> None:
     sock = config.bind_socket()
     assert isinstance(sock, socket.socket)
     sock.close()
-
-
-def has_ipv6(host: str) -> bool:
-    if not socket.has_ipv6:
-        return False  # pragma: no cover
-    try:
-        with socket.socket(socket.AF_INET6) as sock:
-            sock.bind((host, 0))
-    except OSError:  # pragma: no cover
-        return False
-    return True
 
 
 @pytest.mark.skipif(not has_ipv6("::"), reason="IPV6 not enabled")
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -262,6 +262,30 @@ def test_socket_bind() -> None:
     sock.close()
 
 
+def has_ipv6(host: str) -> bool:
+    if not socket.has_ipv6:
+        return False  # pragma: no cover
+    try:
+        with socket.socket(socket.AF_INET6) as sock:
+            sock.bind((host, 0))
+    except OSError:  # pragma: no cover
+        return False
+    return True
+
+
+@pytest.mark.skipif(not has_ipv6("::"), reason="IPV6 not enabled")
+def test_socket_bind_ipv6_only() -> None:
+    config = Config(app=asgi_app, host="::", port=0)
+    config.load()
+    sock = config.bind_socket()
+    assert isinstance(sock, socket.socket)
+    try:
+        assert sock.family == socket.AF_INET6
+        assert sock.getsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY) == 1
+    finally:
+        sock.close()
+
+
 def test_ssl_config(
     tls_ca_certificate_pem_path: str,
     tls_ca_certificate_private_key_path: str,
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -274,14 +274,14 @@ def has_ipv6(host: str) -> bool:
 
 
 @pytest.mark.skipif(not has_ipv6("::"), reason="IPV6 not enabled")
-def test_socket_bind_ipv6_only() -> None:
+def test_socket_bind_ipv6_dual_stack() -> None:
     config = Config(app=asgi_app, host="::", port=0)
     config.load()
     sock = config.bind_socket()
     assert isinstance(sock, socket.socket)
     try:
         assert sock.family == socket.AF_INET6
-        assert sock.getsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY) == 1
+        assert sock.getsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY) == 0
     finally:
         sock.close()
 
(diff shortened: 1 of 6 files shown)


ogajduse:
Thanks for picking this up! I have a concern about the direction of the fix though.

This PR aligns behavior by setting `IPV6_V6ONLY=1` in `bind_socket()`, making multi-worker match single-worker. But the result is that **both** paths now force IPv6-only — there's no way to get dual-stack through uvicorn's configuration in any mode.

Before this PR, `--host '::' --workers 2` gave you dual-stack (IPv4 + IPv6 on one socket) because `bind_socket()` inherited the system default (`IPV6_V6ONLY=0` on Linux). That was the only uvicorn code path that worked for dual-stack. This PR removes it.

The inconsistency is real, but I think it should be resolved in the other direction — by making single-worker also respect the system default (or explicitly set `IPV6_V6ONLY=0`) when only one socket family is being created. The reason asyncio forces `IPV6_V6ONLY=True` is to prevent conflicts when it creates *both* an IPv4 and an IPv6 socket for the same port (e.g., `host=None`). But when `host='::'` is given, `getaddrinfo` returns only the IPv6 entry — asyncio creates one socket, no conflict is possible, yet it still forces IPv6-only.

Since uvicorn can't change asyncio's behavior, the single-worker path could create and bind the socket itself (like `bind_socket()` already does) and pass it via `loop.create_server(sock=sock)`, which bypasses asyncio's socket option override. That would give consistent *and* correct dual-stack behavior across both modes.

The use case: rootless Podman with pasta networking does L4 socket translation — it creates separate IPv4 and IPv6 connections to the container. A server listening on `::` with `IPV6_V6ONLY=1` is unreachable over IPv4. This affects any ASGI app running in a rootless container.

puneetdixit200:
Pushed `4a2c9c7` to switch the fix in the direction suggested here.

What changed:
- `bind_socket()` keeps IPv6 sockets dual-stack with `IPV6_V6ONLY=0`.
- the single-worker `host="::"` path now pre-binds the socket and passes it to `loop.create_server(sock=sock)`, so asyncio does not force IPv6-only.
- added coverage for both `Config.bind_socket()` and `Server.startup()` preserving dual-stack IPv6 sockets.

Verified locally:
- `UV_CACHE_DIR=/tmp/uvicorn-2946-uv-cache uv run --dev pytest -n 0 tests/test_config.py::test_socket_bind_ipv6_dual_stack tests/test_server.py::test_server_uses_dual_stack_ipv6_socket -q`
- `UV_CACHE_DIR=/tmp/uvicorn-2946-uv-cache uv run --dev pytest -n 0 tests/test_main.py::test_exit_on_create_server_with_invalid_host -q`
- `UV_CACHE_DIR=/tmp/uvicorn-2946-uv-cache uv run --dev ruff check uvicorn/config.py uvicorn/server.py tests/test_config.py tests/test_server.py`
- `UV_CACHE_DIR=/tmp/uvicorn-2946-uv-cache uv run --dev ruff format --check uvicorn/config.py uvicorn/server.py tests/test_config.py tests/test_server.py`
- `git diff --check`

I also tried the broader `tests/test_config.py tests/test_server.py` run locally; it hit a machine-local default port conflict on `tests/test_config.py::test_socket_bind` because port 8000 is already in use here, while the focused IPv6 regression and changed-file checks passed.

ogajduse:
Thanks @puneetdixit200 for turning this around so quickly!

I tested the updated branch in a rootless Podman container with pasta networking — the exact environment that was broken before. Built a minimal FastMCP server with `host="::"`, no manual socket workaround, and confirmed both IPv4 and IPv6 connections go through:

```
::ffff:10.x.x.x:50508 - "POST /mcp HTTP/1.1" 200 OK
2620:52:0:xxxx:xxxx:xxxx:xxxx:xxxx:54152 - "POST /mcp HTTP/1.1" 200 OK
```

This is exactly the fix I need.

Minor nit: the `has_ipv6()` helper is duplicated across `test_config.py` and `test_server.py` — might be worth extracting to a shared test utility, but that's up to the maintainers.

puneetdixit200:
Pushed `7b9bd7d` to cover the remaining IPv6 startup log formatting branch that the coverage job reported in `uvicorn/server.py`.

Local verification:
- `UV_CACHE_DIR=/tmp/uvicorn-2946-uv-cache uv run --dev pytest -n 0 tests/test_config.py::test_socket_bind_ipv6_dual_stack tests/test_server.py::test_server_uses_dual_stack_ipv6_socket tests/test_server.py::test_log_started_message_formats_ipv6_host tests/test_main.py::test_exit_on_create_server_with_invalid_host -q`
- `UV_CACHE_DIR=/tmp/uvicorn-2946-uv-cache uv run --dev ruff check uvicorn/config.py uvicorn/server.py tests/test_config.py tests/test_server.py`
- `UV_CACHE_DIR=/tmp/uvicorn-2946-uv-cache uv run --dev ruff format --check uvicorn/config.py uvicorn/server.py tests/test_config.py tests/test_server.py`
- `git diff --check`

The GitHub Actions matrix is green on the new head, including the aggregate `check` job.

puneetdixit200:
Pushed `9ba8d76` to address the IPv6 helper duplication nit. `has_ipv6()` now lives in `tests/utils.py` and is reused from the config, server, and main tests.

Local validation:
- `UV_CACHE_DIR=/tmp/uvicorn-2946-uv-cache uv run --dev pytest -n 0 tests/test_config.py::test_socket_bind_ipv6_dual_stack tests/test_server.py::test_server_uses_dual_stack_ipv6_socket tests/test_server.py::test_log_started_message_formats_ipv6_host 'tests/test_main.py::test_run[ipv6]' -q`\n- `UV_CACHE_DIR=/tmp/uvicorn-2946-uv-cache uv run --dev ruff check tests/utils.py tests/test_config.py tests/test_server.py tests/test_main.py`\n- `UV_CACHE_DIR=/tmp/uvicorn-2946-uv-cache uv run --dev ruff format --check tests/utils.py tests/test_config.py tests/test_server.py tests/test_main.py`\n- `git diff --check`

patyogesh20:
thanks for the fix @puneetdixit200. 
I ran into the same issue on Azure-Linux base image with `net.ipv6.bindv6only=1`. That also confirms the issue is not pasta-specific, it is any host where the kernel default is V6-only.

PR #2947, opened 2026-05-22
Force-kill reload worker that does not exit within grace period

### LINKED ISSUE #2943: BaseReload.restart() / shutdown() hang indefinitely when worker holds a long-lived connection
## Bug

`BaseReload.restart()` and `BaseReload.shutdown()` in `uvicorn/supervisors/basereload.py` call `self.process.join()` without a timeout:

```python
self.process.terminate()
self.process.join()  # blocks forever if worker has open connections
```

If any client holds a long-lived connection (SSE, WebSocket, chunked streaming), the worker will not exit after receiving SIGTERM — it waits for those connections to close first. `process.join()` then blocks indefinitely, the reload path never reaches `get_subprocess(...)`, and no new worker is ever started. From the outside, hot-reload silently hangs.

## Affected versions

0.30.6 through 0.47.0 (current `main` as of May 2026).

## Reproduction

1. Create a minimal FastAPI/Starlette app with an SSE or WebSocket endpoint that holds a connection open.
2. Connect a client so the connection stays open.
3. Edit any watched file to trigger a reload.
4. Observe: `WatchFiles detected changes` is logged, the worker receives SIGTERM and logs its shutdown, but no new worker ever appears.

## Expected behaviour

The old worker is given a grace period to finish. If it has not exited by then it is force-killed, and the new worker is started immediately afterwards.

## Actual behaviour

`process.join()` never returns. The supervisor thread blocks forever. No new worker is started.

## Proposed fix

Reuse `timeout_graceful_shutdown` (already a config option) as the join timeout, falling back to 5s, then force-kill with SIGKILL if the process is still alive:

```python
def restart(self) -> None:
    if sys.platform == "win32":
        self.is_restarting = True
        sys.stdout.flush()
    else:
        self.process.terminate()
        timeout = self.config.timeout_graceful_shutdown or 5
        self.process.join(timeout=timeout)
        if self.process.is_alive():
            logger.warning("Worker did not exit after %ss, sending SIGKILL", timeout)
            self.process.kill()
            self.process.join()
    self.process = get_subprocess(config=self.config, target=self.target, sockets=self.sockets)
    self.process.start()
```

Apply the same pattern to `shutdown()`.

`timeout_graceful_shutdown` is already the configured grace period for lifespan shutdown, so reusing it here is semantically consistent and requires no new config surface.

## Workaround

Add a `threading.Timer` in your app's lifespan shutdown handler that calls `os._exit(0)` after N seconds. This should not be necessary.

ATOM00blue:
# Summary

Fixes #2943.

`BaseReload.restart()` and `BaseReload.shutdown()` call `self.process.join()` with no timeout. A worker that is still serving a long-lived connection (SSE, WebSocket, chunked streaming) does not exit when it receives `SIGTERM` — it waits for those connections to drain first. As a result `join()` blocks indefinitely: the reload path never reaches `get_subprocess(...)`, no new worker is started, and hot-reload silently hangs.

## Reproduce

1. Run an app under `--reload` with an endpoint that holds a connection open (e.g. an SSE stream).
2. Keep a client connected.
3. Edit a watched file.

`WatchFiles detected changes ...` is logged and the worker receives `SIGTERM`, but the new worker never appears — the reloader is stuck on `process.join()`.

## Fix

Wait for the worker for `timeout_graceful_shutdown` seconds (falling back to 5s when it isn't configured) and send `SIGKILL` if it's still alive afterwards. Reusing `timeout_graceful_shutdown` keeps the grace period consistent with the server's own shutdown timeout and adds no new config surface. This mirrors the `terminate()` / `kill()` escalation the multiprocess supervisor already uses.

The join + force-kill logic is shared between `restart()` and `shutdown()` via a small helper.

## Test plan

- [x] Added `test_restart_kills_worker_that_does_not_exit` and `test_shutdown_does_not_kill_worker_that_exits`; both fail on `main` (the old `join()` is unbounded and never escalates to `kill()`).
- [x] `scripts/test` — full suite passes, `basereload.py` stays at 100% coverage.
- [x] `scripts/check` — ruff format, ruff check and mypy all pass.

### DIFF
--- a/tests/supervisors/test_reload.py
+++ b/tests/supervisors/test_reload.py
@@ -399,3 +399,38 @@ def test_base_reloader_closes_sockets_on_shutdown():
     assert sock.fileno() != -1
     reloader.shutdown()
     assert sock.fileno() == -1
+
+
+def test_restart_kills_worker_that_does_not_exit(mocker: MockerFixture):
+    """A worker that doesn't exit after ``terminate()`` is force-killed instead of
+    blocking the reloader forever on an unbounded ``join()``."""
+    config = Config(app="tests.test_config:asgi_app", reload=True, timeout_graceful_shutdown=1)
+    reloader = BaseReload(config, target=run, sockets=[])
+
+    old_process = mocker.MagicMock()
+    old_process.is_alive.return_value = True
+    reloader.process = old_process
+    new_process = mocker.patch("uvicorn.supervisors.basereload.get_subprocess").return_value
+
+    reloader.restart()
+
+    old_process.join.assert_any_call(1)
+    old_process.kill.assert_called_once()
+    # A fresh worker is started after the old one is dealt with.
+    assert reloader.process is new_process
+    new_process.start.assert_called_once()
+
+
+def test_shutdown_does_not_kill_worker_that_exits(mocker: MockerFixture):
+    """When the worker exits within the grace period it is not force-killed."""
+    config = Config(app="tests.test_config:asgi_app", reload=True)
+    reloader = BaseReload(config, target=run, sockets=[])
+
+    process = mocker.MagicMock()
+    process.is_alive.return_value = False
+    reloader.process = process
+
+    reloader.shutdown()
+
+    process.join.assert_called_once_with(5)
+    process.kill.assert_not_called()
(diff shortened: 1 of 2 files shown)


Kludex:
Closing in favor of the existing `--timeout-graceful-shutdown` option, which already bounds the shutdown wait - see #2943.

PR #2955, opened 2026-05-26
Fix: Install signal handlers immediately

whyujjwal:
This PR fixes a bug where CTRL^C would not work while the server was starting up. This was because the signal handlers were not being installed until after the on_startup functions had completed.

This PR fixes the issue by installing the signal handlers immediately, before the on_startup functions are called.


whyujjwal:
Just following up to see if you've had a chance to review this. Let me know if you have any questions or feedback!

PR #2971, opened 2026-06-03
Consume duplicate forwarding headers in ProxyHeadersMiddleware

Kludex:
## Summary

[#2944](https://github.com/Kludex/uvicorn/pull/2944) made `ProxyHeadersMiddleware` ignore `X-Forwarded-For` / `X-Forwarded-Proto` entirely whenever the header appeared more than once. As reported in [#2968](https://github.com/Kludex/uvicorn/discussions/2968), proxy chains where each hop appends its own field (e.g. HAProxy behind an ALB) suddenly stopped having their reverse proxy trusted, silently breaking client-IP logging and rate limiting.

Instead of dropping duplicates, we handle each header according to its actual format.

## Changes

- **`X-Forwarded-For`**: repeated fields are equivalent to a single comma-separated list ([RFC 9110, 5.3](https://www.rfc-editor.org/rfc/rfc9110#section-5.3)), so we join them in order and run the existing reverse trust walk. The leftmost untrusted host is the client, preserving the spoofing protection #2944 was after.
- **`X-Forwarded-Proto`**: we keep the last field, written by the closest (trusted) proxy and the hardest for a client to spoof. This restores the long-standing pre-#2944 behavior (`dict(headers)` kept the last field) without the silent drop on duplicates.
- Updated the duplicate-header regression tests to assert this behavior, and added a test reproducing the HAProxy-behind-ALB scenario from #2968.
- Documented repeated-header handling in the deployment docs.

## Test plan

- [x] `uv run pytest tests/middleware/test_proxy_headers.py` - 265 passed
- [x] `uv run ruff check uvicorn/middleware/proxy_headers.py tests/middleware/test_proxy_headers.py`
- [x] `uv run ruff format --check uvicorn/middleware/proxy_headers.py tests/middleware/test_proxy_headers.py`
- [x] `uv run mypy uvicorn/middleware/proxy_headers.py`

Closes #2968

## AI Disclaimer

### DIFF
--- a/uvicorn/middleware/proxy_headers.py
+++ b/uvicorn/middleware/proxy_headers.py
@@ -32,17 +32,18 @@ class ProxyHeadersMiddleware:
         client_host = client_addr[0] if client_addr else None
 
         if client_host in self.trusted_hosts:
-            # Keep the last X-Forwarded-Proto: it is from the closest proxy and the hardest to spoof.
-            x_forwarded_proto_value: bytes | None = None
+            x_forwarded_proto_values: list[bytes] = []
             x_forwarded_for_values: list[bytes] = []
             for name, value in scope["headers"]:
                 if name == b"x-forwarded-proto":
-                    x_forwarded_proto_value = value
+                    x_forwarded_proto_values.append(value)
                 elif name == b"x-forwarded-for":
                     x_forwarded_for_values.append(value)
 
-            if x_forwarded_proto_value is not None:
-                x_forwarded_proto = x_forwarded_proto_value.decode("latin1").strip()
+            # Repeated fields are equivalent to a single comma-separated list (RFC 9110, 5.3).
+            # The last value is from the closest proxy and the hardest for a client to spoof.
+            if x_forwarded_proto_values:
+                x_forwarded_proto = b",".join(x_forwarded_proto_values).rsplit(b",", 1)[-1].decode("latin1").strip()
 
                 if x_forwarded_proto in {"http", "https", "ws", "wss"}:
                     if scope["type"] == "websocket":
--- a/uvicorn/middleware/proxy_headers.py
+++ b/uvicorn/middleware/proxy_headers.py
@@ -40,10 +40,10 @@ class ProxyHeadersMiddleware:
                 elif name == b"x-forwarded-for":
                     x_forwarded_for_values.append(value)
 
-            # Repeated fields are equivalent to a single comma-separated list (RFC 9110, 5.3),
-            # so the rightmost value comes from the last field, after splitting it on commas.
+            # When a proxy chain sends one field per hop, the last is from the closest (trusted)
+            # proxy, so a client cannot spoof the scheme via an earlier field.
             if x_forwarded_proto_values:
-                x_forwarded_proto = x_forwarded_proto_values[-1].rsplit(b",", 1)[-1].decode("latin1").strip()
+                x_forwarded_proto = x_forwarded_proto_values[-1].decode("latin1").strip()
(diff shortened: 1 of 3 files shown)


uvicorn/middleware/proxy_headers.py
```diff
@@ -40,18 +40,21 @@ async def __call__(self, scope: Scope, receive: ASGIReceiveCallable, send: ASGIS
...
 
-            # Only consume the header when exactly one copy is present to avoid spoofing issues.
-            if len(x_forwarded_proto_values) == 1:
-                x_forwarded_proto = x_forwarded_proto_values[0].decode("latin1").strip()
+            # Repeated fields are equivalent to a single comma-separated list (RFC 9110, 5.3),
+            # so join the values in order before parsing.
+            if x_forwarded_proto_values:
+                joined_proto = b", ".join(x_forwarded_proto_values).decode("latin1")
+                # The leftmost value is the protocol the client used to reach the first proxy.
+                x_forwarded_proto = joined_proto.split(",", 1)[0].strip()
```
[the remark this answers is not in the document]
Kludex: Good catch - fixed in 91f7334. Now uses the rightmost X-Forwarded-Proto value (written by the closest trusted proxy) instead of the leftmost client-controlled one, so a spoofed leftmost no longer rewrites the scheme. Matches Hypercorn's behaviour.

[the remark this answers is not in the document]
dtczest: Right, that makes sense. I was thinking the intermediate proxies might be http, but even if they are, they should be passing on the original x-forwarded-proto.

dtczest (approved):

uvicorn/middleware/proxy_headers.py
```diff
@@ -40,18 +40,19 @@ async def __call__(self, scope: Scope, receive: ASGIReceiveCallable, send: ASGIS
                 elif name == b"x-forwarded-for":
                     x_forwarded_for_values.append(value)
 
-            # Only consume the header when exactly one copy is present to avoid spoofing issues.
-            if len(x_forwarded_proto_values) == 1:
-                x_forwarded_proto = x_forwarded_proto_values[0].decode("latin1").strip()
+            # When a proxy chain sends one field per hop, the last is from the closest (trusted)
+            # proxy, so a client cannot spoof the scheme via an earlier field.
+            if x_forwarded_proto_values:
+                x_forwarded_proto = x_forwarded_proto_values[-1].decode("latin1").strip()
```
dtczest: while uncommon, I think x-forwarded-proto can be comma separated. So if the closest proxy is appending to an existing x-forwarded-proto, this code won't work.

Kludex: any reference?

dtczest: Unfortunately there's no RFC that specifies this, so implementations are varied. Maybe one day we can all move to the "Forwarded" header lol.

This has been discussed for Uvicorn here, previously: https://github.com/Kludex/uvicorn/pull/2104

dtczest (approved):

dtczest:
Thanks for jumping on this so quickly!

dtczest (approved):

uvicorn/middleware/proxy_headers.py
Kludex: I've reverted this. We can do it in another PR if necessary. I'm keeping the previous behavior anyway.

dtczest: I saw, it makes sense. Better to keep this PR focused.

Issue #2639, opened 2025-05-30
Uvicorn --reload-include=.env not working, appears to be overridden by default exclude patterns


adithya04dev:
### Initial Checks

- [x] I confirm this was discussed, and the maintainers suggest I open an issue.
- [x] I'm aware that if I created this issue without a discussion, it may be closed without a response.

### Discussion Link

```Text
https://github.com/encode/uvicorn/issues/2166
```

### Description

When running Uvicorn with hot reload, I need it to detect changes to my .env file, but the --reload-include=.env flag doesn't work as expected. It seems the .env file is still being excluded by Uvicorn's default exclude patterns.

Using --reload-include=.* as a workaround causes Uvicorn to watch all hidden files, not just .env, which is excessive.
What is the correct way to configure Uvicorn to watch only the .env file while maintaining other default exclusions?

### Example Code

```Python
uvicorn backend.app.main:app --reload --reload-include="*.env"
```

### Python, Uvicorn & OS Version

```Text
>>
Running uvicorn 0.34.0 with CPython 3.12.7 on Windows
```

Kludex:
`uvicorn main:app --reload --reload-include=".env"` works just file.

adithya04dev:
ok thanks

mgadd-correla:
I'm not able to watch changes to a `.env` file in my project root when used in combination with `--reload-dir` – which _does_ make sense, thinking about it.

This doesn't work, changing `.env` does nothing:
`uvicorn main:application --reload --reload-dir=apps --reload-include='*.py' --reload-include=.env`

Is there a way to watch a specific file outside of the watched dirs?

PR #2989, opened 2026-06-12
Check directory excludes before the explicit-include shortcut in `FileFilter`

LeSingh1:
When an explicit path (e.g. `.dotted` or `ext/ext.jpg`) is listed in
`--reload-include`, `FileFilter.__call__` matched via `path.match()`
and then hit the `str(path).endswith(include_pattern)` shortcut, which
returned `True` immediately.  The early return skipped the loop that
checks `exclude_dirs`, so any file whose name exactly matched an include
pattern was watched even when it lived inside a directory listed in
`--reload-exclude`.

The `endswith` shortcut was added in c55af77 to let hidden/dotted files
and other non-glob entries bypass the *pattern-level* excludes (e.g.
`.*`).  That intent is correct; the mistake is that it also bypassed
*directory-level* excludes, which is a stronger signal from the user.

The fix moves the `exclude_dirs` loop above the `endswith` shortcut so
that directory exclusions are always evaluated first.  Files that match
an exact include pattern but reside under an excluded directory are now
rejected, while files outside excluded directories continue to bypass
the pattern excludes as before.

A new unit test (pure `FileFilter` instantiation, no file-system
watcher) reproduces the bug and confirms the fix: it asserts that
`vendor/.dotted` is not watched when `.dotted` is in `--reload-include`
and `vendor/` is in `--reload-exclude`.

### DIFF
--- a/tests/supervisors/test_reload.py
+++ b/tests/supervisors/test_reload.py
@@ -409,3 +409,38 @@ def test_base_reloader_closes_sockets_on_shutdown():
     assert sock.fileno() != -1
     reloader.shutdown()
     assert sock.fileno() == -1
+
+
+@pytest.mark.skipif(WatchFilesReload is None, reason="watchfiles not available")
+def test_file_filter_explicit_include_respects_exclude_dirs(tmp_path: Path):
+    """Exact-name entries in --reload-include must still be blocked by --reload-exclude dirs.
+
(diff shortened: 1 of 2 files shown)


Kludex:
There's no issue or discussion reporting this, and the only related reports (#2639, #2166) show users expecting explicit includes to win over excludes - which is the current behavior. Happy to reconsider if a discussion makes the case for the opposite precedence.

PR #3001, opened 2026-06-21
Exit with a dedicated code on startup failure and stop the supervisor when a worker can't boot

Kludex:
## What this does

Keeps the supervisor from restarting a broken worker forever (#2440). Split from the original PR: the eager-import memory fix for #2980 now lives in #3012.

Every startup-failure site - app loading, TLS config, socket bind and lifespan startup - now exits with the existing `STARTUP_FAILURE` (3) sentinel directly where the failure happens, making "exit 3 = startup failure" a uniform contract (previously a mix of 0, 1 and 3). The multiprocess supervisor stops when a worker dies with that code instead of respawning forever. A worker that started and later exits - recycling on `limit_max_requests`, a runtime crash - is restarted as before.

## Behavior changes

- Startup failures that previously exited 1 (bad app import string, loop config errors, socket bind failure) now exit 3, consistent with lifespan startup failure. Anyone scripting against exit code 1 for these cases needs to check 3 instead.
- `Server.serve()` now raises `SystemExit(3)` on lifespan startup failure instead of returning; embedders that call `serve()` directly and relied on the silent return will see the exception.
- Exit code 3 is reserved for startup failure: an app that deliberately exits 3 at runtime would be treated as a boot failure.

## Scope and follow-ups

- TLS errors that raise (e.g. a bad certfile path in `create_ssl_context`) still exit 1 with a traceback and are not detected by the supervisor; wrapping them is left for a follow-up.
- The parent still exits 0 after stopping on a worker boot failure; propagating the exit code is left for a follow-up.
- The `--reload` path does not yet detect a worker that fails to boot.

## AI Disclaimer

### DIFF
--- a/uvicorn/config.py
+++ b/uvicorn/config.py
@@ -76,6 +76,11 @@ INTERFACES: list[InterfaceType] = ["auto", "asgi3", "asgi2", "wsgi"]
 
 SSL_PROTOCOL_VERSION: int = ssl.PROTOCOL_TLS_SERVER
 
+# Exit code for a failure before the server started serving - a broken app, TLS config
+# or socket bind. The multiprocess supervisor stops on it instead of restarting a worker
+# that would fail the same way forever. See https://github.com/encode/uvicorn/discussions/2440.
+STARTUP_FAILURE = 3
+
 LOGGING_CONFIG: dict[str, Any] = {
     "version": 1,
     "disable_existing_loggers": False,
@@ -425,7 +430,7 @@ class Config:
             return import_from_string(self.app)
         except ImportFromStringError as exc:
             logger.error("Error loading ASGI app. %s" % exc)
-            sys.exit(1)
+            sys.exit(STARTUP_FAILURE)
 
     def load(self) -> None:
         assert not self.loaded
@@ -495,7 +500,7 @@ class Config:
         except TypeError as exc:
             if self.factory:
                 logger.error("Error loading ASGI app factory: %s", exc)
-                sys.exit(1)
+                sys.exit(STARTUP_FAILURE)
         else:
             if not self.factory:
                 logger.warning(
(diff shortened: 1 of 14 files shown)


uvicorn/config.py
```diff
@@ -76,6 +76,11 @@ class UvicornDeprecationWarning(UserWarning):
 
 SSL_PROTOCOL_VERSION: int = ssl.PROTOCOL_TLS_SERVER
 
+# Exit code for a failure before the server started serving - a broken app, TLS config
+# or socket bind. The multiprocess supervisor stops on it instead of restarting a worker
+# that would fail the same way forever. See https://github.com/encode/uvicorn/discussions/2440.
```
Kludex: This comment is too verbose. No need to exemplify when. Actually, no need comment at all for this. It's pretty clear.

uvicorn/main.py
```diff
@@ -601,12 +600,14 @@ def run(
...
     )
-    if (config.reload or config.workers > 1) and not isinstance(app, str):
-        logger = logging.getLogger("uvicorn.error")
-        logger.warning("You must pass the application as an import string to enable 'reload' or 'workers'.")
-        sys.exit(1)
+    if config.reload or config.workers > 1:
+        if not isinstance(app, str):
+            logger = logging.getLogger("uvicorn.error")
+            logger.warning("You must pass the application as an import string to enable 'reload' or 'workers'.")
+            sys.exit(1)
```
Kludex: Why is this 1 tho? Shouldn't it be part of startup error?

uvicorn/config.py
Kludex: Removed in ea733a0.

uvicorn/main.py
Kludex: Agreed - changed to `STARTUP_FAILURE` in ea733a0.

uvicorn/config.py
```diff
@@ -425,7 +430,7 @@ def load_app(self) -> Any:
             return import_from_string(self.app)
         except ImportFromStringError as exc:
             logger.error("Error loading ASGI app. %s" % exc)
-            sys.exit(1)
+            sys.exit(STARTUP_FAILURE)
```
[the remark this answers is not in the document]
Kludex: Intentional - exit 3 now uniformly means startup failure, documented in the Behavior changes section. The quoted sentence is from an earlier revision of the description.

PR #3002, opened 2026-06-22
Stringify SERVER_PORT in the WSGI environ

patrickwehbe:
# Summary

`build_environ` assigns the scope's server port straight into the WSGI environ:

```python
```

`server[1]` is an `int` (and can be `None` for a Unix domain socket, per the `server: tuple[str, int | None] | None` type in `uvicorn/_types.py`). PEP 3333 requires every value in the WSGI environ, including the CGI variables like `SERVER_PORT`, to be a native string. Every other CGI variable here (`SERVER_NAME`, `PATH_INFO`, `QUERY_STRING`, ...) is already a string, so `SERVER_PORT` is the odd one out, and a downstream WSGI app that does string work on it gets the wrong type.

The fix casts it, falling back to 80 when the port is `None`, which mirrors what asgiref's WSGI adapter does:

```python
environ["SERVER_PORT"] = str(server[1] or 80)
```

Added a test asserting `SERVER_PORT` is the string `"8000"` for an explicit server, and that the no-server default stays `"80"`.

# Checklist

<a href="https://cubic.dev/pr/Kludex/uvicorn/pull/3002?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/middleware/test_wsgi.py
+++ b/tests/middleware/test_wsgi.py
@@ -136,3 +136,32 @@ def test_build_environ_encoding() -> None:
     assert environ["SCRIPT_NAME"] == "/文".encode().decode("latin-1")
     assert environ["PATH_INFO"] == b"/all".decode("latin-1")
     assert environ["HTTP_KEY"] == "value1,value2"
+    # WSGI requires SERVER_PORT to be a native string (PEP 3333).
+    assert environ["SERVER_PORT"] == "80"
+
+
+def test_build_environ_server_port_is_str() -> None:
+    scope: HTTPScope = {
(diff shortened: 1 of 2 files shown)


Kludex (approved):

Kludex:
That middleware is deprecated.

PR #3011, opened 2026-07-04
Add opt-in dual-stack IPv6 socket support

pi0n00r:
## Summary

Adds an opt-in `--dual-stack` setting for IPv6 socket binding. When enabled with an IPv6 host such as `--host ::`, Uvicorn pre-binds an IPv6 socket with `IPV6_V6ONLY=0` on platforms that support dual-stack sockets, then serves that socket.

The default remains unchanged, so existing deployments that rely on `--host ::` being IPv6-only keep the current behavior.

## Motivation

The current single-worker path delegates host/port binding to `asyncio.create_server()`, which creates an IPv6-only socket. The reload / multi-worker path uses `Config.bind_socket()`, so behavior can differ depending on process mode. This option gives operators an explicit, documented way to request dual-stack behavior without changing default network exposure.

Related: #2945 and discussion #1529.

## Tests

```text
python -m pytest -o addopts="" tests/test_config.py::test_dual_stack_socket_bind tests/test_server.py::test_dual_stack_server_accepts_ipv4_on_ipv6_socket -q
..                                                                       [100%]
2 passed in 0.20s
```

<a href="https://cubic.dev/pr/Kludex/uvicorn/pull/3011?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in `--dual-stack` option, allowing IPv6 sockets to accept IPv4 connections.
  * Dual-stack connections now work over both IPv4 and IPv6 when supported by the platform.
  * Startup fails clearly on platforms without dual-stack IPv6 support.

* **Documentation**
  * Documented the option’s behavior and disabled-by-default setting.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -47,6 +47,7 @@ uvicorn itself.
 * `--port <int>` - Bind to a socket with this port. If set to 0, an available port will be picked. **Default:** *8000*.
 * `--uds <path>` - Bind to a UNIX domain socket, for example `--uds /tmp/uvicorn.sock`. Useful if you want to run Uvicorn behind a reverse proxy.
 * `--fd <int>` - Bind to socket from this file descriptor. Useful if you want to run Uvicorn within a process manager.
+* `--dual-stack` - When binding an IPv6 socket, also accept IPv4 connections using IPv4-mapped IPv6 addresses on platforms that support dual-stack sockets. This changes network exposure for `--host '::'`, so it is opt-in. **Default:** *False*.
 
 ## Development
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -20,7 +20,7 @@ from pytest_mock import MockerFixture
 from tests.custom_loop_utils import CustomLoop
 from tests.utils import as_cwd, get_asyncio_default_loop_per_os
 from uvicorn._types import ASGIApplication, ASGIReceiveCallable, ASGISendCallable, Environ, Scope, StartResponse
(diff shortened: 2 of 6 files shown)


pi0n00r:
Thanks, this was valid. I moved `dual_stack` to the end of `Config.__init__` so existing positional callers keep their current argument mapping, while the CLI and `uvicorn.run(..., dual_stack=True)` keyword path remain unchanged.

I also verified the old positional shape still maps `loop` correctly and reran the focused dual-stack tests:

```text
positional-compat-ok
2 passed in 0.20s
```

pi0n00r:
Rebased onto current main (0.51.0-era, through #3036). The dual-stack path now uses the dedicated startup-failure exit code when the platform cannot provide dual-stack IPv6, with a regression test.\n\nFresh validation: full local suite passed (968 passed, 7 skipped, 100% coverage), formatting/Ruff/mypy passed, and package/docs builds completed. All GitHub matrix legs that reached the code passed; the lone failure was Windows Python 3.14 before tests because setup-uv could not fetch its external version manifest. Windows Python 3.14t passed.

PR #3019, opened 2026-07-05
Split comma-separated `Sec-WebSocket-Protocol` values in the websockets-sansio implementation

Aayush7352:
# Summary

Fixes #3018 (discussion), discovered downstream in falconry/falcon#2678.

Since 0.50.0 the `websockets-sansio` implementation builds the ASGI scope with:

```python
"subprotocols": event.headers.get_all("Sec-WebSocket-Protocol"),
```

`get_all()` returns one string per header *line*, so a client advertising multiple subprotocols in a single comma-separated header (what the `websockets` client library sends) yields `scope["subprotocols"] == ["amqp, wamp"]` instead of `["amqp", "wamp"]`. An app that then accepts `scope["subprotocols"][0]` echoes the malformed value back, and spec-compliant clients reject the handshake with `InvalidHeader: invalid Sec-WebSocket-Protocol header: multiple values: amqp, wamp`.

This PR splits header values on commas, mirroring the legacy `websockets` implementation ([`websockets_impl.py`](https://github.com/Kludex/uvicorn/blob/main/uvicorn/protocols/websockets/websockets_impl.py#L175-L177)). `wsproto` already parses subprotocols natively, so only the sansio implementation needed the change.

The existing `test_subprotocols` test now also asserts the scope contents; it fails on `main` for the sansio implementation (`['proto1, proto2']`) and passes with this fix. Root cause analysis credit: @Deep070203 in falconry/falcon#2678.

# Checklist

- [x] I've updated the documentation accordingly. (No docs change needed — behavior now matches the documented ASGI spec and the other implementations.)

<a href="https://cubic.dev/pr/Kludex/uvicorn/pull/3019?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -658,8 +658,13 @@ async def test_subprotocols(
     subprotocol: str,
     unused_tcp_port: int,
 ):
+    advertised_subprotocols: list[str] | None = None
+
     class App(WebSocketResponse):
         async def websocket_connect(self, message: WebSocketConnectEvent):
+            nonlocal advertised_subprotocols
+            assert self.scope["type"] == "websocket"
+            advertised_subprotocols = list(self.scope["subprotocols"])
             await self.send({"type": "websocket.accept", "subprotocol": subprotocol})
 
     async def get_subprotocol(url: str):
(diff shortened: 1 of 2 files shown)


Kludex (approved):

Kludex:
Thanks!

PR #3024, opened 2026-07-06
Report worker readiness through the healthcheck pong

Kludex:
## Why

The healthcheck pong only proved a worker *booted*: the pong thread starts before `Server.run()`, so a worker answered `is_alive` before its app had imported or lifespan startup had finished. Zero-downtime restart (#3025) needs to know when a worker is actually *serving*, not merely alive.

## How

`Process` owns its `Server` (created lazily in the `server` property) and the pong now reports `Server.started`. The healthcheck splits into two pure queries over one round-trip:

- `ping()` - liveness: the worker answered.
- `is_ready()` - readiness: the worker answered *and* its server finished startup.

No stored `ready` flag; readiness is queried, not latched (keeps command/query separation).

## Is it backwards compatible?

Yes for the public API - the change is internal to `uvicorn.supervisors`. `Process`/`Multiprocess` drop the `target` parameter (each `Process` builds its own `Server` so the pong can read `started`); the documented entrypoints (`uvicorn.run`, the CLI) are unaffected.

### DIFF
--- a/uvicorn/supervisors/multiprocess.py
+++ b/uvicorn/supervisors/multiprocess.py
@@ -35,7 +35,7 @@ class Process:
         self.process = get_subprocess(config, self.target, sockets)
 
     def _healthcheck(self, timeout: float) -> bool | None:
-        """Ping the worker and return its server-started flag, or None if it never answered."""
+        """Return the worker's reported startup flag, or None if it does not answer within timeout seconds."""
         try:
             self.parent_conn.send(b"ping")
             if self.parent_conn.poll(timeout):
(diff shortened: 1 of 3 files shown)


uvicorn/supervisors/multiprocess.py
```diff
@@ -64,8 +68,9 @@ def target(self, sockets: list[socket] | None = None) -> Any:  # pragma: no cove
                 lambda sig, frame: signal.raise_signal(signal.SIGTERM),
             )
 
+        self._server = Server(config=self.config)
```
Kludex: Instead of this, can we have the creation logic inside the property? If _server is none, we do this.

Kludex: Already done in `6f613fc` - the worker `Server` is created lazily inside the `server` property (`if self._server is None: ...`).

uvicorn/supervisors/multiprocess.py
```diff
@@ -27,34 +26,44 @@ class Process:
...
-            return False
+            return None
+
+    def ping(self, timeout: float = 5) -> bool:
+        # Liveness: the worker answered the healthcheck within `timeout`.
+        return self._healthcheck(timeout) is not None
+
+    def is_ready(self, timeout: float = 5) -> bool:
+        # Readiness: the worker answered and its server finished startup.
+        return self._healthcheck(timeout) is True
```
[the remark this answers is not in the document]
Kludex: Not applying this one. `is_ready()` is only ever called from `wait_until_ready()`, which already does `if not self.process.is_alive(): return False` immediately before it - so a dead worker can't be reported ready in practice. Adding the guard here would also break `test_process_ready`, which drives the pipe in-process without starting a real subprocess (so `self.process.is_alive()` is False there).

PR #3041, opened 2026-07-28
Remove duplicate content headers from WebSocket denial responses on websockets-sansio

### LINKED ISSUE #3040: websockets-sansio: WebSocket Denial Response emits duplicate Content-Type / Content-Length
### Summary

On the `websockets-sansio` implementation, a WebSocket Denial Response (`websocket.http.response.*`) is written with uvicorn's own synthesized `Content-Type` and `Content-Length` **plus** the application's headers, rather than the application's alone. The `websockets` and `wsproto` implementations emit the same response correctly.

This matters because `--ws auto` resolves to `WebSocketsSansIOProtocol` whenever `websockets` is installed, so it's the default path for most installs.

### Versions

```
uvicorn    0.51.0
websockets 16.1.1
python     3.12.7
```

### Reproduction

Dependency-free ASGI app that rejects the handshake with a JSON 422:

```python
import base64, os, socket, threading, time
import uvicorn

BODY = b'{"detail":"nope"}'

    await receive()  # websocket.connect
    await send({
        "type": "websocket.http.response.start",
        "status": 422,
        "headers": [(b"content-type", b"application/json")],
    })
    await send({"type": "websocket.http.response.body", "body": BODY})

def serve(port, impl):
    cfg = uvicorn.Config(app, host="127.0.0.1", port=port, log_level="critical", ws=impl)
    s = uvicorn.Server(cfg)
    threading.Thread(target=s.run, daemon=True).start()
    while not s.started:
        time.sleep(0.05)
    return s

def handshake(port):
    key = base64.b64encode(os.urandom(16)).decode()
    req = (f"GET /ws HTTP/1.1\r\nHost: 127.0.0.1:{port}\r\n"
           f"Upgrade: websocket\r\nConnection: Upgrade\r\n"
           f"Sec-WebSocket-Key: {key}\r\nSec-WebSocket-Version: 13\r\n\r\n")
    s = socket.create_connection(("127.0.0.1", port), timeout=5)
    s.sendall(req.encode()); s.settimeout(3)
    chunks = []
    try:
        while True:
            b = s.recv(4096)
            if not b: break
            chunks.append(b)
    except socket.timeout:
        pass
    s.close()
    return b"".join(chunks)

for impl, port in (("websockets", 8811), ("wsproto", 8812), ("websockets-sansio", 8813)):
    srv = serve(port, impl)
    print(f"--- {impl} ---")
    print(handshake(port).split(b"\r\n\r\n")[0].decode())
    print()
    srv.should_exit = True
    time.sleep(0.5)
```

### Actual

```
--- uvicorn --ws websockets ---
HTTP/1.1 422 Unprocessable Entity
content-type: application/json
Date: Mon, 27 Jul 2026 18:37:09 GMT
Content-Length: 17
Connection: close

--- uvicorn --ws wsproto ---
HTTP/1.1 422
content-type: application/json
Transfer-Encoding: chunked

--- uvicorn --ws websockets-sansio ---
HTTP/1.1 422 Unprocessable Entity
Date: Mon, 27 Jul 2026 18:37:10 GMT
Connection: close
Content-Length: 17
Content-Type: text/plain; charset=utf-8      <-- uvicorn's own
content-type: application/json               <-- the application's
```

Note the ordering: a client that takes the first `Content-Type` sees `text/plain; charset=utf-8` for a JSON body.

With a framework response that sets its own `Content-Length` (any Starlette/FastAPI `JSONResponse`, for example), `Content-Length` is duplicated as well:

```
Content-Length: 164
Content-Type: text/plain; charset=utf-8
content-length: 164
content-type: application/json
```

In that form the `websockets` client rejects the response outright with `InvalidMessage` instead of surfacing the status and body.

### Expected

The application's headers only, as the other two implementations produce.

### Cause

`websockets_sansio_impl.py`, in the `websocket.http.response.body` branch:

```python
response = self.conn.reject(self.initial_response[0], body.decode())
response.headers.update(self.initial_response[1])
```

`conn.reject()` builds a complete response, deriving `Content-Type: text/plain; charset=utf-8` and `Content-Length` from the body text. `websockets.datastructures.Headers.update()` then **appends** rather than replaces:

```python
>>> h = Headers()
>>> h["Content-Type"] = "text/plain; charset=utf-8"
>>> h.update([("content-type", "application/json")])
>>> list(h.raw_items())
[('Content-Type', 'text/plain; charset=utf-8'), ('content-type', 'application/json')]
```

so both sets reach the wire. `websockets_impl.py` and `wsproto_impl.py` build the response from the application's headers only, which is why they're unaffected.

### Note

I haven't sent a PR because the right fix depends on a preference I don't want to guess at â€” deleting uvicorn's synthesized content headers before `update()`, versus constructing the rejection so they're never generated. Happy to put one together if you'd like to indicate which.

cyphercodes:
# Summary

On the `websockets-sansio` implementation, a WebSocket Denial Response was built with `conn.reject()`, which synthesizes its own `Content-Length` and `Content-Type` headers, and the application's headers were then appended with `Headers.update()` — which appends rather than replaces. The result was duplicate `Content-Type`/`Content-Length` headers on the wire, an invalid response that strict clients reject outright.

`conn.reject()` is not a good fit for this path: it only accepts a text body (`body.decode()` raised `UnicodeDecodeError` for valid ASGI byte bodies that aren't UTF-8, dropping the connection without sending the denial response at all) and its synthesized headers have to be undone afterwards. The response is now constructed directly from the application's status, headers, and bytes body, with `Date`, `Connection`, `Content-Length`, and `Content-Type` added via `setdefault` only when the application didn't provide them — matching the behavior of the `websockets` and `wsproto` implementations.

Fixes #3040

## Tests

- `test_server_reject_connection_with_custom_content_headers`: application-provided `Content-Type`/`Content-Length` appear exactly once, across all ws/http protocol combinations.
- `test_server_reject_connection_with_non_utf8_body`: a denial response with a non-UTF-8 body is delivered intact.

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -932,6 +932,33 @@ async def test_server_reject_connection_with_non_utf8_body(
         await websocket_session(f"ws://127.0.0.1:{unused_tcp_port}")
 
 
+async def test_server_reject_connection_with_non_standard_status(
+    ws_protocol_cls: WSProtocol, http_protocol_cls: HTTPProtocol, unused_tcp_port: int
+):
+    body = b"custom error"
+
+    async def app(scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable):
+        assert scope["type"] == "websocket"
+        await receive()
+        await send(
+            {
+                "type": "websocket.http.response.start",
+                "status": 599,
+                "headers": [(b"content-type", b"text/plain")],
+            }
+        )
+        await send({"type": "websocket.http.response.body", "body": body})
+
+    async def websocket_session(url: str):
+        response = await wsresponse(url)
+        assert response.status_code == 599
+        assert response.content == body
+
+    config = Config(app=app, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off", port=unused_tcp_port)
+    async with run_server(config):
+        await websocket_session(f"ws://127.0.0.1:{unused_tcp_port}")
+
+
 async def test_server_reject_connection_with_multibody_response(
     ws_protocol_cls: WSProtocol, http_protocol_cls: HTTPProtocol, unused_tcp_port: int
 ):
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -932,33 +932,6 @@ async def test_server_reject_connection_with_non_utf8_body(
         await websocket_session(f"ws://127.0.0.1:{unused_tcp_port}")
 
 
-async def test_server_reject_connection_with_non_standard_status(
-    ws_protocol_cls: WSProtocol, http_protocol_cls: HTTPProtocol, unused_tcp_port: int
-):
-    body = b"custom error"
-
-    async def app(scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable):
-        assert scope["type"] == "websocket"
-        await receive()
-        await send(
-            {
-                "type": "websocket.http.response.start",
-                "status": 599,
-                "headers": [(b"content-type", b"text/plain")],
-            }
-        )
-        await send({"type": "websocket.http.response.body", "body": body})
-
-    async def websocket_session(url: str):
-        response = await wsresponse(url)
-        assert response.status_code == 599
-        assert response.content == body
-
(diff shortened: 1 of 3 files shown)


Kludex (approved):

Kludex:
Thanks!

Issue #3047, opened 2026-07-30
`websockets-sansio` implementation aborts connection with RST on server-initiated close, truncating in-flight data


Kludex:
### Discussed elsewhere

(none)

### Summary

Since 0.50.0, `--ws auto` resolves to `websockets-sansio` when `websockets` is installed. The sansio implementation handles a server-initiated close differently from both the legacy `websockets` implementation and `wsproto`: on `websocket.close` it writes the close frame and calls `transport.close()` immediately, without waiting for the client's close reply.

Per RFC 6455 §5.5.1 a conforming client must reply with its own close frame. That reply now arrives at a socket the server has already closed, so the kernel answers with a TCP RST. Two consequences for the client:

1. It always observes an abnormal close (1006 in browsers) instead of the clean 1000 the server sent.
2. If any payload is still unread in the client's kernel receive buffer (large final message, real network RTT), the RST discards it, so the message is **truncated** even though every server-side `send()` succeeded.

This breaks the common "send final frame(s), then close" one-shot stream pattern. The failure is timing-dependent: on fast loopback with a fast reader, the payload usually drains before the close reply arrives, so in-process tests and small-payload tests pass while real clients over a real network fail.

Both `--ws websockets` (legacy, now deprecated) and `--ws wsproto` complete the closing handshake and deliver everything cleanly, so the deprecation currently points users from a correct implementation to an affected one.

### Reproduction

The client is a raw socket standing in for a real client under network latency: it completes the handshake, delays reading (so in-flight data sits in its kernel buffer), replies to the server's close frame, then reads whatever survived.

```python
import asyncio
import contextlib
import socket
import time

import uvicorn

PAYLOAD_SIZE = 20_000_000  # large enough to exceed kernel socket buffers

HANDSHAKE = (
    b"GET / HTTP/1.1\r\n"
    b"Host: 127.0.0.1\r\n"
    b"Upgrade: websocket\r\n"
    b"Connection: Upgrade\r\n"
    b"Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==\r\n"
    b"Sec-WebSocket-Version: 13\r\n\r\n"
)
CLIENT_CLOSE_FRAME = b"\x88\x82\x00\x00\x00\x00\x03\xe8"  # masked close, code 1000


async def app(scope, receive, send):
    await receive()  # websocket.connect
    await send({"type": "websocket.accept"})
[...]

PR #3048, opened 2026-07-30
Add missing write flow control to `websockets-sansio`

Kludex:
## Summary

Addresses the missing flow-control part of #3047. The complete WebSocket closing-handshake behavior is intentionally split into a separate PR.

`WebSocketsSansIOProtocol.send()` awaits `self.writable.wait()`, but the protocol never implemented `pause_writing()` / `resume_writing()`. The event was therefore never cleared and backpressure never engaged.

## Changes

- Add `pause_writing()` and `resume_writing()` to `WebSocketsSansIOProtocol`, matching `wsproto`.
- Set `writable` in the sansio `connection_lost()` path. asyncio doesn’t call `resume_writing()` when a paused transport is lost, so without this a pending ASGI `send()` would remain blocked indefinitely.
- Add deterministic mock-transport tests; no sockets or sleeps:
  - `send()` blocks while writes are paused and resumes afterward, across all WebSocket/HTTP protocol combinations.
  - A paused sansio sender is released when the connection is lost.

## Scope

This PR deliberately does **not** change WebSocket closing-handshake or TCP-close behavior, and it does not modify `wsproto`. Those changes are being split into focused follow-up PRs.

## Verification

- Targeted tests: 9 passed.
- Ruff and mypy clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket send reliability when transport write buffers are under backpressure.
  * Ensured pending sends are released when a connection is lost.
  * Simplified WebSocket connection shutdown so transports close immediately after ASGI handling or an explicit close request.
* **Tests**
  * Updated WebSocket coverage for backpressure handling and connection-loss behavior.

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -1388,39 +1388,32 @@ async def test_server_keepalive_ping_timeout(
             assert exc_info.value.rcvd.reason == "keepalive ping timeout"
 
 
+class MockWriteTransport:
+    """Minimal transport for driving a websocket protocol's write path in-process."""
+
+    def __init__(self) -> None:
+        self.buffer = b""
+        self.closed = False
+
+    def get_extra_info(self, name: str, default: Any = None) -> Any:
+        return {"sockname": ("127.0.0.1", 8000), "peername": ("127.0.0.1", 8001)}.get(name, default)
+
+    def write(self, data: bytes) -> None:
+        self.buffer += data
+
+    def close(self) -> None:
+        self.closed = True
+
+    def is_closing(self) -> bool:
+        return self.closed
+
+
 async def test_send_respects_write_backpressure(ws_protocol_cls: WSProtocol, http_protocol_cls: HTTPProtocol):
     """Test that `send()` blocks while the transport's write buffer is full.
 
     Without backpressure, a server-initiated close can outrun in-flight data.
     See https://github.com/Kludex/uvicorn/issues/3047.
     """
-
-    class MockTransport:
-        def __init__(self) -> None:
-            self.buffer = b""
-            self.closed = False
-
-        def get_extra_info(self, name: str, default: Any = None) -> Any:
-            return {"sockname": ("127.0.0.1", 8000), "peername": ("127.0.0.1", 8001)}.get(name, default)
-
-        def write(self, data: bytes) -> None:
-            self.buffer += data
-
-        def close(self) -> None:
-            self.closed = True
-
-        def is_closing(self) -> bool:
-            return self.closed
-
-        def pause_reading(self) -> None:
-            pass
-
-        def resume_reading(self) -> None:
-            pass
-
-        def set_write_buffer_limits(self, high: int | None = None, low: int | None = None) -> None:
-            pass
-
     handshake = (
         b"GET / HTTP/1.1\r\n"
         b"Host: 127.0.0.1\r\n"
@@ -1445,7 +1438,7 @@ async def test_send_respects_write_backpressure(ws_protocol_cls: WSProtocol, htt
     config = Config(app=app, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off")
     config.load()
     protocol = ws_protocol_cls(config=config, server_state=ServerState(), app_state={})
-    transport = MockTransport()
+    transport = MockWriteTransport()
     protocol.connection_made(transport)  # type: ignore[arg-type]
     protocol.data_received(handshake)
     await accepted.wait()
@@ -1473,33 +1466,6 @@ async def test_connection_lost_unblocks_paused_send(ws_protocol_cls: WSProtocol,
     asyncio never calls resume_writing() on a transport that is lost while paused, so
     connection_lost() must wake pending senders or the ASGI task leaks forever.
     """
-
-    class MockTransport:
-        def __init__(self) -> None:
-            self.buffer = b""
-            self.closed = False
-
-        def get_extra_info(self, name: str, default: Any = None) -> Any:
-            return {"sockname": ("127.0.0.1", 8000), "peername": ("127.0.0.1", 8001)}.get(name, default)
-
-        def write(self, data: bytes) -> None:
-            self.buffer += data
-
-        def close(self) -> None:
-            self.closed = True
-
-        def is_closing(self) -> bool:
-            return self.closed
-
-        def pause_reading(self) -> None:
-            pass
-
-        def resume_reading(self) -> None:
-            pass
-
-        def set_write_buffer_limits(self, high: int | None = None, low: int | None = None) -> None:
-            pass
-
     handshake = (
         b"GET / HTTP/1.1\r\n"
         b"Host: 127.0.0.1\r\n"
@@ -1526,7 +1492,7 @@ async def test_connection_lost_unblocks_paused_send(ws_protocol_cls: WSProtocol,
     config = Config(app=app, ws=ws_protocol_cls, http=http_protocol_cls, lifespan="off")
     config.load()
     protocol = ws_protocol_cls(config=config, server_state=ServerState(), app_state={})
-    transport = MockTransport()
+    transport = MockWriteTransport()
     protocol.connection_made(transport)  # type: ignore[arg-type]
     protocol.data_received(handshake)
     await accepted.wait()
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -1397,8 +1397,6 @@ WS_HANDSHAKE_REQUEST = (
     b"Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==\r\n"
     b"Sec-WebSocket-Version: 13\r\n\r\n"
 )
-CLIENT_CLOSE_FRAME = b"\x88\x82\x00\x00\x00\x00\x03\xe8"  # masked close, code 1000
-CLIENT_TEXT_FRAME = b"\x81\x81\x00\x00\x00\x00x"  # masked text frame, payload "x"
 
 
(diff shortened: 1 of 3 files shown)


Kludex:
I tested this branch (60d6a4e) and it does fix the case from #3047, but the "send final frame(s), then `websocket.close()`" pattern still ends in an RST for payloads below the backpressure threshold.

The repro in #3047 uses 20 MB, which is big enough that `pause_writing` engages and the app's close is therefore delayed past the client's close reply. With smaller messages the write never pauses, `send()` never blocks, and `websocket.close` still reaches `transport.close()` in the same event-loop iteration that wrote the close frame (`websockets_sansio_impl.py:476`; note `run_asgi` closes unconditionally too). The client's close reply cannot arrive sooner than one RTT, so it lands on a closed socket: RST, 1006, and loss of whatever the client had not yet read.

Same shape as the #3047 repro, parametrized over message size and the client's close-reply latency:

```python
import asyncio, contextlib, socket, struct, time
import uvicorn

SIZES = [9_000, 900_000, 5_000_000]
DELAYS = [0.001, 0.020]  # one RTT: our close reply cannot arrive sooner

HANDSHAKE = (
    b"GET / HTTP/1.1\r\nHost: 127.0.0.1\r\nUpgrade: websocket\r\nConnection: Upgrade\r\n"
    b"Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==\r\nSec-WebSocket-Version: 13\r\n\r\n"
)
CLIENT_CLOSE_FRAME = b"\x88\x82\x00\x00\x00\x00\x03\xe8"  # masked close, code 1000
size = 0

    await receive()
    await send({"type": "websocket.send", "text": "x" * size})
    await send({"type": "websocket.close", "code": 1000})

def parse(buf):
    """(payload bytes, close code) of the frames that survived."""
    i = payload = 0
    code = None
    while i + 2 <= len(buf):
        opcode, length = buf[i] & 0x0F, buf[i + 1] & 0x7F
        i += 2
        if length == 126:
            (length,) = struct.unpack(">H", buf[i : i + 2]); i += 2
        elif length == 127:
            (length,) = struct.unpack(">Q", buf[i : i + 8]); i += 8
        body = buf[i : i + length]
        if len(body) < length:
            break
        i += length
        if opcode == 0x1:
            payload += length
        elif opcode == 0x8 and length >= 2:
            (code,) = struct.unpack(">H", body[:2])
    return payload, code

def run_client(port, delay):
    s = socket.create_connection(("127.0.0.1", port))
    s.setsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF, 65536)
    s.sendall(HANDSHAKE)
    while b"\r\n\r\n" not in s.recv(8192, socket.MSG_PEEK):
        pass
    rest = s.recv(8192).split(b"\r\n\r\n", 1)[1]
    time.sleep(delay)
    s.sendall(CLIENT_CLOSE_FRAME)  # conforming close reply, RFC 6455 5.5.1
    time.sleep(delay)
    buf, reset = bytearray(rest), False
    try:
        while chunk := s.recv(65536):
            buf += chunk
    except ConnectionResetError:
        reset = True
    finally:
        s.close()
    payload, code = parse(bytes(buf))
    observed = 1006 if reset else code  # what a browser reports
    ok = payload >= size and observed == 1000
    print(f"  {'OK  ' if ok else 'FAIL'} size={size:>10,} reply_delay={delay * 1000:>5.1f}ms  "
          f"received={payload:>10,}  close={observed}")

async def main():
    global size
    for ws_impl in ["websockets", "websockets-sansio", "wsproto"]:
        print(f"--- ws={ws_impl} ---")
        server = uvicorn.Server(uvicorn.Config(app, port=0, ws=ws_impl, log_level="critical"))
        while not server.started:
            await asyncio.sleep(0.01)
        port = server.servers[0].sockets[0].getsockname()[1]
        for size in SIZES:
            for delay in DELAYS:
                await asyncio.to_thread(run_client, port, delay)
        server.should_exit = True
        with contextlib.suppress(asyncio.CancelledError, TimeoutError):
            await asyncio.wait_for(task, 5)

asyncio.run(main())
```

On this branch (identical output across 3 runs):

```
--- ws=websockets ---
  OK   size=     9,000 reply_delay=  1.0ms  received=     9,000  close=1000
  OK   size=     9,000 reply_delay= 20.0ms  received=     9,000  close=1000
  OK   size=   900,000 reply_delay=  1.0ms  received=   900,000  close=1000
  OK   size=   900,000 reply_delay= 20.0ms  received=   900,000  close=1000
  OK   size= 5,000,000 reply_delay=  1.0ms  received= 5,000,000  close=1000
  OK   size= 5,000,000 reply_delay= 20.0ms  received= 5,000,000  close=1000
--- ws=websockets-sansio ---
  FAIL size=     9,000 reply_delay=  1.0ms  received=     9,000  close=1006
  FAIL size=     9,000 reply_delay= 20.0ms  received=     9,000  close=1006
  FAIL size=   900,000 reply_delay=  1.0ms  received=         0  close=1006
  FAIL size=   900,000 reply_delay= 20.0ms  received=         0  close=1006
  OK   size= 5,000,000 reply_delay=  1.0ms  received= 5,000,000  close=1000
  OK   size= 5,000,000 reply_delay= 20.0ms  received= 5,000,000  close=1000
--- ws=wsproto ---
  FAIL size=     9,000 reply_delay=  1.0ms  received=     9,000  close=1006
  FAIL size=     9,000 reply_delay= 20.0ms  received=     9,000  close=1006
  FAIL size=   900,000 reply_delay=  1.0ms  received=         0  close=1006
  FAIL size=   900,000 reply_delay= 20.0ms  received=         0  close=1006
  OK   size= 5,000,000 reply_delay=  1.0ms  received= 5,000,000  close=1000
  OK   size= 5,000,000 reply_delay= 20.0ms  received= 5,000,000  close=1000
```

So with the branch, sansio now matches wsproto, which is what the PR set out to do. But both still fail the case legacy `websockets` handles: a 9 KB message is delivered yet the close is abnormal, and at 900 KB with 1 ms of reply latency the whole message is discarded. Only at 5 MB, where backpressure engages, is it clean. Flow control makes the failure size-dependent; it does not remove it.

To confirm what is left, subclassing this branch's protocol so `websocket.close` waits for the peer's close frame before closing the transport makes every row above pass, at all sizes and latencies:

```python
class HandshakeWaitProtocol(WebSocketsSansIOProtocol):
    async def send(self, message):
        if message.get("type") == "websocket.close" and self.handshake_complete and not self.close_sent:
            await self.writable.wait()
            code, reason = message.get("code", 1000), message.get("reason", "") or ""
            self.queue.put_nowait({"type": "websocket.disconnect", "code": code, "reason": reason})
            self.conn.send_close(code, reason)
            self.transport.write(b"".join(self.conn.data_to_send()))
            self.close_sent = True
            with contextlib.suppress(TimeoutError):
                await asyncio.wait_for(self.close_reply_received.wait(), 5.0)
            self.transport.close()
            return
        await super().send(message)

    def handle_close(self, event):
        self.close_reply_received.set()
        super().handle_close(event)

        self.close_reply_received.set()
        super().connection_lost(exc)
```

Worth considering the closing-handshake wait for sansio (and wsproto) as well: as it stands, deprecating `--ws websockets` points users at implementations where any server-initiated close after a final frame is unclean, which browsers report as 1006 and, when data is still in flight, as truncation.

Environment: uvicorn @ 60d6a4e, websockets 16.0, wsproto 1.3.2, CPython 3.14, macOS.

Kludex:
You're right, and thanks for the matrix — I reproduced it exactly on 60d6a4e, including the part I had wrong:

```
--- ws=websockets-sansio ---
  FAIL size=   900,000 reply_delay=  1.0ms  received=         0  close=1006
```

At 900 KB the client loses the **whole** message. Earlier in this PR I argued that below the backpressure threshold the failure was only a cosmetic 1006 because asyncio flushes before FIN — that was wrong, and it's why I dropped the closing-handshake wait. Flow control only defers the close while the write buffer is above the high water mark, so it fixed the 20 MB repro in #3047 and left everything smaller broken.

Pushed fa10f9b, which defers the transport close instead of blocking `send()`:

- After writing the close frame, arm a timer and leave the transport open; `handle_close` cancels it and closes once the peer echoes the frame (RFC 6455 §5.5.1). If the reply never comes, the timer closes anyway after 10s, matching the `websockets` library default and permitted by §7.1.1.
- Reads are resumed when the wait starts: they stay paused until the app consumes a message, so an app that closes without consuming one would otherwise never read the reply it is waiting for (this was also Codex's P2 note).
- `run_asgi` no longer closes the transport while a handshake is pending, and `connection_lost` cancels the timer.

I used a timer rather than awaiting inside `send()` (as in your prototype) so `send()` stays non-blocking: awaiting there holds the ASGI task for the whole handshake, which showed up immediately as every `tests/benchmarks/test_ws.py` case taking 10s instead of 0.02s. With the timer the app returns straight after `websocket.close` and only the transport lingers.

Your matrix on fa10f9b, all 18 rows, all three implementations:

```
--- ws=websockets ---        OK × 6
--- ws=websockets-sansio --- OK × 6
--- ws=wsproto ---           OK × 6
```

The #3047 repro also still passes. wsproto gets the same treatment, since it had the identical race once the payload was below the threshold.

New tests, all mock-transport based and deterministic (the timeout path fires the timer callback directly rather than sleeping): close waits for the client's reply; gives up when the reply never comes; resumes paused reads while waiting; and cancels the pending timer when the connection is lost. Suite is 1215 passed with 100% coverage, and the ws benchmarks are back to 0.02s.

Given this, #3049 (the follow-up I filed for the small-payload race) is now fully covered here — happy to close it as part of this PR, or leave it open to track the `shutdown()` 1012 path, which still closes immediately and is not addressed here.

Issue #3049, opened 2026-07-30
WebSocket server-initiated close races the client's close reply when the write buffer is empty (small payloads, shutdown)


Kludex:
### Discussed elsewhere

Follow-up from #3047 / #3048. Affects both `wsproto` and `websockets-sansio`, and predates 0.50.0.

### Summary

Both implementations write the server's close frame and call `transport.close()` immediately. When the transport's write buffer is under the high-water mark (small or no in-flight payload), nothing delays the close, so a conforming client's close reply (RFC 6455 §5.5.1) arriving one RTT later hits an already-closed socket and the kernel answers with a TCP RST. The client observes an abnormal close (1006 in browsers) instead of the code the server sent.

#3048 fixed the large-payload data-truncation case in `websockets-sansio` by adding the missing write flow control: backpressure delays the close until the buffer drains, matching `wsproto`. But that protection is incidental — it only engages when the write buffer is full. Two paths remain exposed in **both** implementations:

1. **Small payloads + real RTT**: app sends a small message then `websocket.close`; the close frame and `transport.close()` execute back to back. No data is lost (asyncio flushes before FIN), but the client's close reply gets an RST and the client reports 1006 instead of a clean close.
2. **Graceful shutdown**: `shutdown()` sends close code 1012 and closes the transport immediately, same race (noted in review of #3048).

The legacy `websockets` implementation is unaffected: it completes the closing handshake, waiting for the client's close frame with `close_timeout=10`.

### Reproduction

Deterministic with a 5-byte payload — 5/5 RST on `wsproto`, while (post-#3048) `websockets-sansio` only avoids it in the large-payload case:

```python
import asyncio, contextlib, socket, time
import uvicorn

PAYLOAD = "hello"
HANDSHAKE = (b"GET / HTTP/1.1\r\nHost: 127.0.0.1\r\nUpgrade: websocket\r\nConnection: Upgrade\r\n"
    b"Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==\r\nSec-WebSocket-Version: 13\r\n\r\n")
CLOSE = b"\x88\x82\x00\x00\x00\x00\x03\xe8"  # masked close, code 1000

async def app(scope, receive, send):
    await receive()
    await send({"type": "websocket.accept"})
    await send({"type": "websocket.send", "text": PAYLOAD})
    await send({"type": "websocket.close", "code": 1000})

def run_client(port):
    s = socket.create_connection(("127.0.0.1", port))
[...]

PR #3051, opened 2026-07-30
Complete server-initiated WebSocket closing handshakes

Kludex:
## Summary

Stacked on #3048. Fixes the remaining server-initiated close race from #3047 and #3049 independently of the sansio flow-control change.

Both SansIO implementations currently write a close frame and close the transport immediately. The peer’s required close reply then lands on a closed socket, producing a TCP RST that can discard unread payload. This affects payloads below the write-backpressure threshold as well as larger messages.

## Changes

- Keep the transport open after a server-initiated close frame.
- Close immediately when the peer’s close frame arrives.
- Fall back to closing after 10 seconds if the peer never replies, matching the `websockets` library default.
- Resume paused reads while waiting, otherwise a close reply can remain unread behind an unconsumed ASGI receive event.
- Cancel pending close fallback on connection loss.
- Apply the behavior to both `websockets-sansio` and `wsproto`.

## Tests

Deterministic mock-transport coverage for:

- waiting for the peer close reply;
- the bounded no-reply fallback;
- resuming paused reads;
- cancelling the fallback on connection loss.

The payload-size/RTT matrix from the review comment passes all 18 combinations across legacy `websockets`, `websockets-sansio`, and `wsproto`.

## Stack

- #3048: sansio write flow control only.
- #3050: independent wsproto paused-sender cleanup.
- This PR: closing-handshake lifecycle only.

Draft while we decide whether the bounded transport-close logic is the right abstraction.

<a href="https://cubic.dev/pr/Kludex/uvicorn/pull/3051?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/protocols/test_websocket.py
+++ b/tests/protocols/test_websocket.py
@@ -1397,6 +1397,8 @@ WS_HANDSHAKE_REQUEST = (
     b"Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==\r\n"
     b"Sec-WebSocket-Version: 13\r\n\r\n"
 )
+CLIENT_CLOSE_FRAME = b"\x88\x82\x00\x00\x00\x00\x03\xe8"  # masked close, code 1000
+CLIENT_TEXT_FRAME = b"\x81\x81\x00\x00\x00\x00x"  # masked text frame, payload "x"
 
 
 class MockWriteTransport:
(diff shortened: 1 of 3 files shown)


Kludex:
Superseded by a smaller websockets-sansio-only closing-handshake PR, as discussed.

PR #3060, opened 2026-08-08
Add ipv6_v6only config option for consistent dual-stack behavior

### LINKED ISSUE #2945: Inconsistent IPv6 dual-stack behavior between single-worker and multi-worker modes

pranjalm37:
## Summary

Fixes #2945.

Binding an IPv6 host currently behaves inconsistently depending on worker count:

- **Single-worker** (default): `Server.startup()` delegates to `loop.create_server(host=..., port=...)`, and asyncio [unconditionally sets `IPV6_V6ONLY=True`](https://github.com/python/cpython/blob/main/Lib/asyncio/base_events.py) for IPv6 sockets.
- **Multi-worker**: `Config.bind_socket()` creates the socket manually and never touches `IPV6_V6ONLY`, so it inherits the OS default (dual-stack on Linux).

So `uvicorn app:app --host '::'` silently drops IPv4 connectivity by default, but accepts it with `--workers 2+` — as reported, this is a real problem for e.g. rootless Podman/pasta networking where IPv4 forwarding has nowhere to connect if the server is IPv6-only.

## Approach

I deliberately did **not** change either existing default (that's a judgment call — dual-stack-by-default vs IPv6-only-by-default has security/deployment implications either way and isn't mine to make). Instead this adds an explicit, opt-in `ipv6_v6only: bool | None = None` `Config` option (+ `--ipv6-v6only`/`--no-ipv6-v6only` CLI flag) that's a no-op when left unset:

- `bind_socket()` now sets `IPV6_V6ONLY` explicitly whenever `ipv6_v6only is not None` — this already covers the multi-worker path, which calls `bind_socket()` directly.
- The single-worker path in `Server.startup()` now also routes through `bind_socket()` (passing the socket via `sock=` instead of `host=`/`port=`) whenever `ipv6_v6only` is explicitly set, since `asyncio.create_server(host=, port=)` would otherwise silently override any socket option set beforehand.
- `bind_socket()` gained a `log: bool = True` parameter so the single-worker path can pass `log=False` and keep using the existing `_log_started_message()` for its startup banner, avoiding a duplicate log line.

With `ipv6_v6only` left at its default `None`, neither code path's behavior changes at all — purely additive.

## Test plan

- [x] Reproduced the exact bug: single-worker `--host '::'` accepts IPv6 but refuses IPv4 by default; `ipv6_v6only=False` fixes it end-to-end (verified with a real `asyncio.open_connection` over both `127.0.0.1` and `::1`).
- [x] Confirmed the unset default is byte-for-byte unchanged from current behavior (`IPV6_V6ONLY` still ends up `True` for single-worker with no explicit setting).
- [x] Confirmed no duplicate startup log line from routing through `bind_socket(log=False)`.
- [x] Added `test_bind_socket_ipv6_v6only_*` in `tests/test_config.py` (explicit True/False, unset no-op, ignored for IPv4 hosts).
- [x] Added `test_ipv6_v6only_false_accepts_ipv4_in_single_worker_mode` in `tests/test_server.py` — real end-to-end server test over httpx, single-worker mode, both IPv4 and IPv6 clients.
- [x] Full suite: `pytest tests/ --ignore=tests/benchmarks` → 993 passed (up from 988 pre-change), 252 skipped.
- [x] `ruff format --check`, `ruff check`, and `mypy uvicorn tests` all clean.
- [x] Documented the new flag in `docs/settings.md` under Socket Binding.

<a href="https://cubic.dev/pr/Kludex/uvicorn/pull/3060?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added `--ipv6-v6only` and `--no-ipv6-v6only` options to control IPv6 socket behavior.
  * IPv6 servers can explicitly allow or restrict IPv4-compatible connections.
  * Single-worker IPv6 servers can accept both IPv4 and IPv6 requests when IPv6-only mode is disabled.

* **Bug Fixes**
  * Improved handling of explicitly configured IPv6 socket options during server startup.
  * Ensured application shutdown cleanup runs when IPv6 binding fails.

* **Documentation**
  * Documented IPv6-only settings, defaults, and IPv4 compatibility behavior.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -45,7 +45,7 @@ uvicorn itself.
 
 * `--host <str>` - Bind socket to this host. Use `--host 0.0.0.0` to make the application available on your local network. IPv6 addresses are supported, for example: `--host '::'`. **Default:** *'127.0.0.1'*.
 * `--port <int>` - Bind to a socket with this port. If set to 0, an available port will be picked. **Default:** *8000*.
-* `--ipv6-v6only` / `--no-ipv6-v6only` - Explicitly set the `IPV6_V6ONLY` socket option for an IPv6 `--host`, instead of relying on the OS default. Left unset by default, in which case single- and multi-worker mode may bind differently: single-worker mode delegates to `asyncio.loop.create_server()`, which always forces IPv6-only, while multi-worker mode inherits the OS default (dual-stack on Linux). Pass `--no-ipv6-v6only` to make an IPv6 `--host` accept IPv4 connections too, in both modes.
+* `--ipv6-v6only` / `--no-ipv6-v6only` - Explicitly set the `IPV6_V6ONLY` socket option for an IPv6 `--host`, instead of relying on the OS default. Left unset by default, in which case single- and multi-worker mode may bind differently: single-worker mode delegates to `asyncio.loop.create_server()`, which always forces IPv6-only, while multi-worker mode inherits the OS default (commonly, but not always, dual-stack on Linux — configurable via the `net.ipv6.bindv6only` sysctl). Pass `--no-ipv6-v6only` to make an IPv6 `--host` accept IPv4 connections too, in both modes.
 * `--uds <path>` - Bind to a UNIX domain socket, for example `--uds /tmp/uvicorn.sock`. Useful if you want to run Uvicorn behind a reverse proxy.
 * `--fd <int>` - Bind to socket from this file descriptor. Useful if you want to run Uvicorn within a process manager.
 
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -45,6 +45,7 @@ uvicorn itself.
 
 * `--host <str>` - Bind socket to this host. Use `--host 0.0.0.0` to make the application available on your local network. IPv6 addresses are supported, for example: `--host '::'`. **Default:** *'127.0.0.1'*.
 * `--port <int>` - Bind to a socket with this port. If set to 0, an available port will be picked. **Default:** *8000*.
+* `--ipv6-v6only` / `--no-ipv6-v6only` - Explicitly set the `IPV6_V6ONLY` socket option for an IPv6 `--host`, instead of relying on the OS default. Left unset by default, in which case single- and multi-worker mode may bind differently: single-worker mode delegates to `asyncio.loop.create_server()`, which always forces IPv6-only, while multi-worker mode inherits the OS default (dual-stack on Linux). Pass `--no-ipv6-v6only` to make an IPv6 `--host` accept IPv4 connections too, in both modes.
 * `--uds <path>` - Bind to a UNIX domain socket, for example `--uds /tmp/uvicorn.sock`. Useful if you want to run Uvicorn behind a reverse proxy.
 * `--fd <int>` - Bind to socket from this file descriptor. Useful if you want to run Uvicorn within a process manager.
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -276,13 +276,21 @@ def test_bind_socket_ipv6_v6only_unset_leaves_os_default() -> None:
         assert isinstance(sock.getsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY), int)
 
 
-@pytest.mark.skipif(not has_ipv6("::1"), reason="IPV6 not enabled")
+@pytest.mark.skipif(not has_ipv6("::"), reason="IPV6 not enabled")
 @pytest.mark.parametrize("v6only", [True, False])
 def test_bind_socket_ipv6_v6only_explicit(v6only: bool) -> None:
     """
     An explicit `ipv6_v6only` is applied to the bound socket (#2945).
+
+    Uses the wildcard `::` host rather than `::1`: on Linux, binding
+    specifically to the loopback address `::1` silently forces
+    IPV6_V6ONLY=True regardless of the requested socket option, since a
+    loopback-only bind can never be meaningfully dual-stack (there is no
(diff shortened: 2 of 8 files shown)


pranjalm37:
Pushed fixes for everything flagged by review:

**Fixed:**
- **Positional-arg break** (caught independently by both CodeRabbit and cubic — thank you): moved `ipv6_v6only` to the end of `Config.__init__`'s parameter list instead of inserting it after `port`, which was silently shifting every subsequent positional argument. Verified `Config(app, host, port, uds)`-style positional calls now work exactly as before.
- **Missing lifespan cleanup on bind failure**: `config.bind_socket()`'s internal `sys.exit()` wasn't caught by the surrounding `except OSError`, so an occupied IPv6 port skipped `lifespan.shutdown()`. Now catches `SystemExit` too, runs cleanup, then re-raises. Verified with a direct repro (occupied port + fake lifespan tracking whether shutdown fired).
- **Docs wording**: no longer implies Linux is unconditionally dual-stack (it's `net.ipv6.bindv6only`-configurable).
- **Test robustness**: extracted the ad-hoc `_has_ipv6()` from `test_main.py` into a shared `tests/utils.py::has_ipv6()`, used to skip the new IPv6 tests in environments without IPv6.

**CI failure — root cause found and fixed:**
The Linux CI failure on `test_bind_socket_ipv6_v6only_explicit[False]` turned out to be a real, interesting kernel behavior, not a flake. I reproduced it directly on Linux (Docker, `python:3.11-slim`) and isolated it precisely:

```
host='::1' set=False -> before_bind=0 after_bind=1
host='::1' set=True  -> before_bind=1 after_bind=1
host='::'  set=False -> before_bind=0 after_bind=0
host='::'  set=True  -> before_bind=1 after_bind=1
```

Setting `IPV6_V6ONLY=False` takes effect immediately after `setsockopt()`, but the kernel silently resets it back to `True` **at `bind()` time, specifically when binding to `::1`** — not `::`. That's sensible: `::1` is loopback-only and can never be meaningfully dual-stack (there's no IPv4-loopback equivalent of `::1`, unlike `::`/`0.0.0.0`). My test used `::1` and could therefore never observe `ipv6_v6only=False` actually sticking. Switched it to `::`, which is also the actual real-world case from the original bug report. Verified all 6 IPv6-related tests green on real Linux via Docker.

All 993 local tests still pass, `ruff`/`mypy` clean.

pranjalm37:
Looked into the latest CI run — every platform/version job is failing on exactly one **different, unrelated** test each time (e.g. \`test_run_chain_only\` on 3.12-ubuntu, \`test_limit_max_requests_jitter[zttp]\` on 3.12-macos, \`test_request_than_limit_max_requests_warn_log[h11]\` on 3.11-ubuntu, etc.), with the actual functional test counts all passing (1000–1270 tests green in each job). Windows jobs passed cleanly with 0 errors. That pattern — one random unrelated test erroring per run, varying every time — looks like CI resource contention under full parallel load rather than anything in this PR; I don't see how the changes here would cause an unrelated test like `test_run_chain_only` to fail at setup.

I don't have permission to re-run the workflow as an outside contributor. Happy to take another look if a maintainer re-triggers it and something reproducible shows up.

pranjalm37:
Found the real root cause of the CI failures — retracting my earlier "unrelated flakiness" theory, it wasn't that.

`Config.bind_socket()` creates a socket and, on a `bind()` `OSError`, logs the error and calls `sys.exit()` **without closing the socket first** — a pre-existing leak in `uvicorn/config.py`. My new `test_ipv6_v6only_bind_failure_runs_lifespan_shutdown` test is the first test in the suite to actually exercise a real bind failure on this path, so the leaked socket surfaced as a `ResourceWarning` at teardown, which pytest escalates to a hard failure. That's why every job failed identically across all platforms (including Windows, which had been clean before).

Fixed in `293af33`:
- `bind_socket()` now closes `sock` in both `except OSError` branches (UDS and inet) before `sys.exit()`.
- Also addressed cubic's review comment: moved the test's own `blocker` socket setup inside `try/finally` so a bind/listen failure there can't leak either.

Verified locally with `pytest -W error::ResourceWarning` across `test_server.py`, `test_config.py`, `test_main.py` — all pass clean, no warnings.

pranjalm37:
Good catch, and confirmed the hard way: the Windows CI jobs on the un-skip commit hung for ~8.5 minutes and got cancelled rather than failing cleanly — exactly matching your prediction that \`SO_REUSEADDR\` lets both sockets bind on Windows, so \`bind()\` never raises and \`pytest.raises(SystemExit)\` never sees anything.

Reverted in \`7784f59\`: restored the win32 skip on \`test_ipv6_v6only_bind_failure_runs_lifespan_shutdown\`, and instead excluded the corresponding \`except SystemExit\` branch in \`server.py\` from Windows' 100%-coverage requirement via \`# pragma: py-win32\` (same pattern this codebase already uses for other platform-untestable branches, e.g. the UDS path). Verified locally and confirmed Linux/macOS still cover that branch via the non-Windows run of the same test.

PR #3062, opened 2026-08-12
Reject a non-numeric content-length instead of raising ValueError

RavSinghChandan:
`send()` in the httptools implementation validates header names and values and
raises RuntimeError for anything malformed, then hands the content-length
straight to `int()`:

    if HEADER_VALUE_RE.search(value):
        raise RuntimeError("Invalid HTTP header value.")
    ...
    if name == b"content-length" and self.chunked_encoding is None:
        self.expected_content_length = int(value.decode())

HEADER_VALUE_RE only blocks control characters, so `content-length: abc` gets
through it and then blows up as `ValueError: invalid literal for int() with
base 10: 'abc'` from inside the protocol. The client just sees the connection
close — no 500, no response at all.

What made me look: the h11 implementation rejects the exact same value cleanly,
because h11 validates it with `_content_length_re` and raises
`LocalProtocolError: bad Content-Length`. So the two HTTP implementations
behave differently on identical input, and only one of them gives an error you
can act on.

Wrapped the conversion and raised the same RuntimeError already used two lines
above. Behaviour for valid values is unchanged.

Added a parametrised test covering "abc", "" and "1 2". It runs against both
implementations — h11 passes on main already, the three httptools cases fail
on main with the ValueError above. tests/protocols/test_http.py is 172 passed,
ruff check and format clean.

<a href="https://cubic.dev/pr/Kludex/uvicorn/pull/3062?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid `Content-Length` header values—including non-numeric, empty, whitespace, signed, negative, and underscore-containing values—are now rejected consistently.
  - Malformed requests no longer result in an internal server error.
  - Invalid header values now close the connection and are logged appropriately.
  - Protocol errors are reported as clear runtime errors instead of exposing raw parsing exceptions.

### DIFF
--- a/tests/protocols/test_http.py
+++ b/tests/protocols/test_http.py
@@ -370,11 +370,10 @@ async def test_invalid_content_length_value(
     assert b"HTTP/1.1 500 Internal Server Error" not in protocol.transport.buffer
     assert protocol.transport.is_closing()
 
-    errors = [r for r in caplog.records if r.exc_info]
-    assert errors, "the invalid header should have been reported"
-    raised = errors[-1].exc_info[0]
-    assert raised is not ValueError, "content-length must not fail with a bare ValueError"
-    assert issubclass(raised, (RuntimeError, LocalProtocolError))
+    raised = [r.exc_info[1] for r in caplog.records if r.exc_info and r.exc_info[1] is not None]
+    assert raised, "the invalid header should have been reported"
+    assert not isinstance(raised[-1], ValueError), "content-length must not fail with a bare ValueError"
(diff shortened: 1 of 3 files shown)


Kludex:
I don't think this is necessary. It's already rejected.

PR #3069, opened 2026-08-16
Fix Ctrl+C not interrupting a frozen lifespan startup

### LINKED ISSUE #2751: If the lifetime is frozen, the app will not respond to the Ctrl + C
### Initial Checks

### Discussion Link

```Text
https://github.com/Kludex/uvicorn/discussions/2716
```

### Description

When the application hangs during its lifespan, it becomes completely unresponsive to the `Ctrl+C` command (SIGINT), preventing a graceful shutdown. This forces users to resort to using the `kill -9` command.

### Example Code

```Python
import asyncio
from collections.abc import AsyncIterator

import uvicorn
from fastapi import FastAPI

@asynccontextmanager
async def lifespan(_: FastAPI) -> AsyncIterator[None]:
    await asyncio.sleep(5000)  # Simulates hanging operation
    yield

def create_app() -> FastAPI:
    return FastAPI(lifespan=lifespan)

if __name__ == "__main__":
    uvicorn.run(create_app(), host="0.0.0.0", port=8000)
```

```Text
Running uvicorn 0.38.0 with CPython 3.12.11 on Linux
fastapi: 0.121.0
OS: Arch Linux x86_64
Kernel: 6.17.7-arch1-1
```

aleks-drozy:
## Summary

Fixes #2751.

If the ASGI application never yields control back to uvicorn while
processing the `lifespan.startup` event (e.g. it hangs, or sleeps
indefinitely, before its first `yield`), the process becomes
completely unresponsive to `Ctrl+C`/`SIGINT` and can only be killed
with `SIGKILL`.

### Root cause

`Server.startup()` awaits `LifespanOn.startup()`, which blocks on
`await self.startup_event.wait()`. That event is only ever set from
`LifespanOn.send()` (when the app emits `lifespan.startup.complete`/
`.failed`), or from `main()`'s `finally` block once the lifespan task
itself finishes. If the app hangs before yielding, neither of those
ever happens.

Meanwhile, `Server.handle_exit()` (installed via `signal.signal()` in
`capture_signals()`) runs synchronously on `SIGINT`/`SIGTERM`, but it
only ever sets `self.should_exit`/`self.force_exit`. Those flags are
exclusively consulted in `main_loop()`/`on_tick()`, which is never
reached until `startup()` returns — so a signal received during a
frozen startup was silently swallowed.

### Fix

- `LifespanOn` now keeps a reference to its own lifespan task
  (`self.main_lifespan_task`) and exposes `cancel()`, which cancels
  that task only if the startup event hasn't been set yet (i.e.
  startup is still genuinely pending). Cancelling it makes `main()`'s
  existing `except BaseException` / `finally` block run, which already
  sets both `startup_event` and `shutdown_event`, unblocking the
  pending wait.
- `LifespanOff.cancel()` is a no-op, for symmetry.
- `Server.handle_exit()` now calls `self.lifespan.cancel()` (guarded
  for the case a signal arrives before `self.lifespan` is even
  assigned).

Because `cancel()` only acts while `startup_event` isn't set yet, this
is a no-op once startup has completed — the well-established shutdown
signal handling for a running server (main loop / graceful shutdown)
is unaffected.

## Testing

Added `test_handle_exit_cancels_frozen_lifespan_startup` in
`tests/test_server.py`: builds a `Server` whose app hangs forever right
after receiving `lifespan.startup`, drives startup as a background
task, and calls `server.handle_exit(signal.SIGINT, None)` shortly
after — exactly what the real signal handler does. Verified this test
**fails** (`asyncio.TimeoutError` after a bounded 3s wait, proving the
hang) with the source fix reverted and the test unchanged, and
**passes** in ~1-2s with the fix restored.

Ran the directly relevant slice locally (Windows, Python 3.11):

```
pytest tests/test_server.py tests/test_lifespan.py
28 passed, 4 skipped in 8.38s
```

Also ran `ruff check`, `ruff format --check`, and `mypy` against the
changed files — no new issues (the few pre-existing `mypy` findings in
`server.py` are unrelated Windows-platform socket/signal typing issues
present before this change).

### Test plan
- [x] New regression test added, confirmed to fail before the fix and pass after
- [x] `tests/test_server.py` and `tests/test_lifespan.py` pass locally
- [x] `ruff check` / `ruff format --check` clean on changed files
- [x] No CHANGELOG.md entry — per `docs/contributing.md`, that's assembled by maintainers as part of the release PR, not per-contribution

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<a href="https://cubic.dev/pr/Kludex/uvicorn/pull/3069?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Server shutdown now reliably interrupts applications stuck during startup, allowing the server to exit instead of hanging.
  - Signal-triggered shutdown now properly cancels in-progress lifespan startup operations.
  - Interrupted startup is reported as a failure rather than a successful startup, with shutdown completing cleanly.

- **Tests**
  - Added regression coverage for startup cancellation before execution and while blocked.
  - Verified interrupted startup updates server state and completes shutdown within the expected timeout.

### DIFF
--- a/tests/test_lifespan.py
+++ b/tests/test_lifespan.py
@@ -50,6 +50,7 @@ def test_lifespan_off():
 
         await lifespan.startup()
         await lifespan.shutdown()
+        lifespan.cancel()
 
     loop = asyncio.new_event_loop()
     loop.run_until_complete(test())
--- a/tests/test_lifespan.py
+++ b/tests/test_lifespan.py
@@ -1,4 +1,5 @@
 import asyncio
+import logging
 
 import pytest
 
@@ -156,6 +157,70 @@ def test_lifespan_with_failed_startup(mode, raise_exception, caplog):
     ]
     assert "the lifespan event failed" in error_messages.pop(0)
     assert "Application startup failed. Exiting." in error_messages.pop(0)
+
+
+def test_lifespan_cancel_before_main_task_runs():
+    """cancel() must unblock startup() even if the lifespan task is cancelled
+    before its coroutine ever runs, in which case main()'s try/finally never
+    executes. Regression test for https://github.com/encode/uvicorn/issues/2751."""
+
+    async def app(scope, receive, send):  # pragma: no cover
+        pass
+
+    async def test():
+        config = Config(app=app, lifespan="auto")
+        lifespan = LifespanOn(config)
+
+        loop = asyncio.get_event_loop()
+        lifespan.main_lifespan_task = loop.create_task(lifespan.main())
+        # Cancel before the event loop has ever scheduled main() to run.
+        lifespan.cancel()
+
+        # Must complete instead of hanging forever.
+        await asyncio.wait_for(lifespan.startup_event.wait(), timeout=3)
+        assert lifespan.shutdown_event.is_set()
+        assert lifespan.error_occurred
+        assert lifespan.startup_failed
+
+    loop = asyncio.new_event_loop()
+    loop.run_until_complete(test())
+    loop.close()
+
+
+def test_lifespan_cancel_while_frozen_in_startup(caplog):
+    """Cancelling an app frozen inside lifespan startup must report the startup
+    as interrupted/failed, not 'Application startup complete.'."""
+    caplog.set_level(logging.INFO, logger="uvicorn.error")
+
+    async def app(scope, receive, send):
+        message = await receive()
+        assert message["type"] == "lifespan.startup"
+        await asyncio.sleep(30)  # pragma: no cover  # Never yields control back.
+
+    async def test():
+        config = Config(app=app, lifespan="auto")
+        lifespan = LifespanOn(config)
+
+        startup_task = asyncio.create_task(lifespan.startup())
+        # Let the app reach its frozen await inside startup handling.
+        for _ in range(10):
+            await asyncio.sleep(0)
+        assert not lifespan.startup_event.is_set()
+
+        lifespan.cancel()
+        await asyncio.wait_for(startup_task, timeout=3)
+
+        assert lifespan.error_occurred
+        assert lifespan.startup_failed
+        assert lifespan.should_exit
+
+    loop = asyncio.new_event_loop()
+    loop.run_until_complete(test())
+    loop.close()
+    messages = [record.message for record in caplog.records if record.name == "uvicorn.error"]
+    assert "Application startup interrupted." in messages
+    assert "Application startup failed. Exiting." in messages
+    assert "Application startup complete." not in messages
 
 
 def test_lifespan_scope_asgi3app():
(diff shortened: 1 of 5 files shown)


aleks-drozy:
Heads up on the required `check` status above — it's currently red only because the Python 3.10/3.11 matrix jobs on the latest commit (23b8d0f) show `cancelled`, not failed (looks like they got superseded mid-run by the rapid follow-up pushes). Every job that actually ran (3.12/3.13/3.14/3.14t across all three OSes) passed. I don't have permission to re-run the workflow myself — could a maintainer trigger a re-run when convenient?

aleks-drozy:
Correction on my earlier comment: I was wrong that the red check on 23b8d0f was caused by rapid follow-up pushes superseding the run. I pulled the actual CI logs and there was only a single workflow run against that commit; its Python 3.10/3.11 jobs on all three OSes ran for roughly 10 minutes (versus 1-2 minutes for 3.12+) before ending in cancelled right as the required check gate reported failure. That is a hang, not a supersede-by-push.

Root cause: this test uses asyncio.wait_for(server.serve(), timeout=3). On Python <3.12, wait_for() runs the awaited coroutine in a child Task. When server.serve() raises SystemExit (a BaseException) inside that child Task, Task.__step re-raises it a second time into the event loop's callback dispatch, which can race against wait_for()'s own result handling while another task (interrupt_during_startup) is still pending. Python 3.12's wait_for() was rewritten on top of asyncio.timeout() and awaits the coroutine directly on the caller's task, so it isn't exposed to this race. That's a real, version-dependent CPython limitation, not a bug in LifespanOn.cancel(), _on_cancelled_task_done(), or Server.handle_exit() - I verified all three behave correctly on every Python version tested when awaited without an intervening wait_for()-created child Task, including a minimal reproduction independent of any uvicorn code.

Fix (test-only, just pushed): the test no longer wraps server.serve() in asyncio.wait_for(). It instead schedules a call_later() that cancels the current task after 3s and awaits server.serve() directly on the test's own task, so SystemExit propagates through normal coroutine unwinding into pytest.raises. No production code changed.

Verified: fail-before reproduced on the parent commit (SystemExit escapes pytest.raises, matching the CI hang), pass-after confirmed on Python 3.11 and 3.12, and the full tests/test_server.py + tests/test_lifespan.py suite passes (30 passed, 4 skipped, skips are pre-existing and unrelated).

PR #3070, opened 2026-08-16
Fix inconsistent IPv6 dual-stack behavior between single-worker and multi-worker modes

### LINKED ISSUE #2945: Inconsistent IPv6 dual-stack behavior between single-worker and multi-worker modes

aleks-drozy:
## Summary

Fixes #2945.

A server bound to an IPv6 wildcard host (`--host ::`) behaved inconsistently depending on how it was run:

- **Multi-worker / `--reload`**: `Config.bind_socket()` creates the socket manually and only sets `SO_REUSEADDR`. It never sets `IPV6_V6ONLY`, so dual-stack behavior silently depended on the OS default (e.g. Linux's `net.ipv6.bindv6only` sysctl, commonly `0` → dual-stack).
- **Single worker (default)**: `Server.startup()`'s "standard case" delegates to `asyncio.loop.create_server(host=..., port=...)`. Per [`asyncio/base_events.py`](https://github.com/python/cpython/blob/main/Lib/asyncio/base_events.py), asyncio *always* sets `IPV6_V6ONLY=True` on `AF_INET6` sockets it creates through this path, regardless of the platform — making the socket IPv6-only unconditionally.

The net effect: on a host where dual-stack is the OS default (e.g. Linux), a single-worker server bound to `::` silently drops IPv4 connectivity, while the same app run with `--workers 2` accepts IPv4 fine — or vice versa depending on the sysctl. This is confusing and, as reported, breaks real-world setups such as rootless Podman with pasta networking, which relies on the server accepting both address families.

## Root cause

- `uvicorn/config.py`, `Config.bind_socket()`: creates the IPv6 socket but never calls `setsockopt(IPPROTO_IPV6, IPV6_V6ONLY, ...)`.
- `uvicorn/server.py`, `Server.startup()` "standard case": relies on `loop.create_server(host=.., port=..)`, whose internal `getaddrinfo`-based socket creation in CPython unconditionally forces `IPV6_V6ONLY=True` for every `AF_INET6` socket, bypassing any platform default.

## Fix

Make both code paths agree on dual-stack-by-default, explicitly and deterministically (not relying on the OS default):

- `Config.bind_socket()` now explicitly sets `IPV6_V6ONLY=False` right after creating an `AF_INET6` socket.
- `Server.startup()`'s single-worker "standard case" now, for IPv6 hosts, binds its own socket (`SO_REUSEADDR` + `IPV6_V6ONLY=False`) and passes it to `loop.create_server(sock=...)` instead of `host=/port=`, bypassing asyncio's `getaddrinfo` path (and its hardcoded `IPV6_V6ONLY=True`) entirely. This is the same technique already used by the reload/multiprocess supervisors via `bind_socket()`.

Non-IPv6 hosts, UDS, and `--fd` paths are untouched.

## Testing

- Added `test_ipv6_socket_bind_is_dual_stack` in `tests/test_config.py`: binds via `Config.bind_socket()` with `host="::"` and asserts `IPV6_V6ONLY == 0`.
- Added `test_run_ipv6_dual_stack_single_worker` in `tests/test_main.py`: runs a real single-worker server bound to `::` and confirms an IPv4 (`127.0.0.1`) client can connect successfully — this is the exact regression from the issue.
- Verified both new tests **fail** without the fix (confirmed `httpx.ConnectError: All connection attempts failed` for the single-worker case, and `IPV6_V6ONLY == 1` for the `bind_socket()` case) and **pass** with it.
- Ran the full `tests/test_config.py`, `tests/test_main.py`, and `tests/test_server.py` slices locally (Windows): all passing, no regressions.
- `ruff format --check`, `ruff check`, and `mypy` all clean on the changed files (pre-existing `mypy` findings on Windows are unrelated stdlib-stub platform mismatches — e.g. `AF_UNIX`/`SIGHUP` not being present in the `win32` stubs — and not on any line touched by this change).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<a href="https://cubic.dev/pr/Kludex/uvicorn/pull/3070?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * IPv6 wildcard bindings now support dual-stack operation, allowing IPv4 and IPv6 connections through the same server socket.
  * Servers configured with an IPv6 host now start reliably and expose the correct listening socket.
  * Improved handling and reporting of IPv6 socket setup failures, including clean shutdown when startup cannot complete.
  * Added coverage to verify IPv4 and IPv6 connectivity for servers bound to IPv6 addresses.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -262,6 +262,22 @@ def test_socket_bind() -> None:
     sock.close()
 
 
+def test_ipv6_socket_bind_is_dual_stack() -> None:
+    # Regression test for https://github.com/encode/uvicorn/issues/2945
+    # `bind_socket()` (used by the reload/multiprocess workers) must not rely
+    # on the OS default for `IPV6_V6ONLY` (e.g. the Linux `net.ipv6.bindv6only`
+    # sysctl), otherwise it ends up dual-stack or IPv6-only depending on the
+    # host. It should always bind dual-stack, matching the fixed single-worker
+    # path (see `test_run_ipv6_dual_stack_single_worker` in `test_main.py`).
+    config = Config(app=asgi_app, host="::", port=0)
+    config.load()
+    sock = config.bind_socket()
+    try:
+        assert sock.getsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY) == 0
+    finally:
+        sock.close()
+
+
 def test_ssl_config(
     tls_ca_certificate_pem_path: str,
     tls_ca_certificate_private_key_path: str,
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -18,7 +18,7 @@ import yaml
 from pytest_mock import MockerFixture
 
 from tests.custom_loop_utils import CustomLoop
-from tests.utils import as_cwd, get_asyncio_default_loop_per_os
+from tests.utils import as_cwd, get_asyncio_default_loop_per_os, has_ipv6
 from uvicorn._types import ASGIApplication, ASGIReceiveCallable, ASGISendCallable, Environ, Scope, StartResponse
 from uvicorn.config import Config, LoopFactoryType, UvicornDeprecationWarning
 from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
@@ -262,6 +262,7 @@ def test_socket_bind() -> None:
     sock.close()
 
 
+@pytest.mark.skipif(not has_ipv6("::"), reason="IPV6 not enabled")
 def test_ipv6_socket_bind_is_dual_stack() -> None:
     # Regression test for https://github.com/encode/uvicorn/issues/2945
     # `bind_socket()` (used by the reload/multiprocess workers) must not rely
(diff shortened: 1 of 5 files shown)


aleks-drozy:
Noting for the record: the CodSpeed check flags a -10% regression on `test_bench_fragmented_body[httptools]`, but this PR's diff only touches IPv6 socket binding in `config.py`/`server.py` — nothing on the HTTP parsing/body-fragmentation path. Looks like unrelated runner noise rather than a real regression from this change, but flagging in case it's a known flaky benchmark.

PR #3101, opened 2026-08-28
Require explicit opt-in for zttp HTTP/2

Kludex:
# Summary

Keep `zttp` as the only exposed zttp protocol implementation.
Add `--http2` and `http2=True` as explicit opt-ins for HTTP/2 negotiation.
Preserve HTTP/1.1 as the default and update the tests and documentation.

# Checklist

- [x] I understand that this PR may be closed if there was no previous discussion.
- [x] Tests cover each introduced behavior.
- [x] Documentation is updated.

<a href="https://cubic.dev/pr/Kludex/uvicorn/pull/3101?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## AI Disclaimer


tests/test_cli.py
```diff
@@ -151,6 +151,30 @@ def test_cli_event_size() -> None:
...
+
+def test_cli_http2_requires_zttp_protocol() -> None:
+    runner = CliRunner()
+
+    result = runner.invoke(cli, ["tests.test_cli:App", "--http", "h11", "--http2"])
+
+    assert result.exit_code == 2
+    assert "Invalid value for --http2: HTTP/2 requires the `zttp` HTTP protocol" in result.output
+    assert "pip install zttp" in result.output
+    assert "https://uvicorn.dev/concepts/http2/" in result.output
```
Kludex, line 175: Are we testing something useful here, or can we remove those?

uvicorn/config.py
```diff
@@ -203,6 +201,7 @@ def __init__(
         fd: int | None = None,
         loop: LoopFactoryType | str = "auto",
         http: type[asyncio.Protocol] | HTTPProtocolType | str = "auto",
+        http2: bool = False,
         ws: type[asyncio.Protocol] | WSProtocolType | str = "auto",
```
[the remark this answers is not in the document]
Kludex: We are accepting this compatibility break. Given the long `Config` signature, new configuration should use keyword arguments.

uvicorn/config.py
```diff
@@ -439,7 +442,12 @@ def load(self) -> None:
         assert not self.loaded
 
         if isinstance(self.http, str):
-            http_protocol_class = import_from_string(HTTP_PROTOCOLS.get(self.http, self.http))
+            http_protocol = (
+                "uvicorn.protocols.http.auto_zttp_impl:AutoZttpProtocol"
+                if self.http2
+                else HTTP_PROTOCOLS.get(self.http, self.http)
```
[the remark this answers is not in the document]
Kludex: Addressed in `ec7ed0d`: `http2=True` now rejects every `http` value except `"zttp"`, so custom implementations are no longer silently replaced. Custom HTTP/2 protocols remain usable without the flag.

tests/test_cli.py
Kludex: These cover separate public CLI behavior. `test_cli_http2` ensures Click forwards the flag to `run()`. `test_cli_http2_requires_zttp_protocol` ensures invalid combinations produce a usage error with the install command and documentation link instead of a traceback. I would keep both.

uvicorn/main.py
```diff
@@ -437,6 +446,9 @@ def main(
     reset_contextvars: bool,
     factory: bool,
 ) -> None:
+    if http2 and http != "zttp":
+        raise click.BadParameter(HTTP2_REQUIREMENTS, param_hint="--http2")
```
Kludex, line 450: We don't need this here... Only in `Config` is fine.

uvicorn/config.py
```diff
@@ -35,12 +35,17 @@ class UvicornDeprecationWarning(UserWarning):
...
+HTTPProtocolType = Literal["auto", "h11", "httptools", "zttp"]
 WSProtocolType = Literal["auto", "none", "websockets", "websockets-sansio", "wsproto"]
 LifespanType = Literal["auto", "on", "off"]
 LoopFactoryType = Literal["none", "auto", "asyncio", "uvloop"]
 InterfaceType = Literal["auto", "asgi3", "asgi2", "wsgi"]
 
+HTTP2_REQUIREMENTS = (
+    "HTTP/2 requires the `zttp` HTTP protocol. Install it with `pip install zttp`, then select it with "
+    "`--http zttp` or `http='zttp'`. See https://uvicorn.dev/concepts/http2/ for more information."
+)
```
Kludex, line 47: I don't think we need a constant for this. We can inline the error message in the exception.

uvicorn/config.py
```diff
@@ -249,13 +253,17 @@ def __init__(
         h11_max_incomplete_event_size: int | None = None,
         reset_contextvars: bool = False,
     ):
+        if http2 and http != "zttp":
+            raise ValueError(HTTP2_REQUIREMENTS)
+
```
Kludex, line 258: I think there are better places in the code to put this error... This doesn't follow our current pattern!

PR #3134, opened 2026-09-12
Remove races from supervisor tests

Kludex:
## Summary

- Join the multiprocess supervisor thread instead of its potentially uninitialized process list.
- Keep touching watched files until `watchfiles` detects them, avoiding its lazy-startup race.
- Test SIGHUP replacement independently from worker readiness timing, which is covered separately.

## Testing

- `scripts/test` - 1336 passed, 19 skipped, 100% coverage
- Full suite on free-threaded Python 3.14 - 1336 passed, 19 skipped, 100% coverage
- Both newly addressed failures passed 20 consecutive runs on free-threaded Python 3.14

## AI Disclaimer


tests/supervisors/test_multiprocess.py
```diff
@@ -80,9 +80,18 @@ def test_multiprocess_run() -> None:
...
-    supervisor.join_all()
+    thread.join(timeout=10)
+    if thread.is_alive():  # pragma: no cover - only runs after a timeout
+        supervisor.should_exit.set()
+        for process in supervisor.processes:
+            if process.exitcode is None:
+                process.kill()
+                process.join()
+        thread.join(timeout=10)
+    assert not thread.is_alive(), "Supervisor did not shut down in time"
```
Kludex, line 94: This seems too much.

PR #3135, opened 2026-09-13
fix: set IPV6_V6ONLY explicitly when binding an IPv6 socket

### LINKED ISSUE #2945: Inconsistent IPv6 dual-stack behavior between single-worker and multi-worker modes

Selahattinozdmr:
## Summary
Fixes #2945.

`bind_socket()` (the multi-worker path in `config.py`) creates IPv6 sockets without setting `IPV6_V6ONLY`, so the socket silently inherits the platform's dual-stack default (`0`/dual-stack on Linux). The single-worker path delegates to asyncio's `loop.create_server()`, which always binds IPv6 sockets as IPv6-only. This made `--host '::'` accept IPv4 connections when running with multiple workers but reject them with a single worker, with no way to tell from the config which behavior you'd get.

## Fix
Explicitly set `IPV6_V6ONLY=True` in `bind_socket()` so both paths agree, matching asyncio's existing behavior rather than introducing a new default.

This is intentionally the conservative fix: it makes the two code paths consistent without adding a new public config option, so it doesn't foreclose a follow-up discussion about whether uvicorn should expose an opt-in dual-stack setting later.

## Test plan
- [x] Added `test_socket_bind_ipv6_sets_v6only`, which binds an IPv6 socket via `Config(host="::1").bind_socket()` and asserts `IPV6_V6ONLY` is set.
- [x] `pytest tests/test_config.py` — 127 passed
- [x] `ruff format --check` / `ruff check` clean

<a href="https://cubic.dev/pr/Kludex/uvicorn/pull/3135?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Selahattinozdmr:
Good catch — fixed in 1e3a5e2. The unconditional `setsockopt(IPPROTO_IPV6, IPV6_V6ONLY, True)` could raise on platforms where the constants exist but the option itself isn't accepted by the kernel (same edge case `socket.has_dualstack_ipv6()` guards against). Now checks for the constants and catches `OSError` around the call, so an unsupported platform silently skips it instead of failing startup.

Selahattinozdmr:
Addressed both follow-ups in e026fae:
- Log a warning instead of silently swallowing the `OSError` when `IPV6_V6ONLY` is rejected, so operators know the socket fell back to dual-stack.
- Guarded the test's option assertion with the same `hasattr` check, and added a mocked-`OSError` test asserting `bind_socket()` still returns a bound socket when the platform rejects the option.

Selahattinozdmr:
Good point — fixed in e8fd06f: the test now asserts the warning is actually present in `caplog`, not just that the socket still binds.

